PR #467, opened 2017-09-08
Pylint: Disable scoring and excess parallelism. Log timing and version.

markfickett:
On my Google workstation, I had pylint 1.1.x which has scoring off by default and does not recognize the `--score` flag. Upgrade with `sudo pip install --upgrade pylint`. CircleCI uses the latest (1.7.2).

Setting `--score n` reduces log output, but only reduced the test time from 7:30 (previous run) to 7:10; scoring was not causing the slowness.

Switching from `parallel` to passing the file lists directly to `pylint` significantly sped thigns up, especially on CircleCI. Using `pylint -j 0` for pylint's own parallelism speeds things up slightly (10s -> 8s) locally, but slows things down significantly on CircleCI (11s -> 1m). So I think the moral is parallelism in a constrained / VM environment is not a benefit (even if tools think the machine has 32 cores).


markfickett:
Commented!

PR #468, opened 2017-09-08
Adding missing column

danqrodney:
I messed up and forgot to add this Alembic update to the schema, so biobank import is failing in our environments.

Given that the code is already pushed out, I'm just going to apply this directly to staging/stable/prod manually (once this gets successfully applied to test by autopush.)


danqrodney:
(Note to self: we need to have a test that ensures there aren't model changes which aren't reflected in Alembic versions.)

I'm going to submit this TBR so we can get things working again in prod.

danqrodney:
As mentioned offline, our tests use the schema generated directly from our model objects by SQLAlchemy, rather than running Alembic migrations.

At a minimum we should add a test to ensure that running "generate_schema" returns an empty migration before submitting.

PR #470, opened 2017-09-13
[DA-384] Auto-pair UNSET participants based on PMs or orders.

markfickett:
* Require that all PhysicalMeasurements (except amendments)  and BiobankOrders have finalizedSiteId.
* Update tests to include site info.

After this is in, we will need to backfill Participant.hpoId and .providerLink, and ParticipantSummary.hpoId. (BiobankOrder and PhysicalMeasurements finalizedSiteId have already been backfilled.)

### DIFF
--- a/rest-api/dao/participant_dao.py
+++ b/rest-api/dao/participant_dao.py
@@ -248,7 +248,10 @@ def raise_if_withdrawn(obj):
 
 
 def make_primary_provider_link(hpo_name=None, hpo_id=None, hpo=None):
-  """Returns serialized FHIR JSON for a primary provider link based on HPO information."""
+  """Returns serialized FHIR JSON for a primary provider link based on HPO information.
+
+  Exactly one of hpo_name, hpo_id, or hpo (a model.HPO) must be provided.
+  """
   if len([v for v in hpo_name, hpo_id, hpo if v is not None]) != 1:
     raise ValueError(
         'Exactly one of hpo_name=%r hpo_id=%r or hpo=%r must be defined.'
--- a/rest-api/dao/participant_dao.py
+++ b/rest-api/dao/participant_dao.py
@@ -210,7 +210,7 @@ class ParticipantDao(UpdatableDao):
       if site is None:
         raise BadRequest('Invalid siteId reference %r.' % site_id)
 
-      participant = self.get_for_update(participant_id)
+      participant = self.get_for_update(session, participant_id)
       if participant.hpoId != UNSET_HPO_ID:
         return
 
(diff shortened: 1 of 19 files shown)


rest-api/dao/participant_dao.py
```diff
@@ -201,6 +202,26 @@ def from_client_json(self, resource_json, id_=None, expected_version=None, clien
         withdrawalStatus=resource_json.get('withdrawalStatus'),
         suspensionStatus=resource_json.get('suspensionStatus'))
 
+  def add_missing_hpo_from_site(self, participant_id, site_id):
```
danqrodney: session should be passed into this method... everything should happen in the same transaction.

Also, we already have code that updates participant summaries when physical measurements are updated here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/physical_measurements_dao.py#L252

And fetch the participant summary when inserting a biobank order here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/biobank_order_dao.py#L114

It would be better to avoid looking up this data twice, and instead pass in the participant summary, etc. into a common method.

rest-api/dao/participant_dao.py
```diff
@@ -225,3 +246,24 @@ def _get_hpo_name_from_participant(participant):
 def raise_if_withdrawn(obj):
   if obj.withdrawalStatus == WithdrawalStatus.NO_USE:
     raise Forbidden('Participant %d has withdrawn' % obj.participantId)
+
+
+def make_primary_provider_link(hpo_name=None, hpo_id=None, hpo=None):
```
danqrodney: The alternative to this is having make_primary_provider_link() wrapped by make_primary_provider_link_with_hpo_id() and make_primary_provider_link_with_hpo() helper methods that take one of those things...

I prefer that to having this looping conditional logic.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -215,19 +216,24 @@ def set_measurement_ids(physical_measurements):
...
     if obj.logPosition is not None:
       raise BadRequest('%s.logPosition must be auto-generated.' % self.model_type.__name__)
     obj.logPosition = LogPosition()
     obj.final = True
     obj.created = clock.CLOCK.now()
     resource_json = json.loads(obj.resource)
     for extension in resource_json['entry'][0]['resource'].get('extension', []):
-      url = extension.get('url', '')
-      if url != _AMENDMENT_URL:
-        logging.info('Ignoring unsupported extension for PhysicalMeasurements: %r.' % url)
```
danqrodney: So previously were we logging "Ignoring unsupported extension" for the created / finalized extensions?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -241,7 +247,11 @@ def insert_with_session(self, session, obj):
           return measurements
     PhysicalMeasurementsDao.set_measurement_ids(obj)
 
-    super(PhysicalMeasurementsDao, self).insert_with_session(session, obj)
+    inserted_obj = super(PhysicalMeasurementsDao, self).insert_with_session(session, obj)
+    if not is_amendment:  # Amendments aren't expected to have site ID extensions.
+      ParticipantDao().add_missing_hpo_from_site(
```
danqrodney: Are we positive that all new physical measurements will have these extensions? (If you haven't already, please check with Jason Tan.)

rest-api/test/unit_test/api_test/physical_measurements_api_test.py
```diff
@@ -15,7 +15,7 @@ def setUp(self):
     self.participant_id = self.create_participant()
```
danqrodney: Can you add the equivalent of test_auto_pair_updates_participant_and_summary for this test (physical measurements API) as well?

rest-api/dao/participant_dao.py
markfickett: Switched.

rest-api/dao/participant_dao.py
danqrodney: OK. I believe as long as you use the same session, it shouldn't incur extra RPCs to look up the same objects multiple times, so hopefully this is OK.

rest-api/dao/participant_dao.py
```diff
@@ -201,6 +202,25 @@ def from_client_json(self, resource_json, id_=None, expected_version=None, clien
...
+    if participant is None:
+      raise BadRequest('No participant %r for HPO ID udpate.' % participant_id)
+    if participant.hpoId != UNSET_HPO_ID:
+      return
+
+    participant.hpoId = site.hpoId
+    participant.providerLink = make_primary_provider_link_for_id(site.hpoId)
+    if participant.participantSummary is None:
+      raise RuntimeError('No ParticipantSummary available for P%d.' % participant_id)
+    participant.participantSummary.hpoId = site.hpoId
```
danqrodney: Do you need to do session.merge() somewhere? (I guess not?)

markfickett: I mostly didn't do it because it worked. (: But revisiting the docs, merging is only necessary when you create/modify something from outside a session. (Essentially an INSERT OR UPDATE I think.) http://docs.sqlalchemy.org/en/latest/orm/session_state_management.html#unitofwork-merging

PR #471, opened 2017-09-14
[DA-384] SQL to backfill HPO based on BiobankOrder/PhysicalMeasurements.

markfickett:
What do you think of this option?

Example output from the SELECT (demonstrating that the `providerLink` JSON is valid and HPO ID is going from 0 to nonzero):
```
+------------+------------+----------------------------------------------------------------------------+
| old_hpo_id | new_hpo_id | provider_link                                                              |
+------------+------------+----------------------------------------------------------------------------+
|          0 |          1 | [{"primary":true,"organization":{"reference":"Organization/PITT"}}]        |
|          0 |          1 | [{"primary":true,"organization":{"reference":"Organization/PITT"}}]        |
|          0 |          1 | [{"primary":true,"organization":{"reference":"Organization/PITT"}}]        |
...
|          0 |          5 | [{"primary":true,"organization":{"reference":"Organization/COMM_HEALTH"}}] |
|          0 |         12 | [{"primary":true,"organization":{"reference":"Organization/CAL_PMC"}}]     |
|          0 |         12 | [{"primary":true,"organization":{"reference":"Organization/CAL_PMC"}}]     |
|          0 |         14 | [{"primary":true,"organization":{"reference":"Organization/TRANS_AM"}}]    |
+------------+------------+----------------------------------------------------------------------------+
18 rows in set (0.04 sec)
```

Since there are only ~200 participants with no HPO pairing, it would also be easy to write a script that iterates over all them, and for each (within its own transaction) verifies that it's still not paired and updates it using DAO methods.

SQL: Pro: already written / fast to write. Easy to audit by querying before/after state. Con: Slightly less formal.
DAO: Pro: easier to follow logic. Con: More to write.

### DIFF
--- a/rest-api/tools/participant_hpo_backfill.sql
+++ b/rest-api/tools/participant_hpo_backfill.sql
@@ -1,5 +1,7 @@
 # This SELECT mirrors the UPDATE below, to serve as pre-validation.
 # See https://precisionmedicineinitiative.atlassian.net/browse/DA-384 .
+# TODO(markfickett) Delete this file once the backfill has been run (it only
+# serves as a record / for review).
 
 SELECT
   participant.hpo_id old_hpo_id,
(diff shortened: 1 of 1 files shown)


markfickett:
Done with validation on a local db, notes at https://docs.google.com/document/d/1RDiR4jGFe9-JDVohC1U__FQSIaWhFHv8YoZykZ2Tb5w . I'll merge this now and work on the prod update (and send another PR after the update's done to delete this file again).

PR #472, opened 2017-09-15
alpha delivery, ETL for Survey data

atsvetkova-ody:


rest-api/etl/etl.sql
```diff
@@ -0,0 +1,2062 @@
...
+-- -------------------------------------------------------------------
+
+INSERT INTO cdm.measurement
+SELECT DISTINCT
+    NULL                                    AS measurement_id,
+    meas.participant_id                     AS person_id,
+    meas.cv_concept_id                      AS measurement_concept_id,
+    DATE(meas.measurement_time)             AS measurement_date,
+    TIME(meas.measurement_time)             AS measurement_time,
+    44818701                                AS measurement_type_concept_id,
```
jmandel: Can you add comments documenting static codes like `44818701` here?

rest-api/etl/etl.sql
```diff
@@ -0,0 +1,2062 @@
...
+
+-- -------------------------------------------------------------------
+-- Rules all together
+-- -------------------------------------------------------------------
+
+INSERT INTO cdm.tmp_clean_all
+SELECT
+    pa.participant_id               AS participant_id,
+    qr.created                      AS date_of_survey,
+    co_q.value                      AS question_ppi_code,
```
danqrodney: We now have a short_value field on the code table, which will be populated in cases where value > 50 characters. We should do something like IFNULL(co_q.short_value, co_q.value) so that short_value is used when set.

rest-api/etl/etl.sql
```diff
@@ -0,0 +1,2062 @@
...
+    qq.code_id                      AS question_code_id,
+    co_a.value                      AS value_ppi_code,
```
danqrodney: See above, should be something like IFNULL(co_a.short_value, co_a.value)

rest-api/etl/etl.sql
```diff
@@ -0,0 +1,2062 @@
...
+    END                                 AS value_boolean_concept_id,
+    src_c.value_date                    AS value_date,
+    src_c.value_string                  AS value_string,
+    src_c.questionnaire_response_id     AS questionnaire_response_id,
+    src_c.unit_id                       AS unit_id 
+FROM cdm.src_clean src_c
+JOIN cdm.src_participant src_p
+    ON  src_c.participant_id = src_p.participant_id
+LEFT JOIN voc.concept vc1 
+    ON  LEFT(src_c.question_ppi_code, 50) = vc1.concept_code
```
danqrodney: You technically won't need the LEFT() here anymore once using short_value above. Though it doesn't hurt.

rest-api/etl/etl.sql
```diff
@@ -0,0 +1,2062 @@
...
+CREATE TABLE cdm.src_race
+(
+    person_id                   bigint,
+    ppi_code                    varchar(255),
+    race_source_concept_id      bigint,
+    race_target_concept_id      bigint,
+    PRIMARY KEY (person_id)
+);
+
+INSERT INTO cdm.src_race
```
danqrodney: How are people with multiple race answers being handled here?

rest-api/etl/etl.sql
atsvetkova-ody: Dan, thank you for your comments!
Probably it will be good to populate short_value field with all the codes, both short and cut? Then we would just replace "value" with "short_value" everywhere. The reference "value" is being used in qa.sql frequently, replacing "value" with "IFNULL(..)" will make the code less readable. And, likely, slower.

danqrodney: OK. I can look into doing that. Please update this to use short_value.

rest-api/etl/etl.sql
danqrodney: You can remove the LEFT now.

rest-api/etl/etl.sql
```diff
@@ -0,0 +1,2113 @@
...
+    AND vc1.vocabulary_id = 'PPI'
+LEFT JOIN voc.concept_relationship vcr1
+    ON  vc1.concept_id = vcr1.concept_id_1
+    AND vcr1.relationship_id = 'Maps to'
+    AND vcr1.invalid_reason IS NULL
+LEFT JOIN voc.concept vc2
+    ON  vcr1.concept_id_2 = vc2.concept_id
+    AND vc2.standard_concept = 'S'
+LEFT JOIN voc.concept vc3 
+    ON  LEFT(src_c.value_ppi_code, 50) = vc3.concept_code
```
danqrodney: You can remove the LEFT now.

danqrodney:
Anna, I'm going to go ahead and merge this and give it a try on our end. Thanks for the changes!

PR #473, opened 2017-09-15
Adding "short_value" to code (for codes where value > 50 chars.)

danqrodney:
Changing most participant summary fields to be 255 chars, to match PTC expectations.


rest-api/model/code.py
```diff
@@ -42,6 +42,9 @@ class _CodeBase(object):
   codeId = Column('code_id', Integer, primary_key=True)
   system = Column('system', String(255), nullable=False)
   value = Column('value', String(80), nullable=False)
```
markfickett: Should we bump the long `value` to 255, too?

rest-api/model/code.py
```diff
@@ -42,6 +42,9 @@ class _CodeBase(object):
...
+  # OMOP codes are supposed to be at most 50 characters long; for legacy codes that exceeded this
+  # limit, we populate a shortened version for use in OMOP here.
```
markfickett: "Otherwise, `shortValue` remains NULL." right?

rest-api/model/participant_summary.py
```diff
@@ -27,12 +27,12 @@ class ParticipantSummary(Base):
   participantId = Column('participant_id', Integer, ForeignKey('participant.participant_id'),
                          primary_key=True, autoincrement=False)
   biobankId = Column('biobank_id', Integer, nullable=False)
-  firstName = Column('first_name', String(80), nullable=False)
-  middleName = Column('middle_name', String(80))
-  lastName = Column('last_name', String(80), nullable=False)
+  firstName = Column('first_name', String(255), nullable=False)
```
markfickett: Is this just a safeguard (seems sane), or are we matching a specific value in HealthPro or something (in which case could you comment)?

danqrodney: PTC currently doesn't impose *any* limits on most of its string fields (with exceptions for things like zip code); as a result, some data entered by users (albeit with garbage values) is failing to sync to the RDR, since it won't fit into our participant summary. 

We agreed not having any limits was not a good idea; they asked if we could increase our field sizes to 255, and I said that was fine. They probably should choose to limit some of these fields further (having a > 80 character name doesn't really make sense), but if they *do* send us a value that exceeds 80 chars, they want us to accept it. Added a comment.

PR #474, opened 2017-09-18
Updating sites to latest

danqrodney:


markfickett:
Do we expect sites to go away (for example `Northeast,VA Boston Healthcare System,7035759,HPO-VABostonHC,hpo-site-VABostonHC@prod.pmi-ops.org,VA,`)? Or do our scripts only ever add sites?

danqrodney:
At the moment, our site import is purely additive.

At some point, we intend to allow the tools to be run in such a way that sites which no longer show up can be deleted. But we haven't done that work yet.

PR #477, opened 2017-09-21
[DA-402] Changing elapsed time in reconcilation report to be based on creation…

danqrodney:
… time ending rather than

confirmed time.

Adjusting logic to count 24+ hours as late as opposed to 25+ hours.


danqrodney:
README updated.

PR #478, opened 2017-09-21
[DA-401] Changing short value to be the first 50 chars of value...

danqrodney:
when short-code isn't set.

This way we can just use the "short_value" column in our ETL, etc. without conditional logic.


danqrodney:
Done.

PR #480, opened 2017-09-22
[DA-384] Remove SQL script for completed backfill.

markfickett:


markfickett:
> I guess we can always resurrect it if that happens?

Yeah, that's my thinking.

I think we're safe w/re running the same backfill later, though, since we don't support in-place updates for BOs or PMs; so if HealthPro wants to send us new versions with better site info, it'll trigger the automatic pairing.

PR #482, opened 2017-09-26
[DA-405] Log unchanged pairings in hpo_assigner.

markfickett:


markfickett:
Ping -- I found this logging useful last time I was updating pairings. Frequently there are a number which have a re-pairing requested, but actually already have the right pairing.

PR #483, opened 2017-09-26
Getting RDR ETL running locally.

danqrodney:
Adding setup_local_database_for_etl.sh which sets up the databases and loads vocabularies, run_local_etl.sh which runs the ETL SQL.

Fixing issues with the ETL script. Adding missing parent_id column to measurement table.

Changing empty values in source_to_concept_map.csv to be \N, so as to be accepted by mysqlimport.

Fixing typo in setup_database.sh.


rest-api/etl/ddl.sql
```diff
@@ -430,6 +430,11 @@ CREATE TABLE observation
     PRIMARY KEY (observation_id)
 );
 
+CREATE INDEX ux_observation ON cdm.observation (person_id, observation_concept_id, observation_date,
```
jmandel: Are more than the first 2 cols important to index? (I know this is just moving code from another file, but it's causing me to read closer ;-))

rest-api/etl/ddl.sql
```diff
@@ -456,10 +461,18 @@ CREATE TABLE measurement
     unit_source_value varchar(50),
     value_source_value varchar(50),
     meas_id bigint,
+    parent_id bigint,
```
jmandel: This PR introduces `parent_id` but doesn't use it?

rest-api/etl/ddl.sql
```diff
@@ -456,10 +461,18 @@ CREATE TABLE measurement
...
     unit_id varchar(50) NOT NULL,
     PRIMARY KEY (measurement_id)
 );
 
+CREATE INDEX ux_measurement ON cdm.measurement (person_id, measurement_concept_id, measurement_date,
```
jmandel: Same comment about cols indexed here.

rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -0,0 +1,111 @@
...
+mkdir -p ${CSV_DIR}/voc
+
+SERVICE_ACCOUNT=pmi-drc-api-test@appspot.gserviceaccount.com
+
+echo "Activating service account..."
+#gcloud iam service-accounts keys create $CREDS_FILE --iam-account=$SERVICE_ACCOUNT --account=$ACCOUNT
+#gcloud auth activate-service-account pmi-drc-api-test@appspot.gserviceaccount.com --key-file=$CREDS_FILE
+
+echo "Copying vocabulary files from GCS..."
+gsutil cp gs://all-of-us-rdr-vocabulary/vocabularies-2017-09-18/* ${CSV_DIR}/voc
```
jmandel: Might be good to explicitly list filemnames (as documentation if nothing else)

rest-api/etl/ddl.sql
danqrodney: TBH I haven't looked. It does seem kinda unnecessary index *all* of these. :)

I'll ask Anna.

rest-api/etl/ddl.sql
danqrodney: It is implicitly used when inserting into the measurement table. (Without it the insert fails.)

rest-api/etl/ddl.sql
danqrodney: Yep, will ask about that.

rest-api/etl/setup_local_database_for_etl.sh
danqrodney: If it turns out we need more tables in some future iteration, it'd be nice not to have to change this?

If you want documentation, I could add some comment I guess?

jmandel: I'd rather not document in comments only (will just get out of sync). Happy to leave as-is if you prefer not to whitelist.

PR #485, opened 2017-09-28
Generating SQL dumps in GCS for Cloud SQL import.

danqrodney:
Adding setup_database_for_etl.sh which imports the SQL dumps.

Changing setup_local_database_for_etl.sh to also read source_to_concept_map CSV from GCS; removing from github. (It arguably belongs with the other CSV files.)

Removing indexes we don't need from DDL.


rest-api/etl/setup_database_for_etl.sh
```diff
@@ -0,0 +1,49 @@
...
+if [ -z "${PROJECT}" ]
+then
+  echo "Usage: $USAGE"
+  exit 1
+fi
+CREDS_ACCOUNT=${ACCOUNT}
+
+echo "Activating service account..."
+source tools/auth_setup.sh
+gcloud auth activate-service-account circle-deploy@all-of-us-rdr-staging.iam.gserviceaccount.com --key-file=$CREDS_FILE
```
jmandel: Can we avoid naming `circle-deploy@all-of-us-rdr-staging.iam.gserviceaccount.com` in this (otherwise generic) script? This would appear to tie us to the staging env.

rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -15,20 +15,23 @@
...
 # Make sure the user you run the script as has write access to /var/lib/mysql-files.
+# Generates files under /var/lib/mysql-files/cloud-csv (which can be imported into GCS and Cloud
+# SQL.)
 
 source tools/setup_local_vars.sh
 DB_CONNECTION_NAME=
 CREATE_DB_FILE=/tmp/create_dbs.sql
 CSV_DIR=/var/lib/mysql-files/rdr-csv
-CREDS_FILE=/tmp/rdr_creds.json
+OUTPUT_DIR=/var/lib/mysql-files/rdr-sql-dump
```
jmandel: Does this need to be in `/var/lib`? Generally prefer tmp for this stuff if the idea is short-lived -- but maybe they're being re-used somewhere?

rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -54,7 +57,7 @@ set_local_db_connection_string
 function finish {  
   rm -f ${CREATE_DB_FILE}
   rm -rf ${CSV_DIR}
-  rm -f ${CREDS_FILE}
+  cleanup
```
jmandel: Per comment above: would clean up output here too if that's appropriate. (But OK if we need to persist for reuse.)

rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -82,22 +87,27 @@ rm -rf ${CSV_DIR}
...
 mkdir -p ${CSV_DIR}/cdm
 mkdir -p ${CSV_DIR}/voc
+if [ "${GENERATE_SQL_DUMP}" ]
+then
+  mkdir -p ${OUTPUT_DIR}
+  chmod -R 0777 ${OUTPUT_DIR}
+fi
 
-SERVICE_ACCOUNT=pmi-drc-api-test@appspot.gserviceaccount.com
+PROJECT=pmi-drc-api-test
```
jmandel: Again, good to avoid hardcoding an env.

rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -82,22 +87,27 @@ rm -rf ${CSV_DIR}
...
+CREDS_ACCOUNT="${ACCOUNT}"
 
 echo "Activating service account..."
-#gcloud iam service-accounts keys create $CREDS_FILE --iam-account=$SERVICE_ACCOUNT --account=$ACCOUNT
-#gcloud auth activate-service-account pmi-drc-api-test@appspot.gserviceaccount.com --key-file=$CREDS_FILE
+source tools/auth_setup.sh
+gcloud auth activate-service-account circle-deploy@all-of-us-rdr-staging.iam.gserviceaccount.com --key-file=$CREDS_FILE
```
jmandel: Same comment on hardcoding, especially given that project appears to be from "test" and SA is "staging"

rest-api/etl/setup_database_for_etl.sh
danqrodney: The vocabulary files are in a GCS bucket under pmi-drc-api-test (which this service account has access to.) We're going to use those files independent of what environment we're updating... the vocabulary doesn't change from environment to environment.

rest-api/etl/setup_local_database_for_etl.sh
danqrodney: Without some config changes to MySQL, yes, you can only access files under /var/lib/mysql-files.

rest-api/etl/setup_local_database_for_etl.sh
danqrodney: Good call -- done.

rest-api/etl/setup_local_database_for_etl.sh
danqrodney: See above re: test bucket.

rest-api/etl/setup_local_database_for_etl.sh
danqrodney: See above.

rest-api/etl/setup_database_for_etl.sh
jmandel: So I think we're saying "vocab always lives in the test project" --  two notes

1. Referring to `gs://all-of-us-rdr-vocabulary` makes sense, but why do we need to hard-code a reference to the project in which this bucket is housed?

2. Can't we just give access to this bucket to SAs as needed? So if we're working on ETL in staging, we'd ensure the staging SA has access to the `gs://all-of-us-rdr-vocabulary` bucket, but the code here could be generic?

rest-api/etl/setup_local_database_for_etl.sh
jmandel: The code here is just piping out to this file (mysql is unaware of it). Is it being used somewhere else too?

rest-api/etl/setup_database_for_etl.sh
jmandel: (And one more note: if we plan to flow vocab data into prod, it may be best to store our vocab data in an environment with more prod-like access control sensibilities and auditing; might create a new project for this.)

danqrodney: Yeah, changed setup_database_for_etl.sh to use service account for project, added comment for why we use test in setup_local_database_for_etl.sh.

Re: other note. OK. I'll think about this as a follow-up.

rest-api/etl/setup_local_database_for_etl.sh
danqrodney: You're right, OUTPUT_DIR can be under /tmp. Changed.

PR #486, opened 2017-09-29
Adding export tables utility

danqrodney:
To be used after running RDR -> OMOP ETL, and for other purposes (e.g. generating CSV data for PPI review.)

### DIFF
--- a/rest-api/tools/export_tables.py
+++ b/rest-api/tools/export_tables.py
@@ -3,6 +3,7 @@
 
 import logging
 import sys
+import MySQLdb
 
 from time import sleep
 from main_util import get_parser, configure_logging
@@ -28,7 +29,7 @@ def main(args):
           args.database
         ],
         "csvExportOptions": {
-          "selectQuery": "select * from %s" % table
+          "selectQuery": "select * from %s" % MySQLdb.escape_string(table)
         }
       }
     }
@@ -53,23 +54,18 @@ if __name__ == '__main__':
   configure_logging()
   parser = get_parser()
   parser.add_argument('--project',
-                      type=str,
                       help='Project to export data for',
                       required=True)
   parser.add_argument('--creds_file',
-                      type=str,
                       help='Path to credentials JSON file.',
                       required=True)
   parser.add_argument('--output_path',
-                      type=str,
                       help='GCS path to write the output files to.',
                       required=True)
   parser.add_argument('--database',
-                      type=str,
                       help='Name of the database containing the tables to export.',
                       required=True)
   parser.add_argument('--tables',
-                      type=str,
                       help=('Comma separated list of table names to export; ' +
                             ' all columns will be exported, in the order they are defined.'),
                       required=True)
--- a/rest-api/tools/export_tables.py
+++ b/rest-api/tools/export_tables.py
@@ -66,7 +66,7 @@ if __name__ == '__main__':
                       help='Name of the database containing the tables to export.',
                       required=True)
   parser.add_argument('--tables',
-                      help=('Comma separated list of table names to export; ' +
-                            ' all columns will be exported, in the order they are defined.'),
+                      help='Comma separated list of table names to export; '
+                        ' all columns will be exported, in the order they are defined.',
                       required=True)
   main(parser.parse_args())
--- a/rest-api/tools/export_tables.py
+++ b/rest-api/tools/export_tables.py
@@ -1,0 +1,76 @@
+"""Uses Cloud SQL export API to export the contents of one more tables to CSV files.
+"""
+
+import logging
+import sys
+
+from time import sleep
+from main_util import get_parser, configure_logging
+from googleapiclient import discovery
+from oauth2client.service_account import ServiceAccountCredentials
+
+_SCOPE = "https://www.googleapis.com/auth/cloud-platform"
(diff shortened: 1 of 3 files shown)


rest-api/tools/export_tables.py
```diff
@@ -0,0 +1,76 @@
...
+                      required=True)
+  parser.add_argument('--output_path',
+                      type=str,
+                      help='GCS path to write the output files to.',
+                      required=True)
+  parser.add_argument('--database',
+                      type=str,
+                      help='Name of the database containing the tables to export.',
+                      required=True)
+  parser.add_argument('--tables',
```
markfickett: You can use action='append' to make a list directly, for example with `--table one --table two`. https://docs.python.org/3/library/argparse.html#action

rest-api/tools/export_tables.py
```diff
@@ -0,0 +1,76 @@
...
+      logging.error("Errors in response: " % response)
+      sys.exit(-1)
+    else:
+      logging.info('Done exporting %s.' % table)
+
+if __name__ == '__main__':
+  configure_logging()
+  parser = get_parser()
+  parser.add_argument('--project',
+                      type=str,
```
markfickett: `type=str` is the default

rest-api/tools/export_tables.sh
```diff
@@ -0,0 +1,42 @@
...
+  exit 1
+fi
+
+if [ -z "${CREDS_ACCOUNT}" ]
+then
+  CREDS_ACCOUNT="${ACCOUNT}"
+fi
+
+source tools/auth_setup.sh
+SQL_SERVICE_ACCOUNT=`gcloud sql instances describe --project ${PROJECT} --account ${ACCOUNT} \
```
markfickett: This looks important -- bubble it up to the top-level docs? We're taking ownership of the specified bucket?

rest-api/tools/export_tables.sh
```diff
@@ -0,0 +1,42 @@
...
+  rdrmaindb | grep serviceAccountEmailAddress | cut -d: -f2`
+gsutil acl ch -u ${SQL_SERVICE_ACCOUNT}:W gs://${BUCKET}
+
+REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd ../.. && pwd )"
```
markfickett: Please use https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/tools/set_path.sh . (Last time something about the path changed it was a pain to hunt down & centralize path setup.)

rest-api/tools/export_tables.py
```diff
@@ -0,0 +1,76 @@
...
+    request_body = {
+      "exportContext": {
+        "kind": "sql#exportContext",
+        "fileType": "CSV",
+        "uri": uri,
+        "databases": [
+          args.database
+        ],
+        "csvExportOptions": {
+          "selectQuery": "select * from %s" % table
```
markfickett: We should either use a SQL escaping library or make it very clear the table name is going into SQL raw.

rest-api/tools/export_tables.py
danqrodney: Removed.

rest-api/tools/export_tables.py
danqrodney: OK. Given that I have a shell script wrapper for this thing right now, it would get complicated to do this properly. I think I'm gonna keep it comma-separated if that's cool.

rest-api/tools/export_tables.sh
danqrodney: I changed the logic to remove the service account from the ACL when the script finishes. Added a comment at the top.

rest-api/tools/export_tables.sh
danqrodney: Done.

rest-api/tools/export_tables.py
danqrodney: Done.

danqrodney:
Also worth noting: all Cloud SQL exports show up in the Cloud SQL audit logs.

danqrodney:
WRT blacklisting: we don't really want to blacklist export of any particular table or environment, and I'm not sure we have the capability to do so anyway... you either have the ability to run Cloud SQL export or you don't. Project owners always have that capability.

rest-api/tools/export_tables.py
```diff
@@ -0,0 +1,72 @@
...
+                      required=True)
+  parser.add_argument('--output_path',
+                      help='GCS path to write the output files to.',
+                      required=True)
+  parser.add_argument('--database',
+                      help='Name of the database containing the tables to export.',
+                      required=True)
+  parser.add_argument('--tables',
+                      help=('Comma separated list of table names to export; ' +
+                            ' all columns will be exported, in the order they are defined.'),
```
markfickett: I think Python will do string concatenation if you leave off the parens and the plus. For example this works:

```
'+'.join(['abcde'
    'fghij', 'KLMNO'])
'abcdefghij+KLMNO'
```

rest-api/tools/export_tables.sh
markfickett: Thanks!

PR #489, opened 2017-10-02
Adding scripts for exporting ETL results to GCS, transforming them.

danqrodney:
(Due to Cloud SQL export issues, we have to replace "N with empty string in the result files -- we'll get rid of this once Cloud SQL export supports specifying null values.)

Moving DDL for main OMOP CDM tables to etl.sql, so they get recreated each run; dropping columns only used for ETL at the end, so that when we export we get just the columns we want for the CDR.


rest-api/etl/transform_etl_results.sh
```diff
@@ -0,0 +1,58 @@
...
+mkdir -p ${CSV_DIR}/transformed
+
+CLOUD_DIR=gs://${PROJECT}-cdm/${DIRECTORY}
+gsutil cp ${CLOUD_DIR}/*.csv ${CSV_DIR}
+
+echo "Transforming CSV files..."
+for file in ${CSV_DIR}/*.csv
+do
+    filename=$(basename "$file")
+    cat $file | sed s/\"N//g > ${CSV_DIR}/transformed/$filename
```
jmandel: Doesn't this strip `.csv` from the filename" Would be good to preserve.

danqrodney: No, it doesn't. basename includes the extension.

PR #494, opened 2017-10-04
[DA-403] Roll forward of previous data gen fix

danqrodney:
Now data_gen_api passes 1 - fraction_missing into biobank_order_dao.get_ordered_samples_sample() to make sure we don't randomly sample out participants here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/biobank_order_dao.py#L168

This avoids the indeterminism we were seeing earlier.


markfickett:
Ah, I missed that we were dropping some samples in `samples_sample` as well as skipping some with the `random.random()` call subsequent.

I think the `_PARTICIPANTS_WITH_STORED_SAMPLES` constant is now unused & OK to remove.

danqrodney:
Constant removed.

PR #495, opened 2017-10-04
Fixing  questionnaire_response_answer_view to do an inner join (not cross join) on participant_summary.

danqrodney:
(I tested this in the test env.)

### DIFF
--- a/rest-api/alembic/versions/89dd3b8f4447_fix_participant_view.py
+++ b/rest-api/alembic/versions/89dd3b8f4447_fix_participant_view.py
@@ -1,0 +1,64 @@
+"""Fix participant view
+
+Revision ID: 89dd3b8f4447
+Revises: 8b1281fb55ed
+Create Date: 2017-10-04 17:10:35.319343
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '89dd3b8f4447'
+down_revision = '8b1281fb55ed'
+branch_labels = None
+depends_on = None
+
+# Creates a view that can be used by the Vanderbilt team (either directly or via CSV export)
+# to analyze PPI quality.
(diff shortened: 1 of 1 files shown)


rest-api/alembic/versions/89dd3b8f4447_fix_participant_view.py
```diff
@@ -0,0 +1,64 @@
...
+ FROM
+   participant p
+    INNER JOIN questionnaire_response qr ON p.participant_id = qr.participant_id
+    INNER JOIN questionnaire_response_answer qra
+       ON qra.questionnaire_response_id = qr.questionnaire_response_id
+    INNER JOIN questionnaire_question qq ON qra.question_id = qq.questionnaire_question_id
+    INNER JOIN questionnaire q ON qq.questionnaire_id = q.questionnaire_id
+    INNER JOIN code qc ON qq.code_id = qc.code_id
+    INNER JOIN participant_summary ps ON p.participant_id = ps.participant_id
+    LEFT OUTER JOIN hpo ON p.hpo_id = hpo.hpo_id
```
jmandel: For these 4 inner joints:

```
```

... technically the right-hand values should always be present -- might be good to change these to `LEFT OUTER` to expose NULLs anywhere this assumption doesn't hold.

jmandel: @danqrodney points out that the schema already ensures this (so they're not just "assumptions", but documented constraints).

danqrodney: As noted over Slack: I prefer INNER JOIN here for clarity and to give the RDBMS a hint about the execution plan. The not null FK constraints will ensure these joins never fail.

PR #497, opened 2017-10-06
[DA-403] Adding script for generating PPI answer spec.

danqrodney:
Updating questionnaire_response_answer_view to contain questionnaire response ID. (Used for calculating number of questionnaire responses in spec.)

### DIFF
--- a/rest-api/alembic/versions/7e250583b9cb_add_qra_qr_id.py
+++ b/rest-api/alembic/versions/7e250583b9cb_add_qra_qr_id.py
@@ -1,0 +1,65 @@
+"""add_qra_qr_id
+
+Revision ID: 7e250583b9cb
+Revises: 89dd3b8f4447
+Create Date: 2017-10-05 15:06:41.213664
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '7e250583b9cb'
+down_revision = '89dd3b8f4447'
+branch_labels = None
+depends_on = None
+
+# Creates a view that can be used by the Vanderbilt team (either directly or via CSV export)
+# to analyze PPI quality.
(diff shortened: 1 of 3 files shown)


rest-api/alembic/versions/7e250583b9cb_add_qra_qr_id.py
```diff
@@ -0,0 +1,65 @@
...
+    INNER JOIN questionnaire_response qr ON p.participant_id = qr.participant_id
+    INNER JOIN questionnaire_response_answer qra
+       ON qra.questionnaire_response_id = qr.questionnaire_response_id
+    INNER JOIN questionnaire_question qq ON qra.question_id = qq.questionnaire_question_id
+    INNER JOIN questionnaire q ON qq.questionnaire_id = q.questionnaire_id
+    INNER JOIN code qc ON qq.code_id = qc.code_id
+    INNER JOIN participant_summary ps ON p.participant_id = ps.participant_id
+    LEFT OUTER JOIN hpo ON p.hpo_id = hpo.hpo_id
+    LEFT OUTER JOIN code ac ON qra.value_code_id = ac.code_id
+    WHERE (ps.email IS NULL OR ps.email NOT LIKE '%@example.com') AND
```
jmandel: Are there really Questionnaire Answer rows associated with participants where `ps.email IS NULL`?

danqrodney: No, shouldn't be. This is a bit overly defensive I guess. :)

danqrodney: (Or... just unnecessary.)

PR #498, opened 2017-10-06
[DA-403] Synthetic PPI generation based on CSV spec

danqrodney:
Changing import_questionnaires.sh to pull latest questionnaires from github.

### DIFF
--- a/rest-api/data_gen/fake_participant_generator.py
+++ b/rest-api/data_gen/fake_participant_generator.py
@@ -6,6 +6,7 @@ import datetime
 import json
 import logging
 import random
+import string
 
 from concepts import Concept
 from code_constants import PPI_SYSTEM, CONSENT_FOR_STUDY_ENROLLMENT_MODULE
@@ -24,6 +25,7 @@ from code_constants import PMI_PREFER_NOT_TO_ANSWER_CODE, PMI_OTHER_CODE, BIOBAN
 from code_constants import HEALTHPRO_USERNAME_SYSTEM, SITE_ID_SYSTEM
 from field_mappings import QUESTION_CODE_TO_FIELD
 
+from cloudstorage import cloudstorage_api
 from dao.code_dao import CodeDao
 from dao.hpo_dao import HPODao
 from dao.participant_dao import make_primary_provider_link_for_hpo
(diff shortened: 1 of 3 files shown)


rest-api/data_gen/fake_participant_generator.py
```diff
@@ -574,12 +602,69 @@ def _choose_name(self, answer_map):
...
+      return [{'valueDateTime': (min_date +
+                                 datetime.timedelta(seconds=random.randint(0, seconds_diff)))
+                                 .isoformat()}
+              for _ in xrange(answer_count)]
+    if int(answer_spec['boolean_answer_count']) > 0:
+      return [{'valueBoolean': random.random() < 0.5}]
+    if int(answer_spec['string_answer_count']) > 0:
+      return [{'valueString': ''.join([random.choice(string.lowercase) for _ in xrange(20)])}
+              for _ in xrange(answer_count)]
+    if int(answer_spec['uri_answer_count']) > 0:
```
jmandel: Can we make this example.com or something else that obviously won't point to data?

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -574,12 +602,69 @@ def _choose_name(self, answer_map):
...
+      if random.random() > float(num_participants) / float(self._answer_specs_max_participants):
+        continue
+      num_questionnaire_responses = int(answer_spec['num_questionnaire_responses'])
+      num_answers = int(answer_spec['num_answers'])
+      answer_count = 1
+      if num_answers > num_questionnaire_responses:
+        rand_val = random.random() * num_answers / num_questionnaire_responses
+        if rand_val > 1:
+          # Set the answer count to 2 or more
+          answer_count = int(1 + min(1, rand_val))
```
jmandel: The min function here always returns 1, no?

rest-api/data_gen/fake_participant_generator.py
danqrodney: Done.

rest-api/data_gen/fake_participant_generator.py
danqrodney: Oops. Changed to max.

danqrodney:
I had to push with -f, so the old stuff is gone. Responded to your comments.

PR #499, opened 2017-10-10
Fixing data generation to re-fetch participant before updating status; updating metrics pipeline to run at 4 EST.

danqrodney:
Changing metrics pipeline to run at 4 AM, after biobank samples pipeline should have finished.

### DIFF
--- a/rest-api/cron.yaml
+++ b/rest-api/cron.yaml
@@ -1,7 +1,8 @@
 cron:
 - description: Daily metrics
   url: /offline/MetricsRecalculate
-  schedule: every day 01:00
+  schedule: every day 04:00
+  timezone: America/New_York
   target: offline
 - description: Daily Biobank sample import and order reconciliation
   url: /offline/BiobankSamplesImport
--- a/rest-api/data_gen/fake_participant_generator.py
+++ b/rest-api/data_gen/fake_participant_generator.py
@@ -487,8 +487,9 @@ class FakeParticipantGenerator(object):
     result = self._update_participant(change_time, participant_response, participant_id)
     return change_time, result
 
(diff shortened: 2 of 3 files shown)


rest-api/cron.yaml
```diff
@@ -1,7 +1,8 @@
 cron:
 - description: Daily metrics
   url: /offline/MetricsRecalculate
-  schedule: every day 01:00
+  schedule: every day 04:00
```
markfickett: Do you mean for this to be in the same PR? Would you mind either calling it out in the PR title or splitting it up? I'd like it to show up in release notes.

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -489,13 +489,17 @@ def _submit_hpo_changes(self, participant_response, participant_id, consent_time
 
   def _submit_status_changes(self, participant_response, participant_id, last_request_time):
```
markfickett: The participant_response parameter can be removed I think.

markfickett: It looks like this is called immediately after the response is fetched; why does it need to be re-fetched?

markfickett: Chatted offline -- adding a comment will be useful, but makes sense now.

rest-api/cron.yaml
danqrodney: Added to PR title.

rest-api/data_gen/fake_participant_generator.py
danqrodney: Done.

danqrodney: Added comments.

PR #500, opened 2017-10-11
Adding ExportTables endpoint, used to export table contents to Unicode CSV in GCS.

danqrodney:
Removing tools/export_tables.py, which used Cloud SQL export and thus didn't handle null characters or newlines properly.


rest-api/etl/export_etl_results.sh
```diff
@@ -13,18 +13,26 @@ while true; do
...
-if [ -z "${ACCOUNT}" ] || [ -z "${PROJECT}" ] || [ -z "${DIRECTORY}" ]
+if [ -z "${DIRECTORY}" ]
 then
   echo "Usage: $USAGE"
   exit 1
 fi
+PROJECT_AND_ACCOUNT=
+if [ "${PROJECT}" ]
+then
+  PROJECT_AND_ACCOUNT="--project ${PROJECT} --account ${ACCOUNT}"
```
jmandel: Both project and account are required, right? Why not treat them as two arguments below?

rdr_client/export_tables.py
```diff
@@ -0,0 +1,34 @@
+# Exports the entire contents of database tables to Unicode CSV files stored in GCS.
+# Used instead of Cloud SQL export because it handles newlines and null characters properly.
+#
+# Usage: ./run_client.sh --project <PROJECT> --account <ACCOUNT> \
+# --service_account exporter@<PROJECT>.iam.gserviceaccount.com export_tables.py \
+# --database rdr --tables code,participant --directory test_directory
+# If "rdr" is chosen for the database, the data will be written to <ENVIRONMENT>-rdr-export;
+# If "cdm" or "voc" are chosen, the data will be written to <ENVIRONMENT>-cdm.
```
jmandel: Please document the meaning of the `directory` parameter (i.e. a folder within the export bucket).

rest-api/offline/main.py
```diff
@@ -78,6 +81,20 @@ def import_biobank_samples():
...
 
+@api_util.auth_required(EXPORTER)
+def export_tables():
+  resource = request.get_data()
+  resource_json = json.loads(resource)
+  database = resource_json.get('database')
+  tables = resource_json.get('tables')
+  if not database:
+    raise BadRequest("database is required")
+  if not tables or not type(tables) is list:
```
jmandel: `not type(tables)` always resolves to `true`, because all types are truthy :)

Need `type(tables) is not list`.

rest-api/offline/table_exporter.py
```diff
@@ -0,0 +1,33 @@
...
+
+  @staticmethod
+  def export_tables(database, tables, directory):
+    app_id = app_identity.get_application_id()
+    # Determine what GCS bucket to write to based on the environment and database.
+    if app_id == "None":
+      bucket_name = app_identity.get_default_gcs_bucket_name()
+    elif database == 'rdr':
+      bucket_name = '%s-rdr-export' % app_id
+    elif database == 'cdm' or database == 'voc':
```
jmandel: Just a style note: `database in ['cdm', 'voc']`

rest-api/offline/table_exporter.py
```diff
@@ -0,0 +1,33 @@
...
+  def _export_csv(cls, bucket_name, database, directory, table_name):
+    SqlExporter(bucket_name, use_unicode=True).run_export('%s/%s.csv' % (directory, table_name),
+                                                          'SELECT * FROM %s.%s' %
+                                                          (database, table_name))
+
+  @staticmethod
+  def export_tables(database, tables, directory):
+    app_id = app_identity.get_application_id()
+    # Determine what GCS bucket to write to based on the environment and database.
+    if app_id == "None":
```
jmandel: Should have single quotes here (sorry, pedantic).

rest-api/offline/table_exporter.py
```diff
@@ -0,0 +1,33 @@
...
+  """API that exports data from our database to UTF-8 CSV files in GCS.
+
+  Used instead of Cloud SQL export because it handles newlines and null characters in a way that
+  other CSV clients (e.g. BigQuery, Google Sheets) can actually understand.
+  """
+
+  @classmethod
+  def _export_csv(cls, bucket_name, database, directory, table_name):
+    SqlExporter(bucket_name, use_unicode=True).run_export('%s/%s.csv' % (directory, table_name),
+                                                          'SELECT * FROM %s.%s' %
```
jmandel: This is a vector for injections -- obviously only from trusted clients, but this should be sanitized. Maybe an assert listing the allowed values for `database` and `table_name`?

rest-api/unicode_csv.py
```diff
@@ -0,0 +1,67 @@
+import csv
+import codecs
+import cStringIO
+
+# Writes CSV files with Unicode encoding. (Standard CSV libraries don't handle Unicode.)
+# Adapted from https://docs.python.org/2/library/csv.html
```
jmandel: I've used https://pypi.python.org/pypi/unicodecsv/0.9.0 in the past, for this. The functionality below is simple enough, so I don't feel strongly... but see note below.

rest-api/unicode_csv.py
```diff
@@ -0,0 +1,67 @@
...
+  which is encoded in the given encoding.
+  """
+
+  def __init__(self, f, dialect=csv.excel, encoding="utf-8", **kwds):
+    f = UTF8Recoder(f, encoding)
+    self.reader = csv.reader(f, dialect=dialect, **kwds)
+
+  def next(self):
+    row = self.reader.next()
+    return [unicode(s, "utf-8") for s in row]
```
jmandel: e.g. [these edge cases](https://github.com/jdunck/python-unicodecsv/blob/de3b59bb18032aa75681dc4068b92b4023e9d56d/unicodecsv/py2.py#L122-L1255) for float handling aren't reflected here; I don't know if they matter.

rest-api/etl/export_etl_results.sh
danqrodney: No -- you can run this locally without specifying project or account, and it will hit the local server (and dump files to the storage dir on disk.)

I updated the usage to indicate this.

rdr_client/export_tables.py
danqrodney: Done.

rest-api/offline/main.py
danqrodney: Oof. Done. :)

rest-api/offline/table_exporter.py
danqrodney: RDR and table_name are already validated below.

I added logic ensuring table_name matches a regex. I'd rather not hardcode the possible tables here, since that will be a maintenance issue over time.

rest-api/offline/table_exporter.py
danqrodney: Done.

rest-api/offline/table_exporter.py
danqrodney: Done.

rest-api/unicode_csv.py
danqrodney: Ah, yeah, to be honest I'm not using UnicodeReader at this stage, and we may never use it. So I'll remove this for now, since as you point out, it doesn't handle non-strings properly.

rest-api/unicode_csv.py
danqrodney: OK. The license there is a little funny -- not a standard license of a type I'm familiar with. I removed UnicodeReader, since I'm not sure we're going to need it; I think UnicodeWriter will work ok.

danqrodney:
Per Karthik's comments on #curation, I have removed the export of vocabulary tables... we're just supposed to export the CDM tables.

rest-api/offline.yaml
```diff
@@ -28,7 +28,6 @@ handlers:
   login: admin
 - url: /offline/.*
   script: offline.main.app
-  login: admin
```
jmandel: Just had to think about this for a bit -- as long as we keep `admin` as a requirement for the `_ah/...` routes, I think we're okay.

rest-api/unicode_csv.py
jmandel: The [license](https://github.com/jdunck/python-unicodecsv/blob/master/LICENSE) is a standard [BSD 2-Clause](https://opensource.org/licenses/BSD-2-Clause) -- but I don't mind where we've ended up here.

rest-api/offline/table_exporter.py
jmandel: The regex is good enough -- but would rather bake it into this `_export_csv` function directly and apply to the database + table name, since it's so general. This ensures that nobody will accidentally call this function with unsanitized attributes in the future. 

The additional check on `database` below is also good :)

danqrodney:
Added check to _export_csv.

Thanks for the review!

PR #504, opened 2017-10-23
[DA-399] Prefer PMI system when parsing physical measurement codes.

danqrodney:


rest-api/dao/physical_measurements_dao.py
```diff
@@ -328,6 +329,19 @@ def make_measurement_id(physical_measurements_id, measurement_count):
     return (physical_measurements_id * 1000) + measurement_count
 
   @staticmethod
+  def get_preferred_coding(code):
```
jmandel: Would call the arg `codeable_concept` (or `concept`), as in `get_preferred_coding(concept)` -- just to avoid having to say "a code has codings which have codes". (I know that some fields in FHIR are named `code`, like `Observation.code` -- but the *type* of this field is `CodeableConcept`.)

rest-api/test/unit_test/api_test/physical_measurements_api_test.py
```diff
@@ -55,15 +55,15 @@ def test_insert(self):
...
+                      codeValue="systolic-diastolic-blood-pressure-1",
                       measurementTime=now,
                       bodySiteCodeSystem="http://snomed.info/sct",
                       bodySiteCodeValue="368209003")
     bp1 = Measurement(measurementId=pm_id * 1000 + 1,
                       physicalMeasurementsId=pm_id,
-                      codeSystem="http://loinc.org",
-                      codeValue="8480-6",
+                      codeSystem="http://terminology.pmi-ops.org/CodeSystem/physical-measurements",
+                      codeValue="systolic-blood-pressure-1",
```
jmandel: Do we not currently have a test that parses PMs from JSON? It would be good to show the "pick preferred" logic in action.

rest-api/dao/physical_measurements_dao.py
danqrodney: Done.

rest-api/test/unit_test/api_test/physical_measurements_api_test.py
danqrodney: Yes, we do -- this test in particular, does. See measurements-as-fhir.json above.

jmandel: Yes, so we do. (Sorry, that was... pretty clear from the change history.)

PR #505, opened 2017-10-23
[DA-386] Adding new columns to participant_summary, finalized to physical_measurements.

danqrodney:
(I'm working on updating these fields and adding tests.)

Backfill will come in a follow-up PR.


rest-api/participant_enums.py
```diff
@@ -37,6 +37,11 @@ class SampleStatus(messages.Enum):
   UNSET = 0
   RECEIVED = 1
 
+class OrderStatus(messages.Enum):
+  """Status of biobank orders and samples"""
+  UNSET = 0
+  ORDERED = 1
```
jmandel: Consider expanding this to include `UNSET`, `CREATED`, `COLLECTED`, `PROCESSED`, `FINALIZED` -- note that the "realtime" nature of these reports may be a bit misleading (let's discussion on Slack).

danqrodney: Done.

danqrodney:
OK, fixed up the logic, added tests. This should be good to go.

README.md
```diff
@@ -585,3 +614,23 @@ resources. Once these are created, a client can query for available samples:
 
 `POST /ImportCodebook` Imports the latest published Codebook (metadata about
 Questionnaire contents).
+
+## Export API
```
jmandel: Thanks for getting this in here :-)

rest-api/api_util.py
```diff
@@ -241,6 +241,14 @@ def format_json_hpo(obj, hpo_dao, field_name):
   else:
     obj[field_name] = UNSET
 
+def format_json_site(obj, site_dao, field_name):
+  site_id = obj.get(field_name + 'Id')
+  if site_id:
```
jmandel: Safer to say `if field_name + 'Id' in obj` -- otherwise we could get tripped up by present-but-falsey values, like `0`. (Now, our site IDs are auto-increment, and MySql starts these sequences with `1` by default, so there's no bug; but the logic isn't as self-evidently correct as it could be.)

danqrodney: Changed to "site_id is not None".

PR #506, opened 2017-10-24
Altering measurement spec to reflect new approach to PMs

danqrodney:


rest-api/dao/physical_measurements_dao.py
```diff
@@ -219,8 +219,6 @@ def set_measurement_ids(physical_measurements):
 
   def insert_with_session(self, session, obj):
     is_amendment = False
-    if obj.logPosition is not None:
```
danqrodney: Note: we had to do this because obj gets its logPosition field set in here, and if the insert fails due to a PK violation, this method will get called again... it shouldn't fail at that point.

danqrodney:
Added the logic to the fake data generator to use the updated PM spec.

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -295,38 +343,150 @@ def _make_measurement_resource(self, measurement, qualifier_set):
...
+          url = related['target']['reference']
+          if url not in related_urls:
+            related_list.append(related)
+            related_urls.add(url)
+    return related_list
+
+  def _mean(self, values):
+    return round(sum(values) / len(values), 1)
+
+  def _find_closest_two_mean(self, measurement_resources):
```
jmandel: Phew! Not proposing this because it ain't readable, but just found myself compelled to carry out the exercise. (I promise it shines when the # of repetitions is unbounded ;-))

```py
import itertools
def diff(xy): return abs(xy[0]-xy[1])

sorted_pairs = sorted(itertools.combinations(measurement_values, 2), key=diff)
grouped_pairs = itertools.groupby(sorted_pairs, diff)
best_group = next(grouped_pairs)[1]
closest_values = set(itertools.chain(*best_group))
self._mean(closest_values)
```

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -295,38 +343,150 @@ def _make_measurement_resource(self, measurement, qualifier_set):
...
+
+  def _calculate_last_two_mean(self, measurement_resources):
+    measurement_values = self._get_measurement_values(measurement_resources)
+    return self._mean(measurement_values[1:])
+
+  def _find_components(self, measurement_resources, submeasurement):
+    components = []
+    for resource in measurement_resources:
+      subcomponents = resource.get('component')
+      if subcomponents:
```
jmandel: Same comment about defaulting to `[]` in the `get`.

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -295,38 +343,150 @@ def _make_measurement_resource(self, measurement, qualifier_set):
...
+      if delta_0_2 < delta_1_2:
+        return self._mean([measurement_values[0], measurement_values[2]])
+      else:
+        return self._mean(measurement_values[1:])
+
+  def _calculate_last_two_mean(self, measurement_resources):
+    measurement_values = self._get_measurement_values(measurement_resources)
+    return self._mean(measurement_values[1:])
+
+  def _find_components(self, measurement_resources, submeasurement):
```
jmandel: Might be clearer to call this `_find_components_by_code`, and pass in just the `code` instead of a whole submeasurement. (Or better, `_find_components_by_coding`, to match systems too.)

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -295,38 +343,150 @@ def _make_measurement_resource(self, measurement, qualifier_set):
...
+    components = []
+    for resource in measurement_resources:
+      subcomponents = resource.get('component')
+      if subcomponents:
+        for subcomponent in subcomponents:
+          if subcomponent['code']['coding'][0]['code'] == submeasurement['code']['code']:
```
jmandel: I guess it's okay to look at just the 1st coding here because we construct them in a specific order. (But this approach forces a reader to wonder / verify.)

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -295,38 +343,150 @@ def _make_measurement_resource(self, measurement, qualifier_set):
...
       components = []
       for submeasurement in measurement['submeasurements']:
-        if random.random() <= _PHYSICAL_MEASUREMENT_ABSENT:
-          continue
-        components.append(self._make_measurement_resource(submeasurement, qualifier_set))
+        previous_component = None
+        # If there was a previous entry, find the component matching the code for this
+        # submeasurement.
+        if previous_resource:
+          for component in previous_resource['component']:
```
jmandel: Can you replace this with `self._find_components([previous_resource], ....` ?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -21,7 +21,7 @@
 _COMPOSITION_RESOURCE_TYPE = 'Composition'
 _CREATED_LOC_EXTENSION = 'http://terminology.pmi-ops.org/StructureDefinition/authored-location'
 _FINALIZED_LOC_EXTENSION = 'http://terminology.pmi-ops.org/StructureDefinition/finalized-location'
-_PM_SYSTEM = 'http://terminology.pmi-ops.org/CodeSystem/physical-measurements'
```
jt2k: In the changes I am working on right now, we are switching to `http://terminology.pmi-ops.org/CodeSystem/physical-measurements` as the code system.

danqrodney: OK. Are you doing that across the board? (We already have data in the system using physical-evaluation.)

PR #509, opened 2017-10-27
Updating biobank orders, samples, physical measurements to update participant

danqrodney:
summary fields on each update, instead of just the first one.


rest-api/dao/physical_measurements_dao.py
```diff
@@ -340,9 +339,9 @@ def get_preferred_coding(codeable_concept):
     for coding in codeable_concept.coding:
       if pm_coding is None:
         pm_coding = coding
-      elif coding.system == _PM_SYSTEM:
-        if pm_coding.system == _PM_SYSTEM:
-          raise BadRequest('Multiple measurement codes with system %s' % _PM_SYSTEM)
+      elif coding.system.startswith(_PM_SYSTEM_PREFIX):
```
jmandel: This is so we accept both `physical-measurements` and `physical-evaluation`? If so... would be good to note this and add a TODO to lock down to the new value after the transition period is complete.

rest-api/tools/setup_env.sh
```diff
@@ -11,7 +11,7 @@ find . | grep \.pyc | xargs rm -if $*
 
 echo "Installing libs..."
 # If this fails due to missing mysql_config, try `sudo apt-get install libmysqlclient-dev`.
-pip install -r requirements.txt -t lib/
+pip install --system -r requirements.txt -t lib/
```
jmandel: Did the default (no `--system` or `--user`) stop working here?

rest-api/dao/physical_measurements_dao.py
danqrodney: This is intended to switch over to physical-measurements.

PMs using the existing system won't handle this; but we're planning on re-sending PMs for everybody, so I'm not sure it matters?

rest-api/tools/setup_env.sh
danqrodney: Yeah, it didn't work for me when I tried running it (with or without --user). Not sure why.

rest-api/dao/physical_measurements_dao.py
jmandel: This would just be about tightening our validation rules in the future, after the switchover. Your call!

danqrodney: Ah. You mean explicitly reject something that uses physical-evaluation? Yeah, we could do that. Will consider for a follow-up.

PR #512, opened 2017-10-31
[DA-354] [DA-312] Adding is_native_american, notes to reconciliation report.

danqrodney:


rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -235,24 +239,20 @@ def _query_and_write_reports(exporter, now, path_received, path_late, path_missi
        database_factory.get_database().session() as session:
     writer = CompositeSqlExportWriter([received_writer, late_writer, missing_writer])
     exporter.run_export_with_session(writer, session, replace_isodate(_RECONCILIATION_REPORT_SQL),
-                                     {"biobank_id_prefix": get_biobank_id_prefix(),
-                                      "pmi_ops_system": _PMI_OPS_SYSTEM,
-                                      "kit_id_system": _KIT_ID_SYSTEM,
-                                      "tracking_number_system": _TRACKING_NUMBER_SYSTEM})
+                                     {'race_question_code_id': race_question_code.codeId,
+                                      'native_american_race_code_id':
```
calbach: opt_nit: Doesn't this include "alaskan natives" as well (which I gather is distinct from native american)? Should this keep the `aian` naming here and throughout?

rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -235,24 +239,20 @@ def _query_and_write_reports(exporter, now, path_received, path_late, path_missi
...
-  for code_value in config.getSettingList(config.NATIVE_AMERICAN_RACE_CODES):
-    code = code_dao.get_code(PPI_SYSTEM, code_value)
-    race_code_ids.append(str(code.codeId))
-  race_codes_sql, params = get_sql_and_params_for_array(race_code_ids, 'race')
-  withdrawal_sql = _WITHDRAWAL_REPORT_SQL.format(race_codes_sql)
-  params['race_question_code_id'] = race_question_code.codeId
-  params['seven_days_ago'] = now - datetime.timedelta(days=7)
-  params['biobank_id_prefix'] = get_biobank_id_prefix()
-  exporter.run_export(path_withdrawals, replace_isodate(withdrawal_sql), params)
+  params = {'race_question_code_id': race_question_code.codeId,
```
calbach: opt_nit: inline this dict as above?

rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -306,6 +306,18 @@ def _get_hpo_type_sql(hpo_alias):
...
   return result
 
+_NATIVE_AMERICAN_SQL = """
+  (SELECT (CASE WHEN count(*) > 0 THEN 'Y' ELSE 'N' END)
+       FROM questionnaire_response qr
+       INNER JOIN questionnaire_response_answer qra
+         ON qra.questionnaire_response_id = qr.questionnaire_response_id
+       INNER JOIN questionnaire_question qq
+         ON qra.question_id = qq.questionnaire_question_id
+      WHERE qr.participant_id = participant.participant_id
```
calbach: Where did "participant" come from here - is this assuming something in the outer scope? If so, should this string be templated instead to make the dependency clear?

rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -367,7 +383,10 @@ def _get_hpo_type_sql(hpo_alias):
       biobank_stored_sample.confirmed,
       biobank_stored_sample.created,
       kit_id_identifier.value biospecimen_kit_id,
-      tracking_number_identifier.value fedex_tracking_number
+      tracking_number_identifier.value fedex_tracking_number, """ + _NATIVE_AMERICAN_SQL + """,
+      biobank_order.collected_note notes_collected,
```
calbach: Are these "notes_*" things related to the change?

rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -340,7 +352,11 @@ def _get_hpo_type_sql(hpo_alias):
     ISODATE[MAX(created)] 'Sample Family Create Date',
     TIMESTAMPDIFF(HOUR, MAX(collected), MAX(created)) elapsed_hours,
     GROUP_CONCAT(DISTINCT biospecimen_kit_id) biospecimen_kit_id,
-    GROUP_CONCAT(DISTINCT fedex_tracking_number) fedex_tracking_number
+    GROUP_CONCAT(DISTINCT fedex_tracking_number) fedex_tracking_number,
+    GROUP_CONCAT(DISTINCT is_native_american) is_native_american,
```
calbach: Not sure I understand the `DISTINCT` usage here; aren't there only two values `Y`, `N`?

rest-api/offline/biobank_samples_pipeline.py
calbach: Nevermind, reread the PR description.

rest-api/offline/biobank_samples_pipeline.py
danqrodney: Added comment to [DA-354] inquiring about this.

rest-api/offline/biobank_samples_pipeline.py
danqrodney: Done.

rest-api/offline/biobank_samples_pipeline.py
danqrodney: It does assume that you have "participant" as a table name in a join elsewhere. I'm not sure that using a template would make this clearer, though. I'll add a comment.

rest-api/offline/biobank_samples_pipeline.py
danqrodney: The current report returns a single row for a given (participant, biobank sample test) pair; there may be multiple biobank sample orders or biobank stored samples for that combination. So you might get multiple 'Y' or 'N' values in the inner query; DISTINCT makes sure we only return one value.

We are going to change this soon so we're matching individual order samples with individual stored samples (see [DA-418]), and we won't need this GROUP_CONCAT business anymore. But not in this PR.

PR #515, opened 2017-11-02
Split out GAE deps from standard utils; client tools should not require GAE

calbach:
This work is primarily to make local RDR development compatible with virtualenv. Previously, when adding the GAE path into `PYTHONPATH` before running setup commands, this would cause issues with resource loading from a virtualenv:

```
File "/usr/local/google/home/calbach/.virtualenvs/pmi-rdr/bin/alembic", line 5, in <module>
from pkg_resources import load_entry_point
File "/usr/local/google/home/calbach/google-cloud-sdk/platform/google_appengine/lib/setuptools-0.6c11/pkg_resources.py", line 2603, in <module>
working_set.require(_requires_)
File "/usr/local/google/home/calbach/google-cloud-sdk/platform/google_appengine/lib/setuptools-0.6c11/pkg_resources.py", line 666, in require
needed = self.resolve(parse_requirements(requirements))
File "/usr/local/google/home/calbach/google-cloud-sdk/platform/google_appengine/lib/setuptools-0.6c11/pkg_resources.py", line 565, in resolve
raise DistributionNotFound(req) # XXX put more info here
pkg_resources.DistributionNotFound: python-dateutil
```

- moved GAE stuff from `rest-api/api_util` -> `rest-api/app_util`
- moved GAE stuff `rest-api/model/utils.py` -> `rest-api/model/config_utils.py`
- lazy load the `config` module within DAO code, as parts of this library access Datastore
- drop the GAE path hacking which was previously used to load the above GAE code outside of app engine - this was interacting badly with virtualenv

My understanding is that even with the previous GAE path hacking, the datastore accessor code most likely wouldn't have worked properly or connected to anything outside the context of a GAE app (this could be wrong).

Possible future cleanups:
- Refactor direct DB access in command-line tools to access via an API.
- Even with the above change, there would likely still be cases where we'd want command-line tools that reference the DAO code. Instead of the current static access to the `config` package, pass a datastore config object / service as a dependency to the DAO classes that need it, then populate this with a clear stub or stand-in value outside of GAE.


rest-api/tools/set_path.sh
```diff
@@ -1,33 +1,7 @@
...
-    GAEPATH=${module}:${GAEPATH}
-  fi
-done
-unset module
-
 export BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"
 ROOT_REPO_DIR="$( cd "${BASE_DIR}" && cd .. && pwd )"
 # RDR libs should appear first in PYTHONPATH so we can override versions from
 # the GAE SDK. (Specifically, we need oauth2client >= 4.0.0 and GAE uses 1.x.)
-export PYTHONPATH=$PYTHONPATH:${BASE_DIR}:${BASE_DIR}/lib:${ROOT_REPO_DIR}/rdr_client:${ROOT_REPO_DIR}/rdr_common:${GAEPATH}
```
danqrodney: So we no longer have any dependencies on the Google Cloud SDK in our tools?

(We may need to run some of the tools interacting with Cloud environments to verify this...)

calbach:
@danqrodney well, I thought this was do'able, but it turns out there were a few more nested scripts I didn't account for which may have intractable dependencies on GAE, namely `import_hpos.py`, `import_sites.py`. I'm not sure if I had some pip environment in which these installed correctly, or if this just never worked properly.

Ultimately, these scripts depend on the DAO code, which depends on the datastore `ndb` package, so I'm not sure there's anyway to avoid that dependency. As a side-note, I'm surprised that we can run a command-line tool which talks to datastore via the app-engine SDK outside the context of an app engine app.

danqrodney:
It looks like from here:

https://circleci.com/gh/all-of-us/raw-data-repository/1867?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

That we've still got a dependency on NDB stuff from google.appengine.ext. If you run tools/setup_local_database.sh or tools/import_hpos.sh you should probably be able to reproduce it. 

I'm not clear on how to avoid this dependency without using the libraries in the Google Cloud SDK. There's something about that may be relevant here:

https://github.com/GoogleCloudPlatform/datastore-ndb-python

danqrodney:
It's possible that pip install google-cloud would pick this up.

Note that if you add dependencies like that in tools/setup_env.sh you may also need to add them to circle.yml at the top:

https://github.com/all-of-us/raw-data-repository/blob/master/circle.yml#L10

danqrodney:
You may be able to use one of:

https://pypi.python.org/pypi/appengine
https://pypi.python.org/pypi/google-appengine
https://pypi.python.org/pypi/appengine-sdk

None of them seem very official. :(

danqrodney:
It seems that we followed the recommendations here:

https://stackoverflow.com/questions/25525258/add-appengine-sdk-to-python-path

for how to include the SDK in your PYTHONPATH. I don't know if there is a better way.

Hopefully you can find a way to make that work even inside venv?

danqrodney:
So I'm pretty sure we *aren't* actually using any NDB stuff in command line tools. So if this is the only thing, more refactoring should allow us to avoid dragging in this import.

There may be other issues, I don't know.

calbach:
OK - not pretty, but I think this at least covers the standard tooling.

markfickett:
An approach we've taken, to cleanly break dependencies from tools, is to
write them a simple API (if it doesn't exist) and then move the actual tool
to rdr_client. See, for example,
https://github.com/all-of-us/raw-data-repository/blob/master/rdr_client/import_codebook.py
and
https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/api/import_codebook_api.py
.

calbach:
Updated the PR description with more details. I'm not familiar with all of the tools yet but I imagine there will   be cases where it makes sense for command-line tools to directly interact with SQL, so I think it's probably worthwhile to separate the DAO package from the Datastore package. That said, agreed with @markfickett that client access via API should be preferred where possible.

rest-api/tools/set_path.sh
calbach: I've basically only validated that tools called as a result of setup_local_database.sh work (and can resolve imports). Are there any other specific ones I should test out?

PR #516, opened 2017-11-03
Create release tracking ticket in PD project.

markfickett:
We used to use it in DA and then move it, but now we only ever use the release ticket in the PD project, so I think we may as well create it there.


calbach:
I'll leave it to @danqrodney to review the semantics of the change, since I don't really understand what `DA` or `PD` stand for (if anything), or where the scope of those components are defined in JIRA.

danqrodney:
DA stands for "DRC API" (RDR, essentially.)

I don't know what PD stands for, but it's process-like stuff... where get approval for doing things in prod, including pushes.

PR #517, opened 2017-11-03
Changing measurement specs to match latest physical measurements code…

danqrodney:
…book

See: 

https://docs.google.com/spreadsheets/d/1KMPxUMjlQLNK7dO1QMVxDTLJ2N5SksLq2Qdp9MCSOMw/edit#gid=1076265690


calbach:
Was this autogenerated? If so, what command-line produced this change (for my education)?

danqrodney:
So the original spec was generated, by this code which takes a sample of physical measurements actually in use in prod and calculates their range of values:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/tools/validate_or_backfill_measurements.py

However, since then we've updated to a new way of representing physical measurements, that isn't yet reflected in prod (but soon will be.) (Note that the above script won't quite work right with those changes, now that clients are actually going to send PMI-specific and LOINC codes for measurements... at some point we'll need to update the logic it's using.)

This PR is updating our spec to match how the data is going to look with the new measurements, so we can generate realisting synthetic data (for use in generating a CDR.)

danqrodney:
I added one more change... our fake participant generator henceforth will use @fakeexample.com email addresses, instead of @example.com email addresses. (The latter are treated as "test" participants, which are entered in prod by HPOs sometimes, and are filtered out of some views, CDR generation, etc.)

Since the fake data generator only operates in non-prod environments, and we want to simulate non-test participants, we're switching domains.

PR #518, opened 2017-11-03
[DA-423] Set limits on Metrics endpoint

shyamnaru:


rest-api/api/metrics_api.py
```diff
@@ -31,7 +32,12 @@ def post(self):
         try:
           end_date = datetime.datetime.strptime(end_date_str, DATE_FORMAT).date()
         except ValueError:
-          raise BadRequest("Invalid start date: %s" % end_date_str)
+          raise BadRequest("Invalid end date: %s" % end_date_str)
+      if start_date and end_date:
```
danqrodney: I think we need to start requiring start_date and end_date to be set as well.

The underlying API doesn't require them, but if you don't specify them it can hose our server.

(We should confirm that HPRO never queries for metrics without specifying both start and end date.)

shyamnaru: In line 20 & 21 we are setting a default value None for both start_date and end_date variables so I think there is no need to check if the variable is set or not.

danqrodney: My point is if you pass None into the DAO before, it won't set an upper or lower bound on the metrics we're fetching... which can lead to more than 7 days of results, and memory exploding. We shouldn't allow that.

shyamnaru: I think setting `buckets = None` by default and placing `buckets = dao.get_active_buckets(start_date, end_date)` in side `if start_date and end_date:` condition should solve this issue.

danqrodney: It doesn't make sense to return None.

If start_date and end_date are required, we should throw a BadRequest / return 400 if you don't specify them... like you are doing here if they are too far apart. Does that make sense?

rest-api/api/metrics_api.py
```diff
@@ -17,28 +17,27 @@ class MetricsApi(Resource):
...
+      try:
+        end_date = datetime.datetime.strptime(end_date_str, DATE_FORMAT).date()
+      except ValueError:
+        raise BadRequest("Invalid end date: %s" % end_date_str)
+      date_diff = abs((end_date - start_date).days)
+      if date_diff > DAYS_LIMIT:
+        raise BadRequest("Difference between start date and end date "\
+          "should not be greater than %s days" % DAYS_LIMIT)
+      buckets = dao.get_active_buckets(start_date, end_date)
+      if buckets is None:
```
danqrodney: You don't really need this if I think.

danqrodney:
Looks like you'll need to update metrics_api_test.

shyamnaru:
Yes, I'm looking at it.

markfickett:
P.S. Since I'm barging in on your PR:

Our general approach to PR approval is that once you've gotten a review and approval from at least one other owner, you're good to squash and merge your own PR. ("Approve" essentially means "I'm confident that we have discussed anything else you might want to change, so merge whenever you're ready.") So, regarding my comment on `assertRaises` which is coming in after Dan's approval, feel free to use it or not, but still go ahead and merge when you're happy with the state of your PR (specifcally no need for an approve from me).

PR #528, opened 2017-11-14
Whitelisting HPRO app IDs, Vibrent IPs for use in RDR in our non-prod

danqrodney:
environments.

(Once we verify this works, we can configure prod as well.)

### DIFF
--- a/rest-api/config/config_stable.json
+++ b/rest-api/config/config_stable.json
@@ -6,7 +6,7 @@
   ],
   "user_info": {
     "pmi-hpo-test@appspot.gserviceaccount.com": {
-      "roles": ["ptc", "healthpro"],
+      "roles": ["healthpro"],
       "whitelisted_appids": ["pmi-hpo-test"]
     },
     "vibrent-drc-stable@vibrent-drc-stable-165017.iam.gserviceaccount.com": {
--- a/rest-api/config/config_stable.json
+++ b/rest-api/config/config_stable.json
@@ -6,10 +6,14 @@
   ],
   "user_info": {
     "pmi-hpo-test@appspot.gserviceaccount.com": {
-      "roles": ["ptc", "healthpro"]
+      "roles": ["ptc", "healthpro"],
+      "whitelisted_appids": ["pmi-hpo-test"]
     },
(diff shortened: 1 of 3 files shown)


rest-api/config/config_test.json
```diff
@@ -7,10 +7,17 @@
...
     "pmi-hpo-dev@appspot.gserviceaccount.com": {
       "roles": ["healthpro"],
-      "comment": "for HealthPro testing"
+      "comment": "for HealthPro testing",
+      "whitelisted_appids": ["pmi-hpo-dev"],
     },
     "vibrent-drc-qa@vibrent-drc-qa-165017.iam.gserviceaccount.com": {
       "roles": ["ptc"]
+      "whitelisted_ip_ranges": {
+        "ip4": ["34.206.2.13",
```
calbach: For my understanding: what do these IP addresses correspond to? Are they stable? How would one regenerate this list?

danqrodney:
Reverted changes to test, as we want to be able to connect to it from local dev machines.

rest-api/config/config_stable.json
```diff
@@ -6,10 +6,14 @@
   ],
   "user_info": {
     "pmi-hpo-test@appspot.gserviceaccount.com": {
-      "roles": ["ptc", "healthpro"]
+      "roles": ["ptc", "healthpro"],
```
jt2k: While we're updating this, I don't think there's a reason for this user to have the `ptc` role.

rest-api/config/config_test.json
danqrodney: They are IP addresses Vibrent sent us. They are supposed to be stable. If they change, Vibrent would have to tell us.

rest-api/config/config_stable.json
danqrodney: Removed.

(In the distant past, HPRO could create participants via a form and needed to act like PTC, which I'm guessing is why this was here. I presume we don't want to support such shenanigans anymore. :)

danqrodney:
Added healthpro-beta to staging app ID whitelist.

PR #529, opened 2017-11-15
[DA-425] Fix mapping of "none of these" race values; metrics config for "prefer not to answer".

danqrodney:
Removing unused "UNMAPPED" and "SKIPPED" values from metrics config.


danqrodney:
Note: we're going to have to do backfill on participant_summary to populate "none of these" values for the race field after this goes live.

rest-api/alembic/versions/b315fec9aa4e_adding_pm_view.py
```diff
@@ -26,13 +26,13 @@
    p.withdrawal_time participant_withdrawal_time,
    p.suspension_status participant_suspension_status,
    p.suspension_time participant_suspension_time,
-   pm.physical_measurements_id physical_measurements_id,   
+   pm.physical_measurements_id physical_measurements_id,
```
calbach: These look like they are just whitespace changes, but should mutating these alembic migration files be generally avoided as a rule after initial submission?

rest-api/test/unit_test/offline_test/metrics_export_test.py
```diff
@@ -399,15 +404,15 @@ def test_metric_export(self):
...
+        'Participant.race.OTHER_RACE': 1,
         'Participant.biospecimenSummary.SAMPLES_ARRIVED': 2,
         'Participant.consentForStudyEnrollmentAndEHR.UNSET': 2,
         'Participant.numCompletedBaselinePPIModules.1': 2,
         'Participant.samplesToIsolateDNA.UNSET': 1,
         'Participant.samplesToIsolateDNA.RECEIVED': 1,
         'Participant.enrollmentStatus.INTERESTED': 2
     })
-    # At TIME_3, P1 is UNMAPPED race, female gender, and now in PITT HPO;
+    # At TIME_3, P1 is UNSET race, UNMAPPED female gender, and now in PITT HPO;
```
calbach: What does UNMAPPED female gender mean? Is this a typo?

rest-api/participant_enums.py
```diff
@@ -49,8 +50,8 @@ class OrderStatus(messages.Enum):
 # in questionnaire responses.
 class Race(messages.Enum):
```
calbach: To check my understanding, we would need to backfill because these enum conversions are happening as PTC calls into the RDR API with questionnaire responses?

rest-api/alembic/versions/b315fec9aa4e_adding_pm_view.py
danqrodney: Modifying their contents in a significant way should be avoided, yes. Modifying whitespace is no big deal.

(The remove_trailing_whitespace script is what did this, BTW.)

rest-api/participant_enums.py
danqrodney: That is correct.

rest-api/test/unit_test/offline_test/metrics_export_test.py
danqrodney: Here's the deal: we have codes stored in the code table, representing codes for questions and answers in questionnaires and questionnaire responses. Codes that are imported from our official codebook are considered "mapped" -- have their mapped flag set to true. However, there's the possibility that due to some mistake we  could get a questionnaire response from PTC that contains a code for an answer that isn't in the codebook. Rather than rejecting this response, we accept it and add a code to the table where mapped = false, and log the fact that this happened. (The idea being that we could then add the code in question to the codebook or fix the issue in PTC... and that when it comes to data submitted by participants, it's better to accept it with flaws rather than rejecting it, since we don't normally have a way of getting it again if it's rejected.)

When generating metrics on fields with code values, we normally use the value of the code itself. However, for unmapped codes, we use the string "UNMAPPED", to call out the fact that the code in question hasn't been mapped yet.

In practice, this doesn't happen, because PTC never sends us codes that aren't in the codebook, but we have the code to support it in case it does.

Previously, we had an UNMAPPED value as one of the possible values returned in our metric fields API to cover this possible case. However, this winds up showing up in dashboards as an always-zero row, which isn't useful. And arguably dashboards aren't the right way for us to notice there are UNMAPPED values in the system anyway.

So with this PR, we would continue to theoretically have UNMAPPED values show up in metrics buckets, which could help us with debugging unmapped values if they ever occurred in future; but UNMAPPED won't show up in dashboards anymore, since they aren't one of the possible values returned from the metric fields API.

I dropped the UNMAPPED value from the race enum entirely, since that is an enum we expose and map code values to, and it didn't seem helpful to keep it around anymore... I could have instead retained the value and just omitted it from the possible values returned for race in the metric fields API. I'm not sure it matters too much either way.

Sorry this is sorta confusing. Let me know if you have suggestions on how to make it better. :)

calbach: Thanks for the explanation. I probably need a better understanding of the system before I could suggest any improvements. It sounds like everything is reasonable except perhaps the actual place where we're noticing issues - ideally it seems like that would be an alert coming from the RDR itself, e.g. incrementing a metric or some periodic batch job which looks for consistency issues and raises an alert. I don't know if these are easy things to setup or whether there's existing RDR infra to do this, so if not, the current approach seems reasonable.

PR #531, opened 2017-11-15
Whitelisting HealthPro and PTC in production.

danqrodney:


danqrodney:
OK. Applying this change in prod now. Scream if anything breaks.

PR #532, opened 2017-11-16
[DA-452] 500s for sending questionnaire responses to bad participant ID

shyamnaru:


rest-api/dao/participant_dao.py
```diff
@@ -165,6 +165,14 @@ def validate_participant_id(self, session, participant_id):
     raise_if_withdrawn(participant)
     return participant
 
+  def check_participant_exist(self, participant_id):
```
danqrodney: I don't think this quite the way we want to do things... this will do an additional database request in a different session (which doesn't guarantee technically that the participant still exists by the time we get to the insert code.)

We already have validate_participant_reference, which is currently being called during _update_participant_summary here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/questionnaire_response_dao.py#L169

That is unfortunately *after* we reference participant.participantSummary here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/questionnaire_response_dao.py#L154

Which is the source of the error.

So what we probably want to do is just have a check above that line that participant (which we fetched for update) is not None, and throw a BadRequestException if it is. (We've been returning NotFound specifically only when trying to retrieve some resource that doesn't exist, rather than returning it for creating a resource on a participant that doesn't exist.) We can then remove the call to validate_participant_reference in the update code I believe.

rest-api/api/questionnaire_response_api.py
```diff
@@ -15,4 +15,4 @@ def get(self, p_id, id_):
 
   @app_util.auth_required(PTC)
   def post(self, p_id):
-    return super(QuestionnaireResponseApi, self).post(participant_id=p_id)
+  	return super(QuestionnaireResponseApi, self).post(participant_id=p_id)
```
danqrodney: You can revert this. We indent two spaces / don't use tabs.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -151,6 +151,11 @@ def _update_participant_summary(
     """
     # Block on other threads modifying the participant or participant summary.
     participant = ParticipantDao().get_for_update(session, questionnaire_response.participantId)
+
+    if participant is None:
```
danqrodney: With this change, you should be able to remove the call to validate_participant_reference below in this method.

danqrodney:
(Please add a test for this behavior.)

rest-api/dao/questionnaire_response_dao.py
shyamnaru: The validate_participant_reference method is calling the validate_participant_id method

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_dao.py#L157

which checks the withdrawn status, is this check not needed?

danqrodney: That's already being checked here on line 178.

danqrodney: OK, the test failure shows that we need to call raise_if_withdrawn not in an else block... it needs to happen in the insert case, too.

PR #533, opened 2017-11-17
ETL optimizations to improve performance.

danqrodney:
Making ETL scripts extra verbose.

### DIFF
--- a/rest-api/etl/etl.sql
+++ b/rest-api/etl/etl.sql
@@ -445,11 +445,11 @@ CREATE TABLE dose_era
 -- -------------------------------------------------------------------
 
 -- -------------------------------------------------------------------
--- table: tmp_clean_all
+-- table: src_clean
 -- -------------------------------------------------------------------
 
-DROP TABLE IF EXISTS cdm.tmp_clean_all;
-CREATE TABLE cdm.tmp_clean_all
+DROP TABLE IF EXISTS cdm.src_clean;
+CREATE TABLE cdm.src_clean
 (
     participant_id              bigint,
     date_of_survey              datetime,
@@ -470,7 +470,7 @@ CREATE TABLE cdm.tmp_clean_all
 -- Rules all together
 -- -------------------------------------------------------------------
 
-INSERT INTO cdm.tmp_clean_all
+INSERT INTO cdm.src_clean
 SELECT
     pa.participant_id               AS participant_id,
     qr.created                      AS date_of_survey,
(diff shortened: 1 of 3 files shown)


danqrodney:
I made some more changes to avoid joins on visit_occurrence at the end (by giving it physical_measurements_id as its ID.)

rest-api/etl/etl.sql
```diff
@@ -767,58 +716,47 @@ CREATE TABLE cdm.src_person_location
...
-    src.participant_id
-HAVING
-    COUNT(DISTINCT src.address_1) <= 2
-    AND COUNT(DISTINCT src.address_2) <= 2
-    AND COUNT(DISTINCT src.city) <= 2
-    AND COUNT(DISTINCT src.zip) <= 2
-    AND COUNT(DISTINCT src.state_ppi_code) <= 2
-;
+    src_participant.participant_id        AS participant_id,
+    MAX(m_address_1.value_string)         AS address_1,
```
calbach: Existing behavior, but I'm failing to understand why it would make sense to call `MAX` on a column named m_address_1. My only real concerned is the interaction with the DISTINCT address question below.

rest-api/etl/etl.sql
```diff
@@ -767,58 +716,47 @@ CREATE TABLE cdm.src_person_location
...
-    MAX(src.address_2)                          AS address_2,
-    MAX(src.city)                               AS city,
-    MAX(src.zip)                                AS zip,
-    MAX(src.state_ppi_code)                     AS state_ppi_code,
-    MAX(RIGHT(src.state_ppi_code, 2))           AS state
-FROM cdm.temp_src_person_location src
-GROUP BY
-    src.participant_id
-HAVING
-    COUNT(DISTINCT src.address_1) <= 2
```
calbach: I don't see this constraint being applied after your changes, though I could easily have missed it.

danqrodney: Yeah, this is a functional change. I'm checking these changes with the Odysseus team this morning to make sure they look to them.

I think the old logic may have been trying to only give an address to a person if they had only one associated address in our PPI data... but then these should have been < 2 instead of <= 2.

Instead I'm trying to change things to take the last address answer they gave.

rest-api/etl/etl.sql
danqrodney: MySQL has a mode (enabled by default since 5.7.5) called ONLY_FULL_GROUP_BY:

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

which rejects queries with GROUP BY where other columns referenced aren't aggregated in some way.

So even if we expect there will be exactly one row in the results, we still need to aggregate the other columns.

danqrodney:
Regarding the ETL: there are some very minimal tests, and basically no documentation. I am going to encourage the Odysseus team to add more of both. (Sorry.)

rest-api/etl/etl.sql
danqrodney: Odysseus confirmed these changes looked good.

danqrodney:
They looked, and approved verbally this morning. I'm going to go ahead and merge.

Thanks!

PR #538, opened 2017-11-29
[DA-463] Updating schema to support awardee > organization > site hierarchy.

danqrodney:
Adding new site fields. Removing consortium_name.


rest-api/alembic/versions/8786f0a95801_org_metadata.py
```diff
@@ -0,0 +1,84 @@
...
+    sa.Column('organization_id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(length=80), nullable=False),
+    sa.Column('display_name', sa.String(length=255), nullable=False),
+    sa.Column('hpo_id', sa.Integer(), nullable=False),
+    sa.ForeignKeyConstraint(['hpo_id'], ['hpo.hpo_id'], ),
+    sa.PrimaryKeyConstraint('organization_id')
+    )
+    op.add_column(u'site', sa.Column('address_1', sa.String(length=1024), nullable=True))
+    op.add_column(u'site', sa.Column('address_2', sa.String(length=1024), nullable=True))
+    op.add_column(u'site', sa.Column('admin_emails', sa.String(length=4096), nullable=True))
```
calbach: nit: some of these aren't in the design doc

rest-api/alembic/versions/8786f0a95801_org_metadata.py
```diff
@@ -0,0 +1,84 @@
...
+down_revision = 'b315fec9aa4e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('organization',
+    sa.Column('organization_id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(length=80), nullable=False),
```
calbach: Per comment I just left on the doc, I find this column name to be confusing in combination with "display_name".

ALso, is there generally any place in the code where documentation of these columns would live? Would be nice especially for columns like this one.

danqrodney: I renamed 'name' to 'external_id'. Does that work?

Note that on hpo this is called 'name', and on site 'googleGroup', so we're not being very consistent... but changing either of those is a bigger deal, and should be tackled separate from this PR.

I added comments to organization.py on the fields. That's the best place we have for documenting model objects.

rest-api/alembic/versions/8786f0a95801_org_metadata.py
danqrodney: Updated the design doc to try to match this.

Let me know if you spot any differences.

As usually, docs go stale the minute you start coding. :)

rest-api/alembic/versions/8786f0a95801_org_metadata.py
calbach: Thanks, I find this to be much clearer.

PR #540, opened 2017-12-01
Adding .gitattributes

danqrodney:
This should hopefully force github to do SQL diffs as text, handle newlines as LF everywhere.

### DIFF
--- a/rest-api/.gitattributes
+++ b/rest-api/.gitattributes
@@ -1,0 +1,4 @@
+*.py -text eol=lf
+*.sh -text eol=lf
+*.sql -text eol=lf
+*.csv -text eol=lf


calbach:
I don't follow, do you have an example PR where this was an issue? My understanding is that git by default normalizes all end of line characters to lf on commit etc, so that if this is disabled we'd allow people to commit carriage returns - I'm failing to square this with your PR description.

danqrodney:
The goal was to avoid the newline mismatch causing the bad diff here:

https://github.com/all-of-us/raw-data-repository/pull/536

But I think this is probably wrong.

I advised the Odysseus folks (who I suspect are developing on Windows) to run:

git config --global core.autocrlf true

instead. And then I would abandon this PR. Let me know if that sounds right to you.

calbach:
Thanks - that sounds right. I'd guess either they are sending a PR with carriage returns, or the currently checked in SQL has carriage returns - strange diff though. Since your previous modifications worked fine I'd guess it must be something client-side on their end.

PR #541, opened 2017-12-01
[RW-263] Add public_metrics_export endpoint

calbach:
For now, we just compute the metrics and return them directly in an approximation of the proposed metrics API.

Open questions:
- Prefixes on code values, e.g. "PMI_" for gender or "PIIState_" for states  - do we care about stripping them and if yes where can we assume a prefix?
- My solution to computing "years old" for both SQLite/MySQL is verbose using the SQL "NOW()" vs passing as parameter may not be ideal, alternative suggestions welcome
- Filter NULL values? Or pass through with a placeholder?
- Do we need to filter down participants first (e.g. check for consent) - or is presence in the particpant_summary table sufficient to include in the metrics?

Still TODO in follow-up PRs:
- Pending a decision on the [design doc](https://docs.google.com/document/d/1KqtsWU-GWd9tYjf8JbcxnmtlVen7O-PRkhJY-8tX5ig), this will later be modified to persist the output to either Cloud SQL or BigQuery
- Configure the cronjob

### DIFF
--- a/rest-api/offline/public_metrics_export.py
+++ b/rest-api/offline/public_metrics_export.py
@@ -51,7 +51,9 @@ _SQL_AGGREGATIONS = [
       WHERE {filter_test_sql}
       GROUP BY 1;
       """.format(filter_test_sql=_FILTER_TEST_SQL),
-      lambda v: EnrollmentStatus.lookup_by_number(v).name,
+      # Rewrite INTERESTED to CONSENTED, see note above.
+      lambda v: ('CONSENTED' if v is EnrollmentStatus.INTERESTED.number
+                 else EnrollmentStatus.lookup_by_number(v).name),
       None),
   # TODO(calbach): Verify whether we need to be conditionally trimming these
   # prefixes or leaving them unmodified. Unclear if all codes will have prefix
--- a/rest-api/offline/public_metrics_export.py
+++ b/rest-api/offline/public_metrics_export.py
@@ -181,13 +181,14 @@ class PublicMetricsExport(object):
     # Using a session here should put all following SQL invocations into a
     # non-locking read transaction per
     # https://dev.mysql.com/doc/refman/5.7/en/innodb-consistent-read.html
+    now = clock.CLOCK.now()
     test_hpo = HPODao().get_by_name(TEST_HPO_NAME)
     with database_factory.make_server_cursor_database().session() as session:
       for (key, sql, valuef, params) in _SQL_AGGREGATIONS:
         sql = replace_years_old(sql)
         out[key] = []
         p = {
-          'now': clock.CLOCK.now(),
+          'now': now,
           'test_hpo_id': test_hpo.hpoId,
           'test_email_pattern': TEST_EMAIL_PATTERN,
           'not_withdrawn_status': WithdrawalStatus.NOT_WITHDRAWN.number
--- a/rest-api/offline/public_metrics_export.py
+++ b/rest-api/offline/public_metrics_export.py
@@ -1,6 +1,9 @@
+import clock
+
 from sqlalchemy import text
 from dao import database_factory
 from dao.database_utils import replace_years_old
+from dao.hpo_dao import HPODao
 from participant_enums import EnrollmentStatus, OrderStatus, PhysicalMeasurementsStatus
 from participant_enums import Race, QuestionnaireStatus
 from participant_enums import TEST_EMAIL_PATTERN, TEST_HPO_NAME
@@ -21,8 +24,8 @@ def _questionnaire_metric(name, col):
 
 _FILTER_TEST_SQL = """
 (NOT participant_summary.email LIKE '{test_email}'
- AND NOT participant_summary.hpo_id = '{test_hpo}')
-""".format(test_email=TEST_EMAIL_PATTERN, test_hpo=TEST_HPO_NAME)
+ AND NOT participant_summary.hpo_id = :test_hpo_id)
+""".format(test_email=TEST_EMAIL_PATTERN)
 
 # Metrics definitions. 3-tuples of:
 # - (str) aggregation key name
@@ -96,15 +99,15 @@ _SQL_AGGREGATIONS = [
    SELECT
      CASE
       WHEN date_of_birth IS NULL THEN 'UNSET'
-      WHEN YEARS_OLD[date_of_birth] < 0 THEN 'UNSET'
-      WHEN YEARS_OLD[date_of_birth] <= 17 THEN '0-17'
-      WHEN YEARS_OLD[date_of_birth] <= 25 THEN '18-25'
-      WHEN YEARS_OLD[date_of_birth] <= 35 THEN '26-35'
-      WHEN YEARS_OLD[date_of_birth] <= 45 THEN '36-45'
-      WHEN YEARS_OLD[date_of_birth] <= 55 THEN '46-55'
-      WHEN YEARS_OLD[date_of_birth] <= 65 THEN '56-65'
-      WHEN YEARS_OLD[date_of_birth] <= 75 THEN '66-75'
-      WHEN YEARS_OLD[date_of_birth] <= 85 THEN '76-85'
+      WHEN YEARS_OLD[:now, date_of_birth] < 0 THEN 'UNSET'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 17 THEN '0-17'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 25 THEN '18-25'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 35 THEN '26-35'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 45 THEN '36-45'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 55 THEN '46-55'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 65 THEN '56-65'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 75 THEN '66-75'
+      WHEN YEARS_OLD[:now, date_of_birth] <= 85 THEN '76-85'
       ELSE '86+'
      END age_range,
      COUNT(*)
@@ -158,11 +161,15 @@ class PublicMetricsExport(object):
     # Using a session here should put all following SQL invocations into a
     # non-locking read transaction per
     # https://dev.mysql.com/doc/refman/5.7/en/innodb-consistent-read.html
+    test_hpo = HPODao().get_by_name(TEST_HPO_NAME)
     with database_factory.make_server_cursor_database().session() as session:
       for (key, sql, valuef) in _SQL_AGGREGATIONS:
         sql = replace_years_old(sql)
         out[key] = []
-        result = session.execute(text(sql))
+        result = session.execute(text(sql), params={
+            'now': clock.CLOCK.now(),
+            'test_hpo_id': test_hpo.hpoId
+        })
         for row in result:
           v = row.items()[0][1]
           if valuef:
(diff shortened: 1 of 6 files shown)


rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
+from sqlalchemy import text
+from dao import database_factory
+from dao.database_utils import replace_years_old
+from participant_enums import EnrollmentStatus, OrderStatus, PhysicalMeasurementsStatus
+from participant_enums import Race, QuestionnaireStatus
+
+# TODO(calbach): consider whether we need to filter down to consented
```
danqrodney: All of these use participant_summary, and thus only operate on consented participants. (Which I think is totally fine and appropriate.)

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+# individuals only.
+
+def _questionnaire_metric(name, col):
+  """Returns a metrics SQL aggregation tuple for the given key/column."""
+  return (
+      name,
+      """
+      SELECT {}, SUM(1)
```
danqrodney: Can we use count(*) instead of SUM(1)?

GROUP BY 1 also confused me initially... normally I'd expect GROUP BY <FIELD>. But I guess either works.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+   GROUP BY 1;
+   """.format(str(OrderStatus.UNSET.number), str(OrderStatus.CREATED.number)),
+   None),
+  _questionnaire_metric('questionnaireOnOverallHealth', 'questionnaire_on_overall_health'),
+  # Personal habits is a newer naming for lifestyle
+  _questionnaire_metric('questionnaireOnPersonalHabits', 'questionnaire_on_lifestyle'),
+  # Sociodemographics is a newer naming for 'the basics'
+  _questionnaire_metric('questionnaireOnSociodemographics', 'questionnaire_on_the_basics'),
+  _questionnaire_metric('questionnaireOnHealthcareAccess', 'questionnaire_on_healthcare_access'),
+  _questionnaire_metric('questionnaireOnMedicalHistory', 'questionnaire_on_medical_history'),
```
danqrodney: FYI, a number of these questionnaires don't actually exist yet. (We added fields for them a long time back... but the only ones currently in use are:

https://github.com/all-of-us-terminology/codebook-to-fhir/blob/master/config/ppi-codebook.json#L10

Not sure if that means we shouldn't include the other ones in the metrics yet.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+    GROUP BY 1
+   ) ps LEFT JOIN code
+   ON ps.state_id = code.code_id;
+   """.format(len('PIIState_') + 1),
+   None),
+  ('ageRange',
+   """
+   SELECT
+     CASE
+      WHEN YEARS_OLD[date_of_birth] < 0 THEN 'UNSET'
```
danqrodney: It's not obvious how YEARS_OLD(null) is returning < 0 here. Should we have a NULL check somewhere?

Also, I think some of these expressions return UNSET and others don't (e.g. state.) We should probably return UNSET everywhere, and make sure we have tests for it.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+  _questionnaire_metric('questionnaireOnFamilyHealth', 'questionnaire_on_family_health')
+]
+
+
+class PublicMetricsExport(object):
+  """Exports data from the database needed to generate public registration metrics."""
+
+  @staticmethod
+  def export():
+    # TODO(calbach): Write the output to a given target destination rather than
```
danqrodney: It may make sense to have a method like this that generates the JSON we want to write out, and have something else that writes it to the database. (I'm not sure.)

Let's also talk about where these wind up.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+    # returning it.
+    out = {}
+    # Using a session here should put all following SQL invocations into a
+    # non-locking read transaction per
+    # https://dev.mysql.com/doc/refman/5.7/en/innodb-consistent-read.html
+    with database_factory.make_server_cursor_database().session() as session:
+      for (key, sql, valuef) in _SQL_AGGREGATIONS:
+        sql = replace_years_old(sql)
```
danqrodney: On each SQL expression we're going to have to tack on some extra logic that says:

ps.hpo_id != :test_hpoId and ps.email NOT LIKE :test_email_pattern

as shown here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/offline/metrics_export.py#L78

The reason: we have test participants in prod which shouldn't count towards metrics.

And yes, I hate this fact, and no, there isn't anything we can do about it but filter them out. :)

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Metrics definitions. 3-tuples of:
+# - (str) aggregation key name
+# - (str) SQL statement to select value, count for a metric (in that order)
+# - (func(str): str) optional function which takes the value from the above SQL
+#   output and converts it for presentation
+_SQL_AGGREGATIONS = [
+  ('enrollmentStatus',
+   """
+   SELECT enrollment_status, SUM(1)
```
danqrodney: I'm wondering if you should add a helper method that's like:

def ps_aggregation(field):
  return 'SELECT %s, COUNT(*) FROM participant_summary GROUP BY 1' % field

Then I think most of these would just be:

ps_aggregation('enrollment_status')

etc.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,145 @@
...
+    FROM participant_summary
+    WHERE gender_identity_id IS NOT NULL
+    GROUP BY 1
+   ) ps LEFT JOIN code
+   ON ps.gender_identity_id = code.code_id;
+   """,
+   None),
+  ('race',
+   """
+   SELECT race, SUM(1)
```
danqrodney: As noted previously: we don't have a simple way today of generating stuff like our Race enum values from the CDR. Perhaps that should change. 

We could alternatively (as you indicated) produce a data warehouse like this one at the point of exporting the OMOP for the CDR version. If we're going to do that, we need to be fairly sure that no participants in the CDR subsequently get filtered out or have these field values change in the process of merging with EHR data.

rest-api/test/unit_test/offline_test/public_metrics_export_test.py
```diff
@@ -0,0 +1,261 @@
...
+
+    with FakeClock(TIME):
+      # Participant 2 starts out unpaired; later gets paired automatically when their physical
+      # measurements come in.
+      participant2 = Participant(participantId=2, biobankId=3)
+      participant_dao.insert(participant2)
+      self.send_consent('P2', email='bob@fexample.com')
+
+    with FakeClock(TIME):
+      # Test HPO affiliation; this test participant is ignored.
```
danqrodney: It looks like you copy / pasted this from the metrics_export_test code?

We could instead refactor it to a common place... or modify it. Some of the stuff (e.g. what time things happen) doesn't matter much for your metrics.

Note these comments indicate what *should* happen WRT test participants in metrics... but I'm not sure they are.

danqrodney:
I don't have any particular problem with using NOW() for the age stuff. You could alternatively take a bound parameter for the date at the time of running the query... that may be a bit easier to test?

rest-api/offline/public_metrics_export.py
danqrodney: One reason to potentially gather all the metrics in a method like this and then write them in a second one: we can write out all the metrics in a single transaction, updating them in place, and have them be consistent, rather than having to worry about metric versioning like we do with the pipeline. (You could do that in the same transaction as the reads I guess, but it seems better to break them up.)

rest-api/offline/public_metrics_export.py
```diff
@@ -3,74 +3,99 @@
...
       FROM participant_summary
+      WHERE {filter_test_sql}
       GROUP BY 1;
-      """.format(col),
+      """.format(col=col, filter_test_sql=_FILTER_TEST_SQL),
       lambda v: QuestionnaireStatus.lookup_by_number(v).name
   )
 
+_FILTER_TEST_SQL = """
+(NOT participant_summary.email LIKE '{test_email}'
```
danqrodney: Any reason not to use bound parameters here like we are elsewhere?

It doesn't matter provided we don't have weird characters in the e-mail or HPO value, but it'd make me feel slightly safer to do LIKE :test_email

rest-api/offline/public_metrics_export.py
danqrodney: I don't think we want to make the client work out counts by omission.

It should be OK to provide UNSET everywhere, and let the client decide whether or not to render it.

rest-api/offline/public_metrics_export.py
```diff
@@ -82,39 +107,43 @@ def _questionnaire_metric(name, col):
...
+      WHEN biospecimen_status = '{unset}' THEN 'UNSET'
+      WHEN biospecimen_status = '{created}' THEN 'UNSET'
       ELSE 'COLLECTED'
-     END, SUM(1)
+     END, COUNT(*)
    FROM participant_summary
+   WHERE {filter_test_sql}
    GROUP BY 1;
-   """.format(str(OrderStatus.UNSET.number), str(OrderStatus.CREATED.number)),
+   """.format(unset=str(OrderStatus.UNSET.number),
```
danqrodney: See above regarding bound parameters.

rest-api/offline/public_metrics_export.py
calbach: Sure - done.

rest-api/offline/public_metrics_export.py
calbach: Done.

rest-api/offline/public_metrics_export.py
calbach: According to the current spec they wouldn't have to do that since they aren't displaying them in the mock, but yeah - UNSET seems fine. Will update the design doc as well.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,204 @@
...
+    # Using a session here should put all following SQL invocations into a
+    # non-locking read transaction per
+    # https://dev.mysql.com/doc/refman/5.7/en/innodb-consistent-read.html
+    test_hpo = HPODao().get_by_name(TEST_HPO_NAME)
+    with database_factory.make_server_cursor_database().session() as session:
+      for (key, sql, valuef, params) in _SQL_AGGREGATIONS:
+        sql = replace_years_old(sql)
+        out[key] = []
+        p = {
+          'now': clock.CLOCK.now(),
```
danqrodney: Nit: store now() outside this loop and reuse it here, so we're using consistent timestamps across queries.

rest-api/offline/public_metrics_export.py
```diff
@@ -0,0 +1,204 @@
...
+      SELECT {col}, COUNT(*)
+      FROM participant_summary
+      WHERE {filter_test_sql}
+      GROUP BY 1;
+      """.format(col=col, filter_test_sql=_FILTER_TEST_SQL),
+      lambda v: QuestionnaireStatus.lookup_by_number(v).name,
+      None
+  )
+
+_FILTER_TEST_SQL = """
```
danqrodney: We may also want to filter out participants who have withdrawn (participant_summary.withdrawal_status = :no_use)

@jmandel let us know if that sounds right.

rest-api/offline/public_metrics_export.py
calbach: Yikes, good catch.

rest-api/offline/public_metrics_export.py
calbach: I added this to the filter.

PR #542, opened 2017-12-02
Fix enrollment status update bug

calbach:
I'm not sure if this column is used anywhere, seems important.. If yes - likely need to backfill if we care about the test env.

### DIFF
--- a/rest-api/dao/participant_summary_dao.py
+++ b/rest-api/dao/participant_summary_dao.py
@@ -248,7 +248,7 @@ class ParticipantSummaryDao(UpdatableDao):
                                                          summary.numCompletedBaselinePPIModules,
                                                          summary.physicalMeasurementsStatus,
                                                          summary.samplesToIsolateDNA)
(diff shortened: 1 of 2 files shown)


danqrodney:
Nice catch. :) 

We are fortunately mostly saved by this backfill which already runs automatically every night when biobank samples are finished importing:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L46

So I don't think we need to run a separate backfill. But that's just dumb luck really.

Would be nice if Python had some linter option for detecting things like this that you probably didn't intend. I'm not sure if there is...

calbach:
Such a check would be predicated, I think, on having an actual stable static definition of the class and static knowledge of the type of `summary` - which is probably not going to happen in Python unless you go fully towards [type annotations](https://www.python.org/dev/peps/pep-0484/). As usual in Python the only way to validate things will actually work at runtime is test coverage.

PR #543, opened 2017-12-04
[PPI] Updated ETL to OMOP v5.2

iboltaev:


rest-api/etl/etl.sql
```diff
@@ -458,7 +509,7 @@ CREATE TABLE cdm.src_clean
     value_ppi_code              varchar(200),
     topic_value                 varchar(200),
     value_code_id               bigint,
-    value_number                decimal(20,6),
+    value_number                double,
```
danqrodney: In Postgres, it looks like, observation.value_number in OMOP is  supposed to be a decimal:

https://github.com/OHDSI/CommonDataModel/blob/master/PostgreSQL/OMOP%20CDM%20ddl%20-%20PostgreSQL.sql#L490

So it seems like representing it as a double can lead to floating point issues. But in Oracle it's a FLOAT:

https://github.com/OHDSI/CommonDataModel/blob/master/Oracle/OMOP%20CDM%20ddl%20-%20Oracle.sql#L488

So... what is it supposed to be?

rest-api/etl/etl.sql
```diff
@@ -891,7 +948,7 @@ SELECT DISTINCT
     YEAR(b.date_of_birth)                       AS year_of_birth,
     MONTH(b.date_of_birth)                      AS month_of_birth,
     DAY(b.date_of_birth)                        AS day_of_birth,
-    TIMESTAMP(b.date_of_birth)                  AS birth_datetime,
+    NULL                                        AS birth_datetime,
```
danqrodney: Why is this NULL?

rest-api/etl/etl.sql
```diff
@@ -938,7 +997,7 @@ SELECT
     src_m1.participant_id                       AS person_id,
     COALESCE(vc.concept_id, 0)                  AS procedure_concept_id,
     src_m2.value_date                           AS procedure_date,
-    TIMESTAMP(src_m2.value_date)                AS procedure_datetime,
+    NULL                                        AS procedure_datetime,
```
danqrodney: Similarly, why NULL?

rest-api/etl/etl.sql
```diff
@@ -1187,49 +1226,137 @@ TRUNCATE TABLE cdm.care_site;
...
+    src_meas.finalized_site_id
+;
+
+-- -------------------------------------------------------------------
+-- table: cdm.tmp_visits_num
+-- -------------------------------------------------------------------
+DROP TABLE IF EXISTS cdm.tmp_visits_num;
+
+-- -------------------------------------------------------------------
+-- Here we order visits by person and visit_end_time, numbering each
```
danqrodney: I don't understand why we need to do this. Can you elaborate?

rest-api/etl/etl.sql
atsvetkova-ody: In the OMOP specification, field value_as_number is declared as float.
https://github.com/OHDSI/CommonDataModel/wiki/OBSERVATION

But, in practice, it can be decimal or float. I agree that to avoid floating point issues it is a good approach to use decimal.

rest-api/etl/etl.sql
atsvetkova-ody: In the source data there is no birth or procedure time information. But both cases are fine. If you need datetime field, we can change it back.

rest-api/etl/etl.sql
atsvetkova-ody: In cdm v.5.2 in visit_occurrence table there is a new field preceding_visit_occurrence_id. This part of the code calculates the preceding visit. This field is not required, so if you don’t need this field, we can leave it empty.
https://github.com/OHDSI/CommonDataModel/wiki/VISIT_OCCURRENCE

rest-api/etl/etl.sql
danqrodney: I don't know -- it all depends on how this field gets used. My assumption is that some analyses will expect this field to be set and won't bother checking date_of_birth if it isn't?

rest-api/etl/etl.sql
danqrodney: I don't think we need this field. @jmandel let me know if you disagree. We can probably drop whatever logic in the ETL is computing it.

rest-api/etl/etl.sql
danqrodney: @jmandel thoughts on this?

rest-api/etl/etl.sql
atsvetkova-ody: After your question, we have asked about rules for populating *_datetime fields when time is unknown, CDM builder team says that it is fine to populate *_datetime fields when there is no time in the source data. So, we can populate it again.
http://forums.ohdsi.org/t/should--datetime-fields-be-required/3686/7

danqrodney:
@iboltaev can you make the proposed changes above?

danqrodney:
Another issue we've discovered; unit_source_value and unit_concept_id are not being filled in on the observation table:

https://github.com/iboltaev/raw-data-repository/blob/608e88363a7d21862969f1f244ccf1a7edad2e60/rest-api/etl/etl.sql#L1400
https://github.com/iboltaev/raw-data-repository/blob/608e88363a7d21862969f1f244ccf1a7edad2e60/rest-api/etl/etl.sql#L1437

In the first case (observations from questionnaire responses), if there's a numeric value, we may need to have a unit for it.

In the second case (observations from measurements), we definitely need units.

Can you make that change as well?

danqrodney:
I'm going to go ahead and merge this.

PR #544, opened 2017-12-05
[DA-463] New awardee/organization/site importer script.

danqrodney:
Refactoring common logic to CsvImporter class.

New script that exports data from Google Sheets in CSV format and puts in the data dir.

Checking in new fake data files as well; won't be used in prod until we have real data that takes the place of our current data in test/test-data/hpos.csv and sites.csv.

### DIFF
--- a/rest-api/data/awardees.csv
+++ b/rest-api/data/awardees.csv
@@ -23,3 +23,4 @@ TEST,Test organization,UNSET
 SEEC,Miami,HPO
 SOUTHERN,Alabama,HPO
 WISCONSIN,Wisconsin,HPO
+FAKE_AWARDEE,Not a real awardee,FQHC
--- a/rest-api/data/awardees.csv
+++ b/rest-api/data/awardees.csv
@@ -1,0 +1,25 @@
+Awardee ID,Name,Type
+UNSET,No organization set,UNSET
+PITT,Pittsburgh,HPO
+COLUMBIA,New York,HPO
+ILLINOIS,Illinois,HPO
+AZ_TUCSON,Arizona,HPO
+COMM_HEALTH,"Community Health Center, Inc",FQHC
+SAN_YSIDRO,San Ysidro Health Center,FQHC
+CHEROKEE,Cherokee Health Systems,FQHC
+EAU_CLAIRE,Eau Claire Cooperative Health Center,FQHC
+HRHCARE,"HRHCare, Inc.",FQHC
+JACKSON,Jackson-Hinds Comprehensive Health Center,FQHC
+CAL_PMC,California ,HPO
+NE_PMC,New England,HPO
+TRANS_AM,Trans-American Consortium (TAC),HPO
+VA,VA,VA
+QTC,QTC,DV
+WALGREENS,Walgreens,DV
+QUEST,Quest Labs,DV
+EMSI,EMSI,DV
+SDBB,San Diego Blood Bank,DV
+TEST,Test organization,UNSET
+SEEC,Miami,HPO
+SOUTHERN,Alabama,HPO
+WISCONSIN,Wisconsin,HPO
--- a/rest-api/tools/csv_importer.py
+++ b/rest-api/tools/csv_importer.py
@@ -77,20 +77,20 @@ class CsvImporter(object):
       return False
     else:
       logging.info('Updating %s: old = %s, new = %s', self.entity_name, existing_dict,
-                   new_dict.as_dict())
+                   new_dict)
       if not dry_run:
         self._do_update(entity, existing_entity, session)
       return True
 
   def _do_update(self, entity, existing_entity, session):
     """Subclasses must specify how to implement update."""
-    for k, v in entity.as_dict().iteritems():
+    for k, v in entity.asdict().iteritems():
       if k != self.external_id_field and k != self.id_field:
         setattr(existing_entity, k, v)
     self.dao.update_with_session(session, entity)
 
   def _insert_entity(self, entity, existing_map, session, dry_run):
     #pylint: disable=unused-argument
-    logging.info('Inserting %s: %s', self.entity_name, entity.as_dict())
+    logging.info('Inserting %s: %s', self.entity_name, entity.asdict())
     if not dry_run:
       self.dao.insert_with_session(session, entity)
--- a/rest-api/tools/csv_importer.py
+++ b/rest-api/tools/csv_importer.py
@@ -5,7 +5,7 @@ import logging
 
 class CsvImporter(object):
   """Importer for database entities from CSV input.
-  
+
   Subclasses indicate in the constructor the name of the entity (for logging purposes),
   the DAO used to save the entity, the name of the primary key database ID field,
   the name of the external ID (referenced in the CSV file), and columns that must be populated
@@ -23,9 +23,9 @@ class CsvImporter(object):
 
   def run(self, filename, dry_run):
     """Imports entities from the CSV file with the specified name.
-    
+
     When dry_run flag is true, entities are not updated; instead logging indicates what would be
-    updated."""    
+    updated."""
     skip_count = 0
     new_count = 0
     updated_count = 0
@@ -79,7 +79,7 @@ class CsvImporter(object):
       logging.info('Not updating %s.', new_dict[self.external_id_field])
       return False
     else:
-      logging.info('Updating %s%s: old = %s, new = %s', self.entity_name, 
+      logging.info('Updating %s%s: old = %s, new = %s', self.entity_name,
                    ' (dry run)' if dry_run else '', existing_dict, new_dict)
       if not dry_run:
         self._do_update(entity, existing_entity, session)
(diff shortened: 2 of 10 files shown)


rest-api/tools/csv_importer.py
```diff
@@ -0,0 +1,96 @@
...
+  def _update_entity(self, entity, existing_entity, session, dry_run):
+    new_dict = entity.asdict()
+    new_dict[self.id_field] = None
+    existing_dict = existing_entity.asdict()
+    existing_dict[self.id_field] = None
+    if existing_dict == new_dict:
+      logging.info('Not updating %s.', new_dict[self.external_id_field])
+      return False
+    else:
+      logging.info('Updating %s: old = %s, new = %s', self.entity_name, existing_dict,
```
calbach: nit: I'd put that same "(dry run)" disclaimer here, this log statement is a bit scary without it.

rest-api/tools/csv_importer.py
```diff
@@ -0,0 +1,96 @@
...
+                   new_dict)
+      if not dry_run:
+        self._do_update(entity, existing_entity, session)
+      return True
+
+  def _do_update(self, entity, existing_entity, session):
+    """Subclasses must specify how to implement update."""
```
calbach: The interface could be clearer here. I'm confused about this documentation vs the default update implementation you've provided here.

rest-api/tools/csv_importer.py
```diff
@@ -0,0 +1,96 @@
...
+
+They then define _entity_from_row(row) to parse an entity out of a row dictionary.
+
+When dry_run flag is true, entities are not updated; instead logging indicates what would be
+updated.
+"""
+import csv
+import logging
+
+class CsvImporter(object):
```
calbach: Consider using https://docs.python.org/2/library/abc.html#abc.ABCMeta if you want this to be an abstract base class.

rest-api/tools/export_organizations.sh
```diff
@@ -0,0 +1,7 @@
+SHEET_EXPORT_BASE="https://docs.google.com/spreadsheets/d/1CcIGRV0Bd6BIz7PeuvrV6QDGDRQkp83CJUkWAl-fG58/export?format=csv&gid="
```
calbach: `#!/bin/bash` ?

calbach: nit: set -e so the whole script fails if one of these curls fails?

rest-api/tools/csv_importer.py
```diff
@@ -0,0 +1,96 @@
+"""Imports entities into the database based on a CSV file.
+
+Subclasses indicate in the constructor the name of the entity (for logging purposes),
```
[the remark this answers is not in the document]
danqrodney: Done.

rest-api/tools/csv_importer.py
```diff
@@ -0,0 +1,96 @@
...
+the DAO used to save the entity, the name of the primary key database ID field,
+the name of the external ID (referenced in the CSV file), and columns that must be populated
+with some value in the CSV.
+
+They then define _entity_from_row(row) to parse an entity out of a row dictionary.
+
+When dry_run flag is true, entities are not updated; instead logging indicates what would be
```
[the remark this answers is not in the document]
danqrodney: Done.

rest-api/tools/csv_importer.py
danqrodney: Done.

rest-api/tools/csv_importer.py
danqrodney: Interesting. We haven't been using this thus far for abstract classes. I don't necessarily mind introducing it... what would you say are the advantages over this approach?

rest-api/tools/csv_importer.py
danqrodney: Sorry, the documentation was wrong. Subclasses can just use this; they only need to override the method if there's special logic to run on an update.

Removed.

rest-api/tools/export_organizations.sh
danqrodney: Done

danqrodney: Done

rest-api/tools/csv_importer.py
calbach: This was based on my reading of the below which seemed to imply you had abstract methods. Looks like you don't so this isn't needed. The only reason I generally use this (when applicable) is to ensure instantiation failure when subclassing without overriding abstract methods.

rest-api/tools/import_organizations.py
```diff
@@ -0,0 +1,156 @@
...
+  The CSV files originate from the Google Sheets found here:
+
+  https://docs.google.com/spreadsheets/d/1CcIGRV0Bd6BIz7PeuvrV6QDGDRQkp83CJUkWAl-fG58/edit
+"""
+
+import logging
+
+from dateutil.parser import parse
+from tools.csv_importer import CsvImporter
+from dao.hpo_dao import HPODao
```
calbach: I don't suppose there's an API you could hit instead of going directly to the DB?

rest-api/tools/import_organizations.py
```diff
@@ -0,0 +1,156 @@
...
+from model.site import Site
+from model.site_enums import SiteStatus
+from participant_enums import OrganizationType
+from main_util import get_parser, configure_logging
+
+class HPOImporter(CsvImporter):
+
+  def __init__(self):
+    super(HPOImporter, self).__init__('awardee', HPODao(), 'hpoId', 'name',
+                                      ['Awardee ID', 'Name', 'Type'])
```
calbach: nit: constants? seems easy to fatfinger unstructured strings like this (awardee ID in particular)

rest-api/tools/import_organizations.py
```diff
@@ -0,0 +1,156 @@
...
+    except TypeError:
+      logging.info('Invalid organization type %s for awardee %s', type_str, row['Awardee ID'])
+      return None
+    return HPO(name=row['Awardee ID'],
+               displayName=row['Name'],
+               organizationType=organization_type)
+
+  def _insert_entity(self, entity, existing_map, session, dry_run):
+    # HPO IDs are not autoincremented by the database; manually set it here.
+    entity.hpoId = len(existing_map) + self.new_count
```
calbach: Do IDs start at 1 or 0? If 1, this should probably be +1

rest-api/tools/import_organizations.py
```diff
@@ -0,0 +1,156 @@
...
+  def __init__(self):
+    super(OrganizationImporter, self).__init__('organization', OrganizationDao(),
+                                               'organizationId', 'externalId',
+                                               ['Awardee ID', 'Organization ID', 'Name'])
+    self.hpo_dao = HPODao()
+
+  def _entity_from_row(self, row):
+    hpo = self.hpo_dao.get_by_name(row['Awardee ID'])
+    if hpo is None:
+      logging.info('Invalid awardee ID %s importing organization %s', row['Awardee ID'],
```
calbach: nit: Most of these seem like they should be logged at at least warning level

rest-api/tools/import_organizations.py
```diff
@@ -0,0 +1,156 @@
+"""Imports awardees / HPOs, organizations, and sites into the database.
+
+Usage:
```
calbach: May be worth stating whether or not this script is meant to be idempotent.

rest-api/data/awardees.csv
```diff
@@ -0,0 +1,26 @@
+Awardee ID,Name,Type
```
calbach: Double-checking - the contents of all of these lists are public information, right?

danqrodney: Correct.

rest-api/tools/csv_importer.py
danqrodney: There is one: subclasses must override _entity_from_row(self, row) in order for this to work.

That said, if you don't do it, the code will throw a useful error message the first time you call run() on a CSV file, which seems pretty much as good as failing on instantiation in this context.

rest-api/tools/import_organizations.py
danqrodney: Done.

rest-api/tools/import_organizations.py
danqrodney: There is not at present.

We had discussed whether it was worthwhile to add one. This is the only way we intend to write organization metadata, so adding an API around it seemed like extra work we could avoid. If we add one in the future, it hopefully shouldn't be too bad to transition to using it.

rest-api/tools/import_organizations.py
danqrodney: Done.

rest-api/tools/import_organizations.py
danqrodney: They start at 0, yes (0 = UNSET.)

rest-api/tools/import_organizations.py
danqrodney: Done.

PR #546, opened 2017-12-07
[RW-263] Define public metrics API

calbach:
These endpoints are restricted to a newly added `STOREFRONT` role.

See https://docs.google.com/document/d/1KqtsWU-GWd9tYjf8JbcxnmtlVen7O-PRkhJY-8tX5ig for details.


README.md
```diff
@@ -602,6 +602,37 @@ The response body includes:
...
+List all available metric sets and their IDs.
+
+The response body includes:
+
+* `metricSets`: an array of metric sets
+
+```
+{
+  "metricSets": [
+    {"id": "live"},
```
danqrodney: I'm wondering if for the sake of extensibility, we should have two parts to the ID: a type and a version, effectively.

So you might have { "type": "participants", "version": "live" } (or maybe "id": "live"). 

This would leave room for other types of metrics in future.

README.md
```diff
@@ -602,6 +602,37 @@ The response body includes:
...
+    ...
+  ]
+}
+```
+
+#### `GET /PublicMetricSets/:msid/Metrics`
```
danqrodney: (If we did what I was proposing above, the path would contain the type, too.)

rest-api/api_util.py
```diff
@@ -14,9 +14,10 @@
 # Role constants
 PTC = "ptc"
 HEALTHPRO = "healthpro"
+STOREFRONT = "storefront"
```
danqrodney: Note that you'll need to add stuff to config_<env>.json in order to whitelist the appropriate service account for storefront in each environment.

README.md
```diff
@@ -602,6 +602,37 @@ The response body includes:
 }
 ```
 
+## Public Metrics API
```
danqrodney: A general naming nit: calling this "public metrics API" seems weird if the API isn't actually publicly accessible. And it might be more flexible not to imply that the metrics returned by this are always visible through some public UI, either.

You could just call it "MetricSets" or whatever.

calbach: Yeah, I agree. My rationale here was that the structure of the response type is currently very specific to these kinds of single-dimensional broad aggregations, and I also wasn't sure about reconciling this against the existing metrics API.

I changed it to just MetricSets instead.

README.md
calbach: I'd prefer just an ID to a composite, but I added a `type` field. In practice, we'll be defining this values in code so we can choose whatever format we prefer for these IDs (including a type probably makes sense). I just find composite IDs annoying to pass around as it's almost an implementation detail of the resource ID that clients shouldn't necessarily need to parse or reconstruct. WDYT?

PS: my true preference on API IDs (per the Cloud style guide) is using a singular resource `name` to refer to things, e.g. `foos/<food_id>/bars/<bar_id>`; APIs would accept `bar_name: "foos/<food_id>/bars/<bar_id>"` to refer to this. But I won't push for this since it doesn't work very well with Swagger unfortunately.

README.md
calbach: See above response

README.md
```diff
@@ -602,6 +602,41 @@ The response body includes:
...
+  ]
+}
+```
+
+#### `GET /MetricSets/:msid/Metrics`
+
+List all metrics within the given metric set, optionally limited to a subset of
+metric names.
+
+TODO(calbach): Dump example JSON request/response here.
```
stephenyeargin: Flagging this TODO for later.

calbach:
Thanks for the reviews.

PR #547, opened 2017-12-08
[RW-263] Add parallel alembic structure for metrics DB + initial models

calbach:
- Fork the existing rdr alembic versions into metrics/rdr; alembic stores information in a table called "alembic_version" within the DB you are migrating, so it is fairly unavoidable versioning.
- Extend a `MetricsBase` class rather than `Base`, which indicates a model belongs to the metrics DB. We could also add subdirs here to be even more explicit (or perhaps, just a metrics subdir?)
- Add the initial alembic migrations for the `metrics` DB

Downsides to this approach:
- Higher complexity of upgrade process / DB tools
- Will need to be careful on the serve with how you access the associated DAOs; a given session may not be applicable to these models which are homed in another DB.


calbach:
Looking again at the multidb in one versions/ dir approach.

calbach:
It's looking possible now, so will likely drop this PR - please do not review.

calbach:
Dropping in favor of https://github.com/all-of-us/raw-data-repository/pull/549

I will leave the branch undeleted for now, in case we decide for some reason to change our approach.

PR #548, opened 2017-12-09
[RW-263] Support single version history alembic w/ 2 DBs; WIP

calbach:
Sending for initial thoughts - alternative approach to https://github.com/all-of-us/raw-data-repository/pull/547; not fully working yet (see comment inline).

This approach seems to be working for a single sql engine (connected to SQL, not a specific DB) but haven't tested extensively.


rest-api/alembic/versions/508708c67d9f_initial_metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(u'biobank_order_ibfk_4', 'biobank_order', type_='foreignkey')
```
calbach: Still trying to figure out what's going on here @danqrodney ; let me know if you have any ideas.

Seems some kind of bad interaction with setting the schema vs the ORM foreign key stuff we're using.

calbach:
Dropping in favor of https://github.com/all-of-us/raw-data-repository/pull/549

I will leave the branch undeleted for now, in case we decide for some reason to change our approach.

PR #549, opened 2017-12-12
[RW-263] Yet another multidb approach

calbach:
This is option 1c in https://docs.google.com/document/d/1FTmH-DDVlyY7BNsBzj0FV9m_kWQpCjKM__zK0GmIiCc/edit

This approach is slightly inconsistent, in that we specify a schema for non-RDR models, and leave it blank for the rest. We also need to maintain a default DB engine, in addition to a more generic (not USEing any DB) engine. However, it has the advantages that it allows us to leave our existing RDR code unmodified, and does not negatively affect alembic generation.


rest-api/alembic/env.py
```diff
@@ -65,13 +79,49 @@ def run_migrations_offline():
...
+
+
+autogen_blacklist = set(["aggregate_metrics_ibfk_1"])
+
+def include_object_fn(engine_name):
+  def f(obj, name, type_, reflected, compare_to):
+    # pylint: disable=unused-argument
+    # Workaround what appears to be an alembic bug for multi-schema foreign
+    # keys. This should still generate the initial foreign key contraint, but
+    # stops repeated create/destroys of the constraint on subsequent runs.
```
danqrodney: Nice find. :)

rest-api/model/metric_set.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class MetricSet(MetricsBase):
+  """A version containing a set of metrics in the database, generated by a pipeline.
+
+  Contains buckets with metrics grouped by HPO ID and date.
+  """
+  __tablename__ = 'metric_set'
+  metricSetId = Column('metric_set_id', String(50), primary_key=True)
+  metricSetType = Column('metric_set_type', Enum(MetricSetType), nullable=False)
```
danqrodney: We may want a last_modified timestamp on here too, for auditing purposes.

rest-api/dao/database_factory.py
```diff
@@ -19,13 +19,13 @@ def __init__(self, db_name, **kwargs):
...
 def get_database():
-  """Returns a singleton _SqlDatabase."""
+  """Returns a singleton _SqlDatabase which USEs the rdr DB."""
   return singletons.get(SQL_DATABASE_INDEX, _SqlDatabase, db_name='rdr')
 
-def get_metrics_database():
-  """Returns a singleton _SqlDatabase."""
-  return singletons.get(METRICS_SQL_DATABASE_INDEX, _SqlDatabase, db_name=None, execution_options={
-      'schema_translate_map': METRICS_SCHEMA_TRANSLATE_MAP
+def get_generic_database():
```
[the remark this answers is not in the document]
calbach: Done

calbach:
Verified this is working in sandbox, ran `rest-api/test/test_server.sh` against it.

This change will need to be sequenced as follows:
1. This PR, deployed and DBs upgraded in all environments
2. A follow-up PR to change the DB connection string in the config to omit the "rdr" part; db_factory will take care of this.

Metric sets DAOs may or may not fully work until (2) is in place.

calbach:
For future reference, I had to run the following along with submitting this PR (otherwise it fails to deploy on master), as this PR created a new DB.

```
tools/setup_database.sh --project pmi-drc-api-test --account <my_pmi_ops_acct>@pmi-ops.org
```

PR #551, opened 2017-12-13
[RW-263] Add public metrics cron specification

calbach:
Actual offline endpoint does not persist metrics yet, just returns them.

### DIFF
--- a/rest-api/cron.yaml
+++ b/rest-api/cron.yaml
@@ -6,7 +6,7 @@ cron:
   target: offline
 - description: Daily public metrics
   url: /offline/PublicMetricsRecalculate
-  schedule: every day 03:30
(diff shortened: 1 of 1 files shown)


rest-api/cron.yaml
```diff
@@ -4,6 +4,11 @@ cron:
   schedule: every day 04:00
   timezone: America/New_York
   target: offline
+- description: Daily public metrics
+  url: /offline/PublicMetricsRecalculate
+  schedule: every day 03:30
```
danqrodney: You might actually want to push this back a bit... BiobankSamplesImport will influence metrics on who is a full participant, and may not be done by the time this starts.

calbach: Done

PR #552, opened 2017-12-14
[RW-263] Persist aggregate metric computations

calbach:


rest-api/dao/metric_set_dao.py
```diff
@@ -1,17 +1,28 @@
...
 
-class AggregateMetricsDao(BaseDao):
+class AggregateMetricsDao(UpsertableDao):
 
   def __init__(self):
     super(AggregateMetricsDao, self).__init__(
         AggregateMetrics, db=database_factory.get_generic_database())
+
+  def delete_all_for_metric_set_with_session(self, session, ms_id):
+    aggs = (session.query(AggregateMetrics)
```
danqrodney: Rather than doing query followed by delete, I think we should use delete():

http://docs.sqlalchemy.org/en/latest/core/dml.html

Seems more efficient?

rest-api/offline/public_metrics_export.py
```diff
@@ -205,3 +215,25 @@ def export():
...
+  def _save(metric_set_id, metrics):
+    ms = MetricSet(
+        metricSetId=metric_set_id,
+        metricSetType=MetricSetType.PUBLIC_PARTICIPANT_AGGREGATIONS,
+        lastModified=clock.CLOCK.now()
+    )
+    with database_factory.get_generic_database().session() as session:
+      MetricSetDao().upsert_with_session(session, ms)
+      agg_dao = AggregateMetricsDao()
+      agg_dao.delete_all_for_metric_set_with_session(session, metric_set_id)
```
danqrodney: I guess if we had a guarantee that all keys in previous metrics were present in new metrics, we wouldn't technically need this delete (could use an upsert below instead). But I agree this is probably safer.

rest-api/dao/metric_set_dao.py
calbach: Thanks, done. This appears to be the right way to use that in combination with the declarative base model.

rest-api/offline/public_metrics_export.py
calbach: Yes, this could happen if we happened to have an aggregation with maybe ~1 person, then that person withdrew or we updated some information in the RDR which moved them out of that group.

PR #554, opened 2017-12-15
Disable daily metrics in sandbox

calbach:


calbach:
Tried deploying this and failed - apparently `gcloud app deploy` is picky about the cronjob filename being `cron.yaml`. Renamed the default cron file to `cron_default.yaml` to make room.

PR #556, opened 2017-12-15
Fixing typo in ETL

danqrodney:
Changing setup_local_database_for_etl.sh to point to new vocabulary in GCS.

Removing source_to_concept_map.csv (which we take from GCS.)


rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -6,7 +6,7 @@
 # the database. (Note: the loading takes upwards of 10 minutes... grab some coffee.)
 #
 # Vocabulary CSV files are copied from GCS at path
-# gs://all-of-us-rdr-vocabulary/vocabularies-2017-09-18/; these files were originally received
+# gs://all-of-us-rdr-vocabulary/vocabularies-2017-11-27/; these files were originally received
```
calbach: Checking - you intended to update this file without updating `setup_database_for_etl.sh` (which still refers to 9-18)?

danqrodney: Heh. Funny you should mention it... I just did that. :)

We may want to change both of these to read from a file which has this in a single place...

PR #558, opened 2017-12-18
Removing CREATE TABLE... AS SELECT... which doesn't work in Cloud SQL.

danqrodney:
Removing reverse fact_relationship for parent-child measurement and measurement-qualifier relationships.

Changing cdm.measurement to use RDR measurement ID for PK; removing meas_id column.


rest-api/etl/etl.sql
```diff
@@ -1783,53 +1781,20 @@ DROP TABLE IF EXISTS cdm.temp_obs;
...
-    27                              AS domain_concept_id_1,     -- Observation
-    fr.observation_id               AS fact_id_1,
-    21                              AS domain_concept_id_2,     -- Measurement
-    fr.measurement_id               AS fact_id_2,
-    581410                          AS relationship_concept_id,  -- Observation to Measurement
-    'observ.meas2'                  AS unit_id
-FROM cdm.tmp_fact_rel_om fr
+FROM cdm.observation cdm_obs
+INNER JOIN rdr.measurement_to_qualifier mtq
+    ON mtq.qualifier_id = cdm_obs.meas_id
```
calbach: Was there no purpose to the previous approach of autogenerated IDs - seems like that was pretty roundabout? For example, is the 'measurement_id' something that could be an identifier which we wouldn't want to expose in the CDR?

rest-api/etl/etl.sql
```diff
@@ -1941,50 +1906,19 @@ WHERE tmp1.systolic_blood_pressure_ind != 0              -- take only systolic b
...
-;
-
 -- ---------------------------------------------------------------------
 -- Insert into fact_relationship child-to-parent measurements relations
 -- ---------------------------------------------------------------------
 INSERT INTO cdm.fact_relationship
 SELECT
     21                              AS domain_concept_id_1,     -- Measurement
-    fr.measurement_id_child         AS fact_id_1,
+    cdm_meas.parent_id              AS fact_id_1,
```
calbach: These are now reversed `fact_id_1` was child, now its parent. Was that intentional? Table comment seems to indicate child, parent order.

rest-api/etl/etl.sql
danqrodney: In the ultimate CDR, it's going to be important to have different IDs from the IDs in the RDR. But there is going to be a merging process of combining the OMOP from the RDR with OMOP from EHR data, which will have to reassign IDs. So there's no need to do it here for that reason.

The only reason to auto-assign IDs in this ETL is to handle cases where there are inserts into the same CDM table from multiple sources, which might conflict. This is the case for the observation table, I believe, but not the measurement table.

rest-api/etl/etl.sql
danqrodney: No, that was a mistake. Fixed.

PR #559, opened 2017-12-20
Da 465

Seeker1911:
Add latitude and longitude geocoding to Site model when importing new organizations.


rest-api/model/site.py
```diff
@@ -27,6 +27,8 @@ class Site(Base):
   city = Column('city', String(255))
   state = Column('state', String(2))
   zipCode = Column('zip_code', String(10))
+  latitude = Column('latitude', Float)
```
danqrodney: I'm confused by this. latitude and longitude are already here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/model/site.py#L21

What's going on with this?

rest-api/tools/auth_setup.sh
```diff
@@ -50,6 +50,7 @@ fi
 
 REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"
 source ${REPO_ROOT_DIR}/tools/setup_vars.sh
+TMP_GEOCODE_INFO_FILE=${TMP_DIR}/geocode_key.json
```
danqrodney: It might be redundant, but we may want to explicitly rm this in cleanup, like we do for TMP_CREDS_FILE, etc.

rest-api/tools/import_organizations.py
```diff
@@ -173,10 +174,20 @@ def _entity_from_row(self, row):
...
                 state=state,
                 zipCode=zip_code,
+                latitude=latitude,
+                longitude=longitude,
                 phoneNumber=phone,
                 adminEmails=admin_email_addresses,
                 link=link)
 
+  def _get_lat_long_for_site(self, address_1, city, state):
+    api_key = os.environ.get('API_KEY')
```
danqrodney: Raise an error if this isn't set in the environment?

rest-api/tools/import_organizations.py
```diff
@@ -173,10 +174,20 @@ def _entity_from_row(self, row):
...
+    gmaps = googlemaps.Client(key=api_key)
+    geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
```
danqrodney: Presumably this could return an error / something saying that the geocode failed?

We should handle that gracefully -- log a warning and proceed with no lat/lng.

rest-api/tools/auth_setup.sh
```diff
@@ -86,6 +87,13 @@ function get_db_password {
   PASSWORD=`grep db_password $TMP_DB_INFO_FILE | cut -d\" -f4`
 }
 
+function get_geocode_key {
```
danqrodney: Doesn't this need to get called from import_organizations.sh or something?

rest-api/tools/auth_setup.sh
```diff
@@ -86,6 +87,13 @@ function get_db_password {
...
+  echo "Getting geocode api key ..."
+  tools/install_config.sh --key geocode_key --instance $INSTANCE \
```
danqrodney: Do we not need to pass in --project and --account here?

We definitely want to pull this from cloud, and not from local datastore.

We also need to make sure that import_organizations.sh allows you to update the local database, while still geocoding using this key...

rest-api/tools/import_organizations.py
```diff
@@ -156,6 +156,7 @@ def _entity_from_row(self, row):
     city = row.get(SITE_CITY_COLUMN)
     state = row.get(SITE_STATE_COLUMN)
     zip_code = row.get(SITE_ZIP_COLUMN)
+    latitude, longitude = self._get_lat_long_for_site(address_1, city, state)
```
danqrodney: Remove the TODO on line 155.

Also, check for address_1 being none; in that case we shouldn't call this.

rest-api/tools/auth_setup.sh
Seeker1911: We'll need to discuss this.

rest-api/tools/import_organizations.sh
```diff
@@ -25,6 +25,7 @@ then
   source tools/auth_setup.sh
   run_cloud_sql_proxy
   set_db_connection_string
+  get_geocode_key
```
danqrodney: The else block here -- when ${PROJECT} is not set -- is when we're just dealing with the local database. 

It still seems like in that case you'll need to get and use the geocode key?

rest-api/tools/auth_setup.sh
```diff
@@ -94,7 +91,7 @@ function get_db_password {
 function get_geocode_key {
   echo "Getting geocode api key ..."
   tools/install_config.sh --key geocode_key --instance $INSTANCE \
-       --config_output "$TMP_GEOCODE_INFO_FILE"
+       --creds_file ${CREDS_FILE} --project ${PROJECT} --config_output "$TMP_GEOCODE_INFO_FILE"
```
danqrodney: As mentioned over slack -- I think we should just hardcode --project to pmi-drc-api-test here. That way we can use the same geocode key for all cloud environments, as well as for local development.

rest-api/tools/import_organizations.py
```diff
@@ -181,12 +180,16 @@ def _entity_from_row(self, row):
...
-    api_key = os.environ.get('API_KEY')
-    gmaps = googlemaps.Client(key=api_key)
-    geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
-    latitude = geocode_result['geometry']['location']['lat']
-    longitude = geocode_result['geometry']['location']['lng']
-    return latitude, longitude
+    try:
+        api_key = os.environ.get('API_KEY')
+        gmaps = googlemaps.Client(key=api_key)
+        geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
```
danqrodney: Do we not care about address 2 for geocoding purposes?

Does there need to be a comma between city and state, or does it not care?

rest-api/tools/import_organizations.py
```diff
@@ -181,15 +181,15 @@ def _entity_from_row(self, row):
...
     try:
-        api_key = os.environ.get('API_KEY')
-        gmaps = googlemaps.Client(key=api_key)
-        geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
-        latitude = geocode_result['geometry']['location']['lat']
-        longitude = geocode_result['geometry']['location']['lng']
-        return latitude, longitude
+      api_key = os.environ.get('API_KEY')
+      gmaps = googlemaps.Client(key=api_key)
+      geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
```
danqrodney: Based on the documentation, it sounds like if you enter in a bad address here, you'll get back no results. We need to handle that case gracefully, logging a message and returning None from this function... or perhaps raising an exception and bailing out, but in a way that makes it clear that the address is wrong.

We should also check for the presence of geometry, location, and lat/lng below before accessing them... it seems possible these won't always come back even if an entry is present in the results.

rest-api/tools/import_organizations.py
```diff
@@ -181,13 +181,26 @@ def _entity_from_row(self, row):
...
+          try:
+            latitude = geocode_result['geometry']['location']['lat']
+            longitude = geocode_result['geometry']['location']['lng']
+            if latitude != None and longitude != None:
+              return latitude, longitude
+            else:
+              logging.info('Can not find lat/long for %s', full_address) 
+              return None, None
+          except KeyError as e:
+            logging.warn('error for lat/long. There is no %s for this address', e)
```
danqrodney: What is %s supposed to be substituted with here?

If you want to log the exception, you should probably use logging.exception I think...

rest-api/tools/import_organizations.py
```diff
@@ -181,13 +181,26 @@ def _entity_from_row(self, row):
...
-        longitude = geocode_result['geometry']['location']['lng']
-        return latitude, longitude
+        if geocode_result:
+          try:
+            latitude = geocode_result['geometry']['location']['lat']
+            longitude = geocode_result['geometry']['location']['lng']
+            if latitude != None and longitude != None:
+              return latitude, longitude
+            else:
+              logging.info('Can not find lat/long for %s', full_address)
```
danqrodney: This should be logging.warn I think.

rest-api/tools/import_organizations.py
```diff
@@ -181,13 +181,26 @@ def _entity_from_row(self, row):
...
         gmaps = googlemaps.Client(key=api_key)
         geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
-        latitude = geocode_result['geometry']['location']['lat']
-        longitude = geocode_result['geometry']['location']['lng']
-        return latitude, longitude
+        if geocode_result:
+          try:
+            latitude = geocode_result['geometry']['location']['lat']
+            longitude = geocode_result['geometry']['location']['lng']
+            if latitude != None and longitude != None:
```
danqrodney: I don't think this check is necessary -- if ['lat'] and ['lng'] don't cause KeyError, it won't be None.

You could instead do:

geometry = geocode_result.get('geometry')
if geometry:
  location = geometry.get('location')
  if location:
    latitude  = geometry.get('lat')
    longitude = geometry.get('lng')
    if latitude and longitude:
     ...

and log warnings in the else blocks.  Then you don't need to catch KeyError below.
 
Have you tested this logic with bad addresses locally?

rest-api/tools/import_organizations.sh
```diff
@@ -37,3 +49,10 @@ fi
 source tools/set_path.sh
 python tools/import_organizations.py --awardee_file data/awardees.csv \
   --organization_file data/organizations.csv --site_file data/sites.csv $DRY_RUN
+
+function finish {
+  rm -rf ${TMP_GEOCODE_DIR}
```
danqrodney: I believe that the trap below overrides the one inside auth_setup.sh (which calls cleanup.) So you need to do something like:

function cleanup {
}

at the top of the file... and then call cleanup in finish, so if auth_setup overrides it it still gets called.

rest-api/tools/import_organizations.py
danqrodney: Ping?

Seeker1911: No, what the Google API expects is a human readable string (without any additional info such as suite, bldg., room, etc.) i.e. "1600 Amphitheatre Parkway Mountain View CA 94043" 
and the postal_code is optional.

rest-api/tools/import_organizations.py
Seeker1911: The substitution will be the key (lat/long).

danqrodney: I don't see how that works... the arg here is "e" which is a KeyError, not a field name?

Seeker1911: It would say, "There is no lat for this address". Moot point now, as the key error checking has been removed.

rest-api/tools/import_organizations.sh
```diff
@@ -36,3 +56,9 @@ fi
 source tools/set_path.sh
 python tools/import_organizations.py --awardee_file data/awardees.csv \
   --organization_file data/organizations.csv --site_file data/sites.csv $DRY_RUN
+
+function finish {
+  cleanup
```
danqrodney: I don't think this does what you want it to do...

tools/auth_setup.sh above will override cleanup, so that finish here winds up only calling the version in auth_setup.

You also don't need to call trap cleanup above. There can only be one trap function I believe (which is the reason you have to do anything special here.)

Instead what you want is something like:


\# Intentionally does nothing.
function cleanup {
}

...

function finish {
  cleanup
  rm -rf ...
}

trap finish EXIT

This will make finish run, and call cleanup first thing... which if auth_setup.sh overrode it will actually do something.

Does that make sense?

rest-api/tools/import_organizations.py
```diff
@@ -187,22 +187,22 @@ def _get_lat_long_for_site(self, address_1, city, state):
...
+          longitude = location.get('lng')
+        if latitude and longitude:
+          return latitude, longitude
+        else:
+          logging.warn('Can not find lat/long for %s', full_address)
           return None, None
       else:
         logging.warn('Geocode results failed for %s.', full_address)
         return None, None
     except ValueError as e:
```
danqrodney: When do we expect this to occur?

rest-api/tools/import_organizations.py
```diff
@@ -202,7 +202,10 @@ def _get_lat_long_for_site(self, address_1, city, state):
...
-      logging.exception('Invalid geocode key: %s . error: %s', api_key, e)
+      logging.exception('Invalid geocode key: %s. ERROR: %s', api_key, e)
+      return None, None
+    except IndexError as e:
```
danqrodney: Does this still actually get thrown with your above changes for a bad address?

Seeker1911: It is possible.

PR #560, opened 2017-12-21
Deidentify operational SQL views

calbach:
For a usable diff, see the [second commit](https://github.com/all-of-us/raw-data-repository/pull/560/commits/5085e1c9d13ecf5305cad7cd2eecf3e1772ba3c4).

Context: https://docs.google.com/document/d/1FCkC_ocBVlXQeHCeFMPe_p8qteXOu6CZCXZXdevFeWw

Summary:
- For all physical measurements or questionnaire responses, omit string response values (replaced by a boolean column `_value_present`) as these generally contain freeform text
- For all timestamps across these three views, truncate to the year
- Clamp maximum age at 89
- Modify participant zip code according to [this process](https://docs.google.com/document/d/1FCkC_ocBVlXQeHCeFMPe_p8qteXOu6CZCXZXdevFeWw/edit?ts=5a3ac09e#heading=h.tne4qmg15zlw)
- Deidentify participant IDs by obfuscating them
  - At export time, generate a random salt to be used for all exported tables
  - Generate a sha1 hash of participant ID + salt and replace the `participant_id` column with this value
- Extend `/ExportTables` to optionally flag for deidentification; this routes to a separate bucket and triggers the above ID swaps. This also subsets the legal db/tables for export (only those which have been deidentified).

Future work:
- Obfuscated IDs should ideally be stable across exports
- Obfuscation should be reversible by those with RDR production access, e.g. for resolving data quality issues found in the operational export

Example table exports can be found [here](https://docs.google.com/spreadsheets/d/19hWu5nxTPxCMjDdAnD-SUonGzAXniZwvUuAhBLD1zzo/edit#gid=866106040).

CC @danqrodney


calbach:
Meeting tomorrow to confirm this approach, so not fully ready to go yet (but my guess is this is pretty close). Just getting the review going in the meantime.

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
```diff
@@ -0,0 +1,279 @@
...
+CREATE OR REPLACE VIEW participant_view AS
+ SELECT
+   p.participant_id,
+   YEAR(p.sign_up_time) sign_up_year,
+   p.withdrawal_status,
+   YEAR(p.withdrawal_time) withdrawal_year,
+   p.suspension_status,
+   YEAR(p.suspension_time) suspension_year,
+   hpo.name hpo,
+   IF(SUBSTR(ps.zip_code, 1, 3) IN (
```
danqrodney: I'm wondering if the logic could just be if the zip < 10000?

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
```diff
@@ -0,0 +1,279 @@
...
+   YEAR(ps.sample_status_1ur10_time) sample_status_1ur10_year,
+   ps.sample_status_1sal,
+   YEAR(ps.sample_status_1sal_time) sample_status_1sal_year,
+   ps.num_completed_baseline_ppi_modules,
+   ps.num_completed_ppi_modules,
+   ps.num_baseline_samples_arrived,
+   ps.samples_to_isolate_dna,
+   ps.consent_for_cabor,
+   YEAR(ps.consent_for_cabor_time) consent_for_cabor_year,
+   (SELECT IFNULL(GROUP_CONCAT(
```
danqrodney: Not sure if we want to do it in this PR, but we may want to consider dropping race_codes and hispanic from this view... they make the view not perform well with lots of participants (e.g. in sandbox.) To do it properly we'll probably have to denormalize things into participant_summary in future, rather than these joins.

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
```diff
@@ -0,0 +1,279 @@
...
+   YEAR(qr.created) questionnaire_response_submission_year,
+   qc.value question_code,
+   qc.code_id question_code_id,
+   YEAR(qra.end_time) answer_end_year,
+   ac.value answer_code,
+   ac.code_id answer_code_id,
+   qra.value_boolean answer_boolean,
+   qra.value_decimal answer_decimal,
+   qra.value_integer answer_integer,
+   (qra.value_string != '') answer_string_present,
```
danqrodney: From this:

https://stackoverflow.com/questions/3059805/difference-between-mysql-is-not-null-and

It sounds like checking != '' will result in NULL when value_string is NULL, and false when it's ''. 

I think in our data we normally have NULL, rather than '' for strings.

In any case, it's a bit weird to treat empty string and NULL differently here. Perhaps we should use COALESCE or something to avoid this? Definitely let's test the null cases...

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
```diff
@@ -0,0 +1,279 @@
...
+CREATE OR REPLACE VIEW physical_measurements_view AS
+ SELECT
+   p.participant_id participant_id,
+   YEAR(p.sign_up_time) participant_sign_up_year,
+   p.withdrawal_status participant_withdrawal_status,
+   YEAR(p.withdrawal_time) participant_withdrawal_year,
+   p.suspension_status participant_suspension_status,
+   YEAR(p.suspension_time) participant_suspension_year,
+   pm.physical_measurements_id physical_measurements_id,
+   pm.created created,
```
danqrodney: This is a timestamp. Should we use YEAR here too?

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
jmandel: (For what it's worth, I thought @calbach's algorithm matches the doc)

rest-api/offline/table_exporter.py
```diff
@@ -14,20 +26,89 @@ class TableExporter(object):
...
+      def f(row_proxy):
+        out = [v for v in row_proxy]
+        for i, key in enumerate(row_proxy.keys()):
+          if key != 'participant_id':
+            continue
+
+          pmi_id = out[i]
+          if pmi_id not in pmi_to_obfuscated:
+            obf_id = TableExporter._obfuscate_id(pmi_id, deidentify_salt)
+            pmi_to_obfuscated[pmi_id] = obf_id
```
jmandel: Passing this function around having it share the to/from mapping state ... closures are so lovely.

rest-api/offline/main.py
```diff
@@ -112,7 +112,13 @@ def export_tables():
   directory = resource_json.get('directory')
   if not directory:
     raise BadRequest("directory is required")
-  return json.dumps(TableExporter.export_tables(database, tables, directory))
+
+  # Ensure this has a boolean value to avoid downstream issues.
+  deidentified = resource_json.get('deidentified')
```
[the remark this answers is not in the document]
calbach: Done - I think I prefer the default of `false` for now. Setting this changes the output bucket so I don't think there's much risk of misuse.

rest-api/offline/table_exporter.py
```diff
@@ -14,20 +26,89 @@ class TableExporter(object):
...
+    if get_database().db_type == 'sqlite':
+      # No schemas in SQLite.
+      sql_table = table_name
+    SqlExporter(bucket_name, use_unicode=True).run_export(
+        output_path, 'SELECT * FROM {}'.format(sql_table), transformf=transformf)
+    return '%s/%s' % (bucket_name, output_path)
 
   @staticmethod
-  def export_tables(database, tables, directory):
+  def export_tables(database, tables, directory, deidentify):
```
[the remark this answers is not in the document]
calbach: Done.

rest-api/offline/table_exporter.py
```diff
@@ -14,20 +26,89 @@ class TableExporter(object):
...
+
+    A deidentified request outputs exports into a different bucket which may have less restrictive
+    ACLs than the other export buckets; for this reason the tables for these requests are also more
+    restrictive.
+
+    Deidentification also obfuscates participant IDs, as these are known by other systems (e.g.
+    HealthPro, PTC). Currently this ID obfuscation is not reversible and is not stable across
+    separate exports (note: it is stable across multiple tables in a single export request).
+
+    TODO: Address the above shortcomings (possibly via symmetric key encryption).
```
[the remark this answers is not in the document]
calbach: Just removed the TODO if we're not sure yet

rest-api/offline/table_exporter.py
```diff
@@ -37,6 +118,22 @@ def export_tables(database, tables, directory):
...
+        raise BadRequest("deidentified exports are only supported for database: {}".format(
+            _DEIDENTIFIED_DB_TABLE_WHITELIST.keys()))
+      tableset = set(tables)
+      table_whitelist = _DEIDENTIFIED_DB_TABLE_WHITELIST[database]
+      if not tableset.issubset(table_whitelist):
+        raise BadRequest("deidentified exports are unsupported for tables:"
+                         "{} (must be in {})".format(tableset - table_whitelist, table_whitelist))
+      # This salt must be identical across all tables exported, otherwise the exported particpant
+      # IDs will not be consistent. Used with sha1, so ensure this value isn't too short.
+      deidentify_salt = random.randint(1e9, 1e18)
```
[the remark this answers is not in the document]
calbach: Good suggestion - done.

rest-api/offline/main.py
```diff
@@ -112,7 +112,13 @@ def export_tables():
...
+  if deidentified is not True:
+    deidentified = False
+
+  return json.dumps(TableExporter.export_tables(database, tables, directory, deidentified))
```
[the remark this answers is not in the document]
calbach: I'm less confident in my ability to roll out this change if it's touching datastore config - how about a follow-up PR? Added TODO for this.

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
danqrodney: OK, didn't understand that. Thanks!

rest-api/offline/table_exporter.py
```diff
@@ -1,33 +1,114 @@
...
+      - salt (number) must be a long
+
+    Returns: a positive integer
+    """
+    h = hashlib.sha1()
+    h.update(struct.pack('>l', pmi_id))
+    h.update(salt)
+    # Just take the first 4 bytes so that the output ID is an integer, roughly in the same domain as
+    # the input PMI participant ID.
+    b = h.digest()[0:4]
```
danqrodney: According to this:

https://stackoverflow.com/questions/1867191/probability-of-sha1-collisions

With 1M participants and 4 bytes, I think you're almost certainly going to have some collisions.

I'd recommend using 8 bytes instead?

There may be other ways of generating unique random IDs, but I guess they are going to be more complicated than this...

rest-api/offline/table_exporter.py
```diff
@@ -1,33 +1,114 @@
...
+    if deidentify_salt:
+      # Deidentification requested: hash outgoing participant IDs with a consistent salt across this
+      # export. Cache obfuscated participant IDs across row callbacks to avoid recomputation and to
+      # detect collisions.
+      pmi_to_obfuscated = {}
+      obfuscated_to_pmi = {}
+      def f(row_proxy):
+        out = [v for v in row_proxy]
+        for i, key in enumerate(row_proxy.keys()):
+          if key != 'participant_id':
```
danqrodney: Just to confirm: the only thing we need to remap is participant_id, and that's the only column name in all views for it?

(That sounds right, just checking.)

rest-api/test/unit_test/offline_test/table_exporter_test.py
```diff
@@ -0,0 +1,54 @@
...
+    csv_path = deferred.run(tasks[0].payload)
+
+    with cloudstorage_api.open('/' + csv_path, mode='r') as output:
+      reader = csv.reader(output)
+      rows = list(reader)[1:]
+    self.assertEqual(2, len(rows))
+
+    pmi_ids = set([p1.participantId, p2.participantId])
+    obf_ids = set([row[0] for row in rows])
+    self.assertFalse(pmi_ids.intersection(obf_ids),
```
danqrodney: So technically it's just really unlikely to overlap, right? (There *could* be a collision...) You may want to update the message here.

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
danqrodney: So I think from an analysis standpoint, it's probably(?) better to have always a 0 or 1 for these "present" fields... easier to deal with than three possible values.

rest-api/offline/table_exporter.py
```diff
@@ -1,33 +1,114 @@
...
+            continue
+
+          pmi_id = out[i]
+          if pmi_id not in pmi_to_obfuscated:
+            obf_id = TableExporter._obfuscate_id(pmi_id, deidentify_salt)
+            pmi_to_obfuscated[pmi_id] = obf_id
```
danqrodney: One cause for concern: how much memory this map takes up when it has 1M entries in it. Assuming you switch the values to longs, the keys + values will be 16M bytes, but the overhead of map entries could be significant.

calbach: I'm not a python performance expert, but my guess is that I'd rather take the memory hit than the processing time.

This cache is not critical though, so I'm fine to remove it (though it does have a nice benefit that we can detect collisions, in the event of a programming error here). I don't necessarily trust this scheme if we scale it up to 1m participants (e.g. the issues you found in your comments).

Perhaps leave for now, then see what the performance looks like in sandbox (if that's in a good state?)

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
calbach: Maybe, though it makes this inconsistent from the other columns here. FTR there were only two possible output values here, NULL and 1.

I'm not particularly happy with any of the above solutions though, so I updated to 0/1.

rest-api/offline/table_exporter.py
```diff
@@ -45,10 +45,10 @@ def _obfuscate_id(pmi_id, salt):
     h = hashlib.sha1()
     h.update(struct.pack('>l', pmi_id))
     h.update(salt)
-    # Just take the first 4 bytes so that the output ID is an integer, roughly in the same domain as
+    # Just take the first 8 bytes so that the output ID is a long, roughly in the same domain as
     # the input PMI participant ID.
-    b = h.digest()[0:4]
-    return abs(struct.unpack('>l', b)[0])
+    b = h.digest()[0:8]
+    return abs(struct.unpack('>q', b)[0])
```
danqrodney: What's >q for? Add a comment?

rest-api/alembic/versions/04564bf1b5e8_deidentify_ppi_views.py
danqrodney: Followed up with @ssutherland about this. We should keep these columns around for now, even though they don't scale well. We can replace them with a more efficient implementation (reading from new denormalized columns in participant_summary) in future.

rest-api/offline/table_exporter.py
calbach: It's a format string for the struct package, added a comment.

calbach:
Agreed that would be a bit strange; we might want a separate endpoint at that point.

Another option would be to just set up a cronjob that periodically reexported the deidentified views. I'm unfamiliar with the requirements on their end.

PR #563, opened 2018-01-04
Deidentify work address zip code

calbach:
For a readable diff, see https://github.com/all-of-us/raw-data-repository/pull/563/commits/499a73bfb8d360ac979aac5e92b0c122e10dec90


rest-api/alembic/versions/060dba019a3a_deidentify_work_zip.py
```diff
@@ -0,0 +1,106 @@
...
+   qra.value_boolean answer_boolean,
+   qra.value_decimal answer_decimal,
+   IF(qc.value = 'EmploymentWorkAddress_ZipCode',
+      NULL,
+      qra.value_integer) answer_integer,
+   IF(qra.value_string IS NULL, 0, 1) answer_string_present,
+   YEAR(qra.value_date) answer_date_year,
+   YEAR(qra.value_datetime) answer_datetime_year,
+   IF(qra.value_uri IS NULL, 0, 1) answer_uri_present,
+   IF(qc.value != 'EmploymentWorkAddress_ZipCode', NULL,
```
danqrodney: I'm wondering if we can instead of this column just have an answer_integer_present column, which can be used for any integers we don't want to include due to PII concerns.

I don't think we actually need the zip in this table -- we're not doing it for other zip code answers...

danqrodney: Or, as discussed over Slack with Regina: just populate answer_integer with the integer version of this.

Adding a new column just for this particular question seems weird.

calbach: Well, we're adding a new oneoff use column in either case. The zip code column would have applied to any other zip code values we wanted to plumb through deidentified in the future.

That said, I'm happy to go with the simpler option. Done.

calbach: Annnnd... actually done.

PR #565, opened 2018-01-05
Mead/timezone support

Seeker1911:
Added time zone id (e.g. 'America/Los Angeles') to any geocoded Site. Exposed in Awardee API.
The Google Time Zone API has been turned on in test and staging.


rest-api/tools/import_organizations.py
```diff
@@ -176,38 +178,47 @@ def _entity_from_row(self, row):
...
                 link=link)
 
   def _get_lat_long_for_site(self, address_1, city, state):
-    full_address = address_1 + ' ' +  city + ' ' + state
+    self.full_address = address_1 + ' ' +  city + ' ' + state
     try:
-      api_key = os.environ.get('API_KEY')
-      gmaps = googlemaps.Client(key=api_key)
-      geocode_result = gmaps.geocode(address_1 + '' +  city + ' ' +  state)[0]
+      self.api_key = os.environ.get('API_KEY')
```
danqrodney: This doesn't get used outside this function I think, so technically you don't need to set it to a member var.

Seeker1911: Tested. It is required for timezone api.

PR #566, opened 2018-01-08
Making participant_view untruncated; creating truncated ppi_participant_view.

danqrodney:
Making ppi_participant_view the thing that can be exported to deidentified bucket.


rest-api/alembic/versions/106ecd33248b_participant_views.py
```diff
@@ -0,0 +1,321 @@
...
+   p.withdrawal_status,
+   p.withdrawal_time,
+   p.suspension_status,
+   p.suspension_time,
+   hpo.name hpo,
+   ps.zip_code,
+   state_code.value state,
+   recontact_method_code.value recontact_method,
+   language_code.value language,
+   LEAST(89, TIMESTAMPDIFF(YEAR, ps.date_of_birth, CURDATE())) age_years,
```
calbach: This was another transform I made on ppi_particpant_view

calbach: (clamping age at 89); may want to revert that here as well.

danqrodney: Done.

PR #570, opened 2018-01-13
Drop more data from PPI QA views

calbach:
Per usual, see commit ec9b7a1 for a useful diff.

- As requested, drop a few columns / questionnaires from this view
- `answer_decimal_value` -> `answer_decimal_present`; I haven't found this to be populated for any questions, but better to strip it out for now until there's a need.
- Ensure we have a 5 digit zip code before deidentifying
- Add a few comments to the SQL

Also, rename several no-op migration files to `<version>_noop.py` to avoid confusion. Would like to delete these entirely but this seems like a hairier operation.


rest-api/alembic/versions/29e344f31b21_drop_qa_columns.py
```diff
@@ -0,0 +1,263 @@
...
+   q.questionnaire_id questionnaire_id,
+   qr.questionnaire_response_id questionnaire_response_id,
+   YEAR(qr.created) questionnaire_response_submission_year,
+   qc.value question_code,
+   qc.code_id question_code_id,
+   YEAR(qra.end_time) answer_end_year,
+   ac.value answer_code,
+   ac.code_id answer_code_id,
+   qra.value_boolean answer_boolean,
+   IF(qra.value_decimal IS NULL, 0, 1) answer_decimal_present,
```
danqrodney: This seems fine. If at some point we do have decimal answers we'll need to consider having a whitelist for them. My guess is they wouldn't be PII, but better safe than sorry I guess.

PR #571, opened 2018-01-15
Mead/da 478

Seeker1911:
Now you can add siteId/organizationId/awardeeId to /PUT request on participant object.
The logic takes the lowest level sent in request and if it is a new value will use that as the new pairing mechanism. It then finds the true parents if they exist and sets them explicitly (regardless of what was sent in request. ProviderLink is also set if an awardee is being set.

IMPORTANT: metrics_export_test.test_metric_export is still failing. :(
```
{
    "signUpTime": "2018-01-15T18:36:03",
    "meta": {
        "versionId": "W/\"1\""
    },
    "siteId": 29,
    "participantId": "P146632080",
    "organizationId": 1,
    "biobankId": "Z783952757",
    "awardeeId": 4,
    "withdrawalStatus": "NOT_WITHDRAWN",
    "suspensionStatus": "NOT_SUSPENDED",
    "providerLink": [
        {
            "primary": true,
            "organization": {
                "reference": "Organization/AZ_TUCSON"
```


rest-api/model/participant.py
```diff
@@ -49,6 +49,13 @@ class ParticipantBase(object):
   def hpoId(cls):
     return Column('hpo_id', Integer, ForeignKey('hpo.hpo_id'), nullable=False)
 
+  @declared_attr
+  def organizationID(cls):
```
danqrodney: organizationId, not organizationID.

rest-api/model/participant_summary.py
```diff
@@ -181,6 +181,18 @@ class ParticipantSummary(Base):
 
   participant = relationship("Participant", back_populates="participantSummary")
 
+  @declared_attr
+  def hpoId(cls):
+    return Column('hpo_id', Integer, ForeignKey('hpo.hpo_id'), nullable=False)
+
+  @declared_attr
+  def organizationID(cls):
```
danqrodney: See above.

rest-api/dao/participant_dao.py
```diff
@@ -201,7 +280,10 @@ def from_client_json(self, resource_json, id_=None, expected_version=None, clien
         providerLink=json.dumps(resource_json.get('providerLink')),
         clientId=client_id,
         withdrawalStatus=resource_json.get('withdrawalStatus'),
-        suspensionStatus=resource_json.get('suspensionStatus'))
+        suspensionStatus=resource_json.get('suspensionStatus'),
+        organizationID=resource_json.get('organizationID'),
```
danqrodney: To be consistent: organizationId, not organizationID (for both).

The bigger issue though, is: the hpoId, organizationId, and siteId fields in the API representation should be strings like "PITT', "WISC_MAD", "hpo-site-bannerphoenix"... and the IDs on the participant and participant summary objects in the model are foreign keys to the underlying database rows. So you need to do the conversion from strings to IDs here, and IDs to strings in to_client_json, using HPODao, OrganizationDao, and SiteDao.

Note that these don't have to be set, so you'll need to handle the None case here.

For an example of where we do the conversion from ID to string on participant summary today:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L282 @

rest-api/test/unit_test/api_test/participant_api_test.py
```diff
@@ -8,6 +8,7 @@ class ParticipantApiTest(FlaskTestBase):
   def setUp(self):
     super(ParticipantApiTest, self).setUp()
     provider_link = {
+      "awardeeId": 2,
```
danqrodney: provider link shouldn't have an awardeeId field on it... that's supposed to go on participant or participant_summary.

rest-api/dao/participant_dao.py
```diff
@@ -127,6 +144,65 @@ def _do_update(self, session, obj, existing_obj):
...
+    existing_obj_list = [old_site, old_org, old_awardee]
+    # TODO: DO WE WANT TO PREVENT PAIRING IF EXISTING SITE HAS PM/BIO.
+    #PM = existing_obj.participantSummary.physicalMeasurementsCreatedSiteId
+    #BIO = existing_obj.participantSummary.biospecimenCollectedSiteId
+
+    if site is not None and site != old_site:
+      obj_parent_organization = session.query(Site.organizationId).filter_by(siteId=site).first()
+      organization = obj_parent_organization[0] if obj_parent_organization else None
+      if organization is not None:
+        obj_parent_awardee = session.query(Organization.hpoId)\
```
danqrodney: Same as above.

rest-api/dao/participant_dao.py
```diff
@@ -127,6 +144,65 @@ def _do_update(self, session, obj, existing_obj):
...
+                            .filter_by(organizationId=organization).first()
+        awardee = obj_parent_awardee[0] if obj_parent_awardee else None
+      else:
+        obj_parent_awardee = session.query(Site.hpoId)\
```
danqrodney: Same as above.

rest-api/dao/participant_dao.py
```diff
@@ -127,6 +144,65 @@ def _do_update(self, session, obj, existing_obj):
...
+    old_site = existing_obj.siteId
+    old_org = existing_obj.organizationID
+    old_awardee = existing_obj.hpoId
+    existing_obj_list = [old_site, old_org, old_awardee]
+    # TODO: DO WE WANT TO PREVENT PAIRING IF EXISTING SITE HAS PM/BIO.
+    #PM = existing_obj.participantSummary.physicalMeasurementsCreatedSiteId
+    #BIO = existing_obj.participantSummary.biospecimenCollectedSiteId
+
+    if site is not None and site != old_site:
+      obj_parent_organization = session.query(Site.organizationId).filter_by(siteId=site).first()
```
danqrodney: You can instead just do OrganizationDao().get(Site.organizationId) and then compare... that will use our cache, whereas this query call won't.

We could alternatively add parent fields to Organization and Site. Not sure if that will cause any issues.

Or you could just compare IDs? You don't have to fetch the objects themselves...

rest-api/dao/participant_dao.py
```diff
@@ -127,6 +144,65 @@ def _do_update(self, session, obj, existing_obj):
...
+        awardee = obj_parent_awardee[0] if obj_parent_awardee else None
+      else:
+        obj_parent_awardee = session.query(Site.hpoId)\
+                            .filter_by(siteId=site).first()
+        awardee = obj_parent_awardee[0] if obj_parent_awardee else None
+
+    elif site is None or site == old_site:
+      if organization is not None and organization != old_org:
+        if site:
+          parent_organization = session.query(Site.organizationId).filter_by(siteId=site).first()
```
danqrodney: Same as above.

rest-api/dao/participant_dao.py
```diff
@@ -127,6 +144,65 @@ def _do_update(self, session, obj, existing_obj):
...
+          if parent_organization != organization:
+            site = None
+        obj_parent_awardee = session.query(Organization.hpoId).filter_by(organizationId=organization).first()
```
danqrodney: Same as above.

rest-api/api_util.py
```diff
@@ -108,6 +114,30 @@ def format_json_enum(obj, field_name):
   else:
     obj[field_name] = UNSET
 
+def get_site_id_from_google_group(self, obj):
+  if 'siteId' in obj:
+    site = self.site_dao.get_by_google_group(obj['siteId'])
+    if site is not None:
+      obj['siteId'] = site.siteId
+  else:
```
danqrodney: You can remove else: pass here and below.

rest-api/api_util.py
```diff
@@ -108,6 +114,30 @@ def format_json_enum(obj, field_name):
   else:
     obj[field_name] = UNSET
 
+def get_site_id_from_google_group(self, obj):
```
danqrodney: As mentioned previously: you probably don't want to  pass in "self" as a first parameter here, as that's used normally when dealing with member methods on classes.  Just pass in the DAO object after obj, like in format_json_hpo. 

Also, I think rather than modifying the passed in object, it would be better in these cases to return it (or None). The only reason we had format_<STUFF> above modify the object in place is because we often create a dictionary first from a SQLAlchemy object and format certain fields in it before returning it... when converting from resource JSON, it makes more sense to construct a new SQLAlchemy object, and leave the original JSON alone. (Particularly as sometimes we save that JSON in the DB separate from the other stuff.)

rest-api/dao/participant_dao.py
```diff
@@ -194,14 +266,20 @@ def to_client_json(self, model):
...
     # biobankId, lastModified, signUpTime are set by DAO.
     return Participant(
         participantId=id_,
         version=expected_version,
         providerLink=json.dumps(resource_json.get('providerLink')),
         clientId=client_id,
         withdrawalStatus=resource_json.get('withdrawalStatus'),
-        suspensionStatus=resource_json.get('suspensionStatus'))
+        suspensionStatus=resource_json.get('suspensionStatus'),
+        organizationId=resource_json.get('organizationId'),
```
danqrodney: As noted above: you could instead do:

organizationId=get_organization_id(resource_json, self.organization_dao),
siteId=get_site_id(resource_json, self.site_dao),

etc...

Also see above: the API fields could be 'awardee', 'organization', and 'site'... we don't need to accept 'hpoId' for backwards compatibility purposes, since previously the only way to modify pairing was using providerLink.

rest-api/dao/participant_dao.py
```diff
@@ -176,6 +242,9 @@ def get_biobank_ids_sample(self, session, percentage, batch_size):
   def to_client_json(self, model):
     client_json = {
         'participantId': to_client_participant_id(model.participantId),
+        'hpoId': model.hpoId,
```
danqrodney: So, the doc here:

https://docs.google.com/document/d/1l7aJqmUYJ3Or6r3lndtfNQ1grfm2SDT2qFKLu8S3ucA/edit#heading=h.34rqknadf1cz

suggests that the API fields be called "awardee", "organization", and "site". "hpoId" already exists, and we should populate it too (with the same value as "awardee" ) for backwards compatibility, but we might deprecate it.

Let me know if those field names sound appropriate to you.

rest-api/dao/participant_dao.py
```diff
@@ -185,6 +255,9 @@ def to_client_json(self, model):
         'suspensionStatus': model.suspensionStatus,
         'suspensionTime': model.suspensionTime
     }
+    format_json_hpo(client_json, self.hpo_dao, 'awardee'),
```
danqrodney: So it seems like this might need to run once for hpoId and then do client_json['awardee'] = client_json['hpoId'].

Seeker1911:
@danqrodney check that `to_client_json` is of the format you expect in your comment above and check that the change to `metrics_export_test` is acceptable with providerLink change and explicitly passing hpoId in to `.update()`

rest-api/test/unit_test/offline_test/metrics_export_test.py
```diff
@@ -112,8 +112,9 @@ def _create_data(self):
     with FakeClock(TIME_2):
       # This update to participant has no effect, as the HPO ID didn't change.
       participant = self._participant_with_defaults(
-          participantId=1, version=1, biobankId=2,
+          participantId=1, version=1, biobankId=2, hpoId=3, # <<<< Had to add hpoId here.
```
danqrodney: OK, this seems fine.

rest-api/dao/participant_dao.py
```diff
@@ -127,13 +151,55 @@ def _do_update(self, session, obj, existing_obj):
...
+        awardee_id = site.hpoId
+        return site_id, organization_id, awardee_id
+    if organization_id != u'UNSET' and organization_id is not None:
+      if organization_id != old_organization_id:
+        organization = self.organization_dao.get(organization_id)
+        awardee_id = organization.hpoId
+        return None, organization_id, awardee_id
+    return None, None, awardee_id
+
+  @staticmethod
```
danqrodney: Remove this method?

rest-api/api_util.py
```diff
@@ -108,6 +114,24 @@ def format_json_enum(obj, field_name):
   else:
     obj[field_name] = UNSET
 
+def get_site_id_from_google_group(obj, site_dao):
+  if 'siteId' in obj:
+    site = site_dao.get_by_google_group(obj['siteId'])
+    if site is not None:
+      return site.siteId
```
danqrodney: I think it's better to explicitly return None at the end of these methods.

rest-api/test/unit_test/offline_test/metrics_export_test.py
danqrodney: I think the reason this was needed was this line:

https://github.com/all-of-us/raw-data-repository/blob/fb27e8acde5a4a607c657ca7d558b58ca979a4ab/rest-api/test/unit_test/unit_test_util.py#L71

Combined with this line:

https://github.com/all-of-us/raw-data-repository/pull/571/files/fb27e8acde5a4a607c657ca7d558b58ca979a4ab#diff-9bf21946b87489547eac6ffb0cf8e3faR117

I think we could have equally removed the UNSET assignment above (leaving hpoId as None) and it might work fine.

danqrodney:
Also, we need to update the README here:

https://github.com/all-of-us/raw-data-repository/blob/master/README.md#participant-api

and here:

https://github.com/all-of-us/raw-data-repository/blob/master/README.md#participantsummary-api

And (because we forgot to do this before) add time zone here:

https://github.com/all-of-us/raw-data-repository/blob/master/README.md#get-awardee

and here:

https://github.com/all-of-us/raw-data-repository/blob/master/README.md#get-awardeeaid

rest-api/test/unit_test/offline_test/metrics_export_test.py
Seeker1911: If I remember correctly I tried that and a bunch of other tests broke that rely on the UNSET variable

rest-api/dao/participant_dao.py
```diff
@@ -263,6 +244,8 @@ def to_client_json(self, model):
     format_json_date(client_json, 'withdrawalTime')
     format_json_date(client_json, 'suspensionTime')
     client_json['awardee'] = client_json['hpoId']
+    if 'siteId' in client_json:
```
danqrodney: Should format_json_site do this?

rest-api/dao/participant_summary_dao.py
```diff
@@ -163,11 +166,19 @@ def _add_order_by(self, query, order_by, field_names, fields):
...
     """Handle HPO and code values when parsing filter values."""
-    if field_name == 'hpoId':
+    if field_name == 'hpoId' or field_name == 'awardee':
       hpo = self.hpo_dao.get_by_name(value)
       if not hpo:
         raise BadRequest('No HPO found with name %s' % value)
+      if field_name == 'awardee':
+        field_name = 'hpoId'
       return super(ParticipantSummaryDao, self).make_query_filter(field_name, hpo.hpoId)
+    if field_name == 'organization':
```
danqrodney: What about site?

rest-api/dao/participant_summary_dao.py
```diff
@@ -279,7 +290,12 @@ def to_client_json(self, model):
       result['ageRange'] = get_bucketed_age(date_of_birth, clock.CLOCK.now())
     else:
       result['ageRange'] = UNSET
+    if 'organizationId' in result:
```
danqrodney: Use format_json_organization, format_json_site here?

rest-api/dao/participant_summary_dao.py
Seeker1911: Site is covered by the _SITE_FIELDS list

rest-api/dao/participant_dao.py
Seeker1911: it does if site_id is not None, lots of fields get called on format_json_site, I ran into a circumstance where I could still have SiteId here and would return null on client side, this was just to catch that IF it happens.

README.md
```diff
@@ -163,11 +163,9 @@ with `B`.
   * `primary`: `true` | `false`, indicating whether this provider is the "main"
     provider responsible for recruiting a participant and performing physical
     measurements and biospecimen collection
-  * `organization`: Reference to an HPO organization, like `{"reference":
-    "Organization/PITT"}`
-  * `site`: array of
-    [references](http://hl7.org/fhir/references.html#reference) to an HPO site,
-    like `{"reference": "Organization/PITT_SOME_CLINIC"}`
+  * `awardee`: Reference to an awardee  pairing level, like `awardee: AZ_TUCSON`
```
danqrodney: I think you can remove awardee: , organization: , and site: after the "like" here and below... those aren't a part of the value.

Also, you should mention UNSET as a possibility here.

README.md
```diff
@@ -232,6 +230,9 @@ The summary includes the following fields:
 * `physicalMeasurementsFinalizedSite`: indicates the site where physical measurements were finalized for the participant
 * `signUpTime`: the time at which the participant initially signed up for All Of Us
 * `hpoId`: HPO marked as `primary` for this participant, if any (just the resource id, like `PITT` — not a reference like `Organization/PITT`)
+* `awardee`: An awardee a participant is paired with or "unset" if none.
```
danqrodney: Make consistent with participant documentation; and it should be 'UNSET' rather than 'unset' I think?

README.md
```diff
@@ -880,6 +884,7 @@ Example response:
       "id": "AZ_TUCSON_BANNER_HEALTH",
       "sites": [
         {
+	  "timeZoneId": "America/Phoenix",
```
danqrodney: Fix indentation

README.md
```diff
@@ -896,7 +901,9 @@ Example response:
           "notes": "Formerly University of Arizona CATS Research  ",
           "phoneNumber": "666-666-6666",
           "physicalLocationName": "",
-          "siteStatus": "INACTIVE"
+          "siteStatus": "INACTIVE",
+	  "longitude": -110.978,
```
danqrodney: Fix indentation

rest-api/test/unit_test/api_test/participant_api_test.py
```diff
@@ -99,10 +85,35 @@ def test_update_right_ifmatch_specified(self):
...
+  def test_change_pairing_for_org_then_site(self):
+    participant = self.send_post('Participant', self.participant)
+    participant['providerLink'] = [ self.provider_link_2]
+    participant_id = participant['participantId']
+    participant['organization'] = 'AZ_TUCSON_BANNER_HEALTH'
+    path = 'Participant/%s' % participant_id
+    update_awardee = self.send_put(path, participant, headers={'If-Match': 'W/"1"'})
+    update_awardee['site'] = 'hpo-site-bannerphoenix'
+    self.send_put(path, update_awardee, headers={'If-Match': 'W/"2"'})
+    self.assertEqual(update_awardee['site'], 'hpo-site-bannerphoenix')
```
danqrodney: You're not capturing the result of send_put here, so this isn't actually testing that things changed...

rest-api/test/unit_test/api_test/participant_api_test.py
```diff
@@ -99,10 +85,35 @@ def test_update_right_ifmatch_specified(self):
...
+  def test_change_pairing_awardee_and_site(self):
+    participant = self.send_post('Participant', self.participant)
+    participant['providerLink'] = [ self.provider_link_2]
+    participant_id = participant['participantId']
+    participant['awardee'] = 'PITT'
+    participant['site'] = 'hpo-site-monroeville'
+    path = 'Participant/%s' % participant_id
+    update_awardee = self.send_put(path, participant, headers={'If-Match': 'W/"1"'})
+    self.assertEquals(participant['site'], update_awardee['site'])
+    self.assertEquals(participant['awardee'], update_awardee['awardee'])
```
danqrodney: It would be good to test that:

- just passing up site changes organization and awardee to ancestors
- just passing up organization changes awardee to parent, and clears out site
- just passing up awardee changes awardee and clears out site and organization

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -70,23 +72,22 @@ def setUp(self):
...
+      }
     }
 
+  def test_pairing_summary(self):
+    participant = self.send_post('Participant', {"providerLink": [self.provider_link]})
+    participant_id = participant['participantId']
+    path = 'Participant/%s' % participant_id
+    participant['awardee'] = 'PITT'
+    particpant_update = self.send_put(path, participant, headers={'If-Match': 'W/"1"'})
+    self.assertEquals(particpant_update['awardee'], participant['awardee'])
```
danqrodney: This checks the summary for awardee, but isn't checking organization or site; and it's not checking that you can query by organization or site and get back participants.

You can add tests in here:

https://github.com/all-of-us/raw-data-repository/blob/8e897fb23127085dc3ff964ae7e70b75cf7e6ff8/rest-api/test/unit_test/api_test/participant_summary_api_test.py#L471

to do this sort of thing...

rest-api/dao/participant_summary_dao.py
danqrodney: If you were going to follow the pattern of format_json_site:

https://github.com/all-of-us/raw-data-repository/blob/8e897fb23127085dc3ff964ae7e70b75cf7e6ff8/rest-api/api_util.py#L87

then it would take care of setting "site" and deleting "siteId" on the object for you. That seems simpler.

rest-api/test/unit_test/api_test/participant_api_test.py
```diff
@@ -111,9 +111,20 @@ def test_change_pairing_for_org_then_site(self):
...
-    self.send_put(path, update_awardee, headers={'If-Match': 'W/"2"'})
-    self.assertEqual(update_awardee['site'], 'hpo-site-bannerphoenix')
+
+    update_1 = self.send_put(path, participant, headers={'If-Match': 'W/"1"'})
+    participant['site'] = 'hpo-site-bannerphoenix'
+    update_2 = self.send_put(path, participant, headers={'If-Match': 'W/"2"'})
+    self.assertEqual(update_1['site'], 'UNSET')
+    self.assertEqual(update_1['organization'], 'UNSET')
+    self.assertEqual(update_2['site'], 'hpo-site-bannerphoenix')
+    self.assertEqual(update_2['organization'], 'UNSET')
```
danqrodney: This looks wrong. If site is sent for pairing, organization and awardee should be set. If organization is set, awardee should be set (but  site should be UNSET.)

Add more tests and fix behavior?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
danqrodney: Please add tests of the ParticipantSummary endpoint, around where I linked to above.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -477,10 +477,22 @@ def testQuery_manyParticipants(self):
...
+                           [[ps_1, ps_2], [ps_3]])
 
       # Test filtering on fields.
       self.assertResponses('ParticipantSummary?_count=2&firstName=Mary',
                            [[ps_2]])
+      self.assertResponses('ParticipantSummary?_count=2&site=hpo-site-monroeville',
+                           [[ps_3]])
+      self.assertResponses('ParticipantSummary?_count=2&awardee=PITT',
+                           [[ps_1, ps_2], [ps_3]])
+      self.assertResponses('ParticipantSummary?_count=2&organization=AZ_TUCSON_BANNER_HEALTH',
```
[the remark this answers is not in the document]
danqrodney: To be addressed in a follow-up.

PR #572, opened 2018-01-17
Participant view: _year -> _time

calbach:


rest-api/alembic/versions/f81a8b941a77_fix_pv_column_name.py
```diff
@@ -56,7 +56,7 @@
    ps.consent_for_electronic_health_records,
    ps.consent_for_electronic_health_records_time,
    ps.questionnaire_on_overall_health,
-   ps.questionnaire_on_overall_health_time questionnaire_on_overall_health_year,
+   ps.questionnaire_on_overall_health_time questionnaire_on_overall_health_time,
```
danqrodney: Nit: You can just omit the questionnaire_on_overall_health_time, and it will use the name of the column from ps... that's what we do for the other questionnaires.

calbach: Oops - done.

PR #573, opened 2018-01-18
[DA-485] sleep after IAM key creation

calbach:
Local testing showed the following (though highly dependent on *when* I ran the tests):
- no sleep: ~75% failure
- 1s sleep: ~13% failure
- 3s sleep: ~2% failure  <--
- 5s sleep: ~1% failure
- 10s sleep: ~0 failures

10s is too slow for a developer tool and 5s didn't seem much better than 3s. Note: this is hardly scientifically sound as tests were run back-to-back with a sample size of only 30-60 attempts and the cause is entirely environmental.

### DIFF
--- a/rest-api/tools/auth_setup.sh
+++ b/rest-api/tools/auth_setup.sh
@@ -46,6 +46,14 @@ else
   gcloud iam service-accounts keys create $CREDS_FILE \
       --iam-account=$SERVICE_ACCOUNT --account=$CREDS_ACCOUNT
   TMP_PRIVATE_KEY=`grep private_key_id $CREDS_FILE | cut -d\" -f4`
+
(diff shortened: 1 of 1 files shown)


calbach:
RE: buganizer not yet, still working through the mailing lists.

I think since this only happens when creating a key for auth, there's not going to be a situation where it would be favorable to risk it and not sleep, i.e. this should always reduce your expected time to run the tool.  Any time this fails you'll still need to notice that the tool failed and manually rerun it which will cost you time; environment should be irrelevant.

For now, this is just myself and @danqrodney running this so its easy enough to adjust if needed.

calbach:
One argument to make it adjustable might be to make the sleep time **higher** if the issue gets really bad, but hopefully 3s reduces the likelihood sufficiently.

PR #576, opened 2018-01-23
merge alembic heads

Seeker1911:


rest-api/alembic/versions/403ebb6aa7af_.py
```diff
@@ -0,0 +1,50 @@
...
+from participant_enums import PhysicalMeasurementsStatus, QuestionnaireStatus, OrderStatus
+from participant_enums import WithdrawalStatus, SuspensionStatus
+from participant_enums import EnrollmentStatus, Race, SampleStatus, OrganizationType
+from participant_enums import MetricSetType, MetricsKey
+from model.site_enums import SiteStatus
+from model.code import CodeType
+
+# revision identifiers, used by Alembic.
+revision = '403ebb6aa7af'
+down_revision = ('f81a8b941a77', '0185cd07668a')
```
danqrodney: I don't know how this works, but I don't think we want to do things this way.

As I understand it, the issue is we had two different alembic versions submitted with down_revision d2e17663c615. We shouldn't do that. So I'd remove this revision...

https://github.com/all-of-us/raw-data-repository/blob/5b771ca128f41f857ceafbe5d35263cabe275192/rest-api/alembic/versions/0185cd07668a_new_foreign_keys.py

and then regenerate schema.

PR #580, opened 2018-01-25
Mead/support new codes

Seeker1911:


rest-api/dao/participant_summary_dao.py
```diff
@@ -201,9 +204,12 @@ def update_from_biobank_stored_samples(self, participant_id=None):
     If participant_id is provided, only that participant will have their summary updated."""
     baseline_tests_sql, baseline_tests_params = get_sql_and_params_for_array(
         config.getSettingList(config.BASELINE_SAMPLE_TEST_CODES), 'baseline')
+    self._back_fix_new_codes(baseline_tests_params)
```
danqrodney: This seems wrong. Where baseline_tests_sql is used below, we *do* want to count 1PS08, 1PST8, 1SS08, and 1SST8. So I don't think we need to call this method.

rest-api/dao/participant_summary_dao.py
```diff
@@ -201,9 +204,12 @@ def update_from_biobank_stored_samples(self, participant_id=None):
...
     dna_tests_sql, dna_tests_params = get_sql_and_params_for_array(
         config.getSettingList(config.DNA_SAMPLE_TEST_CODES), 'dna')
     sample_sql, sample_params = _get_sample_sql_and_params()
+    self._back_fix_new_codes(sample_params)
```
danqrodney: Similarly we don't need to call this here. (Because of what I mentioned above, and also because 1PST8, etc. aren't in the DNA sample list.)

rest-api/dao/participant_summary_dao.py
```diff
@@ -249,6 +255,12 @@ def update_from_biobank_stored_samples(self, participant_id=None):
   def _get_num_baseline_ppi_modules(self):
     return len(config.getSettingList(config.BASELINE_PPI_QUESTIONNAIRE_FIELDS))
 
+  def _back_fix_new_codes(self, obj):
```
danqrodney: You don't need this function; see above.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_test.py
```diff
@@ -52,6 +52,10 @@ def test_end_to_end(self):
       biobank_ids.append(participant.biobankId)
       self.assertEquals(summary_dao.get(participant.participantId).numBaselineSamplesArrived, 0)
     test1, test2, test3 = random.sample(_BASELINE_TESTS, 3)
+    test_list = [test1, test2, test3]
+    test1, test2, test3 = ['1PST8' if i=='1PS08' else '1SST8' if i == '1SS08' else i for i in \
```
danqrodney: So this would have our tests only run with the old codes. I'm thinking we should either switch the tests to only use the new codes, or have a mix of both?

rest-api/dao/participant_summary_dao.py
```diff
@@ -83,8 +83,11 @@ def _get_sample_sql_and_params():
     sample_param = 'sample%d' % i
     sample_param_ref = ':%s' % sample_param
     lower_test = BIOBANK_TESTS[i].lower()
-    sql += _SAMPLE_SQL % {"test": lower_test, "sample_param_ref": sample_param_ref}
-    params[sample_param] = BIOBANK_TESTS[i]
+    if lower_test == '1ps08' or lower_test == '1ss08':
+      del lower_test
+    else:
+      sql += _SAMPLE_SQL % {"test": lower_test, "sample_param_ref": sample_param_ref}
```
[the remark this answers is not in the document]
Seeker1911: This only pull the items from code_constants list to setup the sql query to insert into the respective columns. Removing the 08's should be all we need for this part.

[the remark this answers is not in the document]
danqrodney: If you look here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L66

and here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L71

We're using the test values to set columns. So sample_status_1pst8 and sample_status_1pst8_time gets set when 1PST8 comes in, etc.

We now need logic that will set them when either 1PST8 or 1PS08 comes in... nothing here does that.

PR #581, opened 2018-01-25
Drop withdrawn participants from PPI QA views

calbach:
As usual, see bd5dc1d for a readable diff.

We probably should consider improving the current migration scheme as there's a fairly large buildup of empty files now. I looked at http://alembic.zzzcomputing.com/en/latest/cookbook.html#the-replaceable-object-structure as one possible alternative for managing these views, but I'm thinking it might be too complicated to justify. Alternatively could look at getting rid of these no-op files somehow.


rest-api/alembic/versions/574efa4de1ba_ppi_drop_withdrawals.py
```diff
@@ -222,27 +213,84 @@ def downgrade(engine_name):
...
+        )
+"""
+
+
+_MEASUREMENTS_VIEW_SQL = """
+CREATE OR REPLACE VIEW physical_measurements_view AS
+ SELECT
+   p.participant_id participant_id,
+   YEAR(p.sign_up_time) participant_sign_up_year,
+   p.withdrawal_status participant_withdrawal_status,
```
danqrodney: You can remove this and the next line now,  and also from questionnaire_response_answer_view.

calbach: Done, thanks.

calbach:
It appears that our tests make HTTP requests to GitHub. Unsurprisingly this flaked. Rerunning.

PR #582, opened 2018-01-25
Mead/add new codes

Seeker1911:
Added new columns to participant_summary for codes 1PS08 and 1SS08


Seeker1911:
@danqrodney check latest alembic revisions

PR #585, opened 2018-01-30
Mead/add order id biobank reconciliation

Seeker1911:
I'll remove the print statements.


rest-api/model/biobank_stored_sample.py
```diff
@@ -21,6 +21,8 @@ class BiobankStoredSample(Base):
   # The participant the sample is associated to. We use Biobank's ID for streamlined importing.
   biobankId = Column('biobank_id', Integer, ForeignKey('participant.biobank_id'))
 
+  # The biobank order identifier. to enable joining on reconciliation report.
+  biobankOrderIdentifier = Column('value', String(80), nullable=False)
```
danqrodney: Let's call this column biobank_order_identifier. The "value" name for the column in biobank_order_identifier table only makes sense given the table name (and presence of "system" column in the same table.)

This will require you to alter or regenerate your schema migration.

rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -183,6 +191,7 @@ def _create_sample_from_row(row, biobank_id_prefix):
   sample = BiobankStoredSample(
       biobankStoredSampleId=row[_Columns.SAMPLE_ID],
       biobankId=biobank_id,
+      biobankOrderIdentifier=row[_Columns.BIOBANK_ORDER_IDENTIFIER],
```
danqrodney: We should add some validation logic above that either bails or returns None if the order ID is not set... we shouldn't import samples without order IDs moving forward.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_mysql_test.py
```diff
@@ -120,11 +120,13 @@ def _insert_order(self, participant, order_id, tests, order_time, finalized_test
           finalized=finalized_time))
     return self.order_dao.insert(order)
 
-  def _insert_samples(self, participant, tests, sample_ids, confirmed_time, created_time):
+  def _insert_samples(self, participant, tests, sample_ids, identifier, confirmed_time,
```
danqrodney: We should add a bit of test logic in here to confirm that sample order identifiers are actually being set properly.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_mysql_test.py
```diff
@@ -160,7 +162,7 @@ def test_reconciliation_query(self):
     self._insert_order(p_on_time, 'GoodOrder', BIOBANK_TESTS[:4], order_time,
                        finalized_tests=BIOBANK_TESTS[:3], kit_id='kit1', tracking_number='t1',
                        collected_note=u'\u2013foo', processed_note='bar', finalized_note='baz')
-    self._insert_samples(p_on_time, BIOBANK_TESTS[:2], ['GoodSample1', 'GoodSample2'],
+    self._insert_samples(p_on_time, BIOBANK_TESTS[:2], ['GoodSample1', 'GoodSample2'], 'KIT',
```
danqrodney: We should use different values for the order identifier for different samples... for now it doesn't make much difference, but when it comes time to do the reconciliation report it will.

rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -60,6 +60,9 @@ def upsert_from_latest_csv():
   """Finds the latest CSV & updates/inserts BiobankStoredSamples from its rows."""
   bucket_name = config.getSetting(config.BIOBANK_SAMPLES_BUCKET_NAME)  # raises if missing
   csv_file, csv_filename = _open_latest_samples_file(bucket_name)
+  print '--------------------------------------------'
```
danqrodney: Please remove the print statements.

rest-api/offline/biobank_samples_pipeline.py
Seeker1911: @danqrodney Just to be clear: by order_id you mean biobank_ordered_sample.order_id ?

danqrodney: No, I mean this field -- biobank_stored_sample.biobank_order_identifier (after it's renamed).

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_mysql_test.py
```diff
@@ -126,7 +126,7 @@ def _insert_samples(self, participant, tests, sample_ids, identifier, confirmed_
       self.sample_dao.insert(BiobankStoredSample(
           biobankStoredSampleId=sample_id,
           biobankId=participant.biobankId,
-          biobankOrderIdentifier=identifier,
+          biobankOrderIdentifier=identifier + sample_id,
```
danqrodney: This will make every sample have a different order identifier... more realistic will be having the values of identifier passed in vary, sometimes being the same for multiple samples... they should mirror the order IDs created for each participant.

(Again if you want, you can deal with it in the next PR.)

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_mysql_test.py
danqrodney: Ping?

rest-api/offline/biobank_samples_pipeline.py
Seeker1911: done.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_mysql_test.py
Seeker1911: addressed below.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_mysql_test.py
Seeker1911: addressed below.

Seeker1911: done

PR #591, opened 2018-02-05
add pairing with sites and orgs to missing hpo

Seeker1911:


rest-api/dao/participant_dao.py
```diff
@@ -274,10 +274,10 @@ def add_missing_hpo_from_site(self, session, participant_id, site_id):
     participant = self.get_for_update(session, participant_id)
     if participant is None:
       raise BadRequest('No participant %r for HPO ID udpate.' % participant_id)
-    if participant.hpoId == site.hpoId:
```
danqrodney: We should probably replace this with
if participant.siteId == site.siteId:
  return

(The idea being we don't have to mess with anything on the participant or summary if the site hasn't changed.)

rest-api/dao/participant_dao.py
```diff
@@ -274,10 +274,10 @@ def add_missing_hpo_from_site(self, session, participant_id, site_id):
...
-      return
 
     participant.hpoId = site.hpoId
+    participant.organizationId = site.organizationId
```
danqrodney: Please add tests to confirm this behavior, around here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/dao_test/participant_dao_test.py#L260

rest-api/dao/participant_dao.py
Seeker1911: done

rest-api/test/client_test/participant_test.py
```diff
@@ -72,7 +72,7 @@ def test_create_and_modify_participant(self):
           headers = { 'If-Match': last_etag})
 
     self.assertEqual(response['biobankId'], biobank_id)
-    self.assertJsonEquals(response['providerLink'], [provider_link, provider_link_2])
+    # self.assertJsonEquals(response['providerLink'], [provider_link, provider_link_2])
```
danqrodney: Why is this commented out?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -485,14 +485,13 @@ def testQuery_manyParticipants(self):
                            [[ps_1, ps_2], [ps_3]])
       self.assertResponses('ParticipantSummary?_count=2&_sort:desc=organization',
                            [[ps_1, ps_2], [ps_3]])
-      self.assertResponses('ParticipantSummary?_count=2&_sort:asc=site',
```
danqrodney: Why was this removed?

rest-api/dao/participant_dao.py
danqrodney: Ping.

rest-api/dao/participant_dao.py
```diff
@@ -114,7 +114,7 @@ def _do_update(self, session, obj, existing_obj):
       need_new_summary = True
 
     if obj.organizationId or obj.siteId or obj.hpoId:
-      site, organization, awardee = self.get_pairing_level(obj, existing_obj)
+      site, organization, awardee = self.get_pairing_level(obj)
```
danqrodney: This change looks like it will work functionally.

In future, we may want to consider not calling _do_update() from add_missing_hpo_from_site(), and instead calling a simpler function that just messes with site, organization, awardee, and lastModified, since those are the only things that should change. (It can call helper functions to borrow common functionality.)

rest-api/api_util.py
```diff
@@ -122,7 +122,7 @@ def get_site_id_from_google_group(obj, site_dao):
   return None
 
 def get_awardee_id_from_name(obj, hpo_dao):
-  if 'awardee' in obj:
+  if 'awardee' in obj and obj['awardee'] != 'UNSET':
```
danqrodney: So: there is an UNSET hpo row in the database, and hpoId on participant is not nullable... we set hpoId to UNSET (0) if a participant is unpaired.

However, organizationId and siteId are nullable and there is no UNSET organization or site.

So I would have maybe expected get_site_id_from_google_group and get_organization_id_from_external_id to do these UNSET checks... but not necessarily this function.

Also, here and below, use the UNSET constant:

https://github.com/all-of-us/raw-data-repository/blob/master/rdr_common/code_constants.py#L4

rather than putting 'UNSET' in the code.

rest-api/dao/participant_dao.py
```diff
@@ -151,25 +151,21 @@ def _do_update(self, session, obj, existing_obj):
...
-  def get_pairing_level(self, obj, existing_obj):
+  def get_pairing_level(self, obj):
     organization_id = obj.organizationId
     site_id = obj.siteId
     awardee_id = obj.hpoId
-    old_site_id = existing_obj.siteId
-    old_organization_id = existing_obj.organizationId
     # TODO: DO WE WANT TO PREVENT PAIRING IF EXISTING SITE HAS PM/BIO.
 
     if site_id != u'UNSET' and site_id is not None:
```
danqrodney: See above re: 'UNSET' constant

rest-api/dao/participant_dao.py
```diff
@@ -151,25 +151,21 @@ def _do_update(self, session, obj, existing_obj):
...
-    if organization_id != u'UNSET' and organization_id is not None:
-      if organization_id != old_organization_id:
-        organization = self.organization_dao.get(organization_id)
-        awardee_id = organization.hpoId
-        return None, organization_id, awardee_id
+      site = self.site_dao.get(site_id)
+      organization_id = site.organizationId
+      awardee_id = site.hpoId
+      return site_id, organization_id, awardee_id
+    elif organization_id != u'UNSET' and organization_id is not None:
```
danqrodney: See above re: 'UNSET' constant

rest-api/dao/participant_dao.py
```diff
@@ -151,25 +151,21 @@ def _do_update(self, session, obj, existing_obj):
...
-        site = self.site_dao.get(site_id)
-        organization_id = site.organizationId
-        awardee_id = site.hpoId
-        return site_id, organization_id, awardee_id
-    if organization_id != u'UNSET' and organization_id is not None:
-      if organization_id != old_organization_id:
-        organization = self.organization_dao.get(organization_id)
-        awardee_id = organization.hpoId
-        return None, organization_id, awardee_id
+      site = self.site_dao.get(site_id)
```
danqrodney: We should throw BadRequest here if site is None. (Somebody could pass us a site we don't know about.)

rest-api/dao/participant_dao.py
```diff
@@ -151,25 +151,21 @@ def _do_update(self, session, obj, existing_obj):
...
+      organization_id = site.organizationId
+      awardee_id = site.hpoId
+      return site_id, organization_id, awardee_id
+    elif organization_id != u'UNSET' and organization_id is not None:
+      organization = self.organization_dao.get(organization_id)
```
danqrodney: Similarly, we should throw BadRequest here if organization is None.

rest-api/tools/import_organizations.sh
```diff
@@ -35,7 +34,12 @@ function get_geocode_key {
 }
 
 CREDS_ACCOUNT="${ACCOUNT}"
+if [ -z "${ACCOUNT}" ]
+then
+echo "Not Geocoding addresses without --account"
```
danqrodney: You need to change this logic:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/tools/import_organizations.py#L210

So if api_key isn't set we don't try to do any geocoding.

rest-api/dao/participant_dao.py
Seeker1911: done.

rest-api/dao/participant_dao.py
Seeker1911: agreed.

rest-api/dao/participant_dao.py
Seeker1911: done

rest-api/dao/participant_dao.py
Seeker1911: done

rest-api/dao/participant_dao.py
Seeker1911: done

rest-api/dao/participant_dao.py
Seeker1911: done

rest-api/test/client_test/participant_test.py
Seeker1911: fixed

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: un-removed

rest-api/api_util.py
Seeker1911: done

rest-api/tools/import_organizations.sh
Seeker1911: in progress, trouble passing the flag into .py

rest-api/test/unit_test/dao_test/participant_dao_test.py
```diff
@@ -271,6 +271,9 @@ def test_pairs_unset(self):
     self.assertEquals(paired.hpoId, self._test_db.hpo_id)
     self.assertEquals(paired.providerLink, make_primary_provider_link_for_id(self._test_db.hpo_id))
     self.assertEquals(self.participant_summary_dao.get(participant_id).hpoId, self._test_db.hpo_id)
+    self.assertEquals(paired.organizationId, self._test_db.organization_id)
```
danqrodney: It'd be good to have some tests somewhere else in here that check that pairing at the awardee, organization, and site level (via dao.update()) works as intended in terms of setting these fields.

(You could for example copy  test_update_right_expected_version but instead of updating providerLink, update hpoId, organizationId, or siteId.)

rest-api/tools/import_organizations.sh
danqrodney: When dealing with boolean flags like this, the flag you pass in is either '--geocode_flag' for true or '--nogeocode_flag' (or simpler, the absence of --geocode_flag) for false. The way you have it now, you're passing in literally --true or --false, which doesn't make sense.

rest-api/api_util.py
danqrodney: I don't think you addressed the first part of this comment... should we be checking for UNSET in get_site_id_from_google_group and get_organization_id_from_external_id too?

PR #593, opened 2018-02-08
Provider link pairing fix

Seeker1911:
* changed logic in pairing _do_update to make provider link the priority
* fixed participant test to reflect changes in awardee.csv with UNSET row

### DIFF
--- a/rest-api/dao/participant_dao.py
+++ b/rest-api/dao/participant_dao.py
@@ -117,7 +117,7 @@ class ParticipantDao(UpdatableDao):
     # If the provider link changes, update the HPO ID on the participant and its summary.
     if obj.hpoId is None:
       obj.hpoId = existing_obj.hpoId
-    if obj.providerLink != existing_obj.providerLink:
+    if obj.providerLink != existing_obj.providerLink and obj.providerLink != 'null':
       new_hpo_id = self._get_hpo_id(obj)
       if new_hpo_id != existing_obj.hpoId:
         obj.hpoId = new_hpo_id
--- a/rest-api/dao/participant_dao.py
+++ b/rest-api/dao/participant_dao.py
@@ -117,9 +117,9 @@ class ParticipantDao(UpdatableDao):
     # If the provider link changes, update the HPO ID on the participant and its summary.
     if obj.hpoId is None:
       obj.hpoId = existing_obj.hpoId
-    if obj.providerLink != existing_obj.providerLink and obj.providerLink != 'null':
+    if obj.providerLink != existing_obj.providerLink and obj.providerLink is not None:
       new_hpo_id = self._get_hpo_id(obj)
-      if new_hpo_id != existing_obj.hpoId:
+      if new_hpo_id != existing_obj.hpoId and new_hpo_id != 0:
         obj.hpoId = new_hpo_id
         need_new_summary = True
 
--- a/rest-api/dao/participant_dao.py
+++ b/rest-api/dao/participant_dao.py
@@ -114,6 +114,15 @@ class ParticipantDao(UpdatableDao):
                             else None)
       need_new_summary = True
 
+    # If the provider link changes, update the HPO ID on the participant and its summary.
+    if obj.hpoId is None:
+      obj.hpoId = existing_obj.hpoId
+    if obj.providerLink != existing_obj.providerLink:
+      new_hpo_id = self._get_hpo_id(obj)
+      if new_hpo_id != existing_obj.hpoId:
+        obj.hpoId = new_hpo_id
+        need_new_summary = True
+
     if obj.organizationId or obj.siteId or obj.hpoId:
       site, organization, awardee = self.get_pairing_level(obj)
       obj.organizationId = organization
@@ -125,14 +134,7 @@ class ParticipantDao(UpdatableDao):
 
       need_new_summary = True
 
-    # If the provider link changes, update the HPO ID on the participant and its summary.
-    if obj.hpoId is None:
-      obj.hpoId = existing_obj.hpoId
-    if obj.providerLink != existing_obj.providerLink:
-      new_hpo_id = self._get_hpo_id(obj)
-      if new_hpo_id != existing_obj.hpoId:
-        obj.hpoId = new_hpo_id
-        need_new_summary = True
+
     if need_new_summary and existing_obj.participantSummary:
       # Copy the existing participant summary, and mutate the fields that
       # come from participant.
--- a/rest-api/dao/participant_dao.py
+++ b/rest-api/dao/participant_dao.py
@@ -113,7 +113,6 @@ class ParticipantDao(UpdatableDao):
       obj.suspensionTime = (obj.lastModified if obj.suspensionStatus == SuspensionStatus.NO_CONTACT
                             else None)
       need_new_summary = True
-
     update_pairing = True
     if obj.siteId is None and obj.organizationId is None and obj.hpoId is None and \
       obj.providerLink == 'null':
@@ -122,7 +121,7 @@ class ParticipantDao(UpdatableDao):
 
     if update_pairing == True:
       # site,org,or awardee is sent in request: Get relationships and try to set provider link.
-      if obj.organizationId or obj.siteId or (obj.hpoId >= 0) and (obj.providerLink ==
+      if (obj.organizationId or obj.siteId or (obj.hpoId >= 0)) and (obj.providerLink ==
                                                             existing_obj.providerLink or
                                                             obj.providerLink == 'null'):
         site, organization, awardee = self.get_pairing_level(obj)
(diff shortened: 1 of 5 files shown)


rest-api/dao/participant_dao.py
```diff
@@ -114,6 +114,15 @@ def _do_update(self, session, obj, existing_obj):
                             else None)
       need_new_summary = True
 
+    # If the provider link changes, update the HPO ID on the participant and its summary.
+    if obj.hpoId is None:
+      obj.hpoId = existing_obj.hpoId
+    if obj.providerLink != existing_obj.providerLink and obj.providerLink != 'null':
```
danqrodney: I think it should be "obj.providerLink is not None"... not != 'null'.

rest-api/test/client_test/participant_test.py
```diff
@@ -21,24 +20,30 @@ def test_create_and_modify_participant(self):
...
+    self.assertJsonEquals(response['providerLink'], [provider_link_2])
     biobank_id = response['biobankId']
 
     participant_id = response['participantId']
 
     # Fetch that participant.
     response = self.client.request_json('Participant/{}'.format(participant_id))
-    last_etag = self.client.last_etag
+    # Test that hpo is == current provider link
+    self.assertEqual(response['hpoId'], 'PITT')
```
danqrodney: May want to check response['awardee'] here and below as well.

rest-api/test/client_test/participant_test.py
```diff
@@ -57,7 +62,8 @@ def test_create_and_modify_participant(self):
           headers = { 'If-Match': last_etag})
     self.assertEqual(response['biobankId'], biobank_id)
 
-    self.assertJsonEquals(response['providerLink'], [ provider_link_2])
+    self.assertJsonEquals(response['providerLink'], [provider_link_2])
```
danqrodney: It's a little confusing to re-post response above like this... might be clearer to modify providerLink on new_response and send that in the PUT?

rest-api/dao/participant_dao.py
Seeker1911: added a check for hpo_id being 0 as well.

rest-api/test/client_test/participant_test.py
Seeker1911: done.

rest-api/test/client_test/participant_test.py
Seeker1911: changed.

rest-api/dao/participant_dao.py
```diff
@@ -117,9 +117,9 @@ def _do_update(self, session, obj, existing_obj):
     # If the provider link changes, update the HPO ID on the participant and its summary.
     if obj.hpoId is None:
       obj.hpoId = existing_obj.hpoId
-    if obj.providerLink != existing_obj.providerLink and obj.providerLink != 'null':
+    if obj.providerLink != existing_obj.providerLink and obj.providerLink is not None:
       new_hpo_id = self._get_hpo_id(obj)
-      if new_hpo_id != existing_obj.hpoId:
+      if new_hpo_id != existing_obj.hpoId and new_hpo_id != 0:
```
danqrodney: So... if they *did* specify Organization/UNSET for the provider link, we *should* unpair the participant. I'm not sure we want this zero check here.

Either way, please add a test for that, too.

danqrodney: Err. Sorry I take it back. They shouldn't set a provider link in this case. This is fine.

danqrodney:
As discussed offline, let's test that passing up "providerLink": [] or "awardee": "UNSET" should unpair participants at all levels.

Otherwise, LGTM.

rest-api/dao/participant_dao.py
```diff
@@ -114,25 +114,43 @@ def _do_update(self, session, obj, existing_obj):
...
-      site, organization, awardee = self.get_pairing_level(obj)
-      obj.organizationId = organization
-      obj.siteId = site
-      obj.hpoId = awardee
-      if awardee:
-        # get provider link for hpo_id (awardee)
-        obj.providerLink = make_primary_provider_link_for_id(awardee)
+    update_pairing = True
+    if obj.siteId is None and obj.organizationId is None and obj.hpoId is None and \
+      obj.providerLink == 'null':
```
danqrodney: obj.providerLink should not equal 'null'. The fact that it does sometimes is a bug in our code here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_dao.py#L258

If you fix that, this check would just be obj.providerLink is None.

If you want you can fix that in a follow-up... but we shouldn't keep setting providerLink to 'null'.

rest-api/dao/participant_dao.py
```diff
@@ -114,25 +114,43 @@ def _do_update(self, session, obj, existing_obj):
...
+      if obj.providerLink != existing_obj.providerLink:
+        new_hpo_id = self._get_hpo_id(obj)
+        if new_hpo_id != existing_obj.hpoId:
+          obj.hpoId = new_hpo_id
+          if obj.hpoId == 0:
+            obj.siteId = None
+            obj.organizationId = None
+          need_new_summary = True
+
+    # No paring updates sent, keep existing values.
```
danqrodney: paring -> pairing

rest-api/test/client_test/participant_test.py
```diff
@@ -52,12 +60,26 @@ def test_create_and_modify_participant(self):
...
           headers = { 'If-Match': last_etag})
-    self.assertEqual(response['biobankId'], biobank_id)
 
-    self.assertJsonEquals(response['providerLink'], [ provider_link_2])
+    self.assertEqual(new_response['biobankId'], biobank_id)
+    self.assertJsonEquals(new_response['providerLink'], [provider_link])
+    self.assertEqual(new_response['hpoId'], 'AZ_TUCSON')
+    self.assertEqual(new_response['awardee'], 'AZ_TUCSON')
+
+    new_response['providerLink'] = []
```
danqrodney: Can you also add a test somewhere in here that shows that setting 'awardee' to 'UNSET' unpairs?

rest-api/dao/participant_dao.py
```diff
@@ -114,25 +114,43 @@ def _do_update(self, session, obj, existing_obj):
...
+      obj.providerLink == 'null':
+      update_pairing = False
+
+    if update_pairing == True:
+      if obj.organizationId or obj.siteId or obj.hpoId:
+        site, organization, awardee = self.get_pairing_level(obj)
+        obj.organizationId = organization
+        obj.siteId = site
+        obj.hpoId = awardee
+        if awardee is not None and (obj.hpoId != existing_obj.hpoId):
```
danqrodney: So the logic if I understand it is: if awardee changes, have that take priority. If it doesn't, but providerLink changes, have that take priority. Is that right?

We should add comments to that effect in here. It's kind of weird behavior. (Given that we patch existing participants, I can see why you'd do it that way, but it needs documentation.)

rest-api/dao/participant_dao.py
```diff
@@ -114,25 +114,43 @@ def _do_update(self, session, obj, existing_obj):
...
-    if obj.hpoId is None:
+    if update_pairing == True:
+      # If the provider link changes, update the HPO ID on the participant and its summary.
+      if obj.hpoId is None:
+        obj.hpoId = existing_obj.hpoId
+      if obj.providerLink != existing_obj.providerLink:
+        new_hpo_id = self._get_hpo_id(obj)
+        if new_hpo_id != existing_obj.hpoId:
+          obj.hpoId = new_hpo_id
+          if obj.hpoId == 0:
```
danqrodney: 1) We should use UNSET, not 0.

2) I don't think this logic is quite right. If you started out with pairing at the org or site level, and then passed up a providerLink for another awardee, site and organization should be cleared out... but they aren't. This could lead you to having a providerLink and hpoId that doesn't match the organizationId and siteId. (Please add a test for this.) You could remove this if, but if you do that you'll wind up clearing these out when you set providerLink on line 130.

I think the right thing to do is check for providerLink being set / changing before you get to line 122, and if it is, skip lines 122 -> 130.

You could alternatively change get_pairing_level so that it looks at providerLink first, before looking at hpoId / organizationId / siteId. This would mean that if you pass up an old providerLink with other values for hpoId / organizationId / siteId, the latter would be ignored -- but I think that's probably OK. I have told them to stop passing up providerLink when they do pairing with the new system.

rest-api/test/client_test/participant_test.py
Seeker1911: done

rest-api/dao/participant_dao.py
```diff
@@ -114,25 +114,49 @@ def _do_update(self, session, obj, existing_obj):
...
-        obj.providerLink = make_primary_provider_link_for_id(awardee)
+    update_pairing = True
+    if obj.siteId is None and obj.organizationId is None and obj.hpoId is None and \
+      obj.providerLink == 'null':
+      # Prevent unpairing if /PUT is sent with no pairing levels.
+      update_pairing = False
+
+    if update_pairing == True:
+      # site,org,or awardee is sent in request: Get relationships and try to set provider link.
+      if obj.organizationId or obj.siteId or (obj.hpoId >= 0) and (obj.providerLink ==
```
danqrodney: Can you add parentheses to indicate order of operations on the and vs. the ors?

rest-api/dao/participant_dao.py
```diff
@@ -114,25 +114,49 @@ def _do_update(self, session, obj, existing_obj):
...
+      # If the provider link changes, update the HPO ID on the participant and its summary.
+      if obj.hpoId is None:
+        obj.hpoId = existing_obj.hpoId
+      if obj.providerLink != existing_obj.providerLink:
+        # check/set hpoId (covers the chance of sending only provider link and not entering logic
+        # above with hpoId.
+        new_hpo_id = self._get_hpo_id(obj)
+        if new_hpo_id != existing_obj.hpoId:
+          obj.hpoId = new_hpo_id
+          if obj.hpoId == UNSET_HPO_ID:
```
danqrodney: So I think actually this and the next two lines are unnecessary, because obj.siteId and obj.organizationId won't have been set to anything above in this case. Is that right?

If they ever *were* initialized to the existing site ID and organization ID values, we'd need to clear them out when the provider link changes... so I'd say at least this if statement should be removed.

Please add a test for the case where you're paired with a site and the a provider link comes in for another awardee.

rest-api/dao/participant_dao.py
Seeker1911: done. assuming I understood your meaning (which I'm not positive on)

rest-api/dao/participant_dao.py
Seeker1911: done.

PR #594, opened 2018-02-08
Excessively long valueString values

johnstill:
Ups the allowed string length for `questionnaire_response_answer.value_string` from 1024 to 64KiB - 1 byte, the max allowed by MySQL VARCHAR (which SQLAlchemy String should translate to).  Additionally, if a too large string is passed to us, we return a 400 informing the client that the string is too long.

Also added a test case that verifies that 65,535 characters is ok but 65,536 is not.


johnstill:
Ok so apparently we're not using the same version of MySQL in CI?

```
OperationalError: (_mysql_exceptions.OperationalError) (1074, "Column length too big for column 'value_string' (max = 21845); use BLOB or TEXT instead")
```

rest-api/model/__init__.py
```diff
@@ -0,0 +1,2 @@
+# This is 64KiB - 1 byte
+MAX_MYSQL_VARCHAR = 65535
```
danqrodney: Let's move this to questionnaire_response.py, and make it specific to the field in question.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -364,6 +365,9 @@ def _populate_codes_and_answers(cls, group, code_map, answers, link_id_to_questi
                 if answer.valueInteger is not None:
                   qr_answer.valueInteger = answer.valueInteger
                 if answer.valueString:
+                  if len(answer.valueString) > MAX_MYSQL_VARCHAR:
+                    errMsg = 'String value too long (len={}); must be less than {}'
```
danqrodney: Nit: we usually just do 'Foo %d bar %d' % (a, b). It's a little more succinct.

rest-api/model/questionnaire_response.py
```diff
@@ -53,7 +54,7 @@ class QuestionnaireResponseAnswer(Base):
   valueDecimal = Column('value_decimal', Float)
   valueInteger = Column('value_integer', Integer)
   # Is this big enough?
-  valueString = Column('value_string', String(1024))
+  valueString = Column('value_string', String(MAX_MYSQL_VARCHAR))
```
danqrodney: Let's try changing this to Text.

Either way, once you've made a model change like this, follow these steps:

1) From rest-api dir, while server is running, run 'tools/setup_local_database.sh'
2) From client dir, run './run_client.sh generate_fake_data.py --num_participants 10'. this should generate fake data for 10 participants. You can look in questionnaire_response_answer using mysql client to confirm rows with value_string set were inserted.
3) From rest-api dir, run tools/generate_schema.sh. This will make a new schema migration in alembic/versions. Make sure it looks reasonable and git add it.
4) From rest-api dir, run tools/upgrade_schema.sh. This will apply the migration you just created; check the database afterwards with mysql client to confirm the data was preserved but the column is bigger now.

rest-api/test/unit_test/api_test/questionnaire_response_api_test.py
```diff
@@ -19,20 +20,50 @@ def _questionnaire_response_url(participant_id):
...
+    with open(data_path('questionnaire_response3.json')) as fd:
+      resource = json.load(fd)
+
+    resource['subject']['reference'] = \
+        resource['subject']['reference'].format(participant_id=participant_id)
+
+    resource['questionnaire']['reference'] = \
+        resource['questionnaire']['reference'].format(questionnaire_id=questionnaire_id)
+
+    answer = resource['group']['group'][0]['group'][0]['question'][0]['answer'][0]
```
danqrodney: Rather than reading in a questionnaire response file and mutating it in this awkward fashion, let's try using:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/unit_test_util.py#L572

And passing in a single string_answer that's long.

rest-api/test/unit_test/api_test/questionnaire_response_api_test.py
```diff
@@ -19,20 +20,50 @@ def _questionnaire_response_url(participant_id):
...
+    # This one should evaluate to a string that is one char too long; i.e. exactly 64KiB
+    answer['valueString'] = 'a' * (MAX_MYSQL_VARCHAR + 1)
+    self.send_post(url, resource, expected_status=httplib.BAD_REQUEST)
 
+    # Check that a string MAX_MYSQL_VARCHAR will post
+    # This one should be exactly long enough to pass
+    # Remember we need to send the consent first
+    self.send_consent(participant_id)
+    answer['valueString'] = 'a' * MAX_MYSQL_VARCHAR
+    self.send_post(url, resource)
```
danqrodney: We should ideally use QuestionnnaireResponseAnswerDao to make sure the row got written out with the full text here.

johnstill: How have you guys been doing that so far?  Writing to the test DB and then reading back to verify, or mocking / monkey-patching the DAO?

danqrodney: So most API tests involve writing something with the API and reading it back out with the API.

In this particular case, we're trying to read something -- questionnaire_response_answer row -- that isn't exposed via our API. So we have to use QuestionnaireResponseAnswerDao() here.

Maybe it would be simpler / cleaner to not bother with that here but instead add a test here of the DAO directly:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/dao_test/questionnaire_response_dao_test.py

rest-api/alembic/versions/c2dd2332a63f_questionnaire_response_value_string_.py
```diff
@@ -0,0 +1,60 @@
...
+
+def upgrade_rdr():
+    op.alter_column('questionnaire_response_answer',
+                    'value_string',
+                    existing_type=sa.String(1024),
+                    type_=sa.Text(),
+                    existing_nullable=True)
+
+
+def downgrade_rdr():
```
danqrodney: FYI, this is fine -- but we don't actually care about /  ever use / trust downgrade migrations. So it's also OK to leave it empty.

danqrodney:
Oh and as you probably know: the github UI is terrible and hides old commented under "Show outdated". So you have to go hunting for them. If you know a way around this, please tell me... it really sucks. :(

rest-api/test/unit_test/api_test/questionnaire_response_api_test.py
johnstill: Actually the POST is returning the object as written to the db, I believe, so we can read the response from `send_post` and compare (which I've done).  This call circumvents the DAO's `insert` method so I've thrown a quick check on `from_client_json`, which is the top level DAO method called by the API I think.

danqrodney: So it's returning the resource written to the questionnaire_response row, which is good. What it doesn't prove is that we wrote anything out to questionnaire_response_answer. (Though it does prove that we didn't write something out that failed.) So adding something to questionnaire_response_dao_test would be nice.

johnstill: Done.

PR #598, opened 2018-02-20
DA-515 ignore un-recognized tests

Seeker1911:
change `raise` to `log`.

### DIFF
--- a/rest-api/dao/biobank_stored_sample_dao.py
+++ b/rest-api/dao/biobank_stored_sample_dao.py
@@ -1,5 +1,6 @@
 from code_constants import BIOBANK_TESTS_SET
 from dao.base_dao import BaseDao
+from flask import logging
 from model.biobank_stored_sample import BiobankStoredSample
 
 
@@ -21,9 +22,7 @@ class BiobankStoredSampleDao(BaseDao):
     with self.session() as session:
       for sample in sample_generator:
         if sample.test not in BIOBANK_TESTS_SET:
-          raise ValueError(
-              'Sample %r has invalid test code %r.'
-              % (sample.test, sample.biobankStoredSampleId))
+          logging.warn('test sample %s not recognized.' % sample.test)
         session.merge(sample)
         written += 1
     return written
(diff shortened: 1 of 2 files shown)


rest-api/dao/biobank_stored_sample_dao.py
```diff
@@ -21,9 +22,7 @@ def upsert_all(self, sample_generator):
     with self.session() as session:
       for sample in sample_generator:
         if sample.test not in BIOBANK_TESTS_SET:
-          raise ValueError(
-              'Sample %r has invalid test code %r.'
-              % (sample.test, sample.biobankStoredSampleId))
+          logging.warn('test sample %s not recognized.' % sample.test)
```
danqrodney: You need an else here; otherwise we're going to import these.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_test.py
```diff
@@ -137,12 +136,24 @@ def test_sample_from_row_invalid(self):
     with self.assertRaises(biobank_samples_pipeline.DataError):
       biobank_samples_pipeline._create_sample_from_row(row, get_biobank_id_prefix())
 
+  def test_sample_from_row_old_test(self):
```
danqrodney: As noted offline: we need to call _upsert_samples_from_csv() instead. This test doesn't necessarily check anything we care about.

In the interests of time we can merge without a test this time and do a manual test (we're going to want to do that anyway.) But let's get this fixed tomorrow.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_test.py
```diff
@@ -102,10 +102,9 @@ def test_find_latest_csv(self):
     self.assertEquals(latest_filename, '/%s/%s' % (_FAKE_BUCKET, created_last))
 
   def test_sample_from_row(self):
-    samples_file = test_data.open_biobank_samples(111, 222, 333)
+    samples_file = test_data.open_biobank_samples(112, 222, 333)
```
danqrodney: Why did this change?

PR #600, opened 2018-02-21
[DA-517] Push back all RDR crons by 3h

calbach:


calbach:
No real need to change the public metrics cron here, but may as well for consistency (we wanted this to run after the normal metrics when we first checked this FWIW).

PR #601, opened 2018-02-21
Allow for cron-only deployment

calbach:


rest-api/tools/deploy_app.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash -e
 # Checks out RDR code from git in the current directory; by default, uses the same version of the
 # app that is currently running in the staging environment.
-# After a Y/N confirmation, upgrades the database, installs the latest config, deploys the code, or
-# (by default) does all three.
+# After a Y/N confirmation, upgrades the database, installs the latest config, deploys the code
+# and crons, or just the crons. By default, does all of the above.
```
danqrodney: You might want to indicate that the cron and queue configuration go together, in case we ever need to mess with queue configuration.

calbach: Done; also included this in the JIRA ticket update.

Note: this means that you'd be running these updates in the context of an existing release, which might limit the usefulness of this tool.

calbach:
Going to merge since it's not substantially different from the original PR. Tested both app and cron mode on sandbox locally, and they worked as expected.

PR #602, opened 2018-02-22
DA-510 Mead/temp filter inactive sites

Seeker1911:


rest-api/api/base_api.py
```diff
@@ -136,11 +136,13 @@ def _make_bundle(self, results, id_field, participant_id):
     if results.pagination_token:
       query_params = request.args.copy()
       query_params['_token'] = results.pagination_token
+
       next_url = main.api.url_for(self.__class__, _external=True, **query_params)
       bundle_dict['link'] = [{"relation": "next", "url": next_url}]
     entries = []
+    inactive_sites = request.args.get('_inactive')
```
danqrodney: This argument is specific to the awardee API. It doesn't make sense to look for it here. (And it doesn't make sense to have to add inactive_sites to every to_client_json method.)

What we discussed yesterday was:

1) Change the call to to_client_json on 145 to call _make_resource instead.
2) Override _make_resource in awardee_api to call to_client_json passing in inactive sites.
3) Change HPODao to_client_json to have the optional parameter.

Then you can leave the other DAOs alone, and only have to make a one line change to base_api.

rest-api/dao/hpo_dao.py
```diff
@@ -54,19 +54,20 @@ def _make_query(self, session, query_def):
...
             _ORDER_BY_ENDING)
 
-  def to_client_json(self, model):
-    return HPODao._to_json(model)
+  def to_client_json(self, model, inactive_sites=False):
+    return HPODao._to_json(model, inactive_sites)
 
   @staticmethod
-  def _to_json(model):
+  def _to_json(model, inactive_sites=False):
```
danqrodney: You probably don't have to make inactive_sites optional for the to_json methods on HPODao and OrganizationDao; nothing else calls them.

rest-api/dao/organization_dao.py
```diff
@@ -31,9 +31,13 @@ def get_by_external_id(self, external_id):
...
-  def _to_json(model):
+  def _to_json(model, inactive_sites=False):
     resource = _FhirOrganization()
     resource.id = model.externalId
     resource.display_name = model.displayName
-    resource.sites = [SiteDao._to_json(site) for site in model.sites]
+    if inactive_sites:
+      resource.sites = [SiteDao._to_json(site) for site in model.sites]
+    else:
+      resource.sites = [SiteDao._to_json(site) for site in model.sites if site.siteStatus.name ==
```
danqrodney: You could instead check site.siteStatus == SiteStatus.ACTIVE I think.

rest-api/test/client_test/participant_test.py
```diff
@@ -121,13 +121,13 @@ def test_create_and_modify_participant(self):
       'Participant/{}'.format(participant_id), 'PUT', updated_response,
       headers={'If-Match': last_etag})
 
-    updated_response = self.client.request_json('Participant/{}'.format(participant_id))
-    self.assertEqual(updated_response['site'], 'hpo-site-monroeville')
+    updated_response_2 = self.client.request_json('Participant/{}'.format(participant_id))
```
danqrodney: This is fine, but did this have to change to get tests to pass? (I hope not?)

rest-api/test/unit_test/api_test/awardee_api_test.py
```diff
@@ -70,7 +81,8 @@ def test_get_awardee_no_organizations(self):
   def test_get_awardee_with_organizations(self):
     self._setup_data()
     result = self.send_get('Awardee/PITT')
-    self.assertEquals(self._make_expected_pitt_awardee_resource(), result)
+    # we are now filtering out 'INACTIVE' sites by default.
+    self.assertNotEqual(self._make_expected_pitt_awardee_resource(), result)
```
danqrodney: We shouldn't be asserting that these aren't equal... we should be asserting what we actually expect to get returned.

I would make _make_expected_pitt_awardee_resource() take a param indicating whether to include inactive sites.

rest-api/test/unit_test/api_test/awardee_api_test.py
```diff
@@ -48,6 +48,17 @@ def setUp(self):
     hpo_dao.insert(HPO(hpoId=AZ_HPO_ID, name='AZ_TUCSON', displayName='Arizona',
                        organizationType=OrganizationType.HPO))
 
+  def test_active_and_inactive_site(self):
+    self._setup_data()
+    result = self.send_get('Awardee')
```
danqrodney: We should assert the full contents of these responses, not just that the first site is active or inactive. See below.

Seeker1911: mostly done 😉

rest-api/requirements.txt
```diff
@@ -1,27 +1,66 @@
-# cloudstorage_api docs are at
-# cloud.google.com/appengine/docs/standard/python/googlecloudstorageclient/read-write-to-cloud-storage
+alembic==0.9.6
+aniso8601==1.3.0
```
danqrodney: Is this just a reordering? How did this change occur?

You may want to revert this.

rest-api/test/unit_test/api_test/awardee_api_test.py
danqrodney: OK, I'm not seeing that yet. Are your changes committed?

I was expected make_expected_pitt_awardee_resource to start taking a flag...

rest-api/requirements.txt
Seeker1911: done

rest-api/test/unit_test/api_test/awardee_api_test.py
Seeker1911: done

rest-api/dao/base_dao.py
```diff
@@ -328,7 +328,8 @@ def count(self):
     with self.session() as session:
       return session.query(self.model_type).count()
 
-  def to_client_json(self, model):
+  def to_client_json(self, model, inactive_sites=False):
```
danqrodney: I don't think you should have to add this argument here?

rest-api/test/unit_test/api_test/awardee_api_test.py
danqrodney: So lines 54 - 62 here aren't asserting the full contents of the response. Can we instead do like below:

 self.assertEquals(_make_awardee_with_resource(self._make_expected_pitt_awardee_resource(
                                                   inactive=False), 'PITT'),  result);
                                                   result['entry'][1])

rest-api/test/unit_test/api_test/awardee_api_test.py
```diff
@@ -69,10 +82,11 @@ def test_get_awardee_no_organizations(self):
 
   def test_get_awardee_with_organizations(self):
     self._setup_data()
-    result = self.send_get('Awardee/PITT')
-    self.assertEquals(self._make_expected_pitt_awardee_resource(), result)
+    result = self.send_get('Awardee/PITT?_inactive=true')
```
danqrodney: We also need a test for Awardee/PITT not passing in _inactive=true; that should equal _make_expected_pitt_awardee_resource(inactive=False)

rest-api/test/unit_test/api_test/awardee_api_test.py
Seeker1911: done

rest-api/test/unit_test/api_test/awardee_api_test.py
Seeker1911: done

PR #604, opened 2018-02-23
Throttle the logging output during unit test runs

johnstill:
The test output can be a bit spammy with logging messages; this should alleviate that.

### DIFF
--- a/rest-api/test/runner.py
+++ b/rest-api/test/runner.py
@@ -73,6 +73,10 @@ def main(sdk_path, test_path, test_pattern):
 
 
 if __name__ == '__main__':
+  # Disables logging for the duration of the tests
+  import logging
+  logging.disable(logging.CRITICAL)
+
   parser = argparse.ArgumentParser(
     description=__doc__,
     formatter_class=argparse.RawDescriptionHelpFormatter)
--- a/rest-api/README.md
+++ b/rest-api/README.md
@@ -89,6 +89,16 @@ test/run_tests.sh -g $sdk_dir
 
 This will run both the unit tests and the client tests.
 
+If a test breaks and you would like to enable the output of Python's `logging`
+module for debugging purposes, you can comment out the following two lines in
+the file `./test/runner.py` to re-enable printing of log statements to the
+console:
+
+```Python
+import logging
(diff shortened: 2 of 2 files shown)


rest-api/test/runner.py
```diff
@@ -73,6 +73,10 @@ def main(sdk_path, test_path, test_pattern):
 
 
 if __name__ == '__main__':
+  # Disables logging for the duration of the tests
+  import logging
```
danqrodney: So my main worry with this is if a test fails, you're going to have to comment this out in order to debug anything... because otherwise you won't get any log statements. (Is that right?)

If it's well documented that's the thing to do, I guess that's OK. I personally haven't had trouble with our test output thus far though.

johnstill: This shouldn't suppress actual errors, just statements filtered through the `logging` module - so if a test fails, `unittest` should continue to report that accurately, including the relevant traceback.  Worst case you comment out a pair of lines in the runner.

The test output isn't a huge deal, but consider as an example running the tests for `participant_summary_api` - there's ~250 lines of output, only about ~15 of which I care about at all in the normal case.  So tbh its driving me a little nuts that, for example, its easier to find the names of the actual methods that were tested by running a regex on the test output or the original test file than by just looking at it.

That said, I could make this a command line flag if you prefer.

Seeker1911: FWIW. I pull the branch and introduced an error. I provided the appropriate stack trace. This seems to work well, and still lists all unit tests as it runs them. So it is not all output being suppressed.

danqrodney: What i meant was, when a test fails, quite often the way to diagnose what's going on is to look at log statements -- either already in the code or added temporarily for the purposes of debugging.

If log statements are suppressed, we won't get those by default... so we'll at a minimum need to make it well documented how to un-suppress them in our README.

johnstill: > make it well documented

I've updated `rest-api/README.md` in the section under running the unit tests.

danqrodney: Cool. At some point I may add a --debug arg to run_tests.sh to control this, so we don't have to monkey with the code to get this behavior.

PR #605, opened 2018-02-23
Mm/work q api

Seeker1911:
README update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,11 +346,6 @@ lookup (for physical measurements and biospecimen donation) as well as a
 Participant Work Queue. Any of the above parameters can be provided as a query parameter to do
 an exact match. Examples:
 
-    GET /ParticipantSummary?awardee=PITT
-    GET /ParticipantSummary?awardee=PITT&state=PIIState_MA
-
-    GET /ParticipantSummary?organization=PITT_UPMC 
-    GET /ParticipantSummary?site=hpo-site-UPMC
 
 The participant summary API supports filtering and sorting on last modified time.
 The default order results are returned is...
@@ -363,6 +358,9 @@ Example:
 Pagination is provided with a token i.e.
 
     GET /ParticipantSummary?awardee=PITT&_sort=lastModified&_token=<token string>
+    
+It is possible to return the same participant data multiple times in a sync.
+
 See FHIR search prefixes below
 
 Synchronize Participant Summary last modified link.
@@ -379,6 +377,15 @@ Otherwise the array will contain a `link` with relation=`sync` that can be used 
 * Awardees are responsible for rotating keys on a three day timeframe.
     ** Permissions will be revoked after this time.
 * Awardee must specify Awardee or Organization/Site of Awardee in call to API.
+
+    GET /ParticipantSummary?awardee=PITT
+    
+    GET /ParticipantSummary?awardee=PITT&state=PIIState_MA
+    
+    GET /ParticipantSummary?organization=PITT_UPMC 
+    
+    GET /ParticipantSummary?site=hpo-site-UPMC
+
 
 For integer and date fields, the following prefixes can be provided for query parameter values to
 indicate inequality searches, as per the [FHIR search spec](https://www.hl7.org/fhir/search.html):
--- a/README.md
+++ b/README.md
@@ -346,8 +346,24 @@ lookup (for physical measurements and biospecimen donation) as well as a
 Participant Work Queue. Any of the above parameters can be provided as a query parameter to do
 an exact match. Examples:
 
-    GET /ParticipantSummary?hpoId=PITT
-    GET /ParticipantSummary?hpoId=PITT&state=PIIState_MA
+    GET /ParticipantSummary?awardee=PITT
+    GET /ParticipantSummary?awardee=PITT&state=PIIState_MA
+
+    GET /ParticipantSummary?organization=PITT_FOO 
+    GET /ParticipantSummary?site=SITE_HPO
+
+The participant summary API supports filtering and sorting on last modified time
+and returns only the participants for the appropriate awardee to a matching Service Account.
+The priority of filter is...
+* last modified time
+* participant ID (ascending)
+
+    GET /ParticipantSummary?awardee=PITT&last_modified=2018-02-29
+
+##### Service Accounts
+* Each awardee partner is issued one service account.
+* authorized users can generate API keys for access.
+* awardees are responsible for rotating keys on a three day timeframe.
 
 For integer and date fields, the following prefixes can be provided for query parameter values to
 indicate inequality searches, as per the [FHIR search spec](https://www.hl7.org/fhir/search.html):
--- a/README.md
+++ b/README.md
@@ -378,13 +378,13 @@ Otherwise the array will contain a `link` with relation=`sync` that can be used 
     ** Permissions will be revoked after this time.
 * Awardee must specify Awardee or Organization/Site of Awardee in call to API.
 
-    GET /ParticipantSummary?awardee=PITT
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -346,8 +346,24 @@ lookup (for physical measurements and biospecimen donation) as well as a
 Participant Work Queue. Any of the above parameters can be provided as a query parameter to do
 an exact match. Examples:
 
-    GET /ParticipantSummary?hpoId=PITT
-    GET /ParticipantSummary?hpoId=PITT&state=PIIState_MA
+    GET /ParticipantSummary?awardee=PITT
+    GET /ParticipantSummary?awardee=PITT&state=PIIState_MA
+
+    GET /ParticipantSummary?organization=PITT_FOO
```
danqrodney: We should probably use real organization and site values for this... you can pull them from:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/data/organizations.csv
https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/data/sites.csv

README.md
```diff
@@ -346,8 +346,24 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
-    GET /ParticipantSummary?hpoId=PITT
-    GET /ParticipantSummary?hpoId=PITT&state=PIIState_MA
+    GET /ParticipantSummary?awardee=PITT
+    GET /ParticipantSummary?awardee=PITT&state=PIIState_MA
+
+    GET /ParticipantSummary?organization=PITT_FOO 
+    GET /ParticipantSummary?site=SITE_HPO
+
+The participant summary API supports filtering and sorting on last modified time
+and returns only the participants for the appropriate awardee to a matching Service Account.
```
danqrodney: So the way I'm looking at this is filtering and sorting on last_modified is orthogonal to service accounts using this API. Let's remove "and returns...." here.

README.md
```diff
@@ -346,8 +346,24 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
+The priority of filter is...
```
danqrodney: Do you mean the order results are returned?

README.md
```diff
@@ -346,8 +346,24 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
+* last modified time
+* participant ID (ascending)
+
+    GET /ParticipantSummary?awardee=PITT&last_modified=2018-02-29
```
danqrodney: last_modified should be a timestamp -- either 2018-02-29 12:34:56  UTC+0000 -- not a date.

And normal usage of this will be with a "ge" prefix to mean "greater than or equal to", e.g. something like

GET /ParticipantSummary?awardee=PITT&last_modified=ge2018-02-29 12:34:56  UTC+0000 

(with the space escaped).

I think you're also going to want to talk about how to use this with syncing: 

- Call first without passing in anything for last_modified
- On subsequent calls, pass in last_modified=<MAX value of last_modified returned before minus a buffer of lets say 5 minutes>. (Note this means you'll get the same results multiple times if you call in quick succession.)
- Keep syncing until last_modified is within the past few minutes. At that point you'll be more-or-less caught up. Then you can call again every 10 minutes (or whatever sync period they want to use) to get new things.

README.md
```diff
@@ -346,8 +346,24 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
+
+##### Service Accounts
+* Each awardee partner is issued one service account.
+* authorized users can generate API keys for access.
+* awardees are responsible for rotating keys on a three day timeframe.
```
danqrodney: We should indicate that we intend to make them stop working (on our end) after that period elapses.

We should also indicate that awardees must specify their awardee (or an organization or site under it) in the request parameters for awardee, organization, or site when calling this API.

Capitalize "awardees".

README.md
```diff
@@ -346,8 +346,36 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
-    GET /ParticipantSummary?hpoId=PITT
-    GET /ParticipantSummary?hpoId=PITT&state=PIIState_MA
+    GET /ParticipantSummary?awardee=PITT
+    GET /ParticipantSummary?awardee=PITT&state=PIIState_MA
+
+    GET /ParticipantSummary?organization=PITT_UPMC 
+    GET /ParticipantSummary?site=hpo-site-UPMC
+
+The participant summary API supports filtering and sorting on last modified time.
+The order results are returned is...
```
danqrodney: Two things: our parameter / field names are normally camel case... so it's lastModified not last_modified.

Secondly, the way to request sorting in this API is passing a _sort parameter.

Thirdly, we may prefer to have the pagination for this API provided with a _token parameter as is currently supported, which will encode the timestamp as an opaque string... 

If that's the case, then the API would look like:

GET /ParticipantSummary?awardee=PITT&_sort=lastModified
GET /ParticipantSummary?awardee=PITT&_sort=lastModified&_token=abciufeshwg

README.md
```diff
@@ -346,8 +346,36 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
+
+    GET /ParticipantSummary?awardee=PITT&last_modified=ge2018-02-29
+See FHIR search prefixes below
+
+Synchronize Participant Summary last modified link.
+This allows Awardees to stay up-to-date
+with newly-arrived summaries. The return value is a FHIR History [Bundle](http://hl7.org/fhir/bundle.html)
+where each entry is a `ParticipantSummary` document.
+
+The Bundle's `link` array will include a link with relation=`next` if more results are available immediately.
```
danqrodney: You could show an example of what this looks like here.

README.md
```diff
@@ -346,8 +346,36 @@ lookup (for physical measurements and biospecimen donation) as well as a
...
+Otherwise the array will contain a `link` with relation=`sync` that can be used to check for new results.
+
+##### Service Accounts
+* Each awardee partner is issued one service account.
+* Authorized users can generate API keys for access.
+* Awardees are responsible for rotating keys on a three day timeframe.
+    ** Permissions will be revoked after this time.
+* Awardee must specify Awardee or Organization/Site of Awardee in call to API.
```
danqrodney: You might list the example API calls (with awardee, organization, or site parameters) from above here instead.

README.md
danqrodney: Note that the normal way of making pagination tokens in this API is here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/base_dao.py#L164

and the way of parsing them is here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/base_dao.py#L259

However you'll need to handle things a bit differently in this API since the timestamp needs to be adjusted backwards to account for clock skew, etc., and you probably only want the adjusted timestamp in the token (and not also a participant ID.)

danqrodney: We will also need to indicate that the same participant summary (with the same data) can come back multiple times when syncing; the client should handle that gracefully.

README.md
```diff
@@ -785,6 +824,10 @@ organizations and sites within them, representing the full hierarchy of
 awardees > organizations > sites. No pagination, syncing, or filtering is currently
 supported on this endpoint.
 
+Excepts an `_inactive=true` parameter to allow the addition of inactive sites to
```
danqrodney: Excepts -> Accepts.

You may want to say first that by default, only sites with a status of ACTIVE are included in the response.

README.md
danqrodney: Ping.

Seeker1911: done

README.md
Seeker1911: done

README.md
Seeker1911: 😆 done.

PR #607, opened 2018-02-26
rm stale scripts

calbach:
Please confirm these are both now entirely unused; and also let me know whether there's other files I should be removing here (e.g. any of the testdata files?).


danqrodney:
Yes, they shouldn't be used anywhere. I believe you can remove test/test-data/hpos.csv and sites.csv as well.

(We'll need to run tests to confirm.)

calbach:
That seems to have worked.

PR #610, opened 2018-02-27
Js/da 509 pmi skip support

johnstill:


rest-api/offline/metrics_pipeline.py
```diff
@@ -271,8 +271,9 @@ def map_answers(reader):
                              question_code != RACE_QUESTION_CODE):
       race_codes = [code_dao.get_code(PPI_SYSTEM, value) for value in race_code_values]
       race = get_race(race_codes)
-      yield(last_participant_id, make_tuple(last_start_time,
-                                               make_metric(RACE_METRIC, str(race))))
+      if race == Race.SKIPPED:
+        race = PMI_SKIP_CODE
```
danqrodney: I assume Scott and Eric are OK with getting back the PMI skip code for race and the other fields?

eweitz: Yes, that works.  (John and I chatted about this yesterday.)

johnstill:
In `test/unit_test/api_test/participant_summary_api_test.py`, tests gender identity, income, etc are in `test_get_summary_with_skip_codes` and there is a special test for race at `test_get_summary_with_skip_code_for_race` - am I missing any tests that should go in this file?

I will add some tests to `questionnaire_response_dao_test.py` - I take it these should write with the Questionnaire DAO and read with the ParticipantSummary DAO, as we are basically targeting `QuestionnaireResponseDao._update_participant_summary`.

PR #611, opened 2018-02-27
[DA-477] API for participant counts over time

eweitz:
This implements a web API for retrieving participant counts over time, part of Metrics API 2.  Subsequent pull requests will be sent for a service class, tests for that, and tests for the API.


rest-api/api/participant_counts_over_time_api.py
```diff
@@ -0,0 +1,102 @@
...
+from werkzeug.exceptions import BadRequest
+
+from api_util import HEALTHPRO
+from api_util import get_awardee_id_from_name
+from app_util import auth_required
+from dao.hpo_dao import HPODao
+from participant_enums import EnrollmentStatus
+
+DATE_FORMAT = '%Y-%m-%d'
+DAYS_LIMIT = 100  # provisional, per design doc
```
danqrodney: If you prefer we can make this a config setting (that can be changed without pushing new code); but this is fine, too.

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -0,0 +1,102 @@
...
+    # Validate awardees, get ID list
+    awardee_ids = []
+    for awardee in awardees:
+      awardee_id = get_awardee_id_from_name({'awardee': awardee}, self.hpo_dao)
+      if awardee_id == None:
+        raise BadRequest('Invalid awardee name: %s' % awardee)
+      awardee_ids.append(awardee_ids)
+    params['awardee_ids'] = awardee_ids
+
+    valid_enrollment_statuses = EnrollmentStatus.to_dict()
```
danqrodney: 1) Do we need to support multiple enrollment statuses in the filter? 

2) You should probably convert the enrollment statuses to enums here... something like:

try:
  params["enrollment_statuses"] = [EnrollmentStatus(val) for val in enrollment_statuses]
except ValueError:
  raise BadRequest('Invalid enrollment status: %s' % enrollment_statuses)

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -0,0 +1,102 @@
...
+    for enrollment_status in enrollment_statuses:
+      if enrollment_status not in valid_enrollment_statuses:
+        raise BadRequest('Invalid enrollment status: %s' % enrollment_status)
+
+    valid_stratifications = ['TOTAL', 'ENROLLMENT_STATUS', 'GENDER_IDENTITY',
```
danqrodney: We probably should have an enum for this, too? See above for handling.

If not, definitely define constants for these strings somewhere.

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class ParticipantCountsOverTimeApi(Resource):
+
+  @auth_required(HEALTHPRO)
+  def get(self):
+    self.hpo_dao = HPODao()
+
+    # TODO: After enrollment status is filterable,
+    # wire in 'organization', 'site', and 'enrollmentStatus'.
```
danqrodney: Other params in the doc not mentioned here: withdrawalStatus, bucketSize. Do you want to add those?

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -0,0 +1,102 @@
...
+      'stratification': stratification,
+      'start_date': start_date,
+      'end_date': end_date
+    }
+
+    # Most parameters accepted by this API can have multiple, comma-delimited
+    # values.  Arrange them into lists.
+    for param in params:
+      value = params[param]
+      if param in ['start_date', 'end_date', 'stratification']:
```
danqrodney: I think it would be better to list explicitly the params that allow for repeated values, rather than listing the ones that don't here. (Maybe just awardee, organization, site?)

Seeker1911: I think doing this will always try to encode these three values. If it's not provided do you get an error?

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -0,0 +1,102 @@
...
+        params[param] = value.encode()
```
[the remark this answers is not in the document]
eweitz: Nothing necessary, you're right.  I thought I encountered an issue that encoding fixed, but I can't reproduce it, so I've removed the unneeded encoding.

rest-api/api/participant_counts_over_time_api.py
eweitz: I thought I encountered an issue that encoding fixed, but I can't reproduce it, so I've removed the unneeded encoding.

rest-api/api/participant_counts_over_time_api.py
eweitz: > Do we need to support multiple enrollment statuses in the filter?

Yes, I think so.  I can envision cases where users would want to apply filters to _exclude_ participants in one enrollment status, which we'd presumably support for passing all enrollment statuses except that one.

> You should probably convert the enrollment statuses to enums here... something like

Good point, done, thanks.  I was using the approach from [metric_sets_api.py](https://github.com/all-of-us/raw-data-repository/blob/40a6a141c780d232feafd8b1b4d7605e29b44b45/rest-api/api/metric_sets_api.py#L26), but yours is more consistent with other validation in this class.

rest-api/api/participant_counts_over_time_api.py
eweitz: I meant "withdrawalStatus" where I where "enrollmentStatus", and you're right about bucketSize.  Added in comment.

rest-api/api/participant_counts_over_time_api.py
eweitz: Actually, @danqrodney,  if we support multiple enrollment statuses in the filter as I think is needed, then my original `EnrollmentStatus.to_dict()` seems preferable, because the `raise BadRequest('Invalid enrollment status: %s' % enrollment_statuses)` statement outputs the list of specified enrollment status filters, even those only one may be invalid.

rest-api/api/participant_counts_over_time_api.py
eweitz: I'm fine keeping this as is for now, since it's consistent in form with [metrics_api.py](https://github.com/all-of-us/raw-data-repository/blob/306ab26977b2fe2594c0344fdde08bd091557ae2/rest-api/api/metrics_api.py#L12).

rest-api/api/participant_counts_over_time_api.py
eweitz: All filters will need to support multiple values, because the UI will implement them as checkboxes per convention.  And most parameters will be filters, so it seems simpler to list parameters that don't fit the common case.

If you still prefer the explicitness for that longer list, though, let me know and I'll do that.

danqrodney: OK, this is fine.

rest-api/api/participant_counts_over_time_api.py
danqrodney: Meh. You don't normally expect the client to send you bad values; if it does it won't be that hard to work out which one it is. I'd stick with the change you made.

rest-api/api/participant_counts_over_time_api.py
eweitz: Done.

eweitz:
Thanks Dan.  I think my next step will be to implement the services class, so we can verify that our draft SQL will work as expected.  I'll likely need your SQL wizardry if they don't :).

Then I'll implement tests for both the API and service classes.

PR #612, opened 2018-02-27
Mm/fix sorted keys bug

Seeker1911:


Seeker1911:
if circle ci fails on that test then I quit.

PR #617, opened 2018-03-05
Tools fix for Rodete?

danqrodney:
I don't know that this will work on Mac / other environments, but changing this seemed to help get things working for me on the latest version of Goobuntu.

Without the --system flag, the pip install -r totally failed.

### DIFF
--- a/rest-api/tools/setup_env.sh
+++ b/rest-api/tools/setup_env.sh
@@ -11,7 +11,7 @@ find . | grep \.pyc | xargs rm -if $*
 
 echo "Installing libs..."
 # If this fails due to missing mysql_config, try `sudo apt-get install libmysqlclient-dev`.
-pip install -r requirements.txt -t lib/
+pip install --system --no-binary MySQL-python -r requirements.txt -t lib/
 
 # Needed to setup the local DB.
 pip install -r ../rdr_client/requirements.txt


rest-api/tools/setup_env.sh
```diff
@@ -11,7 +11,7 @@ find . | grep \.pyc | xargs rm -if $*
 
 echo "Installing libs..."
 # If this fails due to missing mysql_config, try `sudo apt-get install libmysqlclient-dev`.
-pip install -r requirements.txt -t lib/
+pip install --system --no-binary MySQL-python -r requirements.txt -t lib/
```
calbach: I don't claim to understand this fix, but thanks for figuring this out. If you have any more information about why this is needed, please add to the above comment.

johnstill: What version of `pip` is this?

danqrodney: 9.0.1

johnstill: Are you sure this is the standard `pip`?  I've got pip 9.0.1 but I'm getting `no such option: --system`.  Where is the documentation for that flag?

johnstill: I'm on latest OS X btw, working in a virtual env with pip installed via `pip install --upgrade pip`

danqrodney: I don't know... I think this is an Ubuntu-specific thing.

I won't merge this until we have an OS-specific check for this.

calbach: Yeah I recall also seeing this issue, but currently the flag is working for me. Perhaps pip got downgraded with the OS change. Upgrading pip might make this flag unnecessary.

danqrodney: OK. I tried upgrading pip... it said it was up-to-date. If you know how to upgrade it, please let me know.

calbach: I'm seriously confused on this. I have pip version 9.0.1. Inside a virtualenv I see the flag. Outside of a virtualenv I don't see the flag. I can't really find much of any documentation about --system online. They appear to be pointing to the same binary. @johnstill can you confirm whether or not you see this inside of a venv?

Perhaps because the default is --system outside of a venv; they don't even give the option. I have no idea.

johnstill: @calbach I do _not_ see this inside a virtual env.  We're using the same version number of pip; I strongly suspect this is platform dependent (though I guess it could conceivably be a Google specific extension).

We can do some spelunking at https://github.com/pypa/pip or ask on IRC I guess.

johnstill: I asked around NashDev and looks like this is definitely not part of a standard `pip` install (although it could still be part of an _ancient_ pip; older than version 7 for sure).

Try throwing

```
echo `pip --version`
exit $?
```

At the top of the tools file and let's verify that the script is correctly using the `pip` from inside the virtual env - perhaps Goobuntu runs a modified / antique `pip` and the script is picking up on that for some reason rather than the `pip` in the local env.

johnstill: Or maybe ask one of the Goobuntu distro devs - do you guys have access to them?

danqrodney: I'm fairly certain I have hit this (needing to pass in --system in a venv) on my home desktop, which runs Ubuntu, not Goobuntu... FWIW.

So hopefully adding Ubuntu-specific logic here should work.

calbach: Tried this on rodete today, and it complained about not having a `--system` (which I could have sworn I actually did have above).

danqrodney:
Abandoning this.

PR #619, opened 2018-03-06
[DA-477] Adding a calendar table to the RDR DB for use in the new metrics API.

danqrodney:


eweitz:
This is related to DA-477.  If there's no more specific ticket (and I'm not sure we need one), convention would have us put "[DA-477]" in the start of the PR title.

danqrodney:
Updated PR title to mention DA-477.

@eweitz Not sure what you mean about the commits. When I pushed it showed them affiliated with my Github account... what makes it appear that they aren't to you?

eweitz:
> When I pushed it showed them affiliated with my Github account... what makes it appear that they aren't to you?
The commits in https://github.com/all-of-us/raw-data-repository/pull/619/commits don't seem to be linked to your GitHub account, but as we just chatted about they're linked in master.  I haven't seen that before, but it seems the issue resolves itself.

This looks good to me.

danqrodney:
Going to go ahead and merge to unblock Eric; @calbach, if you have comments I will address in a follow-up.

PR #622, opened 2018-03-12
Js/da 508 add count

johnstill:
~~The `_query` method is retrieving the full result set with every query anyways (as opposed to using a limit and offset) there is no need to issue a separate SQL query to get the full count.  As a separate issue / PR, we may want to revise this approach down the line depending upon if (and how badly) performance degrades by retrieving the full result set as opposed to a targeted subset.~~

~~N.B. - as the `total` key is added to a `Bundle` (cf. https://hl7.org/fhir/2018Jan/bundle-definitions.html#Bundle.total) it appears that _every_ test that checks one bundle against another implicitly tests this property.~~

This PR adds a new query parameter to list endpoints, `_includeTotal`.  If `_includeTotal=true` is sent, we will add a key (`total`) to the response which will reflect the total number of results for the query.


johnstill:
I can do that fairly trivially.  What's the motivation?

jt2k:
My thought was that for some API requests, it doesn't make sense to return a total count.  After looking more closely, I realized the `query` method probably only affect lists of things.  I was worried at first that this would add a total field to every API request, including retrieving single items.  If this is only for lists, then ignore my comment above.

rest-api/dao/base_dao.py
```diff
@@ -157,25 +157,24 @@ def query(self, query_def):
...
-          return Results(items,
-                         self._make_pagination_token(items[len(items) - 1].asdict(), field_names),
-                         False)
-        else:
-          # Items but no pagination token, and no more available
-          return Results(items, None, False)
-    else:
+      # We could use query.count() here but since we're executing and iterating the entire result
+      # set every time anyways it should be unnecessary: len(items) ought to be the same thing
+      total = len(items)
```
danqrodney: That isn't correct. We limit results, based on a _count parameter or DEFAULT_MAX_RESULTS (100), here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/api/base_api.py#L116

That's the whole reason we have pagination tokens, too.

So you'll get at most pageSize + 1 results here. To get a total, you will have to run a separate count(*) query. In order to ensure this doesn't hurt performance in cases where we don't need the total, we should make this logic conditional upon the client asking for the total via some new query param (e.g. "_getTotal"=true)

If the client *does* ask for a total, we should always return it, even when there is no pagination token returned.

rest-api/dao/base_dao.py
```diff
@@ -154,28 +154,28 @@ def _from_json_value(self, prop, value):
...
-        if query_def.always_return_token:
-          # Items and pagination token, but no more available
-          return Results(items,
-                         self._make_pagination_token(items[len(items) - 1].asdict(), field_names),
-                         False)
-        else:
-          # Items but no pagination token, and no more available
-          return Results(items, None, False)
+
+      if not items:
```
danqrodney: Two concerns about this:

1) If query_def.pagination_token is not None, we could be returning a 0 total when there clearly are some results. (We paginated through them.)

2) You're returning 0 here even when query_def.include_total is False... in other cases total is None in that situation.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +250,46 @@ def testQuery_noSummaries(self):
     response = self.send_get('ParticipantSummary')
     self.assertBundle([], response)
 
+  def test_get_summary_list_returns_total(self):
+    num_participants = 200
```
danqrodney: This is going to run for a while I think. I'd make num_participants much smaller and then pass up _count=<something smaller than that> to make it go faster.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +250,46 @@ def testQuery_noSummaries(self):
...
+        'recontactMethod': PMI_SKIP_CODE,
+        'language': PMI_SKIP_CODE,
+        'education': PMI_SKIP_CODE,
+        'income': PMI_SKIP_CODE,
+        'dateOfBirth': datetime.date(1978, 10, 9),
+        'CABoRSignature': 'signature.pdf',
+      }
+      self.post_demographics_questionnaire(participant_id, questionnaire_id, **answers)
+
+    response = self.send_get('ParticipantSummary?_includeTotal=true')
```
danqrodney: We should also add tests for:

- Requests with pagination token and includeTotal set still get totals
- Requests without includeToken set do NOT return counts

danqrodney: And:

* When there are no participants, you get a count of zero
* Where there are fewer participants than the page size, you get the number of participants returned.

johnstill: Added.

rest-api/dao/base_dao.py
johnstill: Good call on both counts - 2 I had expected b/c the length of an empty list is 0, but it _is_ inconsistent. Fixed.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +249,73 @@ def testQuery_noSummaries(self):
...
     self.assertBundle([], response)
 
+  def test_get_summary_list_returns_total(self):
+    page_size = 10
+    num_participants = 20
+    SqlTestBase.setup_codes([PMI_SKIP_CODE], code_type=CodeType.ANSWER)
+    questionnaire_id = self.create_demographics_questionnaire()
+
+    # Prove that no results means a total of zero (if requested)
+    response = self.send_get('ParticipantSummary?_count=10&_includeTotal=true')
```
jt2k: Use `page_size` here (and in place of the other hard-coded 10's below)?

johnstill: Done

johnstill:
Offline approval from Mike; merging.

Seeker1911:
approved.

PR #625, opened 2018-03-15
Mm/work queue api

Seeker1911:


rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +249,50 @@ def testQuery_noSummaries(self):
...
+        'recontactMethod': PMI_SKIP_CODE,
+        'language': PMI_SKIP_CODE,
+        'education': PMI_SKIP_CODE,
+        'income': PMI_SKIP_CODE,
+        'dateOfBirth': datetime.date(1978, 10, 9),
+        'CABoRSignature': 'signature.pdf',
+      }
+      self.post_demographics_questionnaire(participant_id, questionnaire_id, **answers)
+
+    response = self.send_get('ParticipantSummary?_sync=true&_count=10&lastModified=gt%s' % TIME_1)
```
johnstill: I'm not sure this is testing the `_sync` param.  We make 20 people, then get two pages of 10 each - so far we're just using the pagination token, if I understand correctly we have to make some more people and then use the link provided by the final page of the paginated resource.  So we could

  * make 10 people
  * get a page of 20 people - it will not have a next page, so the token will be for syncing
  * make 10 more people
  * use the URL given by the last page and verify that we got the second batch of ten people in return

rest-api/model/participant_summary.py
```diff
@@ -240,4 +240,5 @@ def siteId(cls):
 Index('participant_summary_hpo_num_baseline_samples', ParticipantSummary.hpoId,
       ParticipantSummary.numBaselineSamplesArrived)
 Index('participant_summary_hpo_withdrawal_status_time', ParticipantSummary.hpoId,
-      ParticipantSummary.withdrawalStatus, ParticipantSummary.withdrawalTime)
+      ParticipantSummary.withdrawalStatus, ParticipantSummary.withdrawalTime),
```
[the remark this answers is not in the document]
Seeker1911: dang wrapped lines, DONE.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: continuing to discuss offline, as there is an un-identified bug right now.

rest-api/api/participant_summary_api.py
```diff
@@ -1,16 +1,72 @@
...
       return super(ParticipantSummaryApi, self)._query('participantId')
+
+  def _make_query(self):
+    query = super(ParticipantSummaryApi, self)._make_query()
+    if self._is_last_modified_sync():
+      # roll back last modified time
+      for filters in query.field_filters:
+        if filters.field_name == 'lastModified':
+          # set time delta subtract
+          time_delta = filters.value - datetime.timedelta(seconds=300)
```
danqrodney: Please make 300 here be a config setting.

rest-api/api/participant_summary_api.py
```diff
@@ -1,16 +1,72 @@
...
+
+  def _make_bundle(self, results, id_field, participant_id):
+    if self._is_last_modified_sync():
+      return make_sync_results_for_request(self.dao, results)
+    return super(ParticipantSummaryApi, self)._make_bundle(results, id_field, participant_id)
+
+  def _is_last_modified_sync(self):
+    return request.args.get('_sync') == 'true'
+
+  def _make_pagination_token(self, item_dict, field_names):
```
danqrodney: Where does this get used?

Not sure you should have to mess with the pagination token stuff.

rest-api/config/config_prod.json
```diff
@@ -17,7 +17,11 @@
     },
     "exporter@all-of-us-rdr-prod.iam.gserviceaccount.com": {
       "roles": ["exporter"]
-    }
+    },
+    "awardee-pitt@all-of-us-rdr-prod.iam.gserviceaccount.com": {
```
danqrodney: We may want to hold off on making the config changes in prod until the service accounts have been created. I'm not sure which awardees will be going first.

rest-api/config/config_test.json
```diff
@@ -9,6 +9,11 @@
       "roles": ["healthpro"],
       "comment": "for HealthPro testing"
     },
+    "awardee-dev@pmi-drc-api-test.iam.gserviceaccount.com": {
```
danqrodney: Call this awardee-pitt?

rest-api/config/config_dev.json
```diff
@@ -6,11 +6,15 @@
...
   "user_info": {
     "example@example.com": {
-      "roles": ["ptc", "healthpro", "exporter", "storefront"],
+      "roles": ["awardee_sa", "ptc", "healthpro", "exporter", "storefront"],
       "whitelisted_ip_ranges": {
         "ip6": ["::1/64"],
         "ip4": ["127.0.0.1/32"]
       }
+    },
+    "awardee-dev@pmi-drc-api-test.iam.gserviceaccount.com": {
```
danqrodney: Call this awardee-pitt?

rest-api/model/participant_summary.py
```diff
@@ -241,3 +241,4 @@ def siteId(cls):
       ParticipantSummary.numBaselineSamplesArrived)
 Index('participant_summary_hpo_withdrawal_status_time', ParticipantSummary.hpoId,
       ParticipantSummary.withdrawalStatus, ParticipantSummary.withdrawalTime)
+Index('participant_summary_last_modified', ParticipantSummary.lastModified)
```
danqrodney: I think you need to run generate_schema -- there needs to be an alembic migration generated in order for this to do anything.

We might want to consider instead making this index on awardee, lastModified... expected usage is always syncing data for a single awardee.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,89 @@ def testQuery_noSummaries(self):
     response = self.send_get('ParticipantSummary')
     self.assertBundle([], response)
 
+  def test_last_modified_sync(self):
+    SqlTestBase.setup_codes([PMI_SKIP_CODE], code_type=CodeType.ANSWER)
+    questionnaire_id = self.create_demographics_questionnaire()
+    t1 = TIME_1
+    t2 = TIME_1 + datetime.timedelta(seconds=400)
+
+    def setup_participant(when):
```
danqrodney: I didn't know you could define functions within functions in Python. :)

We don't do this in other places in RDR code... any particular reason you want to?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,89 @@ def testQuery_noSummaries(self):
...
+      sorted([p['resource']['participantId'] for p in response3['entry']]),
+    )
+
+    no_count_url = 'ParticipantSummary?lastModified=lt%s&_sync=true&awardee=PITT' % TIME_5
+    no_count_response = self.send_get(no_count_url)
+    total_count = len(no_count_response['entry'])
+
+    url = 'ParticipantSummary?lastModified=lt%s&_count=10&_sync=true&awardee=PITT' % TIME_5
+    response = self.send_get(url)
+    self.assertEqual(len(response['entry']), total_count//2)
```
danqrodney: Just do 10 here instead of total_count//2?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,89 @@ def testQuery_noSummaries(self):
...
+    next_url = response['link'][0]['url']
+    next_10 = self.send_get(next_url[index:])
+    self.assertEqual(len(next_10['entry']), total_count//2)
```
danqrodney: Just do 10 here?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,89 @@ def testQuery_noSummaries(self):
...
+    self.assertEqual(len(response3['entry']), len(second_batch))
+    # With the same ID's (they're the same participants)
+    self.assertEqual(
+      sorted([p['participantId'] for p in second_batch]),
+      sorted([p['resource']['participantId'] for p in response3['entry']]),
+    )
+
+    no_count_url = 'ParticipantSummary?lastModified=lt%s&_sync=true&awardee=PITT' % TIME_5
+    no_count_response = self.send_get(no_count_url)
+    total_count = len(no_count_response['entry'])
```
danqrodney: Assert total_count is 20?

Normal sync usage will not specify lastModified as a request param at all -- the first request will start from the beginning of time and the subsequent requests will pass it up in the token. So where you test passing up lastModified, you probably don't want to send up _sync, and vice-versa.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,89 @@ def testQuery_noSummaries(self):
...
+    index = sync_url.find('ParticipantSummary')
+    self.assertEqual(response['link'][0]['relation'], 'sync')
+
+    # Verify that the next chunk has no results (everyone is accounted for)
+    response2 = self.send_get(sync_url[index:])
+    self.assertEqual(len(response2['entry']), 0)
+    self.assertEqual(response2['entry'], [])
+
+    # Create a second batch
+    second_batch = [setup_participant(t2) for _ in range(10)]
```
danqrodney: Three things this test isn't checking, but probably should:

1) Participants come back in increasing timestamp order. Right now you have only two timestamps, t1 and t2, so we're not checking ordering within them.

2) The buffer logic in the API should mean that participants modified in the past 5 minutes are returned (even if they were returned in the initial response.) We should verify that the buffer logic is working as intended. (See above -- I don't think it is right now.)

3) Participants only come back for the specified awardee. Right now all the participants are paired with PITT, so we don't see this filtering.

rest-api/api/participant_summary_api.py
```diff
@@ -1,16 +1,72 @@
...
       return super(ParticipantSummaryApi, self).get(p_id)
     else:
+      if auth_awardee:
+        # make sure request has awardee
+        requested_awardee = request.args.get('awardee')
+        if requested_awardee != auth_awardee:
+          raise Forbidden
       return super(ParticipantSummaryApi, self)._query('participantId')
+
+  def _make_query(self):
```
danqrodney: I don't think this will work... with the API as-is, the client isn't going to be passing up lastModified as a request param (which would cause field_filters to be populated)... it will instead pass up a pagination token, which is used directly to filter the SQLAlchemy query here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/base_dao.py#L209

So I think what we need to do is override ParticipantSummaryDao._decode_token() -- that's where the last modified timestamp passed up in the pagination token will need to be tweaked.

If somebody *does* explicitly specify lastModified as a request param, we should use the provided request param value as-is without modification.

Sorry if I got this wrong last week. This code is confusing. :)

rest-api/test/unit_test/api_test/participant_summary_api_test.py
johnstill: So FYI, the `def` and the `class` statements are considered valid almost anywhere a statement can be used in Python.  It isn't uncommon to have several layers of internally defined functions in order to take advantage of closures; in fact, that's the key language feature behind how decorators work (unless you go the `__init__` and `__call__` route, using a class as a kind of closed over function).  In fact we use functions inside functions for our decorators; see for example [`auth_required`](https://github.com/all-of-us/raw-data-repository/blob/23c48350f5310800c4c3f1ec60bba70cd28f74b6/rest-api/app_util.py#L173).

In this case I think MM broke the func out just for code organizational purposes; the func closes over `questionnaire_id` but that could easily be a parameter instead and the function could be moved down to the class level.  I'd only bother if we're going to use it in other places though.  Up to ya'll.

danqrodney: Fine by me to do things this way. Just noting we hadn't done it (much) before... and that I am still a Python n00b. :)

rest-api/api/participant_summary_api.py
```diff
@@ -1,16 +1,72 @@
...
+  @auth_required(PTC_HEALTHPRO_AWARDEE)
   def get(self, p_id=None):
+    auth_awardee = None
+    user_email, user_info = get_validated_user_info()
+    if AWARDEE in user_info['roles']:
+      if user_email == DEV_MAIL:
+        auth_awardee = request.args.get('awardee')
+      else:
+        try:
+          if user_info['awardee']:
```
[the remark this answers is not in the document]
Seeker1911: done, I was thinking of something else when I set that up.

rest-api/api/participant_summary_api.py
```diff
@@ -1,16 +1,72 @@
...
+            auth_awardee = user_info['awardee']
+          elif user_info['organization']:
+            auth_awardee = user_info['organization']
+          else:
+            auth_awardee = user_info['site']
+        except KeyError:
+          raise BadRequest("Must supply awardee for request.")
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/api/participant_summary_api.py
Seeker1911: done

rest-api/api/participant_summary_api.py
Seeker1911: done, and changed to 60.

rest-api/api/participant_summary_api.py
Seeker1911: removed.

rest-api/config/config_dev.json
```diff
@@ -6,11 +6,15 @@
   ],
   "user_info": {
     "example@example.com": {
-      "roles": ["ptc", "healthpro", "exporter", "storefront"],
+      "roles": ["awardee_sa", "ptc", "healthpro", "exporter", "storefront"],
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/config/config_dev.json
Seeker1911: ok

rest-api/config/config_prod.json
Seeker1911: removed

rest-api/config/config_test.json
Seeker1911: sure

rest-api/model/participant_summary.py
Seeker1911: added hpo_id, still need to run migration and push it up

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: What he said.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: 1. done
2. working on it, see line 361
3. not done yet.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done, so much for being clever.

rest-api/model/participant_summary.py
Seeker1911: ok done now

rest-api/api/participant_summary_api.py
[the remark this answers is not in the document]
danqrodney: You don't need the if here.

rest-api/dao/participant_summary_dao.py
```diff
@@ -335,6 +335,15 @@ def to_client_json(self, model):
 
     return result
 
+  def _decode_token(self, query_def, fields):
+    decoded_vals = super(ParticipantSummaryDao, self)._decode_token(query_def, fields)
+    if query_def.always_return_token == True and 'lastModified' in fields:
+      last_modified_index = fields.index['lastModified']
+      decoded_vals[last_modified_index] = decoded_vals[last_modified_index] - datetime.timedelta(
+      seconds=LAST_MODIFIED_BUFFER_SECONDS)
```
danqrodney: Please fix indentation.

rest-api/dao/participant_summary_dao.py
```diff
@@ -335,6 +335,15 @@ def to_client_json(self, model):
 
     return result
 
+  def _decode_token(self, query_def, fields):
+    decoded_vals = super(ParticipantSummaryDao, self)._decode_token(query_def, fields)
+    if query_def.always_return_token == True and 'lastModified' in fields:
+      last_modified_index = fields.index['lastModified']
```
danqrodney: I think this should be fields.index('lastModified')? Does this actually work?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,115 @@ def testQuery_noSummaries(self):
...
     self.assertBundle([], response)
 
+  def test_last_modified_sync(self):
+    SqlTestBase.setup_codes([PMI_SKIP_CODE], code_type=CodeType.ANSWER)
+    questionnaire_id = self.create_demographics_questionnaire()
+    t1 = TIME_1
+    t2 = TIME_1 + datetime.timedelta(seconds=400)
+    t3 = t2 + datetime.timedelta(seconds=30)
+    t4 = t3 + datetime.timedelta(seconds=30)
+    # 5 minute buffer
```
danqrodney: The buffer is 60 seconds, not 5 minutes, FWIW.

rest-api/api/participant_summary_api.py
```diff
@@ -1,16 +1,56 @@
...
+from flask import request
+from werkzeug.exceptions import Forbidden, InternalServerError
 
 
 class ParticipantSummaryApi(BaseApi):
   def __init__(self):
     super(ParticipantSummaryApi, self).__init__(ParticipantSummaryDao())
 
-  @auth_required(PTC_AND_HEALTHPRO)
+  @auth_required(PTC_HEALTHPRO_AWARDEE)
```
danqrodney: This logic is getting complicated enough that at some point we probably will want some unit tests of it, or API tests that use a user other than DEV_MAIL.... we should have tests for the other paths.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,115 @@ def testQuery_noSummaries(self):
...
+    for i in first_batch:
+      first_batch_list.append(i['lastModified'])
+
+    self.assertListEqual(last_modified_list, sorted(first_batch_list))
+    # With the same ID's (they're the same participants)
+    self.assertEqual(
+      sorted([p['participantId'] for p in first_batch]),
+      sorted([p['resource']['participantId'] for p in response['entry']]),
+    )
+    self.assertEqual(sorted([p['resource']['lastModified'] for p in response['entry']]),
```
danqrodney: You could also assert that the first two participants in the response have t1 for lastModified, etc...

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,115 @@ def testQuery_noSummaries(self):
...
+                      [p['resource']['lastModified'] for p in response['entry']])
+
+    # Get the next chunk with the sync url
+    # Verify that this is, in fact, a sync URL - not a next
+    sync_url = response['link'][0]['url']
+    index = sync_url.find('ParticipantSummary')
+    self.assertEqual(response['link'][0]['relation'], 'sync')
+
+    # Verify that the next chunk has no results (everyone is accounted for)
+    response2 = self.send_get(sync_url[index:])
```
danqrodney: This seems like it should return some of the participants you got back in the first response... anybody with a lastModified value within 60 seconds of the timestamp encoded in the token.

Can you add some print statements and debug why that's not happening?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,115 @@ def testQuery_noSummaries(self):
...
+    self.assertEqual(len(response2['entry']), 0)
+    self.assertEqual(response2['entry'], [])
+
+    # Create a second batch
+    second_batch = [setup_participant(t4) for _ in range(10)]
+    response3 = self.send_get(sync_url[index:])
```
danqrodney: See above, seems like you should be getting back some of the same participants as before?

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,115 @@ def testQuery_noSummaries(self):
...
+    no_last_modified_url = 'ParticipantSummary?_sync=true&awardee=PITT'
+    no_lm_response = self.send_get(no_last_modified_url)
+    self.assertEquals(len(no_lm_response['entry']), 20)
+    self.assertEquals(no_lm_response['link'][0]['relation'], 'sync')
+    # ensure same participants are returned before 5 min. buffer
+    sync_url = no_lm_response['link'][0]['url']
+    setup_participant(t4)
+    add_one_more = self.send_get(sync_url[index:])
+    print len(add_one_more['entry'])
+    # self.assertEquals(len(add_one_more['entry']), 21)
```
danqrodney: I would have expected this to return some of the results from before (anything modified in the last 60 seconds) plus the new one.

rest-api/api/participant_summary_api.py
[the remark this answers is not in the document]
Seeker1911: done

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: fixed, per our conversation.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,129 @@ def testQuery_noSummaries(self):
...
+
+    # Get the next chunk with the sync url
+    # Verify that this is, in fact, a sync URL - not a next
+    sync_url = response['link'][0]['url']
+    index = sync_url.find('ParticipantSummary')
+    self.assertEqual(response['link'][0]['relation'], 'sync')
+
+    # Verify that the next sync has results from t2 and t3 (within BUFFER).
+    response2 = self.send_get(sync_url[index:])
+    self.assertEqual(len(response2['entry']), 5)
```
danqrodney: Please assert some timestamps or IDs here, too.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,129 @@ def testQuery_noSummaries(self):
...
+
+    self.assertListEqual(last_modified_list, sorted(first_batch_list))
+    # With the same ID's (they're the same participants)
+    self.assertEqual(
+      sorted([p['participantId'] for p in first_batch]),
+      sorted([p['resource']['participantId'] for p in response['entry']]),
+    )
+    self.assertEqual(sorted([p['resource']['lastModified'] for p in response['entry']]),
+                      [p['resource']['lastModified'] for p in response['entry']])
+    self.assertEqual(response['entry'][0]['resource']['lastModified'], t1.strftime(
```
danqrodney: Another option here is to grab the IDs of the participants with t1, t2, t3, etc. in different lists, and assert the IDs that come back.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -249,6 +252,129 @@ def testQuery_noSummaries(self):
...
+    sync_again = self.send_get(sync_url[index:])
+    self.send_get(sort_by_lastmodified)
+    self.assertEquals(len(sync_again['entry']), 14)
+    one_min_modified = list()
+
+    for i in sync_again['entry']:
+      one_min_modified.append(datetime.datetime.strptime(i['resource']['lastModified'],
+                                                            '%Y''-''%m''-''%d''T''%X'))
+
+    # Everything should be within 60 seconds.
```
danqrodney: You could alternatively assert the exact timestamps / IDs that come back.

(Creating fewer participants may make this simpler.)

rest-api/dao/participant_summary_dao.py
```diff
@@ -335,6 +335,15 @@ def to_client_json(self, model):
 
     return result
 
+  def _decode_token(self, query_def, fields):
+    decoded_vals = super(ParticipantSummaryDao, self)._decode_token(query_def, fields)
+    if query_def.order_by and (query_def.order_by.field_name == 'lastModified' and
```
danqrodney: Let's add a comment describing why this logic is here.

rdr_common/code_constants.py
```diff
@@ -6,9 +6,9 @@
 # these codes are ignored by RDR.
 PPI_EXTRA_SYSTEM = "http://terminology.pmi-ops.org/CodeSystem/ppi-extra"
 
+LAST_MODIFIED_BUFFER_SECONDS = 60
```
danqrodney: Nit: code_constants is for constants used in multiple places in the code. You could probably just define this in participant_summary_dao directly.

rest-api/dao/participant_summary_dao.py
Seeker1911: done.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done, in a super clever way 😄

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,11 +336,15 @@ def to_client_json(self, model):
     return result
 
   def _decode_token(self, query_def, fields):
+    """ If token exists in participant_summary api, decode and use lastModified to add a buffer
+    of 60 seconds. This ensures when a _sync link is used no one is missed. This may return
+    participants more than once if it is called to often, which should be handled on the client
```
danqrodney: I would tweak this: we fully expect that at least one participant (the last one returned) will be repeated in the next sync response, along with any participants modified in the previous 60 seconds. (This will happen whether the client polls every 5 minutes or every hour.)

rest-api/test/unit_test/api_test/participant_summary_api_test.py
danqrodney: Sorry I don't see that change?

rest-api/api/participant_summary_api.py
Seeker1911: agreed, I'll make a ticket.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: I'll add another test for 3

rdr_common/code_constants.py
Seeker1911: defined in config.py

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: Line 403 asserts that every participant object is the same for the last 14 elements of `sort_lm_response` and `sync_again`.

danqrodney: OK. Then arguably everything after line 413 is unnecessary?

PR #626, opened 2018-03-15
[DA-477] Service for participant counts over time

eweitz:
This implements a database service for retrieving participant counts over time, part of Metrics API 2. Subsequent pull requests will be sent for a test for this service, and tests for the [API](https://github.com/all-of-us/raw-data-repository/pull/611).

The design document for Metrics API 2 is in https://docs.google.com/document/d/1SfFXaxrbWtAj1q_miezaZP5asORIMXh2a6QxjSVfysc.  This initial pass intentionally only implements functionality necessary to add support for filtering by enrollment status.

After this initial Metrics API 2 feature is done and incorporated in the PMI Dashboard UI, additional passes will be made to support filters by other dimensions, e.g. gender identity, and other bucket sizes, e.g. 7 days.


rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+    filters_sql = self.get_facets_sql(filters)
+
+    if stratification == 'TOTAL':
+      strata = ['TOTAL']
+      sql = """
+        SELECT
+            SUM(ps_sum.cnt * (ps_sum.day <= calendar.day)) registered_count,
+            calendar.day start_date
+        FROM calendar,
+        (SELECT COUNT(*) cnt, DATE(ps.sign_up_time) day
```
danqrodney: As discussed over Slack: this will only return consented participants. If we want to include non-consented ones, we'll need to use participant instead of participant_summary here.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+
+    start_date = start_date.replace('-', '')
+    end_date = end_date.replace('-', '')
+
+    filters_sql = self.get_facets_sql(filters)
+
+    if stratification == 'TOTAL':
+      strata = ['TOTAL']
+      sql = """
```
danqrodney: Are filters not supported on the TOTAL stratification?

If not, please add validation logic to that effect in validate_params.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+        (SELECT COUNT(*) cnt, DATE(ps.sign_up_time) day
+        FROM participant_summary ps
+        GROUP BY day) ps_sum
+        WHERE calendar.day >= :start_date
+        AND calendar.day <= :end_date
+        GROUP BY calendar.day
+        ORDER BY calendar.day;
+      """
+    elif stratification == 'ENROLLMENT_STATUS':
+      strata = [str(EnrollmentStatus(val)) for val in EnrollmentStatus]
```
danqrodney: I don't think you need the EnrollmentStatus(val) here... just use str(val)?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+             ON c2.day = member.day
+           LEFT OUTER JOIN
+            (SELECT COUNT(*) cnt,
+             DATE(CASE WHEN enrollment_status = 3 THEN
+                   GREATEST(consent_for_electronic_health_records,
+                            questionnaire_on_the_basics_time,
+                            questionnaire_on_lifestyle_time,
+                            questionnaire_on_overall_health_time,
+                            physical_measurements_time,
+                            CASE WHEN sample_status_1ed04_time IS NOT NULL
```
danqrodney: This works OK but hardcodes the baseline questionnaires and the fact that 1ED04 and 1SAL are the samples needed to be a full participant into the SQL. Previously this was done with config:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/config/base_config.json#L29

For the samples, MySQL doesn't make it easy to do the LEAST of N nullable values, so this is probably OK... but at a minimum it means we should add some comments about it in here.

For the baseline questionnaires, we might want to use the config to generate the GREATEST statement instead. Or make sure to document this if we're not going to.

Also, GREATEST and LEAST don't work in SQLite. So to unit test this you're going to need to add macros for them, e.g.:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/database_utils.py#L50

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+                   ELSE NULL END) day
+               FROM participant_summary ps
+              WHERE %(filters)s
+           GROUP BY day) full
+             ON c2.day = full.day) day_sums, calendar
+          WHERE calendar.day >= :start_date
+            AND calendar.day <= :end_date
+          GROUP BY calendar.day
+          ORDER BY calendar.day;
+      """ % {'filters': filters_sql}
```
danqrodney: Add an else that throws BadRequest for other stratifications for now?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+      'enrollment_statuses': 'enrollment_status'
+    }
+
+    for facet in facets:
+      filters_sql = []
+      db_field = facet_map[facet]
+      filters = facets[facet]
+      for q_filter in filters:
+        if str(q_filter) != '':
+          filters_sql.append('ps.' + db_field + ' = ' + str(int(q_filter)))
```
danqrodney: Seems like when there are multiple values for a facet, we should use an in clause instead of many ORs?

If you are going to use ORs, you'll want to put parens around the OR expressions, or order of operations may be wrong.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+      if len(filters_sql) > 0:
+        filters_sql = ' OR '.join(filters_sql)
+        facets_sql.append(filters_sql)
+
+    if len(facets_sql) > 0:
+      facets_sql = ' AND '.join(facets_sql)
+    else:
+      facets_sql = '1 = 1'
```
danqrodney: Seems like we could return None here and just not include facets_sql when it's None in the main method?

I have a pet peeve about 1=1 in my SQL statements. :)

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+         calendar.day
+       FROM
+       (SELECT c2.day cnt_day,
+               registered.cnt registered_cnt,
+               member.cnt member_cnt,
+               full.cnt full_cnt
+            FROM calendar c2
+            LEFT OUTER JOIN
+            (SELECT COUNT(*) cnt, DATE(ps.consent_for_study_enrollment_time) day
+               FROM participant_summary ps
```
danqrodney: Similar to above, if we have to treat non-consented participants as REGISTERED, we'll need to use participant LEFT OUTER JOINed to participant_summary, or participant_view, here.

If we can ignore them in this query, no need to change anything.

rest-api/dao/participant_counts_over_time_service.py
danqrodney: Also: we could consider using bound parameters here instead of inlining the values in the SQL. We do that in other places using this function:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/database_utils.py#L16

This may help the SQL perform slightly better since the execution plan for the query can be cached when the only thing changing are the bound params.

danqrodney:
Another note: at some point (probably not right now) we're going to want to have this API run against our replica DB instance instead of the live one. That way using it won't impair live API serving.

eweitz:
Thanks for the review and the meeting, @danqrodney.  I will circle back to this first thing next week, when my regular cycle resumes for All of Us.

rest-api/dao/participant_counts_over_time_service.py
eweitz: Done.  Now using LEFT OUTER JOIN to include non-consented individuals, per our chat.

rest-api/dao/participant_counts_over_time_service.py
eweitz: Done.

rest-api/dao/participant_counts_over_time_service.py
eweitz: > If you are going to use ORs, you'll want to put parens around the OR expressions, or order of operations may be wrong.

Done, thanks for the catch.

rest-api/dao/participant_counts_over_time_service.py
eweitz: This no longer uses 1=1, and instead inlines the WHERE keyword to this method and returns an empty string when there's no facets_sql.

rest-api/dao/participant_counts_over_time_service.py
eweitz: Filters are supported for TOTAL, and now implemented.  Thanks for flagging that.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,156 @@
...
+    """Queries DB, returns results in format consumed by front-end
+
+    :param start_date: Start date string, e.g. '2018-01-01'
+    :param end_date: End date string, e.g. '2018-01-31'
+    :param filters: Objects representing filters specified in UI
+    :param stratification: How to stratify (layer) results, as in a stacked bar chart
+    :return: Filtered, stratified results by date
+    """
+
+    start_date = start_date.replace('-', '')
```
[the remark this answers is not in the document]
eweitz: Done.

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -77,26 +102,28 @@ def validate_params(self, params):
     if date_diff > DAYS_LIMIT:
       raise BadRequest('Difference between start date and end date ' \
                        'should not be greater than %s days' % DAYS_LIMIT)
-    params['start_date'] = start_date
-    params['end_date'] = end_date
+    params['start_date'] = start_date_str
```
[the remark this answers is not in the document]
eweitz: I've updated this to use a Date object, instead of an ISO 8601 date string.

rest-api/dao/participant_counts_over_time_service.py
eweitz: I've documented those comments in the code.

For the GREATEST and LEAST, my plan to implement those macros in my pull request for unit tests.  (In our early meetings on this, you recommended sending four PRs, latter two for tests of API and service, so my plan is to do the macros in my next PR.)

rest-api/dao/participant_counts_over_time_service.py
danqrodney: I think the SQL for using an IN clause with bound parameters would be simpler than this, FWIW. But this works.

rest-api/api/participant_counts_over_time_api.py
```diff
@@ -53,8 +56,30 @@ def get(self):
...
     params = self.validate_params(params)
 
+    start_date = params['start_date']
+    end_date = params['end_date']
+
+    logging.info('params')
+    logging.info(params)
+    filters = params
+
+    del filters['start_date']
```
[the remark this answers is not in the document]
danqrodney: OK. Once you're passing them as separate arguments to get_filtered_results, you may not need to have them in params / filters?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,179 @@
...
+from .participant_summary_dao import ParticipantSummaryDao
+from model.participant_summary import ParticipantSummary
+from participant_enums import EnrollmentStatus
+
+class ParticipantCountsOverTimeService(ParticipantSummaryDao):
+
+  def __init__(self):
+    super(ParticipantSummaryDao, self).__init__(ParticipantSummary)
+
+  def get_filtered_results(self, start_date, end_date, filters, stratification='ENROLLMENT_STATUS'):
```
danqrodney: You may want to break this method up to handle each stratification with a separate helper method... each query is pretty complicated on its own.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,179 @@
...
+         SUM(full_cnt * (cnt_day <= calendar.day)) full_participants,
+         calendar.day
+       FROM
+       (SELECT c2.day cnt_day,
+               registered.cnt registered_cnt,
+               member.cnt member_cnt,
+               full.cnt full_cnt
+            FROM calendar c2
+            LEFT OUTER JOIN
+            (SELECT COUNT(*) cnt, DATE(ps.consent_for_study_enrollment_time) day
```
danqrodney: I think you want to change this to p.sign_up_time... otherwise you won't be capturing participants that register but don't consent in the registered numbers.

If you do this, the left outer join below is technically unnecessary... you're not pulling anything from participant summary.

rest-api/dao/participant_counts_over_time_service.py
eweitz: Agreed.  I've noted the IN clause refactoring as a possible future task.  For now, given time constraints, since this works and isn't especially complex, I propose we use what's implemented here.

rest-api/dao/participant_counts_over_time_service.py
eweitz: I changed it to p.sign_up_time; thanks for the catch.  But I kept the left outer join, as some filters use participant_summary.

rest-api/dao/participant_counts_over_time_service.py
eweitz: Good idea, done.

rest-api/dao/participant_counts_over_time_service.py
danqrodney: You need to change ps.sign_up_time to p.sign_up_time here?

rest-api/api/participant_counts_over_time_api.py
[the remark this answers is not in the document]
danqrodney: Ping?

(Not a huge deal. Just think it would make the code a little simpler if you didn't include these.)

rest-api/dao/participant_counts_over_time_service.py
danqrodney: Ping?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,202 @@
...
+      # the bound params.
+      for q_filter in filters:
+        if str(q_filter) != '':
+          filters_sql.append('ps.' + db_field + ' = ' + str(int(q_filter)))
+      if len(filters_sql) > 0:
+        filters_sql = '(' + ' OR '.join(filters_sql) + ')'
+        facets_sql.append(filters_sql)
+
+    if len(facets_sql) > 0:
+      facets_sql = 'WHERE ' + ' AND '.join(facets_sql)
```
danqrodney: I think we're always going to want a WHERE clause that filters on withdrawal status and (TEST awardee or example.com email address). See this for the latter:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/offline/metrics_export.py#L41

If you don't want to do that now, please file a JIRA issue tracking it and add a TODO.

rest-api/api/participant_counts_over_time_api.py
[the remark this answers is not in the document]
eweitz: > Once you're passing them as separate arguments to get_filtered_results, you may not need to have them in params / filters?

I adjusted the code to address that in a previous commit, but overlooked noting that here, sorry.  See [how that code looks now](https://github.com/all-of-us/raw-data-repository/blob/f79dbab6d6733abeceb4a5233fa6cedd0edccb1c/rest-api/api/participant_counts_over_time_api.py#L53).

rest-api/dao/participant_counts_over_time_service.py
eweitz: I've implemented filtering for the test awardee and email.  Withdrawal status filtering is somewhat different (and desirable), so I created DA-588 to track it.

rest-api/dao/participant_counts_over_time_service.py
eweitz: [Done](https://github.com/all-of-us/raw-data-repository/blob/cf84290ff49f52defa834a180aa0f138b7361aaa/rest-api/dao/participant_counts_over_time_service.py#L34).

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -19,14 +19,16 @@ def get_filtered_results(self, start_date, end_date, filters, stratification='EN
     :return: Filtered, stratified results by date
     """
 
-    filters_sql = self.get_facets_sql(filters)
+    # Filters for participant_summary (ps) and participant (p) table
+    filters_sql_ps = self.get_facets_sql(filters)
+    filters_sql_p = self.get_facets_sql(filters, table_prefix='p')
```
danqrodney: I find this a little confusing... what fields do we filter on for p vs. ps? Shouldn't we just have one method call that gets the facets SQL for both?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,222 @@
...
+        facets_sql.append(filters_sql)
+
+    if len(facets_sql) > 0:
+      facets_sql = ' AND '.join(facets_sql) + ' AND'
+    else:
+      facets_sql = ''
+
+    facets_sql = 'WHERE ' + facets_sql
+
+    facets_sql += ' %(table_prefix)s.hpo_id != %(test_hpo_id)s ' % {
```
danqrodney: p.hpo_id

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,222 @@
...
+
+    :param facets: Object representing facets and filters to apply to query results
+    :param table_prefix: Either 'ps' (for participant_summary) or 'p' (for participant)
+    :return: SQL for 'WHERE' clause, reflecting filters specified in UI
+    """
+
+    facets_sql = []
+
+    facet_map = {
+      'awardee_ids': 'hpo_id',
```
danqrodney: Make this p.hpo_id

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,222 @@
...
+      'enrollment_statuses': 'enrollment_status'
```
danqrodney: Make this ps.enrollment_status

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,222 @@
...
+    :param filters: Objects representing filters specified in UI
+    :param stratification: How to stratify (layer) results, as in a stacked bar chart
+    :return: Filtered, stratified results by date
+    """
+
+    self.test_hpo_id = HPODao().get_by_name(TEST_HPO_NAME).hpoId
+    self.test_email_pattern = TEST_EMAIL_PATTERN
+
+    # Filters for participant_summary (ps) and participant (p) table
+    filters_sql_ps = self.get_facets_sql(filters)
```
danqrodney: Switch back to just calling this once, drop table_prefix as a parameter, and have the facet map contain the table prefix for each column as mentioned above.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -0,0 +1,222 @@
...
+      'awardee_ids': 'hpo_id',
+      'enrollment_statuses': 'enrollment_status'
+    }
+
+    for facet in facets:
+      filters_sql = []
+      db_field = facet_map[facet]
+      filters = facets[facet]
+
+      if db_field == 'enrollment_status':
```
danqrodney: If you take teh approach suggested above, you can get rid of this.

danqrodney:
As discussed offline: ignore my previous comments. :)

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -105,9 +109,17 @@ def get_facets_sql(self, facets, table_prefix='ps'):
         facets_sql.append(filters_sql)
 
     if len(facets_sql) > 0:
-      facets_sql = 'WHERE ' + ' AND '.join(facets_sql)
+      facets_sql = ' AND '.join(facets_sql) + ' AND'
     else:
-      return ''
+      facets_sql = ''
+
+    facets_sql = 'WHERE ' + facets_sql
```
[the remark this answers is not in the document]
eweitz: Good point.  I used that odd sequencing because I was transforming `facets_sql` from a list to a string.  The revised code doesn't do that and uses two variables instead, simplifying the initialization.

eweitz:
@danqrodney, thanks for all your help with complicated SQL and navigating the RDR codebase, in addition to this thorough code review.  Next up: tests.

PR #627, opened 2018-03-19
Js/da 508 docs

johnstill:
Update the README to reflect the addition of the `_includeTotal` query parameter.  The param should work for any paginated resource, but since (I think) its only officially supported for `ParticipantSummary`, I've added the note to that section of the README.

Also removed a nearby section of the README which was duplicated.


johnstill:
Yeah I'm doing that right now.  I blame Monday.

PR #629, opened 2018-03-19
DA-502 refactor to allow pairing at different levels

Seeker1911:


rdr_client/hpo_assigner.py
```diff
@@ -98,4 +99,5 @@ def _get_old_hpo(participant):
   arg_parser = get_parser()
   arg_parser.add_argument('file', help='file containing the list of HPOs and participant IDs')
   arg_parser.add_argument('--dry_run', action='store_true')
+  arg_parser.add_argument('--pairing', help='set level of pairing at site')
```
danqrodney: It looks like this is required above. If so, you may want to add required=True here. Alternatively, make it default to awardee.

rdr_client/hpo_assigner.py
```diff
@@ -64,10 +65,10 @@ def main(client):
       if hpo == 'UNSET':
         participant['providerLink'] = []
       else:
-        participant['providerLink'] = [{'primary': True,
```
danqrodney: _ORG_PREFIX is now unused and can be removed?

rdr_client/hpo_assigner.py
```diff
@@ -25,11 +24,13 @@ def main(client):
   num_no_change = 0
   num_updates = 0
   num_errors = 0
+
   with open(client.args.file) as csvfile:
     reader = csv.reader(csvfile)
     for line in reader:
       try:
         participant_id, hpo = [v.strip() for v in line]
+        pairing = client.args.pairing
```
danqrodney: We should have validation that pairing is one of awardee, organization, and site here.

rdr_client/hpo_assigner.py
```diff
@@ -2,7 +2,7 @@
 
 Usage:
   ./run_client.sh --project all-of-us-rdr-prod --account $USER@pmi-ops.org \
-      hpo_assigner.py participant_ids_and_hpos.csv [--dry_run]
+      hpo_assigner.py participant_ids_and_hpos.csv --pairing [site|organization|awardee] [--dry_run]
```
danqrodney: Once this is merged, make sure to update the oncall playbook:

https://docs.google.com/document/d/1FkzFyc_neXn4gAQfdaCoe91mogD4frIAOFtXDeDM-sU/edit#heading=h.jyaxpj74c83k

rdr_client/hpo_assigner.py
```diff
@@ -64,10 +65,10 @@ def main(client):
       if hpo == 'UNSET':
```
danqrodney: Seems like we may wan to delete providerLink, awardee, hpoId, site, and organization before the if here. That way we're only passing up what we want to change?

Also, the _get_old_hpo logic has to change to use pairing... we're now not necessarily changing awardees, and we need to show the change in organization or site if appropriate.

rdr_client/hpo_assigner.py
danqrodney: This works OK assuming that we are going to get pairing files that look like 

<PARTICIPANT ID>,<AWARDEE/ORGANIZATION/SITE ID>

and that all of the entries have the same level of hierarchy.

That seems OK to me, but we should first check in with Ashley to make sure that ServiceNow is going to start sending us files that look like that.... and that they will actually change to use this.

danqrodney: (Let's hold off on this PR until we hear back about this.)

Seeker1911: spoke with the team, this technique is good.

rdr_client/hpo_assigner.py
Seeker1911: done

rdr_client/hpo_assigner.py
Seeker1911: not done

rdr_client/hpo_assigner.py
Seeker1911: yup

rdr_client/hpo_assigner.py
Seeker1911: done

rdr_client/hpo_assigner.py
danqrodney: So is ServiceNow going to start sending us data at org / site level?

rdr_client/hpo_assigner.py
```diff
@@ -13,31 +13,33 @@
...
   num_errors = 0
+
   with open(client.args.file) as csvfile:
     reader = csv.reader(csvfile)
     for line in reader:
       try:
-        participant_id, hpo = [v.strip() for v in line]
+        participant_id, new_pairing = [v.strip() for v in line]
+        pairing = client.args.pairing
+        assert pairing in ('site', 'organization', 'awardee')
```
danqrodney: We should do this check once, up front, and bail out if it fails... not on every line.

rdr_client/hpo_assigner.py
```diff
@@ -13,31 +13,33 @@
...
+      except AssertionError:
+        logging.error('Pairing must be one of site|organization|awardee')
```
danqrodney: Rather than asserting and catching AssertionError, just do an if.

rdr_client/hpo_assigner.py
Seeker1911: according to Ashley, yes.

rdr_client/hpo_assigner.py
```diff
@@ -13,31 +15,35 @@
...
+  if client.args.pairing not in pairing_list:
+    sys.exit('Pairing must be one of site|organization|awardee')
+
   with open(client.args.file) as csvfile:
     reader = csv.reader(csvfile)
     for line in reader:
       try:
-        participant_id, hpo = [v.strip() for v in line]
+        participant_id, new_pairing = [v.strip() for v in line]
+        pairing = client.args.pairing
```
danqrodney: You can assign this just once, outside the loop.

Also, rename to "pairing_key"?

rdr_client/hpo_assigner.py
```diff
@@ -1,10 +1,12 @@
-"""Assigns participants with the specified IDs to the test HPO.
+"""Assigns participants with the specified IDs to the organization.
 
 Usage:
   ./run_client.sh --project all-of-us-rdr-prod --account $USER@pmi-ops.org \
-      hpo_assigner.py participant_ids_and_hpos.csv [--dry_run]
+      hpo_assigner.py participant_ids_and_hpos.csv --pairing [site|organization|awardee] [--dry_run]
```
danqrodney: We might want to rename this script "pairing_assigner" or whatever.

rdr_client/hpo_assigner.py
```diff
@@ -54,20 +60,21 @@ def main(client):
...
+        logging.info('%s unchanged (already %s)', participant_id, old_pairing)
         continue
 
-      logging.info('%s %s => %s', participant_id, old_hpo, hpo)
-      if hpo == 'UNSET':
-        participant['providerLink'] = []
+      logging.info('%s %s => %s', participant_id, old_pairing, new_pairing)
+      if new_pairing == 'UNSET':
+        for i in pairing_list:
+          participant[i] = 'UNSET'
```
danqrodney: Is it correct to set site and organization to UNSET? Or do we just null them out?

What about providerLink? Shouldn't we null that out?

rdr_client/hpo_assigner.py
```diff
@@ -54,20 +60,21 @@ def main(client):
...
       else:
-        participant['providerLink'] = [{'primary': True,
-                                        'organization': {'reference': _ORG_PREFIX + hpo}}]
+        participant[pairing] = new_pairing
```
danqrodney: Should we delete providerLink and other pairing list entries?

rdr_client/hpo_assigner.py
```diff
@@ -1,10 +1,12 @@
...
 
 Usage:
   ./run_client.sh --project all-of-us-rdr-prod --account $USER@pmi-ops.org \
-      hpo_assigner.py participant_ids_and_hpos.csv [--dry_run]
+      hpo_assigner.py participant_ids_and_hpos.csv --pairing [site|organization|awardee] [--dry_run]
+Where site = google_group, organization = external_id, awardee = name.
 
-Where the CSV contains lines with P12345678,NEW_HPO_ID like:
+The CSV contains lines with P12345678,NEW_ORGANIZATION like:
+Example awardees:
```
danqrodney: We might want an example for sites, too.

Seeker1911:
Spoke with Ashley, the expected behavior is to clear everything if 'UNSET'.

PR #631, opened 2018-03-21
Mm/client work queue api

Seeker1911:
updating configs and making script to call work queue as service account.

### DIFF
--- a/rdr_client/work_queue.py
+++ b/rdr_client/work_queue.py
@@ -1,0 +1,21 @@
+""" Uses the service account 'awardee-pitt' to get participant summaries for work queue api
+    with an awardee of PITT"""
+
+import logging
+import pprint
+from main_util import configure_logging
+from client import Client
+
+
+def main():
+  client = Client()
+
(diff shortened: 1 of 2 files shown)


rdr_client/work_queue.py
```diff
@@ -0,0 +1,21 @@
+""" Uses the service account 'awardee-pitt' to get participant summaries for work queue api
+    with an awardee of PITT"""
```
danqrodney: You might consider making awardee be flag-configurable. Though for now this is probably fine.

rdr_client/work_queue.py
```diff
@@ -0,0 +1,21 @@
...
+from main_util import configure_logging
+from client import Client
+
+
+def main():
+  client = Client()
+
+  response = client.request_json('ParticipantSummary?_sync=true&_sort=lastModified&awardee=PITT',
+                                 'GET')
+  logging.info(pprint.pformat(response))
```
danqrodney: You may want to send another request with the sync URL that comes back?

danqrodney:
Another thing to test -- pass up an awardee OTHER than PITT, or don't specify an awardee, and verify that the requests fail.

PR #634, opened 2018-03-23
Mm/client work queue api

Seeker1911:


rdr_client/work_queue.py
```diff
@@ -3,16 +3,25 @@
...
+                                 '={}'.format(AWARDEE), 'GET')
 
-  response = client.request_json('ParticipantSummary?_sync=true&_sort=lastModified&awardee=PITT',
-                                 'GET')
+  sync_url = response['link'][0]['url']
+  index = sync_url.find('ParticipantSummary')
+  sync_results = client.request_json(sync_url[index:], 'GET')
   logging.info(pprint.pformat(response))
+  print
+  print '--------------      Getting next batch from sync...        -----------------'
```
danqrodney: 1) I'd have a loop where we sync new results every SYNC_PERIOD seconds, and print the new results each time, either for some fixed number of times or indefinitely until the program is killed. For testing purposes you can make this a few seconds, but the version we check in and share with HPOs should set it to 300 seconds.

2) These print statements happen after the sync actually occurs. You should probably put them beforehand?

rdr_client/work_queue.py
```diff
@@ -3,16 +3,25 @@
...
 import logging
 import pprint
+
 from main_util import configure_logging
 from client import Client
 
 
 def main():
   client = Client()
+  AWARDEE = 'PITT'
```
danqrodney: We may want to add comments in here:

- the purpose of this code -- syncing participant summaries from the RDR
- you should call code like this with credentials for a service account that has been set up for your awardee by AllOfUs
- you expect to get the same participant summaries back in multiple responses (anything modified within 60 seconds of the last summary in the previous response)
- you'll get back relation="next" on the link if there are more results available in the RDR that weren't returned vs. relation="sync" if there were. (Though I think we probably want them to sync every five minutes either way.)
- provide a link to the README describing the participant summary fields and the API generally

Then this code can serve as an example to HPOs for how to integrate with the API.

Seeker1911: done

rdr_client/work_queue.py
Seeker1911: done

rdr_client/work_queue.py
```diff
@@ -1,18 +1,50 @@
-""" Uses the service account 'awardee-pitt' to get participant summaries for work queue api
-    with an awardee of PITT"""
+""" Uses a service account to get and sync participant summaries for work queue api
+    You can only return results for the same awardee that the service account is connected to.
+
+    Run this using ./run_client.sh --project all-of-us-rdr-prod --account [your account]
+    --service_account [service account]@all-of-us-rdr-prod.iam.gserviceaccount.com work_queue.py
```
danqrodney: I would remove @all-of-us-rdr-prod.iam.gserviceaccount.com; that's a part of the service account, and might vary depending on how we do things. (You could give an example here afterwards.)

rdr_client/work_queue.py
```diff
@@ -1,18 +1,50 @@
...
+
+    You should call code like this with credentials for a service account that has been set up for
+    your awardee by AllOfUs.
+    Expect to get the same participant summaries back in multiple responses (anything modified
+    within 60 seconds of the last summary in the previous response).
+    you'll get back relation="next" on the link if there are more results available in the RDR
+    that weren't returned and relation="sync" if there were.
+    Use the _count parameter to adjust this number. The default is 100
```
danqrodney: Let's set _count to 5000 in the requests below, and update this comment. That's what we want them to use.

rdr_client/work_queue.py
```diff
@@ -1,18 +1,50 @@
...
 import logging
 import pprint
+import time
 from main_util import configure_logging
 from client import Client
 
 
 def main():
   client = Client()
+  awardee = 'PITT'
```
danqrodney: Add a comment saying you should adjust this to the awardee ID you want data for.

rdr_client/work_queue.py
Seeker1911: done

rdr_client/work_queue.py
Seeker1911: done

rdr_client/work_queue.py
Seeker1911: done

rdr_client/work_queue.py
```diff
@@ -2,15 +2,15 @@
     You can only return results for the same awardee that the service account is connected to.
 
     Run this using ./run_client.sh --project all-of-us-rdr-prod --account [your account]
-    --service_account [service account]@all-of-us-rdr-prod.iam.gserviceaccount.com work_queue.py
+    --service_account [service account]@email.com work_queue.py
```
danqrodney: You can remove @email.com, too. The service account value is the whole thing, include stuff after the at sign.

Seeker1911: done

PR #635, opened 2018-03-23
DA-531 add git status check

Seeker1911:
simple git status check using --porcelain flag.


ci/deploy.sh
```diff
@@ -49,6 +49,11 @@ fi
 echo "Deploying $VERSION to: $PROJECT_ID"
 
 cd rest-api
+if [[ `git status --porcelain` ]]; then
```
danqrodney: Nice! But move this to deploy_app.sh instead.

PR #637, opened 2018-03-23
DA-531 add git status check

Seeker1911:


rest-api/tools/deploy_app.sh
```diff
@@ -35,6 +35,12 @@ then
   usage
 fi
 
+if [[ `git status --porcelain` ]]; then
```
danqrodney: Just to be extra safe, we may want to repeat this check below, at least before the deploy step (line 160).

That way if any files get modified while the DB or config is being updated, we will bail out before deploying.

rest-api/tools/deploy_app.sh
```diff
@@ -35,6 +35,12 @@ then
...
+  # Changes
+  echo "git status must be clean"
+  EXIT 0
```
danqrodney: You should probably exit with an error code, not zero.

rest-api/tools/deploy_app.sh
danqrodney: Er, sorry. This needs to run after line 107... up here the right version of the code hasn't been checked out yet.

rest-api/tools/deploy_app.sh
```diff
@@ -100,6 +106,12 @@ fi
 echo "${BOLD}Checking out code...${NONE}"
 git checkout $VERSION
 
+if [[ $(git status --porcelain) ]]; then
```
danqrodney: Can we add one more just before gcloud app deploy?

Seeker1911: done

PR #641, opened 2018-03-27
Update sites

calbach:


Seeker1911:
My sites.csv has a bad zip code about half way down. Is that fixed ?

calbach:
@Seeker1911 I'd guess not if you don't see it in the diff. What value? What row?

Seeker1911:
Looks like it’s been fixed in master, never mind.

PR #642, opened 2018-03-27
DA-512 log invalid sites

Seeker1911:


rest-api/tools/csv_importer.py
```diff
@@ -64,6 +64,9 @@ def run(self, filename, dry_run):
             self._insert_entity(entity, existing_map, session, dry_run)
             new_count += 1
 
+    if self.errors:
+      for err in self.errors:
+        logging.warn(err)
     logging.info('Done importing %ss%s: %d skipped, %d new, % d updated, %d not changed',
```
danqrodney: Add the number of errors in this summary message.

rest-api/tools/import_organizations.py
```diff
@@ -217,6 +222,13 @@ def _populate_lat_lng_and_time_zone(self, site, existing_site):
         site.longitude = longitude
         if latitude and longitude:
           site.timeZoneId = self._get_time_zone(latitude, longitude)
+    else:
+      if site.googleGroup not in self.status_exception_list:
```
danqrodney: We also need a warning if the site *had* an address, but it wasn't geocodable.

rest-api/tools/import_organizations.py
```diff
@@ -122,6 +124,9 @@ def __init__(self):
     self.organization_dao = OrganizationDao()
     args = parser.parse_args()
     self.geocode_flag = args.geocode_flag
+    self.errors = list()
```
danqrodney: I'd move self.errors into the base class; and add to it in all importers wherever we currently log a warning.

rest-api/tools/import_organizations.py
```diff
@@ -217,6 +222,13 @@ def _populate_lat_lng_and_time_zone(self, site, existing_site):
         site.longitude = longitude
         if latitude and longitude:
           site.timeZoneId = self._get_time_zone(latitude, longitude)
+    else:
+      if site.googleGroup not in self.status_exception_list:
+        if site.siteStatus == self.ACTIVE:
+          if len(self.errors) == 0:
+            self.errors.append('Active sites must have valid addresses')
+            self.errors.append('The following sites were not geocoded')
```
danqrodney: I'd add a single error message per site. This seems to be adding three errors here?

rest-api/tools/import_organizations.py
```diff
@@ -217,6 +222,13 @@ def _populate_lat_lng_and_time_zone(self, site, existing_site):
...
+          self.errors.append('Site: {}, Group: {}'.format(site.siteName, site.googleGroup))
```
danqrodney: This logic doesn't actually prevent the site with the invalid address and active status from being imported. We should skip this site if this happens.

danqrodney:
Please note changes coming in:

https://github.com/all-of-us/raw-data-repository/pull/645

rest-api/tools/csv_importer.py
Seeker1911: done

rest-api/tools/import_organizations.py

rest-api/tools/import_organizations.py

rest-api/tools/import_organizations.py
Seeker1911: it only did it the first iteration, but I changed the message anyway.

rest-api/tools/import_organizations.py

rest-api/tools/import_organizations.py
```diff
@@ -194,10 +212,17 @@ def _entity_from_row(self, row):
 
   def _update_entity(self, entity, existing_entity, session, dry_run):
     self._populate_lat_lng_and_time_zone(entity, existing_entity)
+    if entity.siteStatus == self.ACTIVE and (entity.latitude == None or entity.longitude == None):
+      self.errors.append('Skipped active site without geocoding: {}'.format(entity.googleGroup))
+      self.skip = True
```
danqrodney: Rather than having a "skip" field that you have to set here and clear elsewhere, can we just have this method returning None indicate we skipped the entity?

rest-api/tools/import_organizations.py
```diff
@@ -194,10 +212,17 @@ def _entity_from_row(self, row):
...
+      return
     return super(SiteImporter, self)._update_entity(entity, existing_entity, session, dry_run)
 
   def _insert_entity(self, entity, existing_map, session, dry_run):
     self._populate_lat_lng_and_time_zone(entity, None)
+    if entity.siteStatus == self.ACTIVE and (entity.latitude == None or entity.longitude == None):
+      self.errors.append('Skipped active site without geocoding: {}'.format(entity.googleGroup))
```
danqrodney: See above.

rest-api/tools/csv_importer.py
```diff
@@ -58,15 +60,21 @@ def run(self, filename, dry_run):
             changed = self._update_entity(entity, existing_entity, session, dry_run)
             if changed:
               updated_count += 1
+            elif self.skip is True:
+              skip_count += 1
+              self.skip = False
             else:
               matched_count += 1
           else:
             self._insert_entity(entity, existing_map, session, dry_run)
```
danqrodney: Need to handle skip case here, too.

Seeker1911: regarding above, the skip is only to add to the skipped_count which is called on _entity_from_row (prior to any geocoding). So that the count will be accurate during the final logging.info

rest-api/tools/import_organizations.py
Seeker1911: changed to return a tuple since we used None for match count.

rest-api/tools/import_organizations.py

rest-api/tools/csv_importer.py
```diff
@@ -55,18 +56,23 @@ def run(self, filename, dry_run):
...
-            changed = self._update_entity(entity, existing_entity, session, dry_run)
+            changed, skipped = self._update_entity(entity, existing_entity, session, dry_run)
             if changed:
               updated_count += 1
+            elif skipped:
+              skip_count += 1
             else:
               matched_count += 1
           else:
             self._insert_entity(entity, existing_map, session, dry_run)
```
danqrodney: You need to handle the result here.

rest-api/tools/import_organizations.py
```diff
@@ -194,10 +212,17 @@ def _entity_from_row(self, row):
...
+      print 'I AM IN THE RIGHT PLACE.'
+      return False, True
     return super(SiteImporter, self)._update_entity(entity, existing_entity, session, dry_run)
 
   def _insert_entity(self, entity, existing_map, session, dry_run):
     self._populate_lat_lng_and_time_zone(entity, None)
+    if entity.siteStatus == self.ACTIVE and (entity.latitude == None or entity.longitude == None):
+      self.errors.append('Skipped active site without geocoding: {}'.format(entity.googleGroup))
+      return False, True
     super(SiteImporter, self)._insert_entity(entity, existing_map, session, dry_run)
```
danqrodney: This isn't returning anything right now. It should return True.

rest-api/tools/import_organizations.py
```diff
@@ -194,10 +212,17 @@ def _entity_from_row(self, row):
...
+      self.errors.append('Skipped active site without geocoding: {}'.format(entity.googleGroup))
+      print 'I AM IN THE RIGHT PLACE.'
+      return False, True
     return super(SiteImporter, self)._update_entity(entity, existing_entity, session, dry_run)
 
   def _insert_entity(self, entity, existing_map, session, dry_run):
     self._populate_lat_lng_and_time_zone(entity, None)
+    if entity.siteStatus == self.ACTIVE and (entity.latitude == None or entity.longitude == None):
+      self.errors.append('Skipped active site without geocoding: {}'.format(entity.googleGroup))
+      return False, True
```
danqrodney: This can just return None / False. There's no need to return two values here.

PR #643, opened 2018-03-27
DA-528 Update to Circle 2.0

stephenyeargin:
The 1.0 endpoints will be sunset in August.

Related: https://precisionmedicineinitiative.atlassian.net/browse/DA-528


stephenyeargin:
Note that this was created (and not modified afterward) from the conversion endpoint they provided in the docs. It does appear some pieces around tag matching and the staging server did not make the conversion.

circle.yml
```diff
@@ -1,51 +0,0 @@
...
-deployment:
-  # Deploy master to the pmi-drc-api-test project whenever the tests pass
-  test:
-    branch: master
-    commands:
-      - ./ci/activate_creds.sh ~/gcloud-credentials.key
-      - ./tools/upgrade_database.sh -i https://pmi-drc-api-test.appspot.com --creds_file ~/gcloud-credentials.key:
-          pwd:
-            rest-api
-      - ./ci/deploy.sh pmi-drc-api-test config/config_test.json ~/gcloud-credentials.key
```
calbach: It appears that `deploy.sh` didn't make the cut here (both here and staging); definitely need that to run in the new config.

stephenyeargin: Yeah, it looks like they want it configured in a completely different way. From the inline comments in the new file:

```
    # Deployment
    # Your existing circle.yml file contains deployment steps.
    # The config translation tool does not support translating deployment steps
    # since deployment in CircleCI 2.0 are better handled through workflows.
    # See the documentation for more information https://circleci.com/docs/2.0/workflows/
```

.circleci/config.yml
```diff
@@ -0,0 +1,93 @@
...
+# than the prescribed lifecycle of 1.0. In general, we recommend using this generated
+# configuration as a reference rather than using it in production, though in most
+# cases it should duplicate the execution of your original 1.0 config.
+version: 2
+jobs:
+  build:
+    working_directory: ~/all-of-us/raw-data-repository
+    parallelism: 1
+    shell: /bin/bash --login
+    # CircleCI 2.0 does not support environment variables that refer to each other the same way as 1.0 did.
```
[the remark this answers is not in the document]
stephenyeargin: Yeah, once this is in a ✨ state, I will go in and drop any comments that aren't directly relevant, and bring back some of the ones from the original file.

stephenyeargin:
Going to take a break for a bit on this as it's not a pressing matter. Current hassle is wrapped up in that the "Workflow" process they are advocating does not share data automatically between "jobs". Rather than copy/paste a bunch of setup code and stretch out how long it takes things, I need to do some reading:

Reference: https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs

stephenyeargin:
@calbach I am running into the same trouble as this user:

https://discuss.circleci.com/t/persist-python-dependencies-pip-pipenv-installs-in-workflows-using-workspaces/20194/2

I re-ran the job and SSH'd into the container to verify that that was what was keeping the test suite from running.

The "build" phase ends up being not all that useful, as the environment configuration has to be repeated at each stage. I predict this is going to make the test suites take longer to run.

stephenyeargin:
Linking this here, as it may be useful to borrow similar fields. The big thing to note is the use of a custom Docker image specific to the project, which to my knowledge isn't something that's in use on the RDR.

danqrodney:
@stephenyeargin Is this PR still usable?

I just saw that CircleCI 1.0 will no longer be supported after August, so we do need to migrate to 2.0 soon...

stephenyeargin:
@danqrodney It will run the test suite, but it won't support the current methods of deploying passing builds based on the base branch. I had thought it would be a somewhat trivial task until @calbach pointed out the auto-conversion doesn't bring those pieces with it.

[This ticket related to creating Docker images](https://precisionmedicineinitiative.atlassian.net/browse/DA-536) is considered a blocker for resolving this the "right way," but we may need a different strategy to get things knocked out before the deadline.

My advice at this point would be to close this PR and start fresh. I'm not the right person to get the Python set up correctly.

PR #645, opened 2018-03-27
[DA-567] Output a readable diff on organization import

calbach:


rest-api/tools/csv_importer.py
```diff
@@ -81,8 +81,21 @@ def _update_entity(self, entity, existing_entity, session, dry_run):
       logging.info('Not updating %s.', new_dict[self.external_id_field])
```
calbach: FYI: I'm also considering dropping this line since it's pretty spammy and IMO not very useful as you can infer skips by looking at the "total_skipped" at the end. Anyone feel strongly about this?

danqrodney:
Please note changes coming in:

https://github.com/all-of-us/raw-data-repository/pull/642

rest-api/tools/csv_importer.py
```diff
@@ -81,8 +81,21 @@ def _update_entity(self, entity, existing_entity, session, dry_run):
...
       return False
     else:
-      logging.info('Updating %s%s: old = %s, new = %s', self.entity_name,
-                   ' (dry run)' if dry_run else '', existing_dict, new_dict)
+      changes = {}
+      keys = set(existing_dict.keys() + new_dict.keys())
```
danqrodney: A few thoughts:

1) We may want a helper function to do this sort of diffing.

2) There may be a simpler way of doing this using sets or a third party library... see:

https://stackoverflow.com/questions/32815640/how-to-get-the-difference-between-two-dictionaries-in-python
http://deepdiff.readthedocs.io/en/latest/

3) Is the unicode encoding actually necessary for the comparison? I would have thought the old != new comparison would work without it...

rest-api/tools/csv_importer.py
danqrodney: Seems fine to me. @Seeker1911, WDYT?

rest-api/tools/csv_importer.py
calbach: 1. Done

2.
I'm aware of all the Python techniques used in those answers - not sure I agree those are fundamentally simpler, just fewer lines of code. Anyways, those solutions are not really isomorphic to what I'm trying to do - the goal here is to produce a readable diff to explain what is going to change per-field when applying the import. I also want to ensure I handle missing keys for either (though I'm not sure that will happen currently).

3.

It works, albeit with a Python warning being logged:

```
/usr/local/google/home/calbach/workspace/bio/raw-data-repository/rest-api/tools/csv_importer.py:80: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
```

The reason I actually added it though is that it's required a few lines down where I produce a formatted string. Apparently calling `str()` on unicode fails by default:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 5: ordinal not in range(128)
```

PR #646, opened 2018-03-29
Handle db disconnect in apps as 503

johnstill:
Adds an error handler for `DBAPIError`s that indicate a failed or invalidated connection which will return them to the requestor as `503`s rather than `500`s.  HPRO and cron both should automatically retry on 503, so this doesn't directly address the issue of dropping a db connection but should have the same end result (retry the operation) so long as no endpoint commits multiple times during its processing (in which case, we could have an operation fail but _not_ fully rollback the db to its state prior to attempting the operation, which is its own problem).

I'm not exactly sure how to test this; suggestions are welcome.


rest-api/app_util.py
```diff
@@ -18,13 +18,25 @@
 _GMT = pytz.timezone('GMT')
 SCOPE = 'https://www.googleapis.com/auth/userinfo.email'
 
+def handle_database_disconnect(err):
+  """Intended to catch DBAPIError's thrown during a request cycle and transform them into 503's.
+  If the DBAPIError does not represent an invalidated connection, reraise the error.
+
+  Usage: app.register_error_handler(DBAPIError, handle_database_disconnect)
+  """
+  if err.connection_invalidated:
```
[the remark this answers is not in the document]
johnstill: I think its logged automatically isn't it?  Or maybe not.  If not, we should be able to just change [this line](https://github.com/all-of-us/raw-data-repository/blob/e27183644b270664adbba94fdbd315a82f3f1bdb/rest-api/main.py#L41) to have a type tuple: 

```
if isinstance(exception, (HTTPException, DBAPIError)):
    # logging, etc
```

I guess it would be safest to just add the error in - if we're double logging we can always remove it, right? And double logging is better than no logging?

danqrodney:
Regarding testing: you can try taking your local MySQL instance down while your server is running, or take Cloud SQL down in sandbox after deploying there. I'm not sure if that will trigger this behavior, but it seems like it might.

I believe this will help get our API clients to retry; it won't help us in cases where the database goes away in the middle of a pipeline (what this issue was originally filed for.) For that, we'd need to add logic in the pipeline which deals specifically with these kinds of errors, particularly in places where we do lots of writes to the database, e.g. here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/offline/biobank_samples_pipeline.py#L145

(We could look at other pipelines as well, but this is by orders of magnitude the place where we do the most DB reads and writes right now, so it's the main thing to worry about.)

danqrodney:
(I'm not sure how we would get our in-memory testing framework to simulate the database being gone, but we could look into that to if we wanted to write a unit test for it.)

johnstill:
> I believe this will help get our API clients to retry; it won't help us in cases where the database goes away in the middle of a pipeline (what this issue was originally filed for.) 

What do you mean?  The error handler is attached to both the main API app _and_ the "offline" app, and according to the docs, google's cron will automatically retry tasks that return 503.  Are you saying that the pipelines are executed as multiple transactions, so we may have data corruption issues?

danqrodney:
So the offline endpoints may get re-invoked if a 503 is returned... and tasks kicked off in a task queue should get re-executed as well until they succeed (or your limit kicks in.) That is all good.

I guess my issue is that some places we do a ton of database operations within a single task / endpoint call, but not transactionally, and retrying everything isn't how we ideally would handle things. So like for biobank_samples_pipeline we do an upsert on every row in the manifest file; ideally if the database goes away in the middle of all that, we'll just retry that upsert.

johnstill:
well I can patch `BiobankStoredSampleDao.upsert_all` to retry pretty easily.  Like we talked about however, we're not likely to get a more general solution without some re-architecting.  I won't know if its possible to patch other places in the pipelines (or if there are other places in the pipeline that would need it) until I've done a much more in-depth audit of where we are vs aren't using transactions in the pipeline.  But I can push up some code now that should help with `upsert_all`.

rest-api/dao/biobank_stored_sample_dao.py
```diff
@@ -13,17 +14,32 @@ def get_id(self, obj):
...
+    """Inserts/updates samples. Raises ValueError for invalid samples and re-attempts in case of a
+    database connection lost."""
     # TODO(DA-230) Scale this to handle full study data.
     # SQLAlchemy does not provide batch upserting; individual session.merge() calls as below may be
     # expensive but cannot be effectively batched, see stackoverflow.com/questions/25955200. If this
     # proves to be a bottleneck, we can switch to generating "INSERT .. ON DUPLICATE KEY UPDATE".
-    written = 0
-    with self.session() as session:
+    tries = 0
+    while tries < 10:
```
danqrodney: Define 10 as a constant?

rest-api/dao/biobank_stored_sample_dao.py
```diff
@@ -13,17 +14,32 @@ def get_id(self, obj):
...
       for sample in sample_generator:
         if sample.test not in BIOBANK_TESTS_SET:
           logging.warn('test sample %s not recognized.' % sample.test)
         else:
           session.merge(sample)
           written += 1
-    return written
+      try:
+        session.commit()
+      except DBAPIError as exc:
```
danqrodney: We could consider having a function that does the loop, try, catch, etc. and calls a function passed in as a parameter to perform the actual logic; or you could use a decorator for this purpose. That would make it easier to use this sort of logic elsewhere.

rest-api/dao/biobank_stored_sample_dao.py
johnstill: Done

rest-api/dao/biobank_stored_sample_dao.py
johnstill: This is a good idea but I think it merits further thought.  Making a function that wraps session actually shouldn't be that hard (and if we use something like `tenacity` making it retry shouldn't be that hard either), the trickier question is knowing all the places we need to rewrite to use that function.  If we wanted to use it at the DAO level, for example, there are 8 places in `BaseDao` where `with self.session()` is used, and all 8 would need to be converted from using the context manager to using a function that takes a function.  I vote that we leave this Jira ticket open but merge this PR and make further work subject to a new PR.

danqrodney: Yeah, I wasn't necessarily suggesting we actually use this in other places yet... just that we write the code we can use in future whenever we spot another place where we'd like to retry.

danqrodney: A couple places that we may want to consider reusing this kind of logic would be when writing metrics:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/offline/metrics_pipeline.py#L631
https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/offline/public_metrics_export.py#L227

rest-api/model/database.py
```diff
@@ -67,6 +71,30 @@ def session(self):
...
+  @backoff.on_exception(backoff.expo,
+                        DBAPIError,
+                        max_tries=RETRY_CONNECTION_LIMIT,
+                        giveup=lambda err: not err.connection_invalidated)
+  def autoretry(self, func):
+    """Runs a function of the db session and attempts to commit.  If we encounter a dropped
+    connection, we retry the operation.  The retries are spaced out using exponential backoff with
+    full jitter.  All other errors are propagated.
+    """
+    session = self.make_session()
```
danqrodney: can we not just do:

with self.make_session() as session:
  return func(session)

like we do everywhere else? (This handles rollback, etc. for you due to:

https://github.com/all-of-us/raw-data-repository/blob/618506460889d60f4c4c23902bc3b5130160e8a8/rest-api/model/database.py#L59

johnstill: We actually can, since `self.session()` always re-raises whatever exc it hits.  Nice catch!

johnstill:
Discussed offline with Dan; the decorator approach is more complicated and introduces some magic to the decorated function's signature, so we're not going with it at this time.  This branch is ready for merge.

PR #647, opened 2018-03-29
Set a retry limit on the push queues

johnstill:
Per [the source](https://cloud.google.com/appengine/docs/standard/python/refdocs/modules/google/appengine/ext/deferred/deferred#defer), the deferred library does indeed use a default queue called `default`.  Therefore, adding a retry limit on all the listed queues does in fact limit every one of our task queues.

There are other options for retrying push tasks - I don't know if we would want any of those or if 10 is the right number of times to retry.  This much is intended only to prevent requiring manual intervention to stop a failing task from retrying.  More complex configuration should be informed by our actual usage metrics, but if any of the reviewers have sane defaults in mind I'm listening.


danqrodney:
If you want to write a unit test for it, you can see how we're using self.taskqueue in our current unit tests, e.g. here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/offline_test/metrics_export_test.py#L280

https://github.com/all-of-us/raw-data-repository/blob/7b13e651f5bc5f789fe08659818f6414d275eb28/rest-api/test/unit_test/offline_test/table_exporter_test.py#L18

johnstill:
For posterity - turns out there isn't a way to directly check the taskqueue after all since the stub is not executing anything (as is natural for a stub).  No unit test included, merging.

PR #650, opened 2018-04-02
DA-475 create random passwords and store in NDB

Seeker1911:


rest-api/tools/auth_setup.sh
```diff
@@ -88,10 +88,24 @@ function get_instance_connection_name {
 }
 
 function get_db_password {
+  user=$1
   echo "Getting database password..."
   tools/install_config.sh --key db_config --instance $INSTANCE \
       --creds_file ${CREDS_FILE} --config_output $TMP_DB_INFO_FILE
-  PASSWORD=`grep db_password $TMP_DB_INFO_FILE | cut -d\" -f4`
+  if [ "${user}" == "${RDR_DB_USER}" ]
```
danqrodney: I think it'd be a bit cleaner to set a variable to db_password, read_only_db_password, or root_db_password in here, and have the actual PASSWORD=`...` command happen outside the if.

rest-api/tools/auth_setup.sh
```diff
@@ -88,10 +88,24 @@ function get_instance_connection_name {
...
+  then
+      PASSWORD=`grep db_password $TMP_DB_INFO_FILE | cut -d\" -f4`
+  elif [ "${user}" == "${READONLY_DB_USER}" ]
+  then
+      PASSWORD=`grep read_only_db_password $TMP_DB_INFO_FILE | cut -d\" -f4`
+  elif [ "${user}" == "${ROOT_DB_USER}" ]
+  then
+      PASSWORD=`grep root_db_password $TMP_DB_INFO_FILE | cut -d\" -f4`
+  echo "This is a test --------------------------------------------------"
```
danqrodney: Remove echo statements.

rest-api/tools/auth_setup.sh
```diff
@@ -88,10 +88,24 @@ function get_instance_connection_name {
...
+  echo "user is $user"
+  echo "PASSWORD IS $PASSWORD"
+  echo " END TEST ======================================================="
+  fi
```
danqrodney: Add a else statement that prints an error and exits?

rest-api/tools/auth_setup.sh
```diff
@@ -88,10 +88,24 @@ function get_instance_connection_name {
...
 }
```
danqrodney: We may want to add a check here to make sure PASSWORD is set to something after this runs, and print an error and exit if it isn't. (Particularly since existing db_config entries won't have the new root and readonly password fields.)

rest-api/tools/setup_database.sh
```diff
@@ -50,26 +50,22 @@ fi
 
 source tools/setup_vars.sh
 
+# Get a randomly generated password
+function randpw {
+    new_password=$(< /dev/urandom LC_CTYPE=C tr -dc _A-Z-a-z-0-9_\!\@\#\$\%\^\&\*\(\)-+= | head -c${1:-16};echo;)
```
danqrodney: We need to make sure all these characters are OK when used both in SQL and in our DB config. I'd be worried in particular about the slashes.

I don't think we necessarily need all the punctuation symbols in here. A long password of letters and numbers ought to be OK?

rest-api/tools/setup_database.sh
```diff
@@ -50,26 +50,22 @@ fi
...
 # Prompts the user for a database user password.
 function get_required_password {
   db_user_name=$1
-  echo -n "Store and share this password in Valentine with"
-  echo " name=\"$db_user_name\" and description=\"$PROJECT database\"."
-  read -s -p "$db_user_name password for database: " new_password
+  echo "Setting random password for user: $db_user_name "
+  randpw
   echo
   if [ -z "${new_password}" ]
```
danqrodney: I don't think you need get_required_password anymore. Just call randpw directly below?

rest-api/tools/auth_setup.sh
```diff
@@ -88,10 +88,40 @@ function get_instance_connection_name {
...
 
 function get_db_password {
+  user=$1
   echo "Getting database password..."
   tools/install_config.sh --key db_config --instance $INSTANCE \
       --creds_file ${CREDS_FILE} --config_output $TMP_DB_INFO_FILE
-  PASSWORD=`grep db_password $TMP_DB_INFO_FILE | cut -d\" -f4`
+  if [ "$user" == "$RDR_DB_USER" ]
```
danqrodney: You've got two if statements here with repeated ifs and complicated grep logic repeated... this is more complicated than what you had before...

Why not just:

if [ "$user" == "$RDR_DB_USER" ]
then
  PASSWORD_KEY = "db_password"
elif [ "$user" == $READONLY_DB_USER"]
then
  PASSWORD_KEY = "read_only_db_password"
...
PASSWORD =$(grep $PASSWORD_KEY...)

rest-api/tools/setup_database.sh
```diff
@@ -50,33 +50,16 @@ fi
...
-  echo
-  if [ "${repeat_new_password}" != "${new_password}" ]
-  then
-    echo "Password mismatch; exiting."
-    exit 1
-  fi
-}
+# Get a randomly generated password
+function randpw {
+    new_password=$(< /dev/urandom LC_CTYPE=C tr -dc _A-Z-a-z-0-9 | head -c${1:-16};echo;)
```
danqrodney: Looks like you've got hyphen in there twice?

rest-api/tools/auth_setup.sh
Seeker1911: ah, misunderstood your previous comment. Done.

rest-api/tools/auth_setup.sh
Seeker1911: done

rest-api/tools/auth_setup.sh
Seeker1911: done

rest-api/tools/setup_database.sh
Seeker1911: tested, it could create 2 hyphens. removed.

rest-api/tools/auth_setup.sh
```diff
@@ -88,10 +88,28 @@ function get_instance_connection_name {
...
+  then
+      PASSWORD_KEY="rdr_db_password"
+  elif [ "$user" == "$READONLY_DB_USER" ]
+  then
+      PASSWORD_KEY="read_only_db_password"
+  elif [ "$user" == "$ROOT_DB_USER" ]
+  then
+      PASSWORD_KEY="root_db_password"
+  fi
```
danqrodney: else error and exit?

PR #651, opened 2018-04-02
Js/da 321

johnstill:
Closes DA-321 - use the SQLAlchemy >= 1.2 `create_engine` kwarg `pool_pre_ping` rather than the legacy implementation (listening to an event).


rest-api/model/database.py
```diff
@@ -38,7 +37,6 @@ def __init__(self, url, **kwargs):
     # connections after this period. (See DA-237.) To change the db wait_timeout (seconds), run:
     # gcloud --project <proj> sql instances patch rdrmaindb --database-flags wait_timeout=28800
     self._engine = create_engine(url, pool_recycle=3600, **kwargs)
```
danqrodney: Don't you need to add pool_pre_ping=True?

johnstill: _Monday strikes again_

Yes. Yes you do.

PR #653, opened 2018-04-03
Js/da 274 part b

johnstill:
Removes a bunch of dead code and simplifies the configuration implementation / interface.

### DIFF
--- a/ci/license_exceptions.txt
+++ b/ci/license_exceptions.txt
@@ -11,3 +11,7 @@ MySQL-python
 # chardet and certifi are used in circle indirectly by safety, but not as a compile-time dependency.
 chardet
 certifi
+
+# ordereddict is MIT according to https://pypi.python.org/pypi/ordereddict but
+# apparently the metadata is lacking (it shows up as UNKNOWN)
+ordereddict


johnstill:
I'm not seeing the circle ci failure locally; any suggestions are welcome.

Seeker1911:
Looks the same as the problem I had importing backoff. Not sure if circle-ci runs setup_env.

rest-api/config_api.py
```diff
@@ -64,45 +63,60 @@ def check_config_admin():
...
+    if config.getSettingJson(config.ALLOW_NONPROD_REQUESTS, False):
+      date = request.headers.get('x-pretend-date', None)
+      if date:
+        date = parse_date(date)
+
+    client_id = app_util.get_oauth_id()
+    config.store(model, date=date, client_id=client_id)
+    return self.make_response_for_resource(model.configuration)
+
+  def make_response_for_resource(self, data):
```
danqrodney: You can delete this whole thing and just return model.configuration above. Configs don't have last_modified. (The purpose of this logic was for e-tag checking, but we don't do that on configs, either.)

rest-api/config_api.py
```diff
@@ -64,45 +63,60 @@ def check_config_admin():
...
+    data = model.configuration or {}
+    return self.make_response_for_resource(data)
 
-  def put(self, key=None):
-    return super(ConfigApi, self).put(key or config.CONFIG_SINGLETON_KEY)
-
-  def validate_object(self, obj, a_id=None):
-    super(ConfigApi, self).validate_object(obj, a_id)
-    if obj.key.id() != config.CONFIG_SINGLETON_KEY:
+  def post(self, key=config.CONFIG_SINGLETON_KEY):
```
danqrodney: Seems like post and put should share the same implementation,differing only by the lookup for existing config?

rest-api/config.py
```diff
@@ -46,86 +48,80 @@
...
-      initialize_config()
-      obj = super(ConfigurationDAO, self).load_if_present(id_, ancestor_id)
-    return obj
+class ConfigurationHistory(ndb.Model):
+  date = ndb.DateTimeProperty(auto_now_add=True)
+  obj = ndb.StructuredProperty(Configuration, repeated=False)
+  client_id = ndb.StringProperty()
+
+
+def load(_id, date=None):
```
danqrodney: We need tests of load and store to replace data_access_object_test tests below.

rest-api/config_api.py
johnstill: Nice! I wasn't sure if a stored config might have a last modified tag or not.

rest-api/config_api.py
johnstill: not quite - `put` also does some stuff with `date` and `client_id`.  They are similar though; I've managed to reduce them a little more.

PR #658, opened 2018-04-05
Remove dead code and redundant files

johnstill:
As far as I can tell `field_config` was deprecated and untouched for the past year.  Removing it had no effect on tests and grepping the codebase for `field_config` yields no hits, so this seems safe to remove.

Also I've cleaned up the redundant `.gitignore` files and removed some developer specific IDE files.


johnstill:
Should I make a Jira ticket for this?

johnstill:
Thanks for finding that ref.  My "dead code detection" algorithm is basically (1) grep the project for a symbol and (2) if grep finds nothing, delete it and run the tests without that symbol.  So it has a hard time with self-sustaining circular dependencies of dead code.  Knowing where the big changes occurred may help find more dead code in the future.

> Of course with Python you can never truly be sure..

We can (and unfortunately _do_) obscure where modules are installed by hacking on the `PYTHONPATH`, but it is significantly more effort to obscure where modules are used ;) . So long as we don't do any deep customizations of the actual import machinery (as opposed to the import paths) you can find everywhere a name is used in a project just by grepping for it.

johnstill:
JIRA ref: https://precisionmedicineinitiative.atlassian.net/browse/DA-580

@jt2k Thank you for the clarification.  Do you want me to update the .gitignore?  I figured this might help prevent old clones of the repo from accidentally re-committing the deleted files.

jt2k:
@johnstill I don't think it needs to be changed

PR #659, opened 2018-04-10
Js/da 586 coverage

johnstill:
To check this feature out locally, clone down the branch, pick and open port, and then run from `rest-api`:

```sh
./tools/setup_env.sh
dev_appserver.py test.yaml --require_indexes --enable_console  # in a different terminal or with &
./tools/setup_local_database.sh
./test/run_tests.sh -g $sdk_dir
cd htmlcov
python -m SimpleHTTPServer $port    # assuming python --version is 2.x
```

Then open up a browser a `localhost:$port` and check out the browsable coverage report :)

#### Why?

A few main reasons:  (A) to help up identify code that is vs isn't adequately covered by the test suite.  (B) to help us identify code that is _never_ run - it may be dead.  (C) Once the test runner (and suite) is altered to allow us to run single test cases / methods by name, we'll be able to see _exactly_ what code is run by a given test, giving us some insight into how effective our test is and possibly preventing tests that pass "by accident" (rather than b/c they execute the intended code) or tests that are flaky from making it into the codebase.  That will also help us enormously in trying to debug flaky or otherwise bad test behavior (such as the data gen tests).


rest-api/test/runner.py
```diff
@@ -94,7 +94,19 @@ def main(sdk_path, test_path, test_pattern):
...
 
+  from coverage import Coverage
+  cov = Coverage(omit=[
+    '*/lib/*',
+    '*/appengine-mapreduce/*',
+    '*/site-packages/*',
+    '*/google_appengine/*',
+    '*/test/*',
+  ])
+  cov.start()
```
danqrodney: How much does this slow down tests?

Thank you for adding this capability -- but if it slows down tests appreciably when we're not particularly paying attention to coverage, we should make it flagged off by default.

rest-api/requirements.txt
```diff
@@ -24,3 +24,4 @@ protorpc>=0.12.0
 parameterized>=0.6.1
 googlemaps>=2.5.1
 backoff
+coverage
```
danqrodney: I think we may want to consider installing this separately in setup_env.sh instead of putting it here... we don't need coverage to be installed when deploying to AppEngine, right?

rest-api/test/runner.py
johnstill: It seems to add around 5-15 seconds.  Without it, it looks like the tests run in 55-65s, with, 65-80s.   Bear in mind though that I didn't rigorously profile this, I just ran with and without a few times and observed - also those numbers are only for the unit tests, they don't include the client tests (which are not affected).

rest-api/requirements.txt
johnstill: I had originally put it in `setup_env.sh` but CI wouldn't pick up on it - I suppose I could add it in after [this line](https://github.com/all-of-us/raw-data-repository/blob/2ae703fce7064a972ac4456bb389f6383f6750c1/ci/test.sh#L11).  Attaching it to [the circle config file](https://github.com/all-of-us/raw-data-repository/blob/2ae703fce7064a972ac4456bb389f6383f6750c1/circle.yml#L10) doesn't affect the tests b/c they're run inside a virtualenv inside the ci container.

johnstill: Ok, installing it in `ci/test.sh` definitely works; I've removed it from `requirements.txt`

rest-api/test/runner.py
danqrodney: OK. Let's hide it behind a flag then, and add the flag to your instructions.

johnstill: Done.  The flag is `-c` and it is documented in `rest-api/README.md` with the rest of the testing instructions.

PR #668, opened 2018-04-18
Mm/ops data api readme

Seeker1911:


opsdataAPI.md
```diff
@@ -0,0 +1,274 @@
+## OPS-DATA API
+
+The opsdata API offers access to the ParticipantSummary resource and represents an aggregated view of relevant
```
danqrodney: I would talk to Kelsey about what she wants in this doc.

My guess is it should start with how to get a service account key, how often to rotate it, etc.

Then talk about how to call the sync API using that key repeatedly (with a recommended frequency of every 5 minutes or whatever), linking to your example client code.

You can also link out to our main README for the details of some of the stuff that's in the response.

Seeker1911: After speaking with PM team, they don't want the links in the document, they share other docs with Awardees including how to go about service accounts. We'll probably have to change this and the main README in regards to renewing keys in the near future, as I believe I can recreate them in the script and we can handle key rotation.

PR #676, opened 2018-04-24
Making setup_env work on Ubuntu desktops.

danqrodney:
(OS X should be unaffected.)

### DIFF
--- a/rest-api/requirements.txt
+++ b/rest-api/requirements.txt
@@ -19,7 +19,7 @@ simplejson>=3.6.5
 sqlalchemy>=1.2
 alembic>=0.8.10
 dictalchemy>=0.1.2.7
-MySQL-python>=1.2.5
+MySQL-python>=1.2.5 --no-binary=MySQL-python
 protorpc>=0.12.0
 parameterized>=0.6.1
 googlemaps>=2.5.1
--- a/rest-api/tools/setup_env.sh
+++ b/rest-api/tools/setup_env.sh
@@ -15,9 +15,11 @@ rm -rf htmlcov/
 export PYTHONDONTWRITEBYTECODE=True
 
 SYSTEM_FLAG=
+NO_BINARY_FLAG=
 if [ "${OSTYPE}" = "linux-gnu" ]
 then
   SYSTEM_FLAG="--system "
+  NO_BINARY_FLAG="--no-binary "
 fi
 
 echo "Installing libs..."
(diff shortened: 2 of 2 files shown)


calbach:
I don't think OSTYPE is the right trigger:

```
(try-dan-rodete) [15:52][rest-api((HEAD detached at origin/danrodney/env_fix))]$ tools/setup_env.sh 
Removing old libs...
Cleaning up artifacts (pyo, pyc, etc)...
Installing libs...

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --system
```

danqrodney:
You did this on your Ubuntu box in a venv?

Can you run pip --version?

calbach:
Yep

```
pip --version
pip 10.0.1 from /usr/local/google/home/calbach/.virtualenvs/try-dan-rodete/local/lib/python2.7/site-packages/pip (python 2.7)
```

danqrodney:
Ah. When I do it (in a venv), I get:

pip 9.0.1 from /usr/lib/python2.7/dist-packages (python 2.7)

I don't know why 1) my pip is older or 2) why it isn't grabbing pip from a virtualenv like it is for you. Do you know?

Seeker1911:
check your activate script in bin/ grep for `VIRTUAL_ENV` and see if thats the right path,
otherwise, destroy the venv and recreate it

johnstill:
@danqrodney what is the output of `echo $VIRTUAL_ENV`?  I suspect that (A) the env is activated improperly or inactive or (B) the virtual env creation tool used did not bootstrap `pip` into the virtual environment (some older tools didn't, but they should all do that by default now).

If its the former, you can try de- and then re-activating the environment (or even destroying and recreating it).  If its the latter, I recommend you install `virtualenvwrapper` (you should be able to get it via apt-get).  It will prefix the environment name to your prompt so you always know if the env is activated and gives you a bunch of convenience commands, and it should always bootstrap pip when you create an environment.

johnstill:
Also if OSTYPE doesn't work reliably you should be able to use `[[ $(uname -s) = 'Linux' ]]` (it'll be 'Darwin' for OS X).

danqrodney:
Starting with a new venv fixed my issues. I'm closing this.

PR #677, opened 2018-04-24
add condition to check for circle deploy

Seeker1911:
quick fix for cron job


rest-api/tools/test.py
```diff
@@ -0,0 +1,50 @@
+import config
+from google.appengine.api import app_identity
+from googleapiclient import discovery
+import logging
+from datetime import datetime
+
+"""Deletes service account keys older than 3 days as required by NIH"""
```
danqrodney: Do you want to check in this file?

Similar to before, seems like our server endpoint is taking care of this now... we probably don't need a Python utility for doing it?

Seeker1911: no, that was for me to check the return and make sure account['email'] was correct, thought I deleted it.

Seeker1911: I'll remove and then merge

PR #679, opened 2018-04-25
Mm/filter withdrawn participants

Seeker1911:


rest-api/offline/metrics_export.py
```diff
@@ -58,7 +59,8 @@
    AND NOT EXISTS
     (SELECT * FROM participant_summary ps
       WHERE ps.participant_id = ph.participant_id
-        AND ps.email LIKE :test_email_pattern)
+      AND ps.withdrawal_status != 2
```
danqrodney: This should be instead:

ps.participant_id = ph.participant_id AND 
(ps.withdrawal_status = 2 OR ps.email LIKE :test_email_pattern)

because we're doing a NOT EXISTS here.

rest-api/offline/metrics_export.py
```diff
@@ -79,6 +81,7 @@
    AND NOT EXISTS
     (SELECT * FROM participant_summary ps
       WHERE ps.participant_id = p.participant_id
+        AND ps.withdrawal_status != 2
```
danqrodney: See above.

danqrodney:
(Please add tests.)

rest-api/test/unit_test/offline_test/metrics_export_test.py
```diff
@@ -111,6 +112,11 @@ def _create_data(self):
       participant_dao.insert(participant5)
       self.send_consent('P5', email='larry@gmail.com')
 
+      participant6 = Participant(participantId=6, biobankId=7, providerLink=pl_tucson)
+      participant_dao.insert(participant6)
+      self.send_consent('P6', email='larry@gmail.com')
+      print (dir(participant6))
```
danqrodney: Remove this

PR #680, opened 2018-04-26
DA-594 add 1ed10 to dna sample test code

Seeker1911:
Don't think this needs to be added any where else.

### DIFF
--- a/rest-api/config/base_config.json
+++ b/rest-api/config/base_config.json
@@ -30,6 +30,7 @@
     "2ED10"
   ],
   "dna_sample_test_codes": [
+    "1ED10",
     "1ED04",
     "1SAL",
     "1SAL2"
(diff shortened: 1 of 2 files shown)


rest-api/config/base_config.json
```diff
@@ -30,6 +30,7 @@
     "2ED10"
   ],
   "dna_sample_test_codes": [
+    "1ED10",
```
danqrodney: This logic needs updating as well:

https://github.com/all-of-us/raw-data-repository/blob/5424eed51833a1263dd39321feef704f46aab6d1/rest-api/dao/participant_counts_over_time_service.py#L209

@eweitz

danqrodney: (Doing this is going to be a bit of a pain, may want to use COALESCE... there was a previous conversation with @ssutherland about this.)

danqrodney: Specifically, we said over Slack we'd use this:

               CASE WHEN
                         LEAST(
                            COALESCE(sample_status_1ed04_time, '3000-01-01')
                            , COALESCE(sample_status_1sal_time, '3000-01-01')
                            , COALESCE(sample_status_1sal2_time, '3000-01-01')
                            ) = '3000-01-01' THEN NULL
                    ELSE LEAST(
                            COALESCE(sample_status_1ed04_time, '3000-01-01')
                            , COALESCE(sample_status_1sal_time, '3000-01-01')
                            , COALESCE(sample_status_1sal2_time, '3000-01-01')
                            )    
                    END

PR #681, opened 2018-04-26
DA-534 add a request to end of deploy_app

Seeker1911:


rest-api/tools/deploy_app.sh
```diff
@@ -177,4 +177,12 @@ then
   rm "${tmp_files[@]}"
 fi
 
+test_request=$(curl -s http://localhost:8080/rdr/v1/ | grep version_id)
+if [[ -z "$test_request" ]];
+then
+  echo "${BOLD}Test request failed to return the expected response, something may be wrong with
```
danqrodney: We should also update the JIRA ticket if this happens. Ideally we'd send an alert to Slack, too... if you don't want to do that for now, fine by me.

rest-api/tools/deploy_app.sh
```diff
@@ -182,7 +182,9 @@ if [[ -z "$test_request" ]];
 then
   echo "${BOLD}Test request failed to return the expected response, something may be wrong with
        the deployment${NONE}"
+  $UPDATE_TRACKER --version $VERSION --comment "Test request failed to return a valid response."
```
danqrodney: Can we add something like "!!!!!!!!!!!! REQUEST FAILURE !!!!!!!!!" or whatever to make this stick out?

rest-api/tools/deploy_app.sh
```diff
@@ -177,4 +177,14 @@ then
   rm "${tmp_files[@]}"
 fi
 
+test_request=$(curl -s http://localhost:8080/rdr/v1/ | grep version_id)
```
danqrodney: Sorry, you can't use localhost here. You have to use the host of the server.

danqrodney: (Please test this by deploying to sandbox.)

danqrodney: In particular, replace http://localhost:8080 with https://${PROJECT}.appspot.com. (Note that we always use SSL.)

PR #682, opened 2018-04-27
DA-530 add sleep

Seeker1911:
datagen test occasionally fails. Probably due to reading from a csv and inserting a bunch of random dates and times. Added a 1 second sleep to see if this helps.


danqrodney:
Please don't do this. :)

I'd like to get rid of this flake, too, but the way to do that is to identify the cause.

I don't at this point have any reason to believe that adding a sleep will fix this problem, but if it did,  it could be masking an actual bug in our code.

Can you explain why reading from a CSV and inserting a bunch of random dates and times should cause problems?

PR #684, opened 2018-04-27
DA-606 alter metrics to use pm finalized time

Seeker1911:


rest-api/offline/metrics_config.py
```diff
@@ -146,7 +146,7 @@ def _get_answer_values_func(question_code_value):
 
 # Fields generated by our metrics export.
 BASE_PARTICIPANT_FIELDS = ['participant_id', 'date_of_birth', 'first_order_date',
-                           'first_samples_arrived_date', 'first_physical_measurements_date',
+                           'first_samples_arrived_date', 'final_physical_measurements_date',
```
danqrodney: Let's still call this "first_physical_measurements_date" -- we're still finding the MIN value. What's changed is we are using the timestamp of the PM (stored as "finalized", parsed from the JSON resource) instead of the server time at which the PM was created (stored as "created"). But either way, it's the *first* PM date for this user.

rest-api/offline/metrics_export.py
```diff
@@ -30,8 +30,8 @@
     WHERE bo.participant_id = p.participant_id) first_order_date,
   (SELECT ISODATE[MIN(bs.confirmed)] FROM biobank_stored_sample bs
     WHERE bs.biobank_id = p.biobank_id) first_samples_arrived_date,
-  (SELECT ISODATE[MIN(pm.created)] FROM physical_measurements pm
-    WHERE pm.participant_id = p.participant_id) first_physical_measurements_date,
+  (SELECT ISODATE[MIN(pm.finalized)] FROM physical_measurements pm
+    WHERE pm.participant_id = p.participant_id and pm.finalized is not null) final_physical_measurements_date,
```
danqrodney: Note that after the backfill is done, all physical measurements should have finalized timestamps, so this is not null won't be needed... but it doesn't hurt anything.

rest-api/test/unit_test/offline_test/metrics_export_test.py
```diff
@@ -255,7 +256,7 @@ def test_metric_export(self):
     #  'date_of_birth',
     #  'first_order_date',
     #  'first_samples_arrived_date',
-    #  'first_physical_measurements_date',
+    #  'final_physical_measurements_date',
```
danqrodney: See above re: name.

rest-api/offline/metrics_config.py
Seeker1911: done

rest-api/offline/metrics_export.py
Seeker1911: ok

rest-api/test/unit_test/offline_test/metrics_export_test.py
Seeker1911: done

PR #687, opened 2018-04-30
[DA-609] Refine enrollment filter

eweitz:
This refines the update to the enrollment_status filter just merged in #685.  That change unexpectedly included unconsented-but-registered participants when filtering for only MEMBERS; this fixes that so such participants are excluded in such cases.  

The new test here will prevent against future regressions in this.


rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -111,7 +111,7 @@ def get_facets_sql(self, facets, table_prefix='ps'):
       for q_filter in filters:
         if str(q_filter) != '':
           filter_sql = filter_prefix + '.' + db_field + ' = ' + str(int(q_filter))
-          if allow_null:
+          if allow_null and str(int(q_filter)) == '1':
```
danqrodney: Rather than do this, I'd add the "str(int(q_filter)) == '1'" at line 96 and leave this alone.

danqrodney: Please ignore. :)

eweitz: As discussed on Slack, q_filter isn’t defined up there, only down where I’ve used it.  We agree this works.

PR #692, opened 2018-05-02
DA-620 refactor setup_database to not always update passwords

Seeker1911:


rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,50 @@ function finish {
...
-     ' "backup_db_connection_name": "'$BACKUP_INSTANCE_NAME'", '\
-     ' "db_user": "'$RDR_DB_USER'", '\
-     ' "db_name": "'$DB_NAME'" }' > $TMP_DB_INFO_FILE
+function get_existing_db_config {
+  "Getting database info..."
+  tools/install_config.sh --key db_config --instance $INSTANCE \
+   --creds_file ${CREDS_FILE} --config_output "$TMP_DB_INFO_FILE"
+}
 
 for db_name in "rdr" "metrics"; do
```
danqrodney: So I realized the logic before had a bug here, which we should fix while we're at this... the only time we need to loop through these DBs is specifically when running create_db.sql. For updating passwords there's no point in running things twice. So you can put this loop inside the else block.

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,50 @@ function finish {
...
   if [ "${UPDATE_PASSWORDS}" = "Y" ]
   then
-    cat tools/update_passwords.sql | envsubst > $UPDATE_DB_FILE
+    	echo "Updating database user passwords..."
+	randpw
```
danqrodney: So create_db.sql does reference these passwords:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/tools/create_db.sql#L3

It just doesn't use them if the database already exists.

So I think this password generation stuff needs to run at least in the DB creation case, if not the update the existing DB case.

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,50 @@ function finish {
...
+	     ' "db_name": "'$DB_NAME'" }' > $TMP_DB_INFO_FILE
+
+	echo "Setting root password..."
+	#gcloud sql instances set-root-password $INSTANCE_NAME --password $ROOT_PASSWORD
+	gcloud sql users set-password root % --instance $INSTANCE_NAME --password $ROOT_PASSWORD
+
+        cat tools/update_passwords.sql | envsubst > $UPDATE_DB_FILE
   else
-    cat tools/create_db.sql | envsubst > $UPDATE_DB_FILE
+	get_existing_db_config
```
danqrodney: If we aren't changing the existing config at all, then we don't need to run install_config.sh below, either.

Setting up the logic this way does give us the ability to re-run create_db.sql and have those changes applied to existing databases, which is nice. However, this logic won't give us the ability to update the existing db config without also changing the passwords. Ideally there would be a way of getting the existing db config and then creating a new db config based on the old password values, so you wouldn't have to change the passwords when adding new stuff to db_config. That said -- if you want to make it so that for now you must change passwords to update db_config, that's OK with me... just be aware we will have to do this in a maintenance windo win prod.

rest-api/tools/setup_database.sh
Seeker1911: done

rest-api/tools/setup_database.sh
Seeker1911: done

rest-api/tools/setup_database.sh
Seeker1911: If I understood you correctly, then this is done. Not sure what the use case is for updating config independently of creating/updating passwords.

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,56 @@ function finish {
...
+	if [ "${UPDATE_PASSWORDS}" = "Y" ]
+	    then 
+		echo "updating passwords for database"
+		cat tools/update_passwords.sql | envsubst > $UPDATE_DB_FILE
+	fi
+	if [ "${CREATE_INSTANCE}" = "Y" ]
+	    then
+		echo "creating new database"
+		for db_name in "rdr" "metrics"; do
+	       	   cat tools/create_db.sql | envsubst > $UPDATE_DB_FILE
```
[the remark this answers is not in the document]
Seeker1911: db_name is set to 'rdr' and 'metrics' respectively as it iterates. so isn't that what we're doing here already?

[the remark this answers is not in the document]
danqrodney: In bash variable names are case sensitive... db_name != DB_NAME.

It could be that if you did "for DB_NAME" it would work. It might be better to assign DB_NAME to db_name here.

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,56 @@ function finish {
...
-    echo "Creating empty database..."
-  fi
-  mysql -u "$ROOT_DB_USER" -p"$ROOT_PASSWORD" --host 127.0.0.1 --port ${PORT} < ${UPDATE_DB_FILE}
-done
-
-echo "Setting database configuration..."
-tools/install_config.sh --key db_config --config ${TMP_DB_INFO_FILE} --instance $INSTANCE --update --creds_file ${CREDS_FILE}
+	echo "getting database config..."
+		get_existing_db_config
+		echo "Setting database configuration..."
```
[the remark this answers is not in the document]
danqrodney: In addition to above... the idea would be you'd run create_db.sql and grant_permissions.sql above in a loop, and here just run grant_permissions.sql.  

BTW, you don't actually need to call get_existing_db_config / define it unless you're going to support updating db_config without changing passwords. You're not using db_config anywhere here.

rest-api/tools/setup_database.sh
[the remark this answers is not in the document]
danqrodney: Sorry, I'm an idiot. :) This should work.

rest-api/tools/setup_database.sh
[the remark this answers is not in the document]
Seeker1911: done

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,56 @@ function finish {
...
+	echo "Setting root password..."
+	#gcloud sql instances set-root-password $INSTANCE_NAME --password $ROOT_PASSWORD
+	gcloud sql users set-password root % --instance $INSTANCE_NAME --password $ROOT_PASSWORD
+
+	if [ "${UPDATE_PASSWORDS}" = "Y" ]
+	    then 
+		echo "updating passwords for database"
+		cat tools/update_passwords.sql | envsubst > $UPDATE_DB_FILE
+	fi
+	if [ "${CREATE_INSTANCE}" = "Y" ]
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/tools/setup_database.sh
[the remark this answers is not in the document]
Seeker1911: 😄 done

rest-api/grant_permissions.sql
```diff
@@ -0,0 +1,7 @@
+GRANT SELECT ON ${db_name}.* TO '${READONLY_DB_USER}'@'%';
+GRANT SELECT, INSERT, UPDATE, DELETE, CREATE TEMPORARY TABLES ON ${db_name}.*
+TO '${RDR_DB_USER}'@'%';
```
danqrodney: If you're going to break the lines up, please indent.

rest-api/grant_permissions.sql
```diff
@@ -0,0 +1,7 @@
...
+GRANT SELECT, INSERT, UPDATE, DELETE, DROP, ALTER, CREATE, INDEX, REFERENCES,
+CREATE TEMPORARY TABLES, CREATE VIEW ON ${db_name}.* TO '${ALEMBIC_DB_USER}'@
+'%';
```
danqrodney: This doesn't feel like it should be on its own line. If you need to wrap, take everything after the "TO" with it.

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,51 @@ function finish {
...
+		done
+	fi
+
+	mysql -u "$ROOT_DB_USER" -p"$ROOT_PASSWORD" --host 127.0.0.1 --port ${PORT} < ${UPDATE_DB_FILE}
   else
-    echo "Creating empty database..."
-  fi
-  mysql -u "$ROOT_DB_USER" -p"$ROOT_PASSWORD" --host 127.0.0.1 --port ${PORT} < ${UPDATE_DB_FILE}
-done
+	echo "getting database config..."
```
danqrodney: Update this.

rest-api/tools/setup_database.sh
```diff
@@ -100,36 +91,51 @@ function finish {
...
+	for db_name in "rdr" "metrics"; do
+	   cat tools/grant_permissions.sql | envsubst > $UPDATE_DB_FILE
+	done
+fi
 
 echo "Setting database configuration..."
```
danqrodney: This should only run in the first half of the if statement... we don't want to update the db_config when just granting permissions.

rest-api/grant_permissions.sql
Seeker1911: fixed

rest-api/grant_permissions.sql
Seeker1911: fixed

rest-api/tools/setup_database.sh
Seeker1911: k

rest-api/tools/setup_database.sh
Seeker1911: done

rest-api/tools/grant_permissions.sql
```diff
@@ -0,0 +1,4 @@
+GRANT SELECT ON ${db_name}.* TO '${READONLY_DB_USER}'@'%';
+GRANT SELECT, INSERT, UPDATE, DELETE, CREATE TEMPORARY TABLES ON ${db_name}.*TO '${RDR_DB_USER}'@'%';
```
danqrodney: Please add a space before "TO".

rest-api/tools/grant_permissions.sql
```diff
@@ -0,0 +1,4 @@
...
+GRANT SELECT, INSERT, UPDATE, DELETE, DROP, ALTER, CREATE, INDEX, REFERENCES,
+CREATE TEMPORARY TABLES, CREATE VIEW ON ${db_name}.* TO '${ALEMBIC_DB_USER}'@'%';
```
danqrodney: Indent

PR #694, opened 2018-05-02
Setting a dummy latitude and longitude to all sites when importing

danqrodney:
locally / on a CircleCI box.

### DIFF
--- a/rest-api/tools/import_organizations.py
+++ b/rest-api/tools/import_organizations.py
@@ -352,7 +352,7 @@ class SiteImporter(CsvImporter):
         site.latitude = latitude
         site.longitude = longitude
         if latitude and longitude:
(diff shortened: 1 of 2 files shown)


rest-api/tools/import_organizations.py
```diff
@@ -348,6 +348,11 @@ def _populate_lat_lng_and_time_zone(self, site, existing_site):
         site.longitude = longitude
         if latitude and longitude:
           site.timeZoneId = self._get_time_zone(latitude, longitude)
+      else:
```
calbach: The flag documentation should make mention of this or else flag name changed (e.g. --stub_geocoding). Not obvious from this that `--gecode_flag` would exclusively be passed in non-testing environments.

danqrodney: Switching to --stub_geocoding

PR #695, opened 2018-05-05
[DA-471] Disable flakey data gen test

calbach:
The data gen endpoint appears to be for testing purposes, and looks to be very non-deterministic. Disable this for now - not sure if it is worth fixing.


calbach:
> Does unit test.skip show up in logs so we don’t forget about it ?

Yes.

PR #696, opened 2018-05-08
DA-318 change empty qr response to error.

Seeker1911:


Seeker1911:
@danqrodney Do you know where PTSC is at on this ? This is now so old, We'll probably have problems integrating it into master.

danqrodney:
Nope, I never talked to them. Probably would be a good idea to do so at some point.

PR #697, opened 2018-05-08
[DA-628] Pick up test config changes in the participant summary

calbach:
Split individual test updates out from the aggregated field updates. We want to update aggregation counts regardless of whether there was a modification timestamp change in the individual biobank orders, i.e. if we change the set of baseline/dna tests.

### DIFF
--- a/rest-api/dao/participant_summary_dao.py
+++ b/rest-api/dao/participant_summary_dao.py
@@ -72,8 +72,6 @@ _SAMPLE_SQL = """,
           AND biobank_stored_sample.test = %(sample_param_ref)s)
    """
 
-_PARTICIPANT_ID_FILTER = " WHERE participant_id = :participant_id"
-
 _WHERE_SQL = """
 not sample_status_%(test)s_time <=>
 (SELECT MAX(confirmed) FROM biobank_stored_sample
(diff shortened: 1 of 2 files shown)


rest-api/dao/participant_summary_dao.py
```diff
@@ -215,39 +260,30 @@ def make_query_filter(self, field_name, value):
...
-    sample_sql, sample_params = _get_sample_sql_and_params()
-    sql = """
+    now = clock.CLOCK.now()
+    sample_sql, sample_params = _get_sample_sql_and_params(now)
+
+    baseline_tests_sql, baseline_tests_params = _get_baseline_sql_and_params()
+    dna_tests_sql, dna_tests_params = _get_dna_isolates_sql_and_params()
+
+    counts_sql = """
     UPDATE
```
danqrodney: So I was wondering whether we need to do something similar for  _ENROLLMENT_STATUS_SQL. This would only be necessary when we change enrollment status criteria... but I think we are going to do that in future. 

If you don't want to do it in this PR, sounds fine, but let's file an issue to track it in future.

rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
```diff
@@ -253,6 +253,37 @@ def add_sample(participant, test_code, sample_id):
...
+      TIME = datetime.datetime(2018, 3, 2)
+      sample_dao.insert(BiobankStoredSample(
+          biobankStoredSampleId=sample_id, biobankId=participant.biobankId,
+        biobankOrderIdentifier='KIT', test=test_code, confirmed=TIME))
+
+    add_sample(baseline_tests[0], '11111')
+    add_sample(baseline_tests[1], '22223')
+    self.dao.update_from_biobank_stored_samples()
+    summary = self.dao.get(participant.participantId)
+    init_last_modified = summary.lastModified
```
danqrodney: Can we verify here that lastModified has changed from the insert? And then run update_from_biobank_stored_samples() a second time and verify that it doesn't?

rest-api/dao/participant_summary_dao.py
danqrodney: Not pressing, but please file an issue.

calbach: Done - DA-631

PR #699, opened 2018-05-11
Add back in the SQL proxy to setup_database

calbach:


Seeker1911:
Huh. Thought that was there already. Thanks CH.

PR #700, opened 2018-05-11
move cloud_sql_proxy function before updating passwords to prevent error.

Seeker1911:


danqrodney:
As discussed over Slack, I'm not sure this is really needed. Doesn't hurt anything, either.

PR #701, opened 2018-05-11
Fixing setup_database to set connection string after setting passwords.

danqrodney:
Also fixing run_cloud_sql_proxy to run (and mysql be invoked with root password) in the grant_permissions.sql case.


Seeker1911:
Awesome

PR #704, opened 2018-05-15
DA-593 add raw data views for export.

Seeker1911:


calbach:
Do we need new views? Or should we delete the older views? I changed the original views to "year" rather than time, but this could probably be reverted since we ultimately went with ppi_* variants for deidentified exports.

- questionnaire_response_answer_view
- participant_view

physical measurements would likely need a new view.

Perhaps go with the raw_* as you've done, but also drop questionnaire_response_answer_view, participant_view in this change?

danqrodney:
@calbach Kelsey suggested we add new views and keep the old ones around. (They don't really hurt anything.)

participant_view is actually a view already in use by @ssutherland; assuming we need a new view here (and don't just use that one), we'll need a different name. I'm OK with raw_*.

Seeker1911:
would ...
_IDENTIFIED_PPI_PARTICIPANT_VIEW_SQL
_IDENTIFIED_QUESTIONNAIRE_RESPONSE_ANSWER_VIEW_SQL
_IDENTIFIED_MEASUREMENTS_VIEW_SQL
be better ?

calbach:
raw sounds fine to me. I just think we should avoid having two views serving an identical purpose (in this case, that would be participant_view vs raw_participant_view, and qra_view vs raw_qra_view). This makes it very confusing for a developer, e.g. to know which views need to be updated if a bug is discovered, or to just rationalize about which code affects which clients.

One reasonable plan could be to add raw_* per this PR, then ask @ssutherland to switch once this goes to prod. Then later, drop participant_view and qra_view.

ssutherland:
I'm totally OK if you want to add additional fields to participant_view, but please try to avoid making any changes like turning times into dates. I use this view very extensively and such a change will cascade throughout my scripts.

Also fine if you add want to add new views. Your choice. I primarily use participant_view and occasionally the others, though none of my scripts rely on them.

Seeker1911:
I’m not gonna mess with participant_view. It will stay as all new views for this purpose.

rest-api/alembic/versions/32df5fa890bf_raw_export_api.py
```diff
@@ -0,0 +1,299 @@
...
+    INNER JOIN questionnaire_question qq ON qra.question_id = qq.questionnaire_question_id
+    INNER JOIN questionnaire q ON qq.questionnaire_id = q.questionnaire_id
+    INNER JOIN code qc ON qq.code_id = qc.code_id
+    INNER JOIN participant_summary ps ON p.participant_id = ps.participant_id
+    LEFT OUTER JOIN hpo ON p.hpo_id = hpo.hpo_id
+    LEFT OUTER JOIN code ac ON qra.value_code_id = ac.code_id
+    WHERE p.withdrawal_status = 1 AND # NOT_WITHDRAWN
+        (ps.email IS NULL OR ps.email NOT LIKE '%@example.com') AND
+        (hpo.name IS NULL OR hpo.name != 'TEST') AND
+        /* Blacklist any codes which are superfluous for PPI QA. */
```
calbach: remove or update comment

Seeker1911: to what ? It seems accurate to me.

rest-api/alembic/versions/32df5fa890bf_raw_export_api.py
```diff
@@ -0,0 +1,299 @@
...
+     LEFT OUTER JOIN participant_summary ps ON p.participant_id = ps.participant_id
+     LEFT OUTER JOIN code state_code ON ps.state_id = state_code.code_id
+     LEFT OUTER JOIN code recontact_method_code ON ps.recontact_method_id = recontact_method_code.code_id
+     LEFT OUTER JOIN code language_code ON ps.language_id = language_code.code_id
+     LEFT OUTER JOIN code gender_code ON ps.gender_identity_id = gender_code.code_id
+     LEFT OUTER JOIN code sex_code ON ps.sex_id = sex_code.code_id
+     LEFT OUTER JOIN code sexual_orientation_code ON ps.sexual_orientation_id = sexual_orientation_code.code_id
+     LEFT OUTER JOIN code education_code ON ps.education_id = education_code.code_id
+     LEFT OUTER JOIN code income_code ON ps.income_id = income_code.code_id
+     WHERE p.withdrawal_status = 1 AND # NOT_WITHDRAWN
```
danqrodney: So I think the only real difference between this and participant_view now is that we're filtering on withdrawal_status = 1 here.

That may be a good enough reason to have two views... but if the PPI QA people could handle having withdrawal status in the results and doing their own filtering, we wouldn't necessarily need this view. We should inquire.

rest-api/alembic/versions/32df5fa890bf_raw_export_api.py
danqrodney: I think this is fine.

PR #706, opened 2018-05-17
DA-323 delete old organizations no longer in google sheet.

Seeker1911:


rest-api/tools/import_organizations.py
```diff
@@ -108,6 +108,42 @@ def _insert_entity(self, entity, existing_map, session, dry_run):
...
+          )
+          """.format(str_list=str_list)
+
+    session.execute(sql)
+
+  def _cleanup_old_entities(self, session, row_list):
+    hpo_dao = HPODao()
+    hpo_group_list_from_sheet = []
+    existing_hpos = set(hpo.name for hpo in hpo_dao.get_all())
+    [hpo_group_list_from_sheet.append(row[HPO_AWARDEE_ID_COLUMN].upper()) for row in row_list]
```
danqrodney: I think it would be simpler to remove the [] assignment above and write this as:

hpo_group_list_from_sheet = [row[HPO_AWARDEE_ID_COLUMN].upper() for row in row_list]

rest-api/tools/import_organizations.py
```diff
@@ -204,6 +275,38 @@ def run(self, filename, dry_run):
...
             self._insert_new_participants(self.new_sites_list)
 
+
+  def delete_sql_statement(self, session, site_id_list):
+    str_list = ','.join([str(i) for i in site_id_list])
+    sql = """
+          DELETE FROM site
+          WHERE site_id IN ({str_list}) 
+          AND NOT EXISTS(
+          SELECT * FROM participant p, participant_history ph
```
danqrodney: See above re: NOT_EXISTS. Also, you should check site ID columns on participant_summary, physical_measurements, biobank_order here.

rest-api/tools/import_organizations.py
```diff
@@ -204,6 +275,38 @@ def run(self, filename, dry_run):
...
+          )
+          """.format(str_list=str_list)
+
+    session.execute(sql)
+
+  def _cleanup_old_entities(self, session, row_list):
+    site_dao = SiteDao()
+    site_group_list_from_sheet = []
+    existing_sites = set(site.googleGroup for site in site_dao.get_all())
+    [site_group_list_from_sheet.append(str(row[SITE_SITE_ID_COLUMN].lower())) for row in row_list]
```
danqrodney: See above re: assignment.

rest-api/tools/import_organizations.py
```diff
@@ -108,6 +108,42 @@ def _insert_entity(self, entity, existing_map, session, dry_run):
...
+
+    hpos_to_remove = existing_hpos - set(hpo_group_list_from_sheet)
+    if hpos_to_remove:
+      hpo_id_list = []
+      for hpo in hpos_to_remove:
+        logging.info('Deleting old Awardee no longer in master list: %s', hpo)
+        old_hpo = hpo_dao.get_by_name(hpo)
+        hpo_id_list.append(old_hpo.hpoId)
+
+      self.delete_sql_statement(session, hpo_id_list)
```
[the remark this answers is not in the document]

rest-api/tools/import_organizations.py
```diff
@@ -108,6 +108,42 @@ def _insert_entity(self, entity, existing_map, session, dry_run):
     super(HPOImporter, self)._insert_entity(entity, existing_map, session, dry_run)
 
 
+  def delete_sql_statement(self, session, hpo_id_list):
+    str_list = ','.join([str(i) for i in hpo_id_list])
+    sql = """
+          DELETE FROM hpo
+          WHERE hpo_id IN ({str_list})
+          AND NOT EXISTS(
+          SELECT * FROM participant p, participant_history ph, site s, organization org
```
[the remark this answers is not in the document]

rest-api/tools/import_organizations.py
```diff
@@ -133,6 +169,41 @@ def _entity_from_row(self, row):
                         displayName=row[ORGANIZATION_NAME_COLUMN],
                         hpoId=hpo.hpoId)
 
+  def delete_sql_statement(self, session, org_id_list):
+    str_list = ','.join([str(i) for i in org_id_list])
+    sql = """
+          DELETE FROM organization
+          WHERE organization_id IN ({str_list})
+          AND NOT EXISTS(
+          SELECT * FROM participant p, participant_history ph, site s
```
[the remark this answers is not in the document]

rest-api/tools/import_organizations.py
```diff
@@ -133,6 +169,41 @@ def _entity_from_row(self, row):
...
+
+    orgs_to_remove = existing_orgs - set(org_group_list_from_sheet)
+    if orgs_to_remove:
+      org_id_list = []
+      for org in orgs_to_remove:
+        logging.info('Deleting old Organization no longer in master list: %s', org)
+        old_org = org_dao.get_by_external_id(org)
+        org_id_list.append(old_org.organizationId)
+
+      self.delete_sql_statement(session, org_id_list)
```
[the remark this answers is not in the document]

rest-api/tools/import_organizations.py
Seeker1911: done. and there is no site_id on PM and bio tables

rest-api/tools/import_organizations.py
```diff
@@ -204,6 +275,38 @@ def run(self, filename, dry_run):
...
+
+    sites_to_remove = existing_sites - set(site_group_list_from_sheet)
+    if sites_to_remove:
+      site_id_list = []
+      for site in sites_to_remove:
+        logging.info('Deleting old Site no longer in master list: %s', site)
+        old_site = site_dao.get_by_google_group(site)
+        site_id_list.append(old_site.siteId)
+
+      self.delete_sql_statement(session, site_id_list)
```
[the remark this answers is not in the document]

rest-api/tools/import_organizations.py

rest-api/tools/import_organizations.py
```diff
@@ -133,6 +169,41 @@ def _entity_from_row(self, row):
...
+          )
+          """.format(str_list=str_list)
+
+    session.execute(sql)
+
+  def _cleanup_old_entities(self, session, row_list):
+    org_dao = OrganizationDao()
+    org_group_list_from_sheet = []
+    existing_orgs = set(str(org.externalId) for org in org_dao.get_all())
+    [org_group_list_from_sheet.append(row[ORGANIZATION_ORGANIZATION_ID_COLUMN].upper()) for row in
```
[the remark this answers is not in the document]

rest-api/tools/import_organizations.py

rest-api/data/organizations.csv
```diff
@@ -1,5 +1,5 @@
 Awardee ID,Organization ID,Name
-CAL_PMC,AZ_TUCSON_BANNER_HEALTH,Banner Health ba
```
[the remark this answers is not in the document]
Seeker1911: It should have been AZ. Might have pushed up a change on accident previously. I ran export and had this change. Thought it pushed to push up the fix before it gets forgotten about.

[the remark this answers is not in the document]
danqrodney: OK. Looks like this got committed 4 days ago. The good news is we haven't imported organizations since then. :)

rest-api/tools/import_organizations.py
```diff
@@ -108,6 +108,41 @@ def _insert_entity(self, entity, existing_map, session, dry_run):
...
+  def _cleanup_old_entities(self, session, row_list):
+    self.hpo_dao = HPODao()
+    existing_hpos = set(hpo.name for hpo in self.hpo_dao.get_all())
+    hpo_group_list_from_sheet = [row[HPO_AWARDEE_ID_COLUMN].upper() for row in row_list]
+
+    hpos_to_remove = existing_hpos - set(hpo_group_list_from_sheet)
+    if hpos_to_remove:
+      hpo_id_list = []
+      for hpo in hpos_to_remove:
+        old_hpo = self.hpo_dao.get_by_name(hpo)
```
danqrodney: Add logging here for the awardee being deleted?

rest-api/tools/import_organizations.py
```diff
@@ -108,6 +108,41 @@ def _insert_entity(self, entity, existing_map, session, dry_run):
...
+        hpo_id_list.append(old_hpo.hpoId)
+
+      if hpo_id_list:
+        self.delete_sql_statement(session, hpo_id_list)
```
danqrodney: So the only issue with this is: we log what we are trying to delete, but if there are awardees / orgs / sites that still have references, and thus we don't delete them, that doesn't get logged.

That may be OK... since otherwise we'd log them every time we run this... but we may at some point have a way of noticing sites / orgs / awardees still in the database because they are referenced somewhere but are no longer referenced in our sheets. If there are sites, etc. like this at some point, we may need to flag them as "obsolete" or something so that we can e.g. exclude them in the awardee API.

If you don't want to address this now, that's probably fine, but let's file a TODO issue?

rest-api/tools/import_organizations.py
```diff
@@ -108,6 +108,41 @@ def _insert_entity(self, entity, existing_map, session, dry_run):
...
+    hpo_group_list_from_sheet = [row[HPO_AWARDEE_ID_COLUMN].upper() for row in row_list]
+
+    hpos_to_remove = existing_hpos - set(hpo_group_list_from_sheet)
+    if hpos_to_remove:
+      hpo_id_list = []
+      for hpo in hpos_to_remove:
+        old_hpo = self.hpo_dao.get_by_name(hpo)
+        hpo_id_list.append(old_hpo.hpoId)
+
+      if hpo_id_list:
```
danqrodney: We should probably add deletion_count and log that at the end with new_count, skip_count, etc.

We may even want to add a prompt ("are you sure you want to delete XXX")... though arguably if we do a dry run in prod first that should be enough confirmation.

rest-api/tools/import_organizations.py
Seeker1911: done. Got a little over zealous removing the debugging logging.

rest-api/tools/import_organizations.py
Seeker1911: added a deletion count

rest-api/tools/import_organizations.py
Seeker1911: added a jira ticket

PR #708, opened 2018-05-21
DA-629 change race skipped to pmi_skip for metrics and participant summary

Seeker1911:


rest-api/offline/metrics_pipeline.py
```diff
@@ -272,7 +272,7 @@ def map_answers(reader):
                              question_code != RACE_QUESTION_CODE):
       race_codes = [code_dao.get_code(PPI_SYSTEM, value) for value in race_code_values]
       race = get_race(race_codes)
-      if race == Race.SKIPPED:
+      if race == Race.PMI_Skip:
```
danqrodney: I'm not sure you need this if statement anymore?

rest-api/offline/metrics_pipeline.py
```diff
@@ -309,7 +309,7 @@ def map_answers(reader):
   if race_code_values:
     race_codes = [code_dao.get_code(PPI_SYSTEM, value) for value in race_code_values]
     race = get_race(race_codes)
-    if race == Race.SKIPPED:
+    if race == Race.PMI_Skip:
```
danqrodney: Same, maybe you can remove this?

rest-api/offline/metrics_pipeline.py
Seeker1911: removed, was gonna leave it because we explicitly cast PMI_Skip to string, wasn't 100% sure it wouldn''t effect something.

rest-api/offline/metrics_pipeline.py
Seeker1911: done

Seeker1911:
Does this not suffice for a test ?: 
https://github.com/all-of-us/raw-data-repository/blob/d6ff38246b93a58f1364eae64be7130e9712ae12/rest-api/test/unit_test/api_test/participant_summary_api_test.py#L563

danqrodney:
So now I'm confused. I thought we were actually changing server behavior. Where were we returning SKIPPED before? What functional change does this PR accomplish?

Seeker1911:
It just changes what is returned in the json response to `”PMI_Skip”` as other fields do.

danqrodney:
So that's what I thought... but the test above suggests it was already doing that?

Whatever the functional change is, there should be a test verifying it works.

PR #711, opened 2018-05-22
DA-643 add dry run logic to import

Seeker1911:


rest-api/tools/import_organizations.py
```diff
@@ -142,7 +142,7 @@ def _cleanup_old_entities(self, session, row_list):
           hpo_id_list.append(old_hpo.hpoId)
           self.deletion_count += 1
 
-      if hpo_id_list:
+      if hpo_id_list and not dry_run:
```
danqrodney: Please add (dry run) to log statements in these methods

PR #713, opened 2018-05-23
Added mapping for ethnicity and sex at birth mappings

cukarthik:
@danqrodney  I wasn't sure about the local phi vocabularies used, but i created one called ppi-sex for "SexAtBirth" mappings.

### DIFF
--- a/rest-api/etl/source_to_concept_map.csv
+++ b/rest-api/etl/source_to_concept_map.csv
@@ -1,3 +1,5 @@
+SexAtBirth_Male|1585846|ppi-sex|MALE|8507|Gender|1970-1-1|2099-12-31|\N|1
+SexAtBirth_Female|1585847|ppi-sex|FEMALE|8532|Gender|1970-1-1|2099-12-31|\N|1
 AIANSpecific_AIANNoneOfTheseDescribeMe|1585603|ppi-race|AIAN None of These Describe Me|8657|Race|1970-1-1|2099-12-31|\N|1
 AIANSpecific_AlaskaNative|1585601|ppi-race|Alaska Native|38003573|Race|1970-1-1|2099-12-31|\N|1
 AIANSpecific_AmericanIndian|1585600|ppi-race|American Indian|38003572|Race|1970-1-1|2099-12-31|\N|1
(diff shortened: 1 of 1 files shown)


rest-api/etl/source_to_concept_map.csv
```diff
@@ -24,7 +26,7 @@ BlackSpecific_Liberian|1585618|ppi-race|Liberian|38003598|Race|1970-1-1|2099-12-
 BlackSpecific_Nigerian|1585624|ppi-race|Nigerian|38003598|Race|1970-1-1|2099-12-31|\N|1
 BlackSpecific_Somali|1585619|ppi-race|Somali|38003598|Race|1970-1-1|2099-12-31|\N|1
 BlackSpecific_SouthAfrican|1585620|ppi-race|South African|38003598|Race|1970-1-1|2099-12-31|\N|1
-Hispanic_HispanicSpecific|1586156|ppi-race|Hispanic, Latino, or Spanish (For example: Columbian, Cuban, Dominican, Mexican or Mexican American, Puerto Rican, Salvadoran, etc.) *|0|\N|\N|\N|\N|2
+Hispanic_HispanicSpecific|1586156|ppi-race|Hispanic, Latino, or Spanish (For example: Columbian, Cuban, Dominican, Mexican or Mexican American, Puerto Rican, Salvadoran, etc.) *|38003563|Ethnicity|1970-1-1|2099-12-31|\N|2
 HispanicSpecific_Mexican|1585345|ppi-race|Mexican or Mexican American|0|\N|\N|\N|\N|1
```
danqrodney: Should HispanicSpecific_Mexican and HispanicSpecific_PuertoRican also have ethnicity set?

rest-api/etl/source_to_concept_map.csv
```diff
@@ -1,3 +1,5 @@
+SexAtBirth_Male|1585846|ppi-sex|MALE|8507|Gender|1970-1-1|2099-12-31|\N|1
+SexAtBirth_Female|1585847|ppi-sex|FEMALE|8532|Gender|1970-1-1|2099-12-31|\N|1
```
danqrodney: Do we need to worry about SexAtBirth_Intersex and SexAtBirth_SexAtBirthNoneOfThese?

rest-api/etl/source_to_concept_map.csv
```diff
@@ -24,7 +26,7 @@ BlackSpecific_Liberian|1585618|ppi-race|Liberian|38003598|Race|1970-1-1|2099-12-
 BlackSpecific_Nigerian|1585624|ppi-race|Nigerian|38003598|Race|1970-1-1|2099-12-31|\N|1
 BlackSpecific_Somali|1585619|ppi-race|Somali|38003598|Race|1970-1-1|2099-12-31|\N|1
 BlackSpecific_SouthAfrican|1585620|ppi-race|South African|38003598|Race|1970-1-1|2099-12-31|\N|1
-Hispanic_HispanicSpecific|1586156|ppi-race|Hispanic, Latino, or Spanish (For example: Columbian, Cuban, Dominican, Mexican or Mexican American, Puerto Rican, Salvadoran, etc.) *|0|\N|\N|\N|\N|2
```
danqrodney: So it seems that these are more specific codes for hispanic race, but we are missing two top-level race codes in this file: WhatRaceEthnicity_Hispanic and WhatRaceEthnicity_RaceEthnicityNoneOfThese.

It seems like the former should map to hispanic ethnicity, and should be included in this file. I'm less sure about how to handle the second.

danqrodney: Also: I think(?) we may need to keep this row as it was before, and add additional row(s) with source_vocabulary_id set to "ppi-ethnicity" or something.

We want to be able to set race_source_concept_id to the concept IDs for these rows, while setting ethnicity_concept_id to the right ethnicity... those seem like they have to be separate rows?

rest-api/etl/source_to_concept_map.csv
cukarthik: Right now we don't have a mapping for this. We'll need to introduce them. I'll make that change when we create them.

rest-api/etl/source_to_concept_map.csv
cukarthik: I see. So two rows with different vocabulary ids.

PR #714, opened 2018-05-24
DA-649 add profiling

Seeker1911:


Seeker1911:
has been deployed and tested in sandbox.
https://all-of-us-rdr-sandbox.appspot.com/_ah/stats/

danqrodney:
Is there documentation anywhere suggesting whether it's reasonable to use this in production? (I'm not sure how much of a hit it will incur to performance...)

We may want to hold off on merging this until we can prove it won't hurt performance in prod; or we can make it so it only takes effect in nonprod.

Note that I think this appstats stuff only helps with RPC monitoring, which won't necessarily tell us where all the time is being spent in the participant summary API. (I suspect we're spending a lot of CPU time just sticking things in dictionaries.)

Seeker1911:
So it does have an affect, we could use it in non_prod only I suppose. But the latency hit is usually quite small. You can actually see it's own effect if you drill down to individual requests. If you click on the request you can see the actual functions called and the time for each.
About half way down this post. http://googleappengine.blogspot.com/2010/03/easy-performance-profiling-with.html

ci/license_exceptions.txt
```diff
@@ -15,3 +15,7 @@ certifi
 # ordereddict is MIT according to https://pypi.python.org/pypi/ordereddict but
 # apparently the metadata is lacking (it shows up as UNKNOWN)
 ordereddict
+
+# pyzmq/zmq is BSD. Metadata missing.
+pyzmq
```
danqrodney: These aren't in requirements.txt? Remove them?

rest-api/app_base.yaml
```diff
@@ -8,24 +8,8 @@ api_version: 1
 
 builtins:
 - deferred: on
+- appstats: on
```
danqrodney: Shouldn't this be in app_nonprod.yaml too?

rest-api/appengine_config.py
```diff
@@ -9,3 +9,10 @@
 vendor.add('lib-common')
 vendor.add('appengine-mapreduce/python/src')
 vendor.add('appengine-mapreduce/python/test')
+
+
+def webapp_add_wsgi_middleware(app):
```
danqrodney: I don't fully understand where this gets called from, but I suspect it will get called in prod, too... this is the thing that's potentially going to slow things down in prod, not the actual handler in app.yaml.

We should either add an if statement here to check the environment (and just return app in prod), or add a build step so this code doesn't get included when deploying to prod.

danqrodney: (If this only gets invoked when appstats is set to on, maybe this isn't necessary once you move the appstats: on to app_nonprod.yaml... but I'd want to confirm that's the case.)

ci/license_exceptions.txt
Seeker1911: done

rest-api/app_base.yaml
Seeker1911: no, the files still get concatenated.

Seeker1911: okay, after reading below comment, I moved builtins to the environment files and removed appstats from prod.

rest-api/appengine_config.py
Seeker1911: according to google, appstats looks for this function name, so it shouldnt get called in prod, but I'll work on a safety check to ensure it.

ci/license_exceptions.txt
danqrodney: This is still here. Did you mean to remove it?

rest-api/test.yaml
```diff
@@ -1,43 +1,45 @@
...
-  secure: always
-- url: /_ah/queue/deferred
-  script: google.appengine.ext.deferred.deferred.application
-  login: admin
+threadsafe: true
+runtime: python27
+api_version: 1
+
+builtins:
+- deferred: on
```
danqrodney: If you're not enabling appstats here (which seems fine?) you probably don't need the stats handler below.

ci/license_exceptions.txt
Seeker1911: done

rest-api/test.yaml
Seeker1911: done

PR #716, opened 2018-05-30
[DA-645] Making ETL use same vocabulary that curation uses.

danqrodney:
Stripping concept_relationships down to just "Maps to" since that's all we need for the ETL.

Using same sort of logic for populating gender and ethnicity as we use for race in the ETL.


Seeker1911:
👍🏻

PR #717, opened 2018-05-30
Don't use service account when setting up database for ETL. (pmi-ops …

danqrodney:
…account works fine, and avoids issue in prod.)


danqrodney:
TBH, I'm not sure. But when I ran the ETL script in prod it would tell me that my credentials were invalid. This hasn't happened in the past in prod, and still doesn't happen in sandbox. But I couldn't figure out why -- all the ACLs seem OK. 

Anyway, turns out that using a service account isn't actually important -- I'm not sure why we were using it here in the first place -- so I just removed that.

PR #719, opened 2018-05-31
DA-648 making a csv exporter

Seeker1911:


rest-api/api/base_api.py
```diff
@@ -132,7 +134,7 @@ def _make_query(self):
     return Query(field_filters, order_by, max_results, pagination_token,
                  include_total=include_total)
 
-  def _make_bundle(self, results, id_field, participant_id):
+  def _make_bundle(self, results, id_field, participant_id, csv=False):
```
danqrodney: So _make_bundle constructs a JSON object with an array of entries, each one containing a resource JSON dictionary, which gets returned as JSON to the client. My assumption was that we were going to change the whole response to be CSV, not a JSON object containing CSV... which would be what this code would do I think. So I don't think you want to mess with _make_bundle.

Instead, just have a whole new _make_csv(results) method that makes a CSV writer, calls to_client_csv on each object (passing in the writer), and then returns a Response for the whole thing. You'd call that from participant_summary_api.get() when output=csv.

If you want to stream the results to the client as they are written, this may give some indication of how to do that:

https://stackoverflow.com/questions/32608265/streaming-a-generated-csv-with-flask

A baby step I guess that would be a bit easier to implement would be to have the "resource" in each item be a JSON array of items, rather than a JSON dict like today. I guess we could test how much faster *that* makes things run, if we didn't want to return actual CSV. But it seems a bit kooky. :)

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +380,34 @@ def to_client_json(self, model):
...
+    def iter_csv(data):
+      line = StringIO.StringIO()
+      writer = csv.writer(line)
+      for csv_line in data:
+        writer.writerow(csv_line)
+        # line.seek(0)
+        # yield line.read()
+        # line.truncate(0)
+
+    response = Response(iter_csv(model), mimetype='text/csv')
```
danqrodney: A Response would need to be returned from participant_summary_api.get()... we're not making one Response per result, if that makes sense.

Similarly, there should only be one CSV writer, which would get passed in to to_client_csv for each participant summary.

rest-api/api/base_api.py
```diff
@@ -47,7 +47,9 @@ def get(self, id_=None, participant_id=None):
                        (self.dao.model_type.__name__, id_, participant_id))
     return self._make_response(obj)
 
-  def _make_response(self, obj):
+  def _make_response(self, obj, csv=False):
```
danqrodney: I wouldn't mess with this. See below.

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +380,34 @@ def to_client_json(self, model):
...
+  def to_client_csv(self, model):
+    from flask import Response
+    import StringIO
+    # from unicode_csv import UnicodeWriter
+
+
+    def iter_csv(data):
+      line = StringIO.StringIO()
+      writer = csv.writer(line)
+      for csv_line in data:
```
danqrodney: Each participant summary returned should be a single line of CSV. We shouldn't iterate through multiple lines for a single summary... which is what it sounds like this was going to do.

rest-api/api/base_api.py
danqrodney: Note you'll need the writer to first spit out the column headers, which means figuring out what order columns are going to be returned and making sure that matches between the headers and your to_client_csv code.

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +380,34 @@ def to_client_json(self, model):
...
 
+  def to_client_csv(self, model):
+    from flask import Response
+    import StringIO
+    # from unicode_csv import UnicodeWriter
+
+
+    def iter_csv(data):
+      line = StringIO.StringIO()
+      writer = csv.writer(line)
```
danqrodney: Another thing is: the default CSV libraries don't read or write Unicode data. We need to return Unicode properly... see:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/unicode_csv.py#L12

for how we're handling that in other contexts.

Seeker1911:
@danqrodney have a look but it's not done yet, I mostly need @jt2k to look at the `to_client_csv` and see if that will work for him. Lot's of `#TODO's` which are mostly questions. Do you want the binary or do you want human readable `SUBMITTED` etc. I'll attach an example output from test data as well.

Seeker1911:
[test.txt](https://github.com/all-of-us/raw-data-repository/files/2082088/test.txt)

rdr_common/code_constants.py
```diff
@@ -74,3 +74,94 @@
...
+ 'Hist PPI Survey Completion Date',
+ 'Meds PPI Survey Complete',
+ 'Meds PPI Survey Completion Date',
+ 'Family PPI Survey Complete',
+ 'Family PPI Survey Completion Date',
+ 'Access PPI Survey Complete',
+ 'Access PPI Survey Completion Date',
+ 'Physical Measurements Status',
+ 'Physical Measurements Completion Date',
+ 'Paired Site',
```
danqrodney: So awardee isn't needed? (I guess it's unnecessary if we only return one awardee's data at a time?)

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    else:
+      line.append(UNSET)
+    line.append(row.samplesToIsolateDNA)
+    line.append(row.biospecimenStatus)  #TODO: CHECK 'BIOSPECIMEN'
+    for i in ps_sample_status_collection:
+      line.append(getattr(row, i))
+      line.append(getattr(row, i+'Time'))
+    if row.biospecimenSourceSiteId:  #TODO: IS BIOSOURCESITEID CORRECT ?
+      site = self.site_dao.get(row.biospecimenSourceSiteId)
+      line.append(site.siteName)  #TODO: SITENAME OR GOOGLEGROUP
```
danqrodney: googleGroup

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -217,6 +217,50 @@ def post_demographics_questionnaire(self,
...
+      self.post_demographics_questionnaire(participant_id, questionnaire_id, time=when, **answers)
+      return participant
+
+    [setup_participant(t1) for _ in range(5)]
+    response = self.send_get_csv('ParticipantSummary?_output=csv')
+    response = response.split(',')
+    headers = response[:73]
+    # remove '/r/n' from end of list the easy way.
+    headers[72] = 'Biospecimens Site'
+    self.assertEqual(headers, ps_full_data_headers)
```
danqrodney: Add tests on the results.

In particular: I would make sure we have at least one row where each field is set to something. To check to make sure the headers match up with the values, I would use a CSV dict reader to read the response, and verify that it produces the expected dictionaries for whatever rows are returned.

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    line.append(row.firstName)
+    line.append(row.dateOfBirth)
+    if row.languageId:
+      language = self.code_dao.get(row.languageId).value
+      line.append(language)
+    else:
+      line.append(UNSET)
+    line.append(row.enrollmentStatus)
+    if row.consentForStudyEnrollment == 'SUBMITTED':
+      line.append(1)
```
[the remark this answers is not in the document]
Seeker1911: HPRO only uses the binary 1 or 0. This is where I needed @jt2k advice. It may be better to just return the string values (I think we are in `to_client_json` and let client alter data as they see fit.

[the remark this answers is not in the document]
Seeker1911: @shyamnaru opinion?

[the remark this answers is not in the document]
shyamnaru: If we are in `to_client_json` returning string values is fine.

FYI: If you want to return 1 or 0, for this perticular case healthpro checks for `SUBMITTED` string and asigns a value of 1 and for all others it's 0. https://github.com/vanderbilt/pmi-drc-hpo/blob/develop/src/Pmi/WorkQueue/WorkQueue.php#L364

rdr_common/code_constants.py
```diff
@@ -74,3 +74,94 @@
 UNSET = 'UNSET'
 UNMAPPED = 'UNMAPPED'
 BASE_VALUES = [UNSET, UNMAPPED, PMI_SKIP_CODE]
+
+ps_full_data_headers = [
```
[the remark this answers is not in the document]
Seeker1911: done

rdr_common/code_constants.py
Seeker1911: no, and HPRO doesnt return awardee

rest-api/dao/participant_summary_dao.py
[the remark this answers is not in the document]
Seeker1911: returning string values iot be explicit

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    else:
+      line.append(0)
+    line.append(row.consentForStudyEnrollmentTime)
+    line.append(row.consentForElectronicHealthRecords)  #TODO: EHR CONSENT ( 1 || 0)
+    line.append(row.consentForElectronicHealthRecordsTime)
+    line.append(row.consentForCABoR)  #TODO: CABOR (1 || 0)
+    line.append(row.consentForCABoRTime)
+    if (row.withdrawalStatus == WithdrawalStatus.NO_USE or
```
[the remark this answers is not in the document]
Seeker1911: fixed

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    line.append(row.zipCode)
+    line.append(row.email)
+    line.append(row.phoneNumber)
+    if row.sexId:
+      sex = self.code_dao.get(row.sexId).value
+      line.append(sex)
+    else:
+      line.append(UNSET)
+
+    line.append(row.genderIdentityId)
```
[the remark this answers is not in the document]
Seeker1911: yup

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    line.append(row.questionnaireOnMedicationsTime)
+    line.append(row.questionnaireOnFamilyHealth)
+    line.append(row.questionnaireOnFamilyHealthTime)
+    line.append(row.questionnaireOnHealthcareAccess)
+    line.append(row.questionnaireOnHealthcareAccessTime)
+    line.append(row.physicalMeasurementsStatus)  #TODO: IS STATUS THE CORRECT FIELD
+    line.append(row.physicalMeasurementsTime)
+    if row.siteId:
+      site = self.site_dao.get(row.siteId)
+      line.append(site.siteName)  #TODO: SITENAME OR GOOGLEGROUP
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    else:
+      line.append(UNSET)
+    if row.organizationId:
+      organization = self.organization_dao.get(row.organizationId)
+      line.append(organization.displayName)  #TODO: DISPLAY NAME OR EXTERNALID
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,118 @@ def to_client_json(self, model):
...
+    else:
+      line.append(UNSET)
+    if row.physicalMeasurementsFinalizedSiteId:
+      site = self.site_dao.get(row.physicalMeasurementsFinalizedSiteId)
+      line.append(site.siteName)  #TODO: SITENAME OR GOOGLEGROUP
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/dao/participant_summary_dao.py
Seeker1911: done

rest-api/test/unit_test/api_test/participant_summary_api_test.py
Seeker1911: done

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,117 @@ def to_client_json(self, model):
...
+    writer = unicode_csv.UnicodeWriter(csv_data)
+    headers = ps_full_data_headers
+    writer.writerow(headers)
+    for row in results.items:
+      line = self.to_client_csv(row)
+      writer.writerow(line)
+    return csv_data.getvalue()
+
+  def to_client_csv(self, row):
+    """this method must append values in the same order as CODE_CONSTANTS.PS_FULL_HEADERS"""
```
danqrodney: Nit: "this" can be capitalized... CODE_CONSTANTS.PS_FULL_HEADERS shouldn't be.

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,117 @@ def to_client_json(self, model):
...
+      line.append(education)
+    else:
+      line.append(UNSET)
+    if row.numCompletedBaselinePPIModules == 3:
+      line.append(1)
+    else:
+      line.append(0)
+    line.append(row.numCompletedPPIModules)
+    line.append(row.questionnaireOnTheBasics)
+    line.append(row.questionnaireOnTheBasicsTime)
```
danqrodney: What format do all these timestamps get returned as? You may want to be explicit about formatting them... e.g.:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L367

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -217,6 +219,59 @@ def post_demographics_questionnaire(self,
...
+
+    [setup_participant(t1) for _ in range(5)]
+    response = self.send_get_csv('ParticipantSummary?_output=csv')
+    reader = StringIO.StringIO(response)
+    row = UnicodeDictReader(reader)
+    row = row.next()
+    headers = [k for k, _ in row.iteritems()]
+    for i in headers:
+      assert i in ps_full_data_headers
+    self.assertEqual(row['City'], 'Austin')
```
danqrodney: You should be able to just make a big expectedRow dict with exactly what you expect in the response, and assert it matches row here... or if you prefer, read in CSV from a file. Either way, it will need to check all of the columns.

Seeker1911: done

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,77 @@ def to_client_json(self, model):
...
+      else:
+        for k in line.keys():
+          if k not in headers_list:
+            headers_list.append(k)
+            write_headers = True
+
+      for i in headers_list:
+        values_list.append(line.get(i))
+
+      if write_headers:
```
danqrodney: Why not just do this check outside the loop and write the row before you get in here?

We'd want to return the headers even if we had no results... (we should have a test for that.)

rest-api/dao/participant_summary_dao.py
```diff
@@ -379,6 +381,77 @@ def to_client_json(self, model):
...
+    write_headers = True
+    for row in results.items:
+      line = self.to_client_csv(row)
+      values_list = []
+
+      if len(headers_list) == 0:
+        for k in line.keys():
+          headers_list.append(k)
+      else:
+        for k in line.keys():
```
danqrodney: So my assumption has been that a significant chunk of what's slowing down our current JSON generation is hashing dozens of keys for every result returned from our API.My hope was by returning CSV this wouldn't be necessary anymore -- we'd just append values in a fixed order that matches the headers, and have tests to verify we did it right. Does that work? Have you tested how things perform this way vs. the other way?

But if we were going to do things this way, it seems like we could just call to_client_json and get rid of to_client_csv, and then look things up in the dict when appending to the CSV. Or barring that, reuse format_json_* instead of making format_csv_*....

PR #726, opened 2018-06-15
PD-1752 update stable for awardee service account

Seeker1911:


rest-api/config/config_stable.json
```diff
@@ -9,6 +9,13 @@
       "roles": ["healthpro"],
       "whitelisted_appids": ["pmi-hpo-test"]
     },
+    "awardee-pitt@pmi-drc-api-test.iam.gserviceaccount.com": {
```
danqrodney: Is this the SA that's going to get used in stable, too? (Just wondering since it's pmi-drc-api-test)

Seeker1911: Good point, let me go fix that.

PR #734, opened 2018-06-27
DA-639 add spanish fields

Seeker1911:


rest-api/model/site.py
```diff
@@ -19,8 +19,10 @@ class Site(Base):
   enrollingStatus = Column('enrolling_status', Enum(EnrollingStatus))
   digitalSchedulingStatus = Column('digital_scheduling_status', Enum(DigitalSchedulingStatus))
   scheduleInstructions = Column('schedule_instructions', String(2048))
+  scheduleInstructions_ES = Column('schedule_instructions_es', String(2048))
```
calbach: Seems this should probably be lower camelCased like the other fields

Seeker1911: I considered that, but I don't want the `es` part to be lost in the name. I think it needs to be a little obvious. Ideas ?

calbach:
scheduleInstructionsES also seems ok to me.

PR #736, opened 2018-07-05
[DA-659] Changing CSV importer to use UnicodeDictReader;

danqrodney:
changing site_dao to not cast instructions to ascii strings; updating sites.

Adding test of awardee endpoint to automated tests.


danqrodney:
In the interest of getting this out into staging, I'm going to submit this TBR.

calbach:
lgtm

PR #748, opened 2018-07-18
setup alembic migration for invalid  questionnaire

Seeker1911:


rest-api/model/questionnaire.py
```diff
@@ -15,6 +16,8 @@ class QuestionnaireBase(object):
   # The JSON representation of the questionnaire provided by the client.
   # Concepts and questions can be be parsed out of this for use in querying.
   resource = Column('resource', BLOB, nullable=False)
+  isValid = Column('is_valid', Enum(ValidQuestionnaireStatus),
```
danqrodney: You could rename this to just "status" and maybe rename ValidQuestionnaireStatus to QuestionnaireDefinitionStatus I guess. I could imagine other statuses for questionnaires besides INVALID  / VALID.

Alternatively, just make it a Boolean "invalid" flag.

rest-api/participant_enums.py
```diff
@@ -37,6 +37,11 @@ class QuestionnaireStatus(messages.Enum):
   SUBMITTED_NO_CONSENT = 2
   SUBMITTED_NOT_SURE = 3
```
danqrodney: Do you want to add SUBMITTED_INVALID here while you're at it?

PR #752, opened 2018-07-24
DA-675 set and reject invalid questionnaires

Seeker1911:


Seeker1911:
@danqrodney what I'm trying to produce with the new test (test_backfill_invalidates_consent) is create multiple versions of ehr_consent questionnaire where one is valid and one is invalid and then send consent for one participant for both quesitonnaires and one participant for only the invalid one and then add the backfill sql we plan to run to test that only the participant that signed the bad questionnaire only will have ehr_consent set to submitted.invalid

Seeker1911:
update: removing the backfill tests to just manually test the sql in lesser environments as it's a one-off job

PR #757, opened 2018-08-02
DA-688 ehr upload consent docs

Seeker1911:
Pending one question on what we will do with participants who are not paired at site level, but other than that, this is done.

### DIFF
--- a/rest-api/tools/ehr_upload.py
+++ b/rest-api/tools/ehr_upload.py
@@ -3,6 +3,7 @@ and upload to a new organization bucket by the same name.
 eg: tools/ehr_upload.sh --organization AZ_TUCSON --bucket ptc-uploads-pmi-drc-api-prod
 """
 import subprocess
+import shlex
 from dao import database_factory
 from main_util import get_parser, configure_logging
 from sqlalchemy import text
@@ -35,13 +36,14 @@ def get_participants_under_sites(session, organization, source_bucket, destinati
           participant) + "/* " + "gs://" + destination_bucket + "/Participant/" + \
                  google_group + "/P" + str(participant) + "/"
 
-        system_call = subprocess.call(gsutil, shell=True)
-        if system_call == 0:
-          print "Successfully moved folder " + google_group + '/' + str(participant)
-        elif system_call == 1:
-          print "There was an error moving folder " + google_group + '/' + str(participant)
+        gsutil = shlex.split(str(gsutil))
+        system_call = subprocess.Popen(gsutil)
+        system_call.communicate()[0]
+        if system_call.returncode == 0:
+          print "Successfully moved folder " + google_group + '/P' + str(participant)
         else:
-          print "System error message: " + system_call
+          print "There was an error moving folder " + google_group + '/P' + str(participant)
+          print "return code is : " + str(system_call.returncode)
     else:
       print "No participants paired with sites found for organization: " + organization
   finally:
@@ -64,13 +66,14 @@ def get_participants_without_site_pairing(session, organization, source_bucket, 
           participant) + "/* " + "gs://" + destination_bucket + "/Participant/" + \
                  "no_site_pairing" + "/P" + str(participant) + "/"
 
-        system_call = subprocess.call(gsutil, shell=True)
-        if system_call == 0:
+        gsutil = shlex.split(str(gsutil))
+        system_call = subprocess.Popen(gsutil)
+        system_call.communicate()[0]
+        if system_call.returncode == 0:
           print "Successfully moved folder " + "no_site_pairing" + '/P' + str(participant)
-        elif system_call == 1:
-          print "There was an error moving folder " + "no_site_pairing" + '/P' + str(participant)
         else:
-          print "System error message: " + system_call
+          print "There was an error moving folder " + "no_site_pairing" + '/P' + str(participant)
+          print "return code is : " + str(system_call.returncode)
     else:
       print "No participants found that are not paired with sites and paired with organization: " \
             + organization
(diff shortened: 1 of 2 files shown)


Seeker1911:
@calbach have a look when you get a chance. One thing, configurator service account doesn't authenticate for me. Might want to try it yourself. 
```--organization
AZ_TUCSON_BANNER_HEALTH
--destination_bucket
ptc-uploads-pmi-drc-api-sandbox
--source_bucket
ptc-uploads-pmi-drc-api-sandbox
--account
michael.mead@pmi-ops.org
--project
all-of-us-rdr-sandbox```

**Comment by calbach** [comment]
Per slack I tried this 10AM PDT Friday and also got a 403 - has the code changed since then?

**Comment by Seeker1911** [comment]
The 403 on Friday was for `/run_client` although the two may be connected.

File: rest-api/tools/ehr_upload.py [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+  finally:
+    cursor.close()
+    get_participants_without_site_pairing(session, organization, source_bucket, destination_bucket)
+
+
+def get_participants_without_site_pairing(session, organization, source_bucket, destination_bucket):
+  sql = """
+    select participant_id from participant p
+    where p.site_id is NULL and p.organization_id in (
+    select organization_id from organization where external_id = '{}');
```
calbach: Out of curiosity, why a nested select rather than a right JOIN here?

rest-api/tools/ehr_upload.py
```diff
@@ -0,0 +1,78 @@
...
+   """.format(organization)
+  cursor = session.execute(text(sql))
+  try:
+    results = cursor.fetchall()
+    if results:
+      results = [int(i) for i, in results]
+      for participant in results:
+        gsutil = "gsutil -m cp gs://" + source_bucket + "/Participant/P" + str(
+          participant) + "/* " + "gs://" + destination_bucket + "/Participant/" + \
+                 "no_site_pairing" + "/P" + str(participant)
```
calbach: same as above

rest-api/tools/ehr_upload.py
```diff
@@ -0,0 +1,78 @@
...
+    cursor.close()
+
+
+if __name__ == '__main__':
+  configure_logging()
+  parser = get_parser()
+  parser.add_argument('--organization', help='The organization to find participants and sites for',
+                      required=True)
+  parser.add_argument('--source_bucket', help='The bucket to read from in one env.'
+                      ' i.e. ptc-uploads-pmi-drc-api-prod', required=True)
```
calbach: Is it ok for these to have the same value? Is that typical?

rest-api/tools/ehr_upload.py
```diff
@@ -71,7 +84,7 @@ def get_participants_without_site_pairing(session, organization, source_bucket,
   parser.add_argument('--organization', help='The organization to find participants and sites for',
                       required=True)
   parser.add_argument('--source_bucket', help='The bucket to read from in one env.'
-                      ' i.e. ptc-uploads-pmi-drc-api-prod', required=True)
+                      ' i.e. ptc-uploads-pmi-drc-api-sandbox', required=True)
   parser.add_argument('--destination_bucket', help='The bucket to write to in one env.'
                       'i.e. ptc-uploads-pmi-drc-api-prod', required=True)
```
calbach: The example values here should probably be for the same environment right?

calbach:
Also please check out the the codacy complaints, they look legitimate.

rest-api/tools/ehr_upload.py
```diff
@@ -0,0 +1,78 @@
...
+    if results:
+      for participant, google_group in results:
+        gsutil = "gsutil -m cp gs://" + source_bucket + "/Participant/P" + str(
+          participant) + "/* " + "gs://" + destination_bucket + "/Participant/" + \
+          google_group + "/P" + str(participant)
+
+        system_call = subprocess.call(gsutil, shell=True)
+        if system_call == 0:
+          print "Successfully moved folder " + google_group + '/' + str(participant)
+        elif system_call == 1:
```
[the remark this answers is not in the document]
Seeker1911: done

rest-api/tools/ehr_upload.py
Seeker1911: sure

rest-api/tools/ehr_upload.py
Seeker1911: I changed it.

Seeker1911:
regarding subprocess.call security:
A. it's only a security issue when accepting untrusted user input (i.e. command line app)
B. I changed it anyway using a combination of Popen/communicate  and using the shlex module to interpolate the string in a friendly way for subprocess. This no longer creates a shell to run the command and is safer. The communicate function allows us to read the returncode from Popen.

Seeker1911:
@calbach I know you've given approval but considering the amount of refactoring I did, may be best practice to take a 2nd look.

rest-api/tools/ehr_upload.py
```diff
@@ -35,13 +36,14 @@ def get_participants_under_sites(session, organization, source_bucket, destinati
           participant) + "/* " + "gs://" + destination_bucket + "/Participant/" + \
                  google_group + "/P" + str(participant) + "/"
 
-        system_call = subprocess.call(gsutil, shell=True)
-        if system_call == 0:
-          print "Successfully moved folder " + google_group + '/' + str(participant)
-        elif system_call == 1:
-          print "There was an error moving folder " + google_group + '/' + str(participant)
+        gsutil = shlex.split(str(gsutil))
```
calbach: Starting with a string and then parsing is a bit roundabout, since you define the string. I'd just define the process arguments as an array to begin with.

PR #759, opened 2018-08-02
Upgrade from unsafe cryptography package

calbach:
This package comes in automatically with pyOpenSSL. I couldn't figure out where that package actually comes from (it may already be installed on the OS). That particular package doesn't have a better version, so we need to explicitly upgrade this dep to avoid the vulnerability in 2.2.2.


calbach:
Likely dropping this, rerunning without the cache worked so I'm hoping subsequent runs will pass.

calbach:
This worked.

PR #760, opened 2018-08-03
[DA-699] Update sites

calbach:


Seeker1911:
Do we want the html in the sheets?

calbach:
According to Ashley this is expected - I don't know where it's being rendered.

PR #767, opened 2018-08-13
Mm/edit biobank orders table

Seeker1911:
update db to allow for editing/cancelling biobank orders.


Seeker1911:
https://precisionmedicineinitiative.atlassian.net/browse/DA-700
@danqrodney

rest-api/model/biobank_order.py
```diff
@@ -50,6 +54,12 @@ class BiobankOrder(Base):
   # in the resulting JSON.
   finalizedUsername = Column('finalized_username', String(255))
 
+  # cancelled finalized order may still be shipped to biobank for destruction
+  orderStatus = Column('order_status', Enum(BiobankOrderStatus))
+  cancelledTime = Column('cancelled_time', UTCDateTime)
```
danqrodney: Change to last_modified. This should get updated whenever we modify an order in place.

rest-api/model/biobank_order.py
```diff
@@ -50,6 +54,12 @@ class BiobankOrder(Base):
...
+  # a cancelled or edited order must have a reason
```
danqrodney: You should indicate whether this gets set on the new or the old order.

rest-api/participant_enums.py
```diff
@@ -62,6 +62,13 @@ class OrderStatus(messages.Enum):
   PROCESSED = 3
   FINALIZED = 4
 
+class BiobankOrderStatus(messages.Enum):
+  """ The status of a biobank order: finalized/amended/cancelled """
+  UNSET = 0
+  FINALIZED = 1
```
danqrodney: So we may instead of UNSET = 0 want to have FINALIZED = 0 and get rid of UNSET... we need to make sure that existing orders are marked as FINALIZED.

rest-api/model/biobank_order.py
Seeker1911: done

rest-api/model/biobank_order.py
Seeker1911: done

rest-api/participant_enums.py
Seeker1911: as discussed, default to null, unset is back.

rest-api/model/biobank_order.py
```diff
@@ -50,6 +54,14 @@ class BiobankOrder(Base):
   # in the resulting JSON.
   finalizedUsername = Column('finalized_username', String(255))
 
+  # cancelled finalized order may still be shipped to biobank for destruction
```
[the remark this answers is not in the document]
Seeker1911: done

PR #768, opened 2018-08-16
Mm/amend biobank orders

Seeker1911:
some minor changes to table and versioning before we move on to cancelling/amending


rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,28 @@ def setUp(self):
...
+                                         filename='biobank_order_2.json')
+    result = self.send_post(self.path, order_json)
+    full_order_json = load_biobank_order_json(self.participant.participantId,
+                                              filename='biobank_order_1.json')
+    _strip_fields(result)
+    _strip_fields(full_order_json)
+    self.assertEquals(full_order_json, result)
+
+
+    # @TODO: in progress.
```
danqrodney: We definitely should have tests of the new PUT method before checking this in. You could just merge this in with the next PR I guess, so you can test out the version stuff too.

rest-api/dao/biobank_order_dao.py
```diff
@@ -94,6 +94,8 @@ def _order_as_dict(self, order):
     return result
 
   def insert_with_session(self, session, obj):
+    if obj.version == None:
+      obj.version = 1
```
danqrodney: I would make this unconditional. Clients shouldn't be able to dictate the version.

You're also going to need some logic to bump the version during updates. (Though you could add this later.)

rest-api/alembic/versions/0e92151ebd4a_amend_biobank_orders.py
```diff
@@ -48,6 +48,7 @@ def upgrade_rdr():
     op.add_column('biobank_order', sa.Column('restored_site_id', sa.Integer(), nullable=True))
     op.add_column('biobank_order', sa.Column('restored_time', model.utils.UTCDateTime(), nullable=True))
     op.add_column('biobank_order', sa.Column('restored_username', sa.String(length=255), nullable=True))
+    op.add_column('biobank_order', sa.Column('version', sa.Integer(), nullable=False))
```
danqrodney: So since all of our existing DBs should probably have run through these migrations, the change you're doing here -- deleting one migration and merging it into here -- is *probably* safe. But I wouldn't recommend it. If you don't have any schema changes to make, I'd just revert all the Alembic changes in this PR.

rest-api/alembic/versions/e44e303ae759_add_biobank_version.py
```diff
@@ -1,58 +0,0 @@
-"""add biobank version
```
danqrodney: See above, I'm not sure I'd delete this.

rest-api/alembic/versions/4aa79d5ac86e_add_biobank_history.py
```diff
@@ -1,7 +1,7 @@
 """add biobank history
 
 Revision ID: 4aa79d5ac86e
-Revises: e44e303ae759
+Revises: 0e92151ebd4a
```
danqrodney: See above.

rest-api/api/base_api.py
```diff
@@ -206,9 +211,27 @@ def put(self, id_, participant_id=None):
     self._do_update(m)
     return self._make_response(m)
 
+  def patch(self, id_, participant_id=None):
+    """Handles a PATCH request; the current object must exist, and will be amended
+
+    Args:
+      id: The id of the object to update.
+      participant_id: The ancestor id (if applicable). # @TODO: REVIEW THIS <
```
danqrodney: Address TODO? This looks OK I guess.

rest-api/api/base_api.py
```diff
@@ -206,9 +211,27 @@ def put(self, id_, participant_id=None):
...
+    """
+    resource = request.get_json(force=True)
+    etag = request.headers.get('If-Match')
+    if not etag:
+      raise BadRequest("If-Match is missing for PUT request")
```
danqrodney: PUT -> PATCH

rest-api/api/base_api.py
```diff
@@ -179,6 +179,12 @@ def _get_model_to_update(self, resource, id_, expected_version, participant_id=N
       return self.dao.from_client_json(
           resource, id_=id_, expected_version=expected_version, client_id=app_util.get_oauth_id())
 
+  def _get_patch_model_to_update(self, resource, id_, expected_version, participant_id=None):
+    #pylint: disable=unused-argument
+    # Children of participants accept a participant_id parameter to from_client_json; others don't.
+    if participant_id is not None:
+      return self.dao.get_with_children(id_)
```
danqrodney: What's the point of the participant_id is not None check? Should there be an else?

It's worth noting that _get_model_to_update parses JSON resources from the request, rather than reading from the database... I'm not sure why we need another method here, but this is doing something pretty different and might be named differently?

rest-api/api/biobank_order_api.py
```diff
@@ -15,3 +15,11 @@ def post(self, p_id):
   @auth_required(PTC_AND_HEALTHPRO)
   def get(self, p_id, bo_id=None):  # pylint: disable=unused-argument
     return super(BiobankOrderApi, self).get(bo_id)
+
+  @auth_required(HEALTHPRO)
```
danqrodney: So we're supporting PUT and PATCH? Which one will get used under what circumstances?

rest-api/api_util.py
```diff
@@ -105,6 +105,15 @@ def get_site_id_from_google_group(obj, site_dao):
       return site.siteId
   return None
 
+def get_site_id_by_site_value(obj):
```
danqrodney: Unused?

rest-api/dao/biobank_order_dao.py
```diff
@@ -265,8 +276,43 @@ def from_client_json(self, resource_json, participant_id=None, client_id=None):
...
+      order.orderStatus = BiobankOrderStatus.CANCELLED
+      site = site_dao.get_by_google_group(resource.cancelledInfo.site.value)
+      if site:
+        order.cancelledSiteId = site.siteId
+      else:
+        raise BadRequest('A valid google_group is required for cancelledInfo.site.value')
+      order.cancelledUsername = resource.cancelledInfo.author.value
+      order.cancelledTime = clock.CLOCK.now()
+      order.amendedReason = resource.amendedReason
+    elif order.status.upper() == 'RESTORED':
```
danqrodney: One thing that's a little weird here is we're going to have "RESTORED" in our API coming in for restore operations, resulting in "UNSET" in the response... is that OK?

rest-api/dao/biobank_order_dao.py
```diff
@@ -265,8 +276,43 @@ def from_client_json(self, resource_json, participant_id=None, client_id=None):
...
+    if resource.amendedInfo:
+      order.amendedUsername, order.amendedSiteId = self._parse_handling_info(resource.amendedInfo)
+    order.version = expected_version
     return order
 
+  @classmethod
+  def _handle_new_status(cls, order, resource):
+    site_dao = SiteDao()
+    order.status = resource.status
+    if order.status.upper() == 'CANCELLED':
```
danqrodney: Can we use enum value names for these instead?

rest-api/dao/biobank_order_dao.py
```diff
@@ -265,8 +276,43 @@ def from_client_json(self, resource_json, participant_id=None, client_id=None):
           % (participant_id, resource.subject, self._participant_id_to_subject(participant_id)))
     self._add_identifiers_and_main_id(order, resource)
     self._add_samples(order, resource)
+    if resource.status and not resource.amendedInfo:
```
danqrodney: What if there's amendedInfo and resource.status has changed? If that's not valid, we should probably throw an exception... if it is valid, we should handle it.

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+
     return client_json
+
+  def _do_update(self, session, order, resource):
+    order.lastModified = clock.CLOCK.now()
+    order.biobankOrderId = resource.biobankOrderId
+    order.orderStatus = BiobankOrderStatus.AMENDED
+    order.amendedTime = clock.CLOCK.now()
+    order.logPosition = LogPosition()
+    order.logPositionId = resource.logPositionId   #TODO: check this
```
danqrodney: This looks wrong? The resource won't have a logPositionId, and you want a new one anyway so it goes up each time?

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+    # Ensure that if an order was previously cancelled/restored those columns are removed.
+    self._clear_cancelled_and_restored_fields(order)
+    self._update_history(session, order)
+    self._update_identifier_history(session, order)
+    self._update_sample_history(session, order)
+    super(BiobankOrderDao, self)._do_update(session, order, resource)
+
+  def _do_update_with_patch(self, session, order, resource):
+    order.lastModified = clock.CLOCK.now()
```
danqrodney: Set logPosition to LogPosition() here?

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+    if resource['status'] == 'cancelled':
+      order.amendedReason = resource['amendedReason']
+      order.cancelledUsername = resource['cancelledInfo']['author']['value']
+      order.cancelledSiteId = get_site(resource['cancelledInfo'])
+      order.cancelledTime = clock.CLOCK.now()
+      order.orderStatus = BiobankOrderStatus.CANCELLED
+    if resource['status'] == 'restored':
```
danqrodney: elif?

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+      order.amendedReason = resource['amendedReason']
+      order.restoredUsername = resource['restoredInfo']['author']['value']
+      order.restoredSiteId = get_site(resource['restoredInfo'])
+      order.restoredTime = clock.CLOCK.now()
+      order.orderStatus = BiobankOrderStatus.UNSET
```
danqrodney: Is it valid to do another kind of patch? Or do we only do PATCH for changing statuses?

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+    self._clear_cancelled_and_restored_fields(order)
+    self._update_history(session, order)
+    self._update_identifier_history(session, order)
+    self._update_sample_history(session, order)
+    super(BiobankOrderDao, self)._do_update(session, order, resource)
+
+  def _do_update_with_patch(self, session, order, resource):
+    order.lastModified = clock.CLOCK.now()
+    if resource['status'] == 'cancelled':
+      order.amendedReason = resource['amendedReason']
```
danqrodney: What if the existing status was already cancelled; would we still set all this?

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+      order.cancelledUsername = resource['cancelledInfo']['author']['value']
+      order.cancelledSiteId = get_site(resource['cancelledInfo'])
+      order.cancelledTime = clock.CLOCK.now()
+      order.orderStatus = BiobankOrderStatus.CANCELLED
+    if resource['status'] == 'restored':
+      order.amendedReason = resource['amendedReason']
```
danqrodney: See above: what if the existing status was already UNSET; would we still set all this?

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+      raise BadRequest('status of cancelled/restored is required')
+
+    if resource['status'] == 'cancelled':
+      for field in required_cancelled_fields:
+        if field not in resource:
+          raise BadRequest('%s is required for a cancelled biobank order' % field)
+      if 'site' not in resource['cancelledInfo'] or 'author' not in resource['cancelledInfo']:
+        raise BadRequest('author and site are required for cancelledInfo')
+
+    if resource['status'] == 'restored':
```
danqrodney: elif

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +386,119 @@ def to_client_json(self, model):
...
+      client_json['amendedReason'] = model.amendedReason
+      client_json['amendedSiteId'] = model.amendedSiteId
+      client_json['amendedUsername'] = model.amendedUsername
+      client_json['version'] = model.version
+
+
     return client_json
+
+  def _do_update(self, session, order, resource):
+    order.lastModified = clock.CLOCK.now()
```
danqrodney: Does this actually grab new identifiers / samples from the resource?

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,205 @@ def setUp(self):
...
+    cancelled_order = self.send_patch(path, request_data=request_data,
+                                      headers={'If-Match': 'W/"1"'})
+    get_cancelled_order = self.send_get(path)
+    self.assertEqual(cancelled_order, get_cancelled_order)
+    self.assertEqual(get_cancelled_order['status'], 'CANCELLED')
+    self.assertEqual(get_cancelled_order['amendedReason'], 'Its all wrong')
+    self.assertEqual(get_cancelled_order['cancelledUsername'], 'fred@pmi-ops.org')
+    self.assertEqual(get_cancelled_order['cancelledSiteId'], 1)
+
+  def test_retore_an_order(self):
```
danqrodney: retore->restore

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,205 @@ def setUp(self):
...
+    self.summary_dao.insert(self.participant_summary(self.participant))
+    order_json = load_biobank_order_json(self.participant.participantId,
+                                         filename='biobank_order_2.json')
+    result = self.send_post(self.path, order_json)
+    load_biobank_order_json(self.participant.participantId,
+                            filename='biobank_order_1.json')
+
+    biobank_order_id = result['identifier'][1]['value']
+    path = self.path + '/' + biobank_order_id
+    request_data = {
```
danqrodney: Make sure to modify either the identifiers or samples in each amendment, and verify the result... that's the only reason to do an amendment in the first place.

rest-api/dao/biobank_order_dao.py
danqrodney: Yeah, that should work. Thanks for clarifying.

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +391,140 @@ def to_client_json(self, model):
...
+      raise BadRequest('status must be restored or cancelled for patch request.')
+
+    self._update_history(session, order)
+    self._update_identifier_history(session, order)
+    self._update_sample_history(session, order)
+    super(BiobankOrderDao, self)._do_update(session, order, resource)
+    return order
+
+  def _validate_patch_update(self, session, model, resource, expected_version):
+    required_cancelled_fields = ['amendedReason', 'cancelledInfo', 'status']
```
danqrodney: We could make a _FhirBiobankOrderPatch class that is used for parsing this stuff, like _FhirBiobankOrder... I don't really mind either way... just noting.

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +391,140 @@ def to_client_json(self, model):
...
+        raise BadRequest('author and site are required for cancelledInfo')
+
+    elif resource['status'] == 'restored':
+      for field in required_restored_fields:
+        if field not in resource:
+          raise BadRequest('%s is required for a cancelled biobank order' % field)
+      if 'site' not in resource['restoredInfo'] or 'author' not in resource['restoredInfo']:
+        raise BadRequest('author and site are required for restoredInfo')
+
+    super(BiobankOrderDao, self)._validate_patch_update(session, model, resource, expected_version)
```
danqrodney: See above; we can just inline the logic from base_dao.validate_patch_update here.

We should also validate that you can only cancel an order that isn't canceled, and only restore an order that is.

rest-api/dao/biobank_order_dao.py
```diff
@@ -339,4 +391,140 @@ def to_client_json(self, model):
...
+      history = BiobankOrderedSampleHistory()
+      history.fromdict(sample.asdict(), allow_pk=True)
+      history.version = order.version
+      history.biobankOrderId = order.biobankOrderId
+      session.add(history)
+
+  @staticmethod
+  def _clear_cancelled_and_restored_fields(order):
+    #pylint: disable=unused-argument
+    """ Just in case these fields have values, we don't want them in the most recent record,
```
danqrodney: I don't know if this clearing is strictly necessary; having the latest row maintain the users, timestamp, and site when an order was restored, cancelled, etc. seems OK, too. (We'll have the status field to tell us what state the order is in.)

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,205 @@ def setUp(self):
...
+      "status": "cancelled"
+    }
+    cancelled_order = self.send_patch(path, request_data=request_data,
+                                      headers={'If-Match': 'W/"1"'})
+    get_cancelled_order = self.send_get(path)
+    self.assertEqual(cancelled_order, get_cancelled_order)
+    self.assertEqual(get_cancelled_order['status'], 'CANCELLED')
+    self.assertEqual(get_cancelled_order['amendedReason'], 'Its all wrong')
+    self.assertEqual(get_cancelled_order['cancelledUsername'], 'fred@pmi-ops.org')
+    self.assertEqual(get_cancelled_order['cancelledSiteId'], 1)
```
danqrodney: See above, this should be an identifier.

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,205 @@ def setUp(self):
...
+        }
+      },
+      "status": "restored"
+    }
+
+    self.send_patch(path, request_data=request_data, headers={'If-Match': 'W/"2"'})
+
+    request_data = {
+      "amendedReason": "Its all better",
+      "amendedInfo": {
```
danqrodney: Mutate the samples and identifiers here, too, and verify they come back correctly below.

Seeker1911: done

rest-api/dao/biobank_order_dao.py
danqrodney: OK. Doesn't this apply to amendedInfo, too?

I guess I'm confused... you don't actually call this method on PATCH ever, right? So I'm not sure why you need to handle new statuses here, if we only do cancel / restore on PATCH calls?

rest-api/dao/biobank_order_dao.py
danqrodney: You can use BiobankOrderStatus.CANCELLED.name or str(BiobankOrderStatus.CANCELLED), and compare that to order.status. Or you can make a BiobankOrderStatus out of order.status here, and compare the enums.

(Comparisons of strings and enums directly will fail.)

rest-api/dao/biobank_order_dao.py
danqrodney: Then this should be UNSET in the request, too, right? The request and response should match...

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,275 @@ def setUp(self):
...
+          "value": "hpo-site-bannerphoenix"
+        }
+      }
+    }
+    get_order = self.send_get(path)
+    full_order = get_order.copy()
+    full_order.update(request_data)
+
+    self.assertEqual(len(full_order['samples']), 16)
+    del get_order['samples'][0]
```
danqrodney: I don't get how deleting from get_order here winds up mutating full_order... I guess because get_order.copy isn't a deep copy? 

Why not delete from full_order instead?

rest-api/test/unit_test/api_test/biobank_order_api_test.py
danqrodney: I don't see any amendments of identifiers in the tests?

rest-api/dao/biobank_order_dao.py
Seeker1911: You're correct, at one time this was called here (before using `handling_info`). I don't think this is required at all now. Removed.

rest-api/dao/biobank_order_dao.py
Seeker1911: function removed as per above.

rest-api/dao/biobank_order_dao.py
Seeker1911: Spoke with HPRO team again, it makes sense to them the way it is now, personally I think sending `unset` in the request would be confusing, if anything I think we should return 'restored' in the payload. I'll leave for now, you and I can discuss further, it will be an easy change whichever we pick.

rest-api/test/unit_test/api_test/biobank_order_api_test.py
Seeker1911: It changes the site and removes one sample, would you like more ?

rest-api/test/unit_test/api_test/biobank_order_api_test.py
Seeker1911: done

rest-api/test/unit_test/api_test/biobank_order_api_test.py
danqrodney: I was talking about biobank order identifiers, specifically. Are the tests modifying that and I missed it?

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,290 @@ def setUp(self):
...
+          "value": "fred@pmi-ops.org"
+        },
+        "site": {
+          "system": "https://www.pmi-ops.org/site-id",
+          "value": "hpo-site-bannerphoenix"
+        }
+      }
+    }
+
+    biobank_order_identifiers = {"created": "2018-02-21T16:25:12",
```
danqrodney: This sets the identifiers... but the code below doesn't actually check that they got updated. Can you add a test for that?

rest-api/api/base_api.py
```diff
@@ -206,9 +205,25 @@ def put(self, id_, participant_id=None):
     self._do_update(m)
     return self._make_response(m)
 
+  def patch(self, id_):
```
danqrodney: I'd either move this to biobank_order_dao, or define an update_with_patch method here that at a minimum throws an exception saying "implement me"... the call to update_with_patch below with no implementation here is weird.

rest-api/dao/biobank_order_dao.py
```diff
@@ -201,6 +220,37 @@ def _update_participant_summary(self, session, obj):
...
       setattr(participant_summary, status_field + 'Time', time)
 
+  def _get_non_cancelled_biobank_orders(self, session, participantId):
+    # look up latest order without cancelled status
+    return session.query(BiobankOrder).filter(BiobankOrder.participantId ==
+                                              participantId).filter(or_(BiobankOrder.orderStatus !=
+                                                                    BiobankOrderStatus.CANCELLED,
+                                                                    BiobankOrder.orderStatus == None
+                                                                    )).order_by(
+      BiobankOrder.created.desc()).all()
```
danqrodney: I would get rid of the desc(); that way below you're going through orders in ascending timestamp order, which is the right behavior.

rest-api/dao/biobank_order_dao.py
```diff
@@ -333,10 +390,150 @@ def to_client_json(self, model):
...
     resource.finalized_info = self._to_handling_info(model.finalizedUsername, model.finalizedSiteId)
+    resource.amendedReason = model.amendedReason
+
+    if model.orderStatus == BiobankOrderStatus.CANCELLED:
+      resource.status = str(BiobankOrderStatus.CANCELLED)
+      resource.cancelledInfo = self._to_handling_info(model.cancelledUsername,
+                                                       model.cancelledSiteId)
+
+    restored = getattr(model, 'restoredSiteId')
+    if restored:
```
danqrodney: So what happens if an order gets restored and then canceled again? This logic will set its status to UNSET.

I'm not clear on whether we care about including cancelledInfo / restoredInfo / amendedInfo in the response when the status doesn't match. If we don't, then we can just change this if and the next one to be in else blocks.  If we do, then you probably need to handle resource.status separate from cancelledInfo/restoredInfo/amendedInfo.

rest-api/dao/biobank_order_dao.py
```diff
@@ -333,10 +390,150 @@ def to_client_json(self, model):
...
+      resource.status = str(BiobankOrderStatus.UNSET)
+      resource.restoredInfo = self._to_handling_info(model.restoredUsername,
+                                                       model.restoredSiteId)
+
+    if model.orderStatus == BiobankOrderStatus.AMENDED:
```
danqrodney: See above; the status bit at least could be in an else block.

rest-api/dao/biobank_order_dao.py
```diff
@@ -333,10 +390,150 @@ def to_client_json(self, model):
...
+      order.restoredSiteId = get_site(resource['restoredInfo'])
+      order.restoredTime = clock.CLOCK.now()
+      order.orderStatus = BiobankOrderStatus.UNSET
+    else:
+      raise BadRequest('status must be restored or cancelled for patch request.')
+
+    self._update_history(session, order)
+    self._update_identifier_history(session, order)
+    self._update_sample_history(session, order)
+    super(BiobankOrderDao, self)._do_update(session, order, resource)
```
danqrodney: I noticed that we do super._do_update() in _do_update() before update_history(), etc. and here we do it afterwards. Not sure if it matters...  but we might want to be consistent.

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,315 @@ def setUp(self):
...
+        }
+      },
+      "status": "cancelled"
+    }
+    cancelled_order = self.send_patch(path, request_data=request_data,
+                                      headers={'If-Match': 'W/"1"'})
+    get_cancelled_order = self.send_get(path)
+    get_summary = self.summary_dao.get(self.participant.participantId)
+
+    self.assertEqual(get_summary.biospecimenSourceSiteId, None)
```
danqrodney: You could assert that participant summary sample statuses and timestamps are unset here, too. (I know you've got DAO tests for that, too, but API tests are nice.)

rest-api/test/unit_test/api_test/biobank_order_api_test.py
danqrodney: Ping? Still not checking identifiers after the amendment?

rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -17,9 +19,315 @@ def setUp(self):
...
+      }
+    }
+    get_order = self.send_get(path)
+    full_order = get_order.copy()
+    full_order.update(request_data)
+    self.send_put(path, request_data=full_order, headers={'If-Match': 'W/"3"'})
+
+    get_amended_order = self.send_get(path)
+    self.assertEqual(len(get_amended_order['samples']), 2)
+    self.assertEqual(get_amended_order['amendedInfo']['author']['value'], 'mike@pmi-ops.org')
```
danqrodney: Check that the status here is AMENDED?

rest-api/test/unit_test/dao_test/biobank_order_dao_test.py
```diff
@@ -118,3 +172,139 @@ def test_store_invalid_test(self):
...
+    cancelled_request = self._get_cancel_patch()
+    ps_dao = ParticipantSummaryDao().get(self.participant.participantId)
+
+    self.assertEqual(ps_dao.sampleOrderStatus1ED10, OrderStatus.CREATED)
+    self.assertEqual(ps_dao.sampleOrderStatus2ED10, OrderStatus.CREATED)
+    self.dao.update_with_patch(order_1.biobankOrderId, cancelled_request,
+                               order_1.version)
+    ps_dao = ParticipantSummaryDao().get(self.participant.participantId)
+
+    self.assertEqual(ps_dao.sampleOrderStatus1ED10, None)
```
danqrodney: Check the timestamp values, too?

You can do:

 with FakeClock(TIME_1):
  <insert order 1>
with FakeClock(TIME_2):
  <insert order 2>

to make sure the timestamps differ.

Seeker1911: done

rest-api/api/base_api.py
```diff
@@ -206,9 +205,29 @@ def put(self, id_, participant_id=None):
...
+    """
+    resource = request.get_json(force=True)
+    etag = request.headers.get('If-Match')
+    if not etag:
+      raise BadRequest("If-Match is missing for PATCH request")
+    expected_version = _parse_etag(etag)
+    order = self.dao.update_with_patch(id_, resource, expected_version)
+    return self._make_response(order)
+
+  def update_with_patch(self):
```
danqrodney: Add the expected parameters here?

Seeker1911: done

PR #772, opened 2018-08-27
Circleci 20 test

Seeker1911:


.circleci/config.yml
```diff
@@ -78,7 +92,23 @@ jobs:
...
+        - v1-dep-
+    - run: sudo chmod o+w /opt
+    - run: pip install --upgrade pip
+    - run: pip install --upgrade httplib2 jira oauth2client parallel pylint
+    - run: pip install -I alembic==0.9.1
+    - run:
+        working_directory: ~/all-of-us/raw-data-repository
+        command: 'sudo service mysql status || sudo service mysql start; '
     - run: ./ci/activate_creds.sh ~/gcloud-credentials.key
+    - run: ./ci/test.sh
```
danqrodney: You shouldn't need to run test.sh here; build_and_test already ran before this.

.circleci/config.yml
```diff
@@ -60,13 +60,27 @@ jobs:
...
+          - v1-dep-
+      - run: sudo chmod o+w /opt
+      - run: pip install --upgrade pip
+      - run: pip install --upgrade httplib2 jira oauth2client parallel pylint
+      - run: pip install -I alembic==0.9.1
       - run:
           working_directory: ~/all-of-us/raw-data-repository
           command: 'sudo service mysql status || sudo service mysql start; '
       - run: ./ci/activate_creds.sh ~/gcloud-credentials.key
+      - run: ./ci/test.sh
```
danqrodney: Same here

.circleci/config.yml
```diff
@@ -63,10 +63,24 @@ jobs:
       - run:
           working_directory: ~/all-of-us/raw-data-repository
           command: 'sudo service mysql status || sudo service mysql start; '
+      - restore_cache:
```
danqrodney: So for restore_cache / save_cache I think we want something like this:

https://github.com/all-of-us/pyclient/blob/master/.circleci/config.yml#L15

https://github.com/all-of-us/pyclient/blob/master/.circleci/config.yml#L28

So that we save off venv and don't have to fetch all the dependencies every time. (If you want to do that in a follow-up, that's fine.) I'm not sure that the current code with restore_cache without a corresponding save_cache is actually doing anything...

ci/setup.sh
```diff
@@ -0,0 +1,32 @@
...
+  # The API server doesn't expect to be in venv, it just wants a lib/.
+  ln -s venv/lib/python*/site-packages/ lib
+  # Install coverage so we can collect test suite metadata
+  pip install coverage
+}
+
+cd rest-api
+activate_local_venv
+git submodule update --init
+export CLOUDSDK_CORE_DISABLE_PROMPTS=1
```
danqrodney: I think the lines after this should be in test.sh?

.circleci/config.yml
```diff
@@ -35,18 +35,16 @@ jobs:
...
           command: 'sudo service mysql status || sudo service mysql start; '
       - restore_cache:
-          keys:
-          # This branch if available
-          - v1-dep-{{ .Branch }}-
-          # Default branch if not
-          - v1-dep-master-
-          # Any branch if there are none on the default branch - this should be unnecessary if you have your default branch configured correctly
-          - v1-dep-
+          key: deps-{{ .Branch }}-{{ checksum "rest-api/requirements.txt" }}
```
danqrodney: Nice. :)

.circleci/config.yml
Seeker1911: done

ci/setup.sh
Seeker1911: probably right, since I added it as a test after the upgrade_db issue started. But that seemed to be something else.

.circleci/config.yml
Seeker1911: all done

PR #775, opened 2018-08-31
add history tables for biobank order identifier and samples

Seeker1911:


rest-api/model/biobank_order.py
```diff
@@ -125,41 +124,78 @@ class BiobankOrder(BiobankOrderBase, Base):
   samples = relationship('BiobankOrderedSample', cascade='all, delete-orphan')
 
 
-class BiobankOrderIdentifier(Base):
+class BiobankOrderIdentifierBase(object):
+  @declared_attr
```
danqrodney: So I believe the only fields you need to do the @declared_attr thing on are foreign keys and relationships. You can leave the other columns the way they were..

rest-api/alembic/versions/45d77ae13992_add_id_and_sample_history_tables.py
```diff
@@ -0,0 +1,78 @@
...
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('biobank_order_identifier_history',
+    sa.Column('version', sa.Integer(), nullable=False),
+    sa.Column('system', sa.String(length=80), nullable=False),
+    sa.Column('value', sa.String(length=80), nullable=False),
+    sa.Column('biobank_order_id', sa.String(length=80), nullable=False),
+    sa.ForeignKeyConstraint(['biobank_order_id'], ['biobank_order.biobank_order_id'], ),
+    sa.PrimaryKeyConstraint('version', 'system', 'value')
```
danqrodney: The primary key is in the wrong order here; should be system, value, version. (Which kinda matters for querying on this.) I believe if you get rid of the @declared_attr stuff below it should work. You *may* need to add a version column to BiobankOrderIdentifierBase or BiobankOrderedSampleBase, or override all the primary key columns in the subclasses to get SQLAlchemy to handle this appropriately... I'm not sure.

Alternatively, you can mess with the generated code here directly... just make sure that SQLAlchemy doesn't do something crazy like try to change the primary key if you run generate_schema again afterwards.

rest-api/alembic/versions/45d77ae13992_add_id_and_sample_history_tables.py
```diff
@@ -0,0 +1,78 @@
...
+    sa.Column('version', sa.Integer(), nullable=False),
+    sa.Column('description', sa.UnicodeText(), nullable=False),
+    sa.Column('finalized', model.utils.UTCDateTime(), nullable=True),
+    sa.Column('test', sa.String(length=80), nullable=False),
+    sa.Column('processed', model.utils.UTCDateTime(), nullable=True),
+    sa.Column('processing_required', sa.Boolean(), nullable=False),
+    sa.Column('order_id', sa.String(length=80), nullable=False),
+    sa.Column('collected', model.utils.UTCDateTime(), nullable=True),
+    sa.ForeignKeyConstraint(['order_id'], ['biobank_order.biobank_order_id'], ),
+    sa.PrimaryKeyConstraint('version', 'test', 'order_id')
```
danqrodney: Same as above, primary key should be order_id, test, version.

rest-api/alembic/versions/45d77ae13992_add_id_and_sample_history_tables.py
Seeker1911: done

rest-api/alembic/versions/45d77ae13992_add_id_and_sample_history_tables.py
Seeker1911: done

rest-api/model/biobank_order.py
Seeker1911: done

rest-api/alembic/versions/b4f6eb55d503_add_sample_and_identifier_history.py
```diff
@@ -0,0 +1,78 @@
...
+    sa.Column('test', sa.String(length=80), nullable=False),
+    sa.Column('description', sa.UnicodeText(), nullable=False),
+    sa.Column('processing_required', sa.Boolean(), nullable=False),
+    sa.Column('collected', model.utils.UTCDateTime(), nullable=True),
+    sa.Column('processed', model.utils.UTCDateTime(), nullable=True),
+    sa.Column('finalized', model.utils.UTCDateTime(), nullable=True),
+    sa.Column('version', sa.Integer(), nullable=False),
+    sa.Column('order_id', sa.String(length=80), nullable=False),
+    sa.ForeignKeyConstraint(['order_id'], ['biobank_order.biobank_order_id'], ),
+    sa.PrimaryKeyConstraint('test', 'version', 'order_id')
```
danqrodney: This should be order_id, test, version, or order_id, version, test... Putting order_id first matters if you want to make scanning the samples for an order efficient.

It looks like you can get SQLAlchemy to do this properly using __table_args__, instead of primary_key=True, as described here:

https://stackoverflow.com/questions/9034271/sqlalchemy-orm-how-to-declare-a-table-class-that-contains-multi-column-primary

PR #777, opened 2018-09-06
add staging option to deploy app to reflect prod

Seeker1911:


rest-api/tools/deploy_app.sh
```diff
@@ -42,7 +42,7 @@ if [[ $(git status --porcelain) ]]; then
 fi
 
 UPDATE_TRACKER=tools/update_release_tracker.py
-if [ "${PROJECT}" == "all-of-us-rdr-prod" ]
+if [ "${PROJECT}" == "all-of-us-rdr-prod" ] || [ "${PROJECT}" == "all-of-us-rdr-staging" ]
```
danqrodney: This should change to use config_staging.json... we don't want to use prod config, just app_prod.yaml.

rest-api/tools/deploy_app.sh
```diff
@@ -154,7 +154,7 @@ then
 
   if [ "$TARGET" == "app" ] || [ "$TARGET" == "all" ]
   then
-    if [ "${PROJECT}" = "all-of-us-rdr-prod" ]
+    if [ "${PROJECT}" = "all-of-us-rdr-prod" ] || [ "${PROJECT}" = "all-of-us-rdr-staging" ]
```
danqrodney: I'm not sure I'd merge this into master, as we don't want staging normally to use app_prod.yaml... only during load tests.

If we do want to merge something into master here, we should probably add a --use_prod_yaml flag that we check here instead.

rest-api/tools/deploy_app.sh
Seeker1911: done

rest-api/tools/deploy_app.sh
Seeker1911: done

rest-api/tools/deploy_app.sh
```diff
@@ -18,6 +18,7 @@ while true; do
     --version) VERSION=$2; shift 2;;
     --deploy_as_version) DEPLOY_AS_VERSION=$2; shift 2;;
     --target) TARGET=$2; shift 2;;
+    --use_prod_yaml) USE_PROD_YAML='true'; shift 2;;
```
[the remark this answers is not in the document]
Seeker1911: done

PR #781, opened 2018-09-17
DA-563 fixing fake participant importer

Seeker1911:


rest-api/tools/auth_setup.sh
```diff
@@ -61,7 +61,7 @@ source ${REPO_ROOT_DIR}/tools/setup_vars.sh
 TMP_DB_INFO_FILE=${TMP_DIR}/db_info.json
 PORT=3308
 CLOUD_PROXY_PID=
-
+export "GOOGLE_APPLICATION_CREDENTIALS"=$CREDS_FILE
```
danqrodney: Nit: I don't think you need the quotes here.

rest-api/tools/import_organizations.py
```diff
@@ -584,7 +584,7 @@ def main(args):
   HPODao()._invalidate_cache()
   OrganizationImporter().run(args.organization_file, args.dry_run)
   OrganizationDao()._invalidate_cache()
-  SiteImporter().run(args.site_file, args.dry_run)
+  SiteImporter().run(args.site_file, args.dry_run, args.creds_file)
```
danqrodney: I don't think you should pass in creds_file anymore? Does this actually work?

rest-api/tools/auth_setup.sh
Seeker1911: You're right

rest-api/tools/import_organizations.py
Seeker1911: Nope, fixed.

PR #784, opened 2018-09-21
[DA-728] Add exception list for service account from key deletion

wangy70:
Add key deletion exception for redlock-read-only-stable and circle-deploy in staging config.
Add key deletion exception for circle-deploy in test config.

### DIFF
--- a/rest-api/config.py
+++ b/rest-api/config.py
@@ -51,6 +51,9 @@ REQUIRED_CONFIG_KEYS = [BIOBANK_SAMPLES_BUCKET_NAME]
 
 DAYS_TO_DELETE_KEYS = "days_to_delete_keys"
 
+# service accounts exception from key deletion
+SERVICE_ACCOUNT_KEYS = "service_account_keys"
+
 # Overrides for testing scenarios
 CONFIG_OVERRIDES = {}
 
--- a/rest-api/config.py
+++ b/rest-api/config.py
@@ -52,7 +52,7 @@ REQUIRED_CONFIG_KEYS = [BIOBANK_SAMPLES_BUCKET_NAME]
 DAYS_TO_DELETE_KEYS = "days_to_delete_keys"
 
 # service accounts exception from key deletion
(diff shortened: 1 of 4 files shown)


rest-api/config.py
```diff
@@ -51,6 +51,9 @@
 
 DAYS_TO_DELETE_KEYS = "days_to_delete_keys"
 
+# service accounts exception from key deletion
+SERVICE_ACCOUNT_KEYS = "service_account_keys"
```
danqrodney: I would rename the constant name and value to "service_accounts_with_long_lived_keys" or something, to make it clear what this means.

rest-api/offline/sa_key_remove.py
```diff
@@ -9,6 +9,7 @@
 
 def delete_service_account_keys():
   days_to_delete = config.getSetting(config.DAYS_TO_DELETE_KEYS)
+  service_account_keys = config.getSetting(config.SERVICE_ACCOUNT_KEYS)
```
danqrodney: I believe this should be:

service_account_keys = config.getSettingList(config.SERVICE_ACCOUNT_KEYS, default=[])

getSettingList() allows us to return lists instead of individual values; default ensures this doesn't fail if the config setting is missing. (Which it would be for everything but test and staging.)

rest-api/offline/sa_key_remove.py
```diff
@@ -31,6 +32,8 @@ def delete_service_account_keys():
 
           for key in keys:
             keyname = key['name']
+            if keyname in service_account_keys:
```
danqrodney: So I may be wrong here, but I think what we're looking to compare is account['email'], not key['name']. You should replace the logic on line 25 above with this.

Seeker1911: correct. key['name'] is the "human readable" form i.e. `configurator` or `App Engine default`

wangy70:
It's the codacy review, line too long. I just fixed and submitted.

PR #790, opened 2018-09-25
[DA-743] add awardee description to ops api readme

wangy70:


opsdataAPI.md
```diff
@@ -161,7 +161,41 @@ For enumeration fields, the following values are defined:
         `HLS_AND_ONE_OTHER_RACE`, `HLS_AND_MORE_THAN_ONE_OTHER_RACE`, `MORE_THAN_ONE_RACE`,
   		`OTHER_RACE`, `PREFER_NOT_TO_SAY`
 
-Note: hpoId maps to awardee. If awardee is set in a request, hpoId is updated and vice versa.
+`awardee`: `UNSET` , `AZ_TUCSON` , `CAL_PMC` , `CHEROKEE` , `COLUMBIA` , `COMM_HEALTH` , `EAU_CLAIRE` , `EMSI` ,
```
danqrodney: My only concern with this is that when we add awardees in future, this documentation is going to have to get updated. I'm curious whether we could instead link out to:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/data/awardees.csv

wangy70: Yes, that will save a lot of work, I will update it.

PR #791, opened 2018-09-25
Mm/physical measurement db changes

Seeker1911:


rest-api/alembic/versions/ceb7e48928a9_add_cancelled_to_pm.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('physical_measurements', sa.Column('cancelled_status', model.utils.Enum(PhysicalMeasurementsStatus), nullable=True))
```
danqrodney: Should this just be "status" with a default of COMPLETED? What will it be set to if the PM isn't cancelled? What will it be set to if the PM gets amended?

Note there's also the existing "final" field on PMs -- when you amend PMs, that gets set to false on the original row. Do we want to keep it set to true for cancelled PMs?

rest-api/model/measurements.py
```diff
@@ -33,6 +34,9 @@ class PhysicalMeasurements(Base):
   finalizedUsername = Column('finalized_username', String(255))
   logPosition = relationship('LogPosition')
   finalized = Column('finalized', UTCDateTime)
+  cancelledStatus = Column('cancelled_status', Enum(PhysicalMeasurementsStatus))
+  cancelledUsername = Column('cancelled_username', String(255))
```
danqrodney: Probably add cancelledSiteId here too?

Seeker1911: done

rest-api/alembic/versions/ceb7e48928a9_add_cancelled_to_pm.py
Seeker1911: When a PM is amended we set 'status' = 'amended'. I was trying to avoid any possible key errors with this. I do think final would be true on cancelled PM's. Do you think we should just use status here (it was my first choice) until I saw the _update_amended function in the DAO. I could make sure there are no conflicts in tests. What are your thoughts ?

rest-api/model/measurements.py
```diff
@@ -33,6 +34,10 @@ class PhysicalMeasurements(Base):
...
+  cancelledSiteId = Column('cancelled_site_id', Integer, ForeignKey('site.site_id'))
```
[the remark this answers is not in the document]
Seeker1911: done, and changed cancelled_status to status

rest-api/participant_enums.py
```diff
@@ -26,9 +26,12 @@
 TEST_HPO_ID = 19
 
 class PhysicalMeasurementsStatus(messages.Enum):
-  """The state of the participant's physical measurements"""
+  """The state of the participant's physical measurements.
+  Amended measurements will be null, a restored measuremnent will be UNSET."""
```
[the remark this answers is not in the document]
Seeker1911: done

PR #796, opened 2018-09-28
DA-741 mm/cancel physical measurements

Seeker1911:
Added the ability to cancel or restore a physical measurement order.


rest-api/dao/physical_measurements_dao.py
```diff
@@ -46,13 +48,18 @@ def get_with_session(self, session, obj_id, **kwargs):
...
     with self.session() as session:
       query = session.query(PhysicalMeasurements) \
           .options(subqueryload(PhysicalMeasurements.measurements).subqueryload(
               Measurement.measurements)) \
           .options(subqueryload(PhysicalMeasurements.measurements).subqueryload(
               Measurement.qualifiers))
+
+      if for_update:
+        query = query.with_for_update()
+        return session, query.get(physical_measurements_id)
```
danqrodney: You shouldn't be returning session from here. It doesn't match the return type below, and the session is useless once you exit here anyway. (The with self.session() as session will close the session when you exit the block.)

Instead you need a version of this that takes session as an argument... call with self.session() as session outside of the method and pass it in.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -284,12 +292,21 @@ def _update_participant_summary(self, session, obj):
     participant_summary.physicalMeasurementsFinalizedTime = obj.finalized
     participant_summary.physicalMeasurementsCreatedSiteId = obj.createdSiteId
     participant_summary.physicalMeasurementsFinalizedSiteId = obj.finalizedSiteId
+    if 'physicalMeasurementsStatus' in obj:
```
danqrodney: This logic looks wrong... you'll set participant_summary.physicalMeasurementStatus to COMPLETED here, and then the if block below that updates enrollment status won't get run?

We should make sure that update_enrollment_status gets run whenever the physical measurement status changes... that's what the previous logic was trying to accomplish.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -284,12 +292,21 @@ def _update_participant_summary(self, session, obj):
...
-    if participant_summary.physicalMeasurementsStatus != PhysicalMeasurementsStatus.COMPLETED:
+
+    if participant_summary.physicalMeasurementsStatus != PhysicalMeasurementsStatus.COMPLETED and \
+            participant_summary.physicalMeasurementsStatus != PhysicalMeasurementsStatus.CANCELLED:
+
       participant_summary.physicalMeasurementsStatus = PhysicalMeasurementsStatus.COMPLETED
       participant_summary_dao.update_enrollment_status(participant_summary)
       session.merge(participant_summary)
 
+    if participant_summary.physicalMeasurementsStatus == PhysicalMeasurementsStatus.CANCELLED:
```
danqrodney: I think you can combine this with the if above; just make sure this runs whenever the participant summary physical measurement status changes.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -327,6 +344,32 @@ def _update_amended(self, obj, extension, url, session):
...
     obj.amendedMeasurementsId = amended_measurement_id
 
+  def update_with_patch(self, id_, resource):
+    session, measurement = self.get_with_children(id_, for_update=True)
+    return self._do_update_with_patch(session, measurement, resource)
+
+  def _do_update_with_patch(self, session, measurement, resource):
+    self._validate_patch_update(measurement, resource)
+    site_id, author = self._validate_and_get_author_and_site(resource)
+    measurement.reason = resource['reason']  # @TODO: ADD REASON TO DB
```
danqrodney: What's up with the TODO? Isn't this storing the reason?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -327,6 +344,32 @@ def _update_amended(self, obj, extension, url, session):
...
+      measurement.cancelledTime = clock.CLOCK.now()
+      measurement.status = PhysicalMeasurementsStatus.CANCELLED
+    if resource['status'].lower() == 'restored':
+      measurement.cancelledUsername = None
+      measurement.cancelledSiteId = None
+      measurement.cancelledTime = None
+      measurement.status = PhysicalMeasurementsStatus.UNSET
+
+    payload = self.add_root_fields_to_resource(measurement)
+    # @TODO: ADD LOGGING !!!
```
danqrodney: What logging do you want to add here?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -327,6 +344,32 @@ def _update_amended(self, obj, extension, url, session):
...
+      measurement.cancelledSiteId = site_id
+      measurement.cancelledTime = clock.CLOCK.now()
+      measurement.status = PhysicalMeasurementsStatus.CANCELLED
+    if resource['status'].lower() == 'restored':
+      measurement.cancelledUsername = None
+      measurement.cancelledSiteId = None
+      measurement.cancelledTime = None
+      measurement.status = PhysicalMeasurementsStatus.UNSET
+
+    payload = self.add_root_fields_to_resource(measurement)
```
danqrodney: So site_id, author, and reason all came from the resource to begin with.The only thing that seems like it might need updating is status and cancelledTime?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -553,3 +597,75 @@ def from_client_json(resource_json, participant_id=None, **unused_kwargs):
...
+    order_resource = json.loads(order.resource)
+    order_resource['reason'] = order.reason
+    if order.status == PhysicalMeasurementsStatus.CANCELLED:
+      order_resource['status'] = 'cancelled'
+      order_resource['cancelledUsername'] = order.cancelledUsername
+      order_resource['cancelledSiteId'] = order.cancelledSiteId
+      order_resource['cancelledTime'] = str(order.cancelledTime)
+
+    if order.status == PhysicalMeasurementsStatus.UNSET:
+      order_resource['status'] = 'restored'
```
danqrodney: Do we intend to exposed 'restored' in the response to clients? Or just clear it out?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -553,3 +597,75 @@ def from_client_json(resource_json, participant_id=None, **unused_kwargs):
                                 createdUsername=created_username, finalizedSiteId=finalized_site_id,
                                 finalizedUsername=finalized_username)
 
+  @staticmethod
+  def _validate_patch_update(measurement, resource):
+    """validates request of resource"""
+    cancelled_required_fields = ['status', 'reason', 'cancelledInfo']
+    restored_required_fields = ['status', 'reason', 'restoredInfo']
+
+    if resource['status'].lower() == 'cancelled':
```
danqrodney: If status is missing this will give a KeyError. I think you need to use get() here  and check for status first?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -553,3 +597,75 @@ def from_client_json(resource_json, participant_id=None, **unused_kwargs):
...
+      site_id = self.get_location_site_id(_LOCATION_PREFIX + resource['restoredInfo']['site'][
+        'value'])
+      author = self.get_author_username(_AUTHOR_PREFIX + resource['restoredInfo']['author'][
+        'value'])
+
+    return site_id, author
+
+  def patch(self, id_, p_id):
+    # pylint: disable=unused-argument
+    resource = request.get_json(force=True)
```
danqrodney: This is a DAO. We don't normally access request in DAOs, because we want to be able to use them outside of the context of an inbound request. We should follow the pattern of other DAOs, and pass in the resource, e.g.

https://github.com/all-of-us/raw-data-repository/blob/f05196a89dbcfe631c601929ff632430c303659a/rest-api/dao/biobank_order_dao.py#L437

rest-api/dao/physical_measurements_dao.py
```diff
@@ -46,13 +47,18 @@ def get_with_session(self, session, obj_id, **kwargs):
...
     with self.session() as session:
       query = session.query(PhysicalMeasurements) \
           .options(subqueryload(PhysicalMeasurements.measurements).subqueryload(
               Measurement.measurements)) \
           .options(subqueryload(PhysicalMeasurements.measurements).subqueryload(
               Measurement.qualifiers))
+
+      if for_update:
+        query = query.with_for_update()
+        return query.get(physical_measurements_id)
```
danqrodney: The with_for_update() is only going to work if you pass session into this method and use it here. Otherwise the transaction is over as soon as you leave this block.

You may want to make this get_with_children_with_session or whatever.

rest-api/dao/physical_measurements_dao.py
danqrodney: Sorry forgot this was a patch.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -327,6 +340,35 @@ def _update_amended(self, obj, extension, url, session):
...
+      measurement.cancelledUsername = None
+      measurement.cancelledSiteId = None
+      measurement.cancelledTime = None
+      measurement.status = PhysicalMeasurementsStatus.UNSET
+
+    logging.info('%s %s physical measuremnt %s.', author, resource['status'],
+                 measurement.physicalMeasurementsId)
+    payload = self.add_root_fields_to_resource(measurement)
+    super(PhysicalMeasurementsDao, self)._do_update(session, payload, payload)
+    logging.info('updating participant summary to %s for participant: %s', payload.status,
```
danqrodney: Do we need to keep this logging around?

rest-api/dao/physical_measurements_dao.py
```diff
@@ -327,6 +340,35 @@ def _update_amended(self, obj, extension, url, session):
     session.merge(amended_measurement)
     obj.amendedMeasurementsId = amended_measurement_id
 
+  def update_with_patch(self, id_, session, resource):
+    measurement = self.get_with_children(id_, for_update=True)
```
danqrodney: As noted above: pass in session here.

rest-api/dao/physical_measurements_dao.py
Seeker1911: done.

rest-api/dao/physical_measurements_dao.py
Seeker1911: done

rest-api/dao/physical_measurements_dao.py
Seeker1911: removed.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -46,15 +47,33 @@ def get_with_session(self, session, obj_id, **kwargs):
...
+  def get_with_children_with_session(self, session, physical_measurements_id, for_update=False):
+    query = session.query(PhysicalMeasurements) \
+      .options(subqueryload(PhysicalMeasurements.measurements).subqueryload(
+        Measurement.measurements)) \
+      .options(subqueryload(PhysicalMeasurements.measurements).subqueryload(
+        Measurement.qualifiers))
+
+    if for_update:
+      query = query.with_for_update()
       return query.get(physical_measurements_id)
```
danqrodney: Nit: you could remove this line and the method still works.

rest-api/dao/physical_measurements_dao.py
```diff
@@ -46,15 +47,33 @@ def get_with_session(self, session, obj_id, **kwargs):
       ParticipantDao().validate_participant_reference(session, result)
     return result
 
-  def get_with_children(self, physical_measurements_id):
+  def get_with_children(self, physical_measurements_id, for_update=False):
     with self.session() as session:
       query = session.query(PhysicalMeasurements) \
```
danqrodney: You could just have:

return self.get_with_children_with_session(session, physical_measurements_id, for_update)

Then you don't duplicate the logic below.

Seeker1911: done

rest-api/dao/physical_measurements_dao.py
Seeker1911: done

PR #800, opened 2018-10-03
[DA-692] Fix participant summary update bug in participant_dao.

danqrodney:


rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -851,6 +851,37 @@ def test_dvehr_consent(self):
...
+      self.send_put('Participant/%s' % participant_id_1, participant_1,
+                    headers={ 'If-Match':'W/"1"'})
+
+    ps_1 = self.send_get('Participant/%s/Summary' % participant_id_1)
+    self.assertEquals('NO_USE', ps_1['withdrawalStatus'])
+    self.assertEquals('PITT', ps_1['awardee'])
+    self.assertEquals(TIME_2.isoformat(), ps_1.get('withdrawalTime'))
+
+    with FakeClock(TIME_3):
+      participant_1['providerLink'] = [self.az_provider_link]
```
Seeker1911: Does it seem right that we can repair a participant after a `NO_USE` withdrawal ?

danqrodney: Yeah, that's a requirement... in particular, PTSC needs to be able to administratively re-pair withdrawn participants with TEST or whatever.

danqrodney:
Please ignore earlier comment / code. The issue was we had onupdate handlers on withdrawal status and suspension status in participant summary for some reason... this set things to NOT_WITHDRAWN whenever the summary gets updated... which is wrong. So I got rid of those.

PR #803, opened 2018-10-05
[risk=low] Fix prod configs

calbach:
These don't go in the "user_info" section


rest-api/config/config_prod.json
```diff
@@ -50,20 +65,5 @@
       },
       "description": "Using subnet mask these ranges cover 66.189.127.0-32 and 68.184.42.162.160-191"
     },
```
Seeker1911: can this comma be here ?

PR #809, opened 2018-10-15
[DA-671] reconcile m2api to support querying by both sample status time and sa…

wangy70:
reconcile m2api to support querying by both sample status time and sample order status time

### DIFF
--- a/rest-api/dao/participant_counts_over_time_service.py
+++ b/rest-api/dao/participant_counts_over_time_service.py
@@ -30,7 +30,7 @@ class ParticipantCountsOverTimeService(BaseDao):
     filters_sql_ps = self.get_facets_sql(filters)
     filters_sql_p = self.get_facets_sql(filters, table_prefix='p')
 
-    # check if request stratification is for enromment status
+    # check if request stratification is for enrollment status
     query_for_enrollment_status = False
 
     target_enrollment_statuses = []
@@ -60,7 +60,7 @@ class ParticipantCountsOverTimeService(BaseDao):
 
     # Iterate through each result (by date), transforming tabular SQL results
     # into expected list-of-dictionaries response format
-    # set value = 0 if request stratification is for enromment status
+    # set value = 0 if request stratification is for enrollment status
     # but key is not in target_enrollment_statuses
     try:
       results = cursor.fetchall()
@@ -208,9 +208,7 @@ class ParticipantCountsOverTimeService(BaseDao):
           from
           (select a.participant_id,a.register_day,COALESCE(b.member_day,'3000-01-01') member_day,COALESCE(b.core_day,'3000-01-01') core_day from
                 (SELECT p.participant_id,
-                       CASE
-                         WHEN (enrollment_status IS NULL or enrollment_status >= 1) THEN DATE(p.sign_up_time)
-                         ELSE Date('3000-01-01') END register_day
+                       COALESCE(p.sign_up_time, '3000-01-01') as register_day
                 FROM participant p
                        LEFT OUTER JOIN participant_summary ps ON p.participant_id = ps.participant_id
                 %(filters_p)s) a left join
@@ -296,9 +294,7 @@ class ParticipantCountsOverTimeService(BaseDao):
           from
           (select a.participant_id,a.register_day,COALESCE(b.member_day,'3000-01-01') member_day,COALESCE(b.core_day,'3000-01-01') core_day from
                 (SELECT p.participant_id,
-                       CASE
-                         WHEN (enrollment_status IS NULL or enrollment_status >= 1) THEN DATE(p.sign_up_time)
-                         ELSE Date('3000-01-01') END register_day
+                       COALESCE(p.sign_up_time, '3000-01-01') as register_day
                 FROM participant p
                        LEFT OUTER JOIN participant_summary ps ON p.participant_id = ps.participant_id
                 %(filters_p)s) a left join
--- a/rest-api/dao/participant_counts_over_time_service.py
+++ b/rest-api/dao/participant_counts_over_time_service.py
@@ -30,12 +30,24 @@ class ParticipantCountsOverTimeService(BaseDao):
     filters_sql_ps = self.get_facets_sql(filters)
     filters_sql_p = self.get_facets_sql(filters, table_prefix='p')
 
+    # check if request stratification is for enromment status
+    query_for_enrollment_status = False
+
+    target_enrollment_statuses = []
+    if 'enrollment_statuses' in filters:
+      target_enrollment_statuses = [str(val) for val in filters['enrollment_statuses']]
+
     if str(stratification) == 'TOTAL':
       strata = ['TOTAL']
       sql = self.get_total_sql(filters_sql_ps)
     elif str(stratification) == 'ENROLLMENT_STATUS':
+      query_for_enrollment_status = True
       strata = [str(val) for val in EnrollmentStatus]
       sql = self.get_enrollment_status_sql(filters_sql_ps, filters_sql_p)
+    elif str(stratification) == 'ENROLLMENT_STATUS_BY_ORDER_TIME':
+      query_for_enrollment_status = True
+      strata = [str(val) for val in EnrollmentStatus]
+      sql = self.get_enrollment_status_by_order_time_sql(filters_sql_ps, filters_sql_p)
     else:
       raise BadRequest('Invalid stratification: %s' % stratification)
 
@@ -48,6 +60,8 @@ class ParticipantCountsOverTimeService(BaseDao):
 
     # Iterate through each result (by date), transforming tabular SQL results
     # into expected list-of-dictionaries response format
+    # set value = 0 if request stratification is for enromment status
+    # but key is not in target_enrollment_statuses
     try:
       results = cursor.fetchall()
       for result in results:
@@ -56,7 +70,9 @@ class ParticipantCountsOverTimeService(BaseDao):
         values = result[:-1]
         for i, value in enumerate(values):
           key = strata[i]
-          if value == None:
+          if value is None or (query_for_enrollment_status
+                               and 'enrollment_statuses' in filters
+                               and key not in target_enrollment_statuses):
             value = 0
           metrics[key] = int(value)
         results_by_date.append({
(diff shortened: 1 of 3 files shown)


rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -30,12 +30,24 @@ def get_filtered_results(self, start_date, end_date, filters, stratification='EN
     filters_sql_ps = self.get_facets_sql(filters)
     filters_sql_p = self.get_facets_sql(filters, table_prefix='p')
 
+    # check if request stratification is for enromment status
```
danqrodney: enromment -> enrollment

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -48,6 +60,8 @@ def get_filtered_results(self, start_date, end_date, filters, stratification='EN
 
     # Iterate through each result (by date), transforming tabular SQL results
     # into expected list-of-dictionaries response format
+    # set value = 0 if request stratification is for enromment status
```
danqrodney: enromment -> enrollment

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -30,12 +30,24 @@ def get_filtered_results(self, start_date, end_date, filters, stratification='EN
     filters_sql_ps = self.get_facets_sql(filters)
     filters_sql_p = self.get_facets_sql(filters, table_prefix='p')
 
+    # check if request stratification is for enromment status
+    query_for_enrollment_status = False
+
+    target_enrollment_statuses = []
+    if 'enrollment_statuses' in filters:
```
danqrodney: Is this a new requirement? Why do we need to be able to specify what enrollment statuses to return?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -56,7 +70,9 @@ def get_filtered_results(self, start_date, end_date, filters, stratification='EN
         values = result[:-1]
         for i, value in enumerate(values):
           key = strata[i]
-          if value == None:
+          if value is None or (query_for_enrollment_status
```
danqrodney: If we're going to support not getting all enrollment statuses back, then we probably shouldn't run the SQL necessary to calculate them, or populate anything in the response for them.

But I think it's OK for us to always return all enrollment statuses?

wangy70:
Do you know how to handle this Codacy/PR Quality issue ?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -176,70 +192,163 @@ def get_enrollment_status_sql(self, filters_sql_ps, filters_sql_p):
...
+    sql_by_sample_status_time = """
+      select sum(is_register) as registered_participants,
+       sum(is_member) as member_participants,
+       sum(is_core) as full_participants,
+       z.day
+      from
+      (
+           select
+            x.participant_id,
+            case when register_day<=y.day and member_day>y.day then 1 else 0 end as is_register,
```
danqrodney: So I believe that previously, registered included member, and member included full (as registered participants are a superset of members, and members a supserset of full participants.) I think we want to preserve that, which makes this logic incorrect?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -176,70 +192,163 @@ def get_enrollment_status_sql(self, filters_sql_ps, filters_sql_p):
...
-               registered.cnt registered_cnt,
-               member.cnt member_cnt,
-               full.cnt full_cnt
-            FROM calendar c2
-            LEFT OUTER JOIN
-            (SELECT COUNT(*) cnt,
-                (CASE WHEN (enrollment_status = 1 OR enrollment_status IS NULL) THEN
-                  DATE(p.sign_up_time)
-                  ELSE NULL END) day
+    sql_by_sample_status_time = """
```
danqrodney: Have you run this SQL against sandbox or prod to see how well it performs at scale?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -176,70 +192,163 @@ def get_enrollment_status_sql(self, filters_sql_ps, filters_sql_p):
...
+            x.participant_id,
+            case when register_day<=y.day and member_day>y.day then 1 else 0 end as is_register,
+            case when member_day<=y.day and core_day > y.day then 1 else 0 end as is_member,
+            case when core_day<=y.day then 1 else 0 end as is_core,
+            y.day
+          from
+          (select a.participant_id,a.register_day,COALESCE(b.member_day,'3000-01-01') member_day,COALESCE(b.core_day,'3000-01-01') core_day from
+                (SELECT p.participant_id,
+                       CASE
+                         WHEN (enrollment_status IS NULL or enrollment_status >= 1) THEN DATE(p.sign_up_time)
```
danqrodney: I realized that I think this can just be COALESCE(DATE(p.sign_up_time), '3000-01-01') unconditionally... sign_up_time won't be set for anyone who hasn't signed up.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -176,70 +192,163 @@ def get_enrollment_status_sql(self, filters_sql_ps, filters_sql_p):
...
+                             COALESCE(sample_status_1sal2_time, '3000-01-01')
+                               ))
+                       )
+                       else
+                           Date('3000-01-01')
+                       end core_day
+               from participant_summary ps
+               %(filters_ps)s) b
+               on b.participant_id=a.participant_id
+          where register_day<=:end_date or register_day='3000-01-01'
```
danqrodney: Why is register_day = '3000-01-01' needed?

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -176,70 +192,163 @@ def get_enrollment_status_sql(self, filters_sql_ps, filters_sql_p):
...
+                           GREATEST(
+                             COALESCE(sign_up_time, '1000-01-01'),
+                             COALESCE(consent_for_electronic_health_records_time, '1000-01-01'),
+                             COALESCE(questionnaire_on_the_basics_time, '1000-01-01'),
+                             COALESCE(questionnaire_on_lifestyle_time, '1000-01-01'),
+                             COALESCE(questionnaire_on_overall_health_time, '1000-01-01'),
+                             COALESCE(physical_measurements_finalized_time, '1000-01-01')
+                               ),
+                           LEAST(
+                             COALESCE(sample_status_1ed04_time, '3000-01-01'),
```
danqrodney: If the only difference between this SQL and the SQL below is sample_status_* instead of sample_order_status_*, we should have a shared SQL template and just substitute in one or the other, rather than having everything else in two places.

rest-api/dao/participant_counts_over_time_service.py
```diff
@@ -176,70 +192,163 @@ def get_enrollment_status_sql(self, filters_sql_ps, filters_sql_p):
...
+                       LEFT OUTER JOIN participant_summary ps ON p.participant_id = ps.participant_id
+                %(filters_p)s) a left join
+               (select participant_id,
+                      (
+                          CASE
+                            WHEN enrollment_status>=2 and ps.consent_for_electronic_health_records = 1 THEN DATE(COALESCE(ps.consent_for_electronic_health_records_time, '3000-01-01'))
+                            ELSE Date('3000-01-01') END
+                          ) member_day,
+                       case when
+                           enrollment_status>=2
```
danqrodney: Shouldn't this be >= 3? Or = 3?

danqrodney:
You can ignore the Codacy thing here. We are constructing the filters that are substituted in so there should be no risk of SQL injection.

rest-api/dao/participant_counts_over_time_service.py
wangy70: In the new sql, we depend on the register time, member time and core time, so for some case even the enrollment_status is 3, but it's still in member status for that query date. In this case when we query for core participants, since for that query date it's still in member status, so there will not be 0 in the member count in the result. That's why I need to check it here.

rest-api/dao/participant_counts_over_time_service.py
wangy70: It's not a new requirement, just for later use to check the sql execution result.

rest-api/dao/participant_counts_over_time_service.py
wangy70: If a participant is already a member, we don't want to count it for registered, same for core, right? So in this SQL, if a participant is not a member, the member_day will be 3001-01-01, same for core_day.

rest-api/dao/participant_counts_over_time_service.py
wangy70: Yes, it usually return in 1 or 2 secs. And I run it in prod with a time range of 8 month of all data, it will return in secs.

rest-api/dao/participant_counts_over_time_service.py
wangy70: I left join the register ones with member and core, so we need to keep all the participants, if remove this, there will be only register ones.

rest-api/dao/participant_counts_over_time_service.py
wangy70: Actually, I try to do this. There is another different place that make me give up. For the one use sample status time, when I try to get the  core_day, I use enrollment_status=3; For the other one, I use enrollment_status>=2, because for some records with order time, the enrollment status is still 2, but we want to count it as core, so that's why I need it as enrollment_status>=2.

rest-api/dao/participant_counts_over_time_service.py
wangy70: The answer is same as the above one.

rest-api/dao/participant_counts_over_time_service.py
wangy70: Wow, good point.

rest-api/dao/participant_counts_over_time_service.py
wangy70: I just checked the dashboard UI, registered, member and core are separated.  Do you mean they separate them at the front end?

rest-api/dao/participant_counts_over_time_service.py
wangy70: fixed

rest-api/dao/participant_counts_over_time_service.py
wangy70: fixed

rest-api/dao/participant_counts_over_time_service.py
wangy70: done

wangy70:
[DA-671] is resolved by using enrollment_status_member_time, enrollment_status_core_ordered_sample_time and enrollment_status_core_stored_sample_time. Close this PR.

PR #810, opened 2018-10-15
DA-758 add support for suspended participation

Seeker1911:
Added feature to use existing suspension_status to filter out contact fields only while allowing the continued use of PM&B.

### DIFF
--- a/rest-api/dao/participant_summary_dao.py
+++ b/rest-api/dao/participant_summary_dao.py
@@ -14,7 +14,7 @@ from dao.organization_dao import OrganizationDao
 from dao.site_dao import SiteDao
 from model.config_utils import to_client_biobank_id
 from model.participant_summary import ParticipantSummary, WITHDRAWN_PARTICIPANT_FIELDS, \
-  WITHDRAWN_PARTICIPANT_VISIBILITY_TIME
+  WITHDRAWN_PARTICIPANT_VISIBILITY_TIME, SUSPENDED_PARTICIPANT_FIELDS
 from model.utils import to_client_participant_id, get_property_type
 from participant_enums import QuestionnaireStatus, PhysicalMeasurementsStatus, SampleStatus, \
   EnrollmentStatus, SuspensionStatus, WithdrawalStatus, get_bucketed_age
@@ -344,6 +344,11 @@ class ParticipantSummaryDao(UpdatableDao):
         (model.withdrawalTime is None or
          model.withdrawalTime < clock.CLOCK.now() - WITHDRAWN_PARTICIPANT_VISIBILITY_TIME)):
       result = {k: result.get(k) for k in WITHDRAWN_PARTICIPANT_FIELDS}
+
+    elif model.withdrawalStatus != WithdrawalStatus.NO_USE and \
+      model.suspensionStatus == SuspensionStatus.NO_CONTACT:
+      for i in SUSPENDED_PARTICIPANT_FIELDS:
+        result[i] = UNSET
 
     result['participantId'] = to_client_participant_id(model.participantId)
     biobank_id = result.get('biobankId')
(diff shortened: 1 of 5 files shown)


rest-api/model/participant_summary.py
```diff
@@ -26,6 +25,11 @@
 # queries that don't ask for withdrawn participants.
 WITHDRAWN_PARTICIPANT_VISIBILITY_TIME = datetime.timedelta(days=2)
 
+# suspended participants don't allow contact but can still use samples. These fields
+# will not be returned when queried on suspended participant.
+SUSPENDED_PARTICIPANT_FIELDS = ['zipCode', 'city', 'streetAddress', 'phoneNumber',
```
calbach: Where did this list of fields come from? I didn't see it in the attached ticket or document.

danqrodney: +1. Barring any requirements otherwise, I'd assume we can return everything for suspended participants... it's only withdrawn participants that we need to filter stuff out for.

Seeker1911: There was no strict requirement to remove those fields specifically. But there is a no_contact requirement similar to withdrawn. I just found it prudent, if you both disagree I'll remove it.

danqrodney: My understanding is withdrawn means "I'm not in the program anymore -- stop using my data". So after 48 hours, their data  (apart from a few fields) disappears from everywhere.

Suspended just means "don't contact me". So there's no reason we should filter their data out. Though you should confirm with HPRO folks.

Basically I think this requirement is a no-op; we already support what they want.

Seeker1911: OK. Spoke with the PM team for RDR and HPRO, they do want this left in.

rest-api/dao/participant_summary_dao.py
```diff
@@ -345,6 +345,11 @@ def to_client_json(self, model):
          model.withdrawalTime < clock.CLOCK.now() - WITHDRAWN_PARTICIPANT_VISIBILITY_TIME)):
       result = {k: result.get(k) for k in WITHDRAWN_PARTICIPANT_FIELDS}
 
+    elif model.withdrawalStatus != WithdrawalStatus.NO_USE and \
+      model.suspensionStatus == SuspensionStatus.NO_CONTACT:
+      for i in SUSPENDED_PARTICIPANT_FIELDS:
+        result[i] = UNSET
```
danqrodney: Just confirming that as soon as someone is suspended, these will get cleared out (unlike withdrawn participants, that will show up with everything for 48 hours.) So there isn't going to be an opportunity for folks at sites to reach out to them using their contact info after they suspend their account. (That may be totally fine... let's just make sure that's the intention.)

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -287,6 +287,56 @@ def test_admin_withdrawal_returns_right_info(self):
...
+      self.post_demographics_questionnaire(participant_id, questionnaire_id, **answers)
+
+    with FakeClock(TIME_2):
+      path = 'Participant/%s' % participant_id
+      participant['suspensionStatus'] = "NO_CONTACT"
+      self.send_put(path, participant, headers={'If-Match': 'W/"1"'})
+
+    with FakeClock(TIME_3):
+      response = self.send_get('Participant/%s/Summary' % participant_id)
+      self.assertEqual(response['email'], 'UNSET')
```
danqrodney: You should probably check some other fields to make they are *not* UNSET.

rest-api/dao/participant_summary_dao.py
danqrodney: Where does it say that? That's not how things are implemented now... suspended currently just means "don't contact me", not "don't accept data for this person" (unlike withdrawal)

PR #812, opened 2018-10-16
[DA-492] parse primary language from questionnaire response and store to ps

wangy70:
parse primary language from questionnaire response extension and store it to participant summary primaryLanguage field


rest-api/dao/questionnaire_response_dao.py
```diff
@@ -228,6 +229,15 @@ def _update_participant_summary(
         participant_summary.race = race
         something_changed = True
 
+    # if primary language provided in the response, set the new value.
+    resource_json = json.loads(questionnaire_response.resource)
```
danqrodney: Rather than re-parsing resource_json here, we should have _update_participant_summary take either the JSON or the language as a parameter.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -228,6 +229,15 @@ def _update_participant_summary(
...
+    for extension in resource_json.get('extension', []):
+      if extension['url'] == _LANGUAGE_EXTENSION:
+        if participant_summary.primaryLanguage != extension['valueCode']:
+          participant_summary.primaryLanguage = extension['valueCode']
```
danqrodney: I'm wondering if we need some validation here to make sure it's a valid ISO language code?

rest-api/model/participant_summary.py
```diff
@@ -47,6 +47,7 @@ class ParticipantSummary(Base):
   phoneNumber = Column('phone_number', String(80))
   loginPhoneNumber = Column('login_phone_number', String(80))
   email = Column('email', String(255))
+  primaryLanguage = Column('primary_language', String(80))
   recontactMethodId = Column('recontact_method_id', Integer, ForeignKey('code.code_id'))
   languageId = Column('language_id', Integer, ForeignKey('code.code_id'))
```
danqrodney: We at a minimum should deprecate this field; we may be able to drop the column from the database and API entirely. We should ask HPRO folks about it.

rest-api/test/unit_test/api_test/questionnaire_response_api_test.py
```diff
@@ -87,6 +87,11 @@ def test_insert(self):
     get_response = self.send_get(_questionnaire_response_url(participant_id) + "/" + response['id'])
     self.assertJsonResponseMatches(resource, get_response)
 
+    # verify the primary language data is stored in participant summary
+    ps = self.send_get('Participant/%s/Summary' % participant_id)
+    primary_language = resource.get('extension',[])
+    self.assertEquals(primary_language[0]['valueCode'], ps['primaryLanguage'])
```
danqrodney: Rather than asserting they are equal, can you assert the expected value? Is the questionnaire response above actually setting this?

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -228,6 +229,15 @@ def _update_participant_summary(
         participant_summary.race = race
         something_changed = True
 
+    # if primary language provided in the response, set the new value.
+    resource_json = json.loads(questionnaire_response.resource)
+    for extension in resource_json.get('extension', []):
+      if extension['url'] == _LANGUAGE_EXTENSION:
+        if participant_summary.primaryLanguage != extension['valueCode']:
```
danqrodney: I'm not sure whether we should change primaryLanguage on participant summary  to match the latest questionnaire response, or just for the initial consent. We should inquire.

rest-api/alembic/versions/be736e26410b_add_primary_language_to_participant_.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('participant_summary', sa.Column('primary_language', sa.String(length=80), nullable=True))
```
Seeker1911: It may be better to do these in one Alembic migration. Probably doesn't matter though.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -145,13 +147,15 @@ def _update_field(self, participant_summary, field_name, field_type, answer):
     return False
 
   def _update_participant_summary(
-      self, session, questionnaire_response, code_ids, questions, questionnaire_history):
+      self, session, questionnaire_response, code_ids, questions, questionnaire_history,
+    resource_json):
```
Seeker1911: please fix hanging indent
see: https://www.python.org/dev/peps/pep-0008/#id17

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -241,6 +244,15 @@ def _update_participant_summary(
...
             new_status = dvehr_consent
+          elif code.value == CONSENT_FOR_STUDY_ENROLLMENT_MODULE:
+            # set language of consent to participant summary
+            for extension in resource_json.get('extension', []):
+              if extension['url'] == _LANGUAGE_EXTENSION and \
+                extension['valueCode'] in LANGUAGE_OF_CONSENT:
+                if participant_summary.primaryLanguage != extension['valueCode']:
+                  participant_summary.primaryLanguage = extension['valueCode']
+                  something_changed = True
+                break
```
Seeker1911: Are you sure you want a `break` here ? It will exit the entire for loop. Not sure what all resource_json could contain at this point. It may be right or you may want `continue`.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -790,7 +782,6 @@ def testQuery_oneParticipant(self):
       'sexualOrientation': 'straight',
       'phoneNumber': '512-555-5555',
       'recontactMethod': 'email_code',
-      'language': 'en',
```
Seeker1911: we should still test `primary_language` somewhere for participant summary.

rest-api/alembic/versions/be736e26410b_add_primary_language_to_participant_.py
wangy70: Do you mean submit the Alembic change in a separate branch?

rest-api/dao/questionnaire_response_dao.py
wangy70: Yes, it's break here. It just needs the first matched value.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
wangy70: Good point. I added it.

rest-api/alembic/versions/be736e26410b_add_primary_language_to_participant_.py
Seeker1911: generally that is considered a best practice, also, we can probably remove the original languageId from the original alembic file instead of creating it and then deleting it everytime. Maybe Dan can comment on that.

danqrodney: We should not remove the language ID from the original alembic file; if we want to drop the column we want that applied to existing databases.

We certainly can merge together these two migrations; delete  both of them, reset your database, and run generate_schema again, and Alembic should do the drop and create together.

rdr_common/code_constants.py
```diff
@@ -82,3 +82,5 @@
 UNSET = 'UNSET'
 UNMAPPED = 'UNMAPPED'
 BASE_VALUES = [UNSET, UNMAPPED, PMI_SKIP_CODE]
+
+LANGUAGE_OF_CONSENT = ['en', 'es']
```
danqrodney: We may want to add a comment saying English and Spanish are the only accepted languages for now.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -241,6 +244,15 @@ def _update_participant_summary(
             new_status = QuestionnaireStatus.SUBMITTED_NO_CONSENT
           elif code.value == CONSENT_FOR_DVEHR_MODULE:
             new_status = dvehr_consent
+          elif code.value == CONSENT_FOR_STUDY_ENROLLMENT_MODULE:
+            # set language of consent to participant summary
+            for extension in resource_json.get('extension', []):
+              if extension['url'] == _LANGUAGE_EXTENSION and \
```
danqrodney: You may want to use extension.get() for these, just in case there's some weird extension that doesn't contain url, or has the right url but no valueCode. (You'll get a KeyError if you use [].)

rest-api/test/unit_test/unit_test_util.py
```diff
@@ -577,7 +578,8 @@ def sort_lists(obj):
 
 
 def make_questionnaire_response_json(participant_id, questionnaire_id, code_answers=None,
-                                string_answers=None, date_answers=None, uri_answers=None):
+                                string_answers=None, date_answers=None, uri_answers=None,
+                                     consent_language=None):
```
danqrodney: You might want to call this "language" rather than "consent_language"... it will be present on other questionnaire responses that aren't about consent.

rest-api/test/unit_test/unit_test_util.py
```diff
@@ -613,6 +615,8 @@ def make_questionnaire_response_json(participant_id, questionnaire_id, code_answ
   return {"resourceType": "QuestionnaireResponse",
           "status": "completed",
           "subject": { "reference": "Patient/{}".format(participant_id) },
+          "extension":[{"url":"http://hl7.org/fhir/StructureDefinition/iso21090-ST-language",
```
danqrodney: You should only add this when language is not None.

rest-api/test/unit_test/api_test/questionnaire_response_api_test.py
```diff
@@ -225,6 +224,99 @@ def test_demographic_questionnaire_responses(self):
               }
     self.assertJsonResponseMatches(expected, summary)
 
+  def test_consent_with_extension_language(self):
+    with FakeClock(TIME_1):
+      participant_id = self.create_participant()
+      self.send_consent(participant_id, consent_language='es')
```
danqrodney: We should also have a test that verifies when you submit a questionnaire response that is *not* consent, that the language doesn't change.

rest-api/model/participant_summary.py
```diff
@@ -47,8 +47,8 @@ class ParticipantSummary(Base):
   phoneNumber = Column('phone_number', String(80))
   loginPhoneNumber = Column('login_phone_number', String(80))
   email = Column('email', String(255))
+  primaryLanguage = Column('primary_language', String(80))
   recontactMethodId = Column('recontact_method_id', Integer, ForeignKey('code.code_id'))
-  languageId = Column('language_id', Integer, ForeignKey('code.code_id'))
```
danqrodney: So I believe removing this field will change our responses so that instead of having "language": "UNSET" for every participant in prod, there simply won't be a language field. Which is technically a breaking change; we need clients to migrate away from using language before we can get rid of it entirely.

The simplest thing is to keep this field around until clients have migrated to use primaryLanguage, and then get rid of it. Alternatively, you change participant_summary_dao.to_client_json to always set language to "UNSET".

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -241,6 +244,15 @@ def _update_participant_summary(
             new_status = QuestionnaireStatus.SUBMITTED_NO_CONSENT
           elif code.value == CONSENT_FOR_DVEHR_MODULE:
             new_status = dvehr_consent
+          elif code.value == CONSENT_FOR_STUDY_ENROLLMENT_MODULE:
+            # set language of consent to participant summary
+            for extension in resource_json.get('extension', []):
+              if extension.get('url') == _LANGUAGE_EXTENSION and \
```
wangy70: change to get() to avoid KeyError

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -241,6 +244,15 @@ def _update_participant_summary(
...
+                extension.get('valueCode') in LANGUAGE_OF_CONSENT:
```
danqrodney: if extension.get('valueCode') is NOT in LANGUAGE_OF_CONSENT, we may want to log a warning here...

rest-api/model/participant_summary.py
danqrodney: Let's add a comment saying this is deprecated?

wangy70: done

rest-api/dao/questionnaire_response_dao.py
wangy70: done

PR #816, opened 2018-10-18
Add external ID to participant model.

Seeker1911:


rest-api/alembic/versions/6e5980478a0e_add_external_id.py
```diff
@@ -0,0 +1,64 @@
...
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('participant', sa.Column('external_id', sa.BigInteger(), nullable=True))
+    op.create_unique_constraint(None, 'participant', ['external_id'])
+    op.add_column('participant_history', sa.Column('external_id', sa.BigInteger(), nullable=True))
+    op.create_unique_constraint(None, 'participant_history', ['external_id'])
```
danqrodney: external_id can't be unique in history, as there will be multiple versions of the same participant... you'll need to add the unique constraint in participant only.

Seeker1911: 🤦‍♂️ yep, fixed.

PR #817, opened 2018-10-18
Fix stable config errors.

Seeker1911:


rest-api/config/config_stable.json
```diff
@@ -53,25 +53,18 @@
       },
       "description": "Massive subnet mask. see https://precisionmedicineinitiative.atlassian.net/browse/PD-2412"
     },
-    "awardee-ucirvine@all-of-us-ops-data-api-stable.iam.gserviceaccount.com": {
```
danqrodney: I'm assuming this is intentionally being removed?

Seeker1911: yes. The ip was moved to awardee-california. Changes were made to service accounts without us being told.

PR #820, opened 2018-10-24
[DA-769] add member time,dna sample order time and dna sample status time fiel…

wangy70:
add member time,dna sample order time and dna sample status time fields to ps


rdr_common/code_constants.py
```diff
@@ -79,6 +79,8 @@
                  '1PXR2']
 BIOBANK_TESTS_SET = frozenset(BIOBANK_TESTS)
 
+BIOBANK_DNA_TESTS = ['1ED10', '2ED10', '1ED04', '1SAL', '1SAL2']
```
danqrodney: This is already in our config, here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/config/base_config.json#L32

We should avoid duplicating this in constants if possible.

rest-api/dao/biobank_order_dao.py
```diff
@@ -214,11 +215,21 @@ def _set_participant_summary_fields(self, obj, participant_summary):
...
+
+    has_dna_test = False
     for sample in obj.samples:
+      if sample.test in BIOBANK_DNA_TESTS:
+        has_dna_test = True
       status_field = 'sampleOrderStatus' + sample.test
       status, time = self._get_order_status_and_time(sample, obj)
       setattr(participant_summary, status_field, status)
       setattr(participant_summary, status_field + 'Time', time)
+    if has_dna_test:
```
danqrodney: The logic for being a core participant is more complicated than this... it's not just having DNA samples stored / ordered, it's also having completed EHR consent, physical measurements, and baseline questionnaires. See:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L45

(for SQL run from biobank samples pipeline) and

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/participant_summary_dao.py#L327

for Python logic run when updating the participant summary. This is what needs to be updated.

rest-api/dao/participant_summary_dao.py
```diff
@@ -150,6 +150,48 @@ def _get_dna_isolates_sql_and_params():
...
+        AND biobank_stored_sample.test IN {tests_sql}
+      ) X  
+  )
+  """.format(confirm_value_sql=confirm_value_sql, tests_sql=tests_sql)
+
+  sql = """
+    UPDATE
+      participant_summary
+    SET
+      enrollment_status_core_stored_sample_time = {confirm_time_sql}
```
danqrodney: We may want to have a WHERE clause so that this only gets set for participant summaries where it hasn't been set already. 

You may also be able to combine this with the enrollment status updating SQL... not sure if it matters.

danqrodney: Yeah, that rename is fine by me. (A little less explicit, but the official definition of "core" is based on stored sample time, so this is fine.)

rest-api/test/unit_test/dao_test/biobank_order_dao_test.py
```diff
@@ -15,9 +15,10 @@
 
 
 class BiobankOrderDaoTest(SqlTestBase):
+  config.getSettingList(config.DNA_SAMPLE_TEST_CODES)
   _A_TEST = BIOBANK_TESTS[0]
   _B_TEST = BIOBANK_TESTS[1]
-  _DNA_TEST = BIOBANK_DNA_TESTS[0]
+  _DNA_TEST = config.getSettingList(config.DNA_SAMPLE_TEST_CODES)[0]
```
danqrodney: Try just hardcoding this to the first value from the config, and see if it passes.

You can also try inlining the import of config, instead of importing it at the top of the file.

(I believe the problem here is that config is getting imported before memcache is set up for the test.)

rest-api/dao/participant_summary_dao.py
```diff
@@ -322,6 +375,11 @@ def update_enrollment_status(self, summary):
                                                          summary.numCompletedBaselinePPIModules,
                                                          summary.physicalMeasurementsStatus,
                                                          summary.samplesToIsolateDNA)
+    summary.enrollmentStatusMemberTime = self.calculate_member_time(consent, summary)
+    summary.enrollmentStatusCoreOrderedSampleTime = self.calculate_core_ordered_sample_time(
```
wangy70: Update member time, ordered time and stored time in update_enrollment_status, so that when questionnaire response/PM/BO comes, it can be called and set those timestamps.

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +394,67 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+    elif consent_for_study_enrollment_and_ehr:
+      return participant_summary.consentForElectronicHealthRecordsTime
+    else:
+      return None
+
+  def calculate_core_stored_sample_time(self, consent_for_study_enrollment_and_ehr,
+                                        participant_summary):
+    if consent_for_study_enrollment_and_ehr and \
+      participant_summary.numCompletedBaselinePPIModules == \
+      self._get_num_baseline_ppi_modules() and \
```
wangy70: Check the relevant inputs first, only calculate sample times when all other criteria meet.

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +394,67 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+      participant_summary.physicalMeasurementsStatus == PhysicalMeasurementsStatus.COMPLETED and \
+      participant_summary.samplesToIsolateDNA == SampleStatus.RECEIVED:
```
wangy70: Similar logic with ordered time, just add one more criteria samplesToIsolateDNA

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -1086,6 +1089,56 @@ def testWithdrawThenPair(self):
     self.assertEquals('NO_USE', ps_1['withdrawalStatus'])
     self.assertEquals(TIME_2.isoformat(), ps_1.get('withdrawalTime'))
 
+  def test_member_ordered_stored_times(self):
+    questionnaire_id = self.create_questionnaire('questionnaire3.json')
+    questionnaire_id_1 = self.create_questionnaire('all_consents_questionnaire.json')
```
wangy70: Add a new test case for enrollmentStatusMemberTime, enrollmentStatusCoreOrderedSampleTime and enrollmentStatusCoreStoredSampleTime.

rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
```diff
@@ -292,6 +294,36 @@ def add_sample(test_code, sample_id):
     self.assertEquals(summary.numBaselineSamplesArrived, 1)
     self.assertNotEqual(init_last_modified, summary.lastModified)
 
+  def test_only_update_dna_sample(self):
+    dna_tests = ["1ED10", "1SAL2"]
```
wangy70: Add a new test case, make sure only update dna sample will not update enrollmentStatusCoreStoredSampleTime, because other criteria not meet.

rest-api/dao/biobank_order_dao.py
```diff
@@ -205,6 +205,7 @@ def _update_participant_summary(self, session, obj):
                        obj.participantId)
     raise_if_withdrawn(participant_summary)
     self._set_participant_summary_fields(obj, participant_summary)
+    participant_summary_dao.update_enrollment_status(participant_summary)
```
wangy70: When a biobank order comes, call update_enrollment_status to update.

rest-api/model/participant_summary.py
```diff
@@ -78,6 +78,15 @@ class ParticipantSummary(Base):
 
   signUpTime = Column('sign_up_time', UTCDateTime)
 
+  # The time that this participant become a member
+  enrollmentStatusMemberTime = Column('enrollment_status_member_time', UTCDateTime)
+  # The time when we get the first stored sample
+  enrollmentStatusCoreStoredSampleTime = Column('enrollment_status_core_stored_sample_time',
```
wangy70: Keep the field name not change, because if change this to core_time, the enrollmentStatusCoreOrderedSampleTime will become more confuse.

rest-api/dao/participant_summary_dao.py
```diff
@@ -150,6 +150,53 @@ def _get_dna_isolates_sql_and_params():
...
+
+  status_time_sql = '%s' % ','.join(["""COALESCE(sample_status_%s_time, '3000-01-01')""" % item
+                                     for item in dns_test_list])
+
+  sub_sql = """
+  (
+    SELECT 
+      MAX(
+        consent_for_electronic_health_records_time,
+        questionnaire_on_the_basics_time,
```
danqrodney: You may want to generate the SQL for these questionnaire fields based on config (similar to status_time_sql), like we do when running Python here:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/dao/questionnaire_response_dao.py#L41

That way when the config changes we won't have to update the code.

rest-api/dao/participant_summary_dao.py
```diff
@@ -150,6 +150,53 @@ def _get_dna_isolates_sql_and_params():
...
+                ) = '3001-01-01' THEN NULL
+            ELSE MIN(
+                {status_time_sql}
+                )
+        END
+      )
+    FROM
+      participant_summary ps
+    WHERE
+      ps.enrollment_status = 3
```
danqrodney: I'm not sure you need this WHERE condition or the one on line 186; they duplicate the WHERE conditions of the outer SQL.

rest-api/dao/participant_summary_dao.py
```diff
@@ -150,6 +150,53 @@ def _get_dna_isolates_sql_and_params():
       params
   )
 
+def _get_sample_status_time_sql_and_params():
```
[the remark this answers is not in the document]
danqrodney: I think this is probably fine. The only issue is we have a potentially big database transaction open while these statements are running... but I'm not sure it'd be any better if we put everything into one query really.

rest-api/dao/participant_summary_dao.py
danqrodney: It'd be nice if we could have a helper function that both calculate_core_stored_sample_time and calculate_core_ordered_sample_time call, with the shared logic.

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +394,67 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+
+      keys = ['sampleStatus%sTime' % test
+              for test in config.getSettingList(config.DNA_SAMPLE_TEST_CODES)]
+      sample_stored_time_list = \
+        [v for k, v in participant_summary if k in keys and v is not None]
+
+      sample_stored_time = min(sample_stored_time_list) if sample_stored_time_list else None
+
+      if sample_stored_time is not None:
+        return max([sample_stored_time,
```
danqrodney: I would add:

if participant_summary.enrollmentStatusCoreStoredSampleTime:
  return participant_summary.enrollmentStatusCoreStoredSampleTime 

Basically, we only want to set this field if it isn't set already. This is because the timestamps we are deriving this from get updated when repeated data comes in (e.g. we get physical measurements or the same questionnaire response twice.) See https://precisionmedicineinitiative.atlassian.net/browse/DA-607.

We should *clear* the field if all the criteria aren't met anymore -- that can happen if something gets canceled, etc. -- but we shouldn't set the field to a new value if it's already set.

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +394,67 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+      participant_summary.physicalMeasurementsStatus == PhysicalMeasurementsStatus.COMPLETED:
+      keys = ['sampleOrderStatus%sTime' % test
+              for test in config.getSettingList(config.DNA_SAMPLE_TEST_CODES)]
+      sample_ordered_time_list = \
+        [v for k, v in participant_summary if k in keys and v is not None]
+
+      sample_ordered_time = min(sample_ordered_time_list) if sample_ordered_time_list else None
+
+      if sample_ordered_time is not None:
+        return max([sample_ordered_time,
```
danqrodney: Similar to above, I'd add:

if participant_summary.enrollmentStatusCoreOrderedSampleTime:
return participant_summary.enrollmentStatusCoreOrderedSampleTime

rest-api/test/unit_test/api_test/participant_summary_api_test.py
danqrodney: We may want to add some tests showing that these fields get cleared when physical measurements or biobank orders get canceled, or somebody withdraws EHR consent.

rest-api/dao/participant_summary_dao.py
```diff
@@ -150,6 +150,54 @@ def _get_dna_isolates_sql_and_params():
...
+def _get_sample_status_time_sql_and_params():
+  """Gets SQL that to update enrollmentStatusCoreStoredSampleTime field
+  on the participant summary.
+  """
+
+  dns_test_list = config.getSettingList(config.DNA_SAMPLE_TEST_CODES)
+
+  status_time_sql = '%s' % ','.join(["""COALESCE(sample_status_%s_time, '3000-01-01')""" % item
+                                     for item in dns_test_list])
+  baseline_ppi_module_fields = config.getSettingList(config.BASELINE_PPI_QUESTIONNAIRE_FIELDS, [])
```
wangy70: generate sql with baseline ppi module config

rest-api/dao/participant_summary_dao.py
```diff
@@ -150,6 +150,54 @@ def _get_dna_isolates_sql_and_params():
...
+                ) = '3001-01-01' THEN NULL
+            ELSE MIN(
+                {status_time_sql}
+                )
+        END
+      )
+    FROM
+      participant_summary ps
+    WHERE
+      ps.participant_id = participant_summary.participant_id
```
wangy70: remove dup criteria

rest-api/dao/participant_summary_dao.py
wangy70: add a function for the dup logic

rest-api/dao/physical_measurements_dao.py
```diff
@@ -296,9 +296,10 @@ def _update_participant_summary(self, session, obj):
     participant_summary.physicalMeasurementsCreatedSiteId = obj.createdSiteId
     participant_summary.physicalMeasurementsFinalizedSiteId = obj.finalizedSiteId
     participant_summary.lastModified = clock.CLOCK.now()
-    if 'physicalMeasurementsStatus' in obj:
-      # If this is an update it should have this field. New inserts may not.
-      participant_summary.physicalMeasurementsStatus = obj.physicalMeasurementsStatus
+
+    if obj.status and obj.status == PhysicalMeasurementsStatus.CANCELLED:
```
wangy70: fix a bug, update the participant summary when a physical measurement cancel request comes

rest-api/test/unit_test/api_test/participant_summary_api_test.py
wangy70: Done

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +395,62 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+
+  def calculate_core_stored_sample_time(self, consent_for_study_enrollment_and_ehr,
+                                        participant_summary):
+    if consent_for_study_enrollment_and_ehr and \
+      participant_summary.numCompletedBaselinePPIModules == \
+      self._get_num_baseline_ppi_modules() and \
+      participant_summary.physicalMeasurementsStatus == PhysicalMeasurementsStatus.COMPLETED and \
+      participant_summary.samplesToIsolateDNA == SampleStatus.RECEIVED:
+
+      if participant_summary.enrollmentStatusCoreStoredSampleTime:
```
wangy70: add this logic to "stored" but not for "ordered", because when a biobank order cancel request comes, we need to re-calculate the order sample timestamp.

danqrodney: So even in that circumstance, I think we want this kind of logic... what you could do is calculate the the max core sample time, and if it's not None, then use the existing value if there is one, otherwise use the new value.

rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
danqrodney: We should have a similar test that the CoreOrderedSampleTime doesn't update once it's set (except to be set to None). See above.

rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
```diff
@@ -292,6 +294,36 @@ def add_sample(test_code, sample_id):
...
+
+    confirmed_time_0 = datetime.datetime(2018, 3, 1)
+    add_sample(p_dna_samples, dna_tests[0], '11111', confirmed_time_0)
+
+    self.dao.update_from_biobank_stored_samples()
+
+    self.assertEquals(self.dao.get(p_dna_samples.participantId).samplesToIsolateDNA,
+                      SampleStatus.RECEIVED)
+    # only update dna sample will not update enrollmentStatusCoreStoredSampleTime
+    self.assertEquals(
```
danqrodney: We should probably have tests somewhere that check that enrollment status gets updated appropriately when canceling PMs or biobank orders... maybe updating:

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/api_test/biobank_order_api_test.py#L58

https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/api_test/physical_measurements_api_test.py#L266

wangy70: These two test cases are already in participant_summary_api_test.py

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -1086,6 +1090,260 @@ def testWithdrawThenPair(self):
     self.assertEquals('NO_USE', ps_1['withdrawalStatus'])
     self.assertEquals(TIME_2.isoformat(), ps_1.get('withdrawalTime'))
 
+  def test_member_ordered_stored_times_for_multi_biobank_order(self):
+    questionnaire_id = self.create_questionnaire('questionnaire3.json')
+    questionnaire_id_1 = self.create_questionnaire('all_consents_questionnaire.json')
```
wangy70: Add test case for multiple biobank order, make sure the enrollmentStatusCoreOrderedSampleTime is not updated after set.

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -1086,6 +1090,260 @@ def testWithdrawThenPair(self):
...
+    self._send_biobank_order(participant_id_1, order_json2, time=TIME_2)
+    # make sure enrollmentStatusCoreOrderedSampleTime is not changed
+    ps_1 = self.send_get('Participant/%s/Summary' % participant_id_1)
+    self.assertEquals(TIME_6.isoformat(), ps_1.get('enrollmentStatusMemberTime'))
+    self.assertEquals('2016-01-04T10:55:41', ps_1.get('enrollmentStatusCoreOrderedSampleTime'))
+    self.assertIsNone(ps_1.get('enrollmentStatusCoreStoredSampleTime'))
+
+  def test_member_ordered_stored_times_for_biobank_order_cancel(self):
+    questionnaire_id = self.create_questionnaire('questionnaire3.json')
+    questionnaire_id_1 = self.create_questionnaire('all_consents_questionnaire.json')
```
wangy70: Here is the test case for biobank order cancel

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -1086,6 +1090,260 @@ def testWithdrawThenPair(self):
...
+    self.send_patch(path, request_data=request_data, headers={'If-Match': 'W/"1"'})
+
+    ps_1 = self.send_get('Participant/%s/Summary' % participant_id_1)
+    self.assertIsNone(ps_1.get('enrollmentStatusCoreOrderedSampleTime'))
+    self.assertIsNone(ps_1.get('sampleOrderStatus2ED10Time'))
+    self.assertEquals(ps_1.get('sampleOrderStatus2ED10'), 'UNSET')
+
+  def test_member_ordered_stored_times_for_consent_withdraw(self):
+    questionnaire_id = self.create_questionnaire('questionnaire3.json')
+    questionnaire_id_1 = self.create_questionnaire('all_consents_questionnaire.json')
```
wangy70: Here is the test case for consent withdraw

rest-api/test/unit_test/api_test/participant_summary_api_test.py
```diff
@@ -1086,6 +1090,260 @@ def testWithdrawThenPair(self):
...
+    # two days after withdraw
+    with FakeClock(TIME_5):
+      ps_1 = self.send_get('Participant/%s/Summary' % participant_id_1)
+    self.assertIsNone(ps_1.get('enrollmentStatusMemberTime'))
+    self.assertIsNone(ps_1.get('enrollmentStatusCoreOrderedSampleTime'))
+    self.assertIsNone(ps_1.get('enrollmentStatusCoreStoredSampleTime'))
+
+  def test_member_ordered_stored_times_for_physical_measurement_cancel(self):
+    questionnaire_id = self.create_questionnaire('questionnaire3.json')
+    questionnaire_id_1 = self.create_questionnaire('all_consents_questionnaire.json')
```
wangy70: Here is the test case for physical measurement cancel

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +395,69 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+      participant_summary.physicalMeasurementsStatus == PhysicalMeasurementsStatus.COMPLETED and \
+      participant_summary.samplesToIsolateDNA == SampleStatus.RECEIVED:
+
+      max_core_sample_time = self.calculate_max_core_sample_time(participant_summary,
+                                                                 field_name_prefix='sampleStatus')
+
+      if max_core_sample_time and participant_summary.enrollmentStatusCoreStoredSampleTime:
+        return participant_summary.enrollmentStatusCoreStoredSampleTime
+      else:
+        return max_core_sample_time
```
wangy70: change the logic here, if max_core_sample_time is not None, return the existing value.

rest-api/dao/participant_summary_dao.py
```diff
@@ -336,6 +395,69 @@ def calculate_enrollment_status(self, consent_for_study_enrollment_and_ehr,
...
+      self._get_num_baseline_ppi_modules() and \
+      participant_summary.physicalMeasurementsStatus == PhysicalMeasurementsStatus.COMPLETED:
+
+      max_core_sample_time = self.calculate_max_core_sample_time(
+        participant_summary, field_name_prefix='sampleOrderStatus')
+
+      if max_core_sample_time and participant_summary.enrollmentStatusCoreOrderedSampleTime:
+        return participant_summary.enrollmentStatusCoreOrderedSampleTime
+      else:
+        return max_core_sample_time
```
wangy70: change the logic here, if max_core_sample_time is not None, return the existing value.

PR #823, opened 2018-10-25
DA-768 fix case of physical measurement json.

Seeker1911:


Seeker1911:
Good catch CH but no, that is only used to make a PUT request. Case doesn't matter then. We convert to upper.

PR #827, opened 2018-10-31
DA-783 add biobank oder to history on insert.

Seeker1911:
Add biobank order to history on initial insert
Add test for new orders inserted to history

### DIFF
--- a/rest-api/dao/biobank_order_dao.py
+++ b/rest-api/dao/biobank_order_dao.py
@@ -124,6 +124,9 @@ class BiobankOrderDao(UpdatableDao):
     inserted_obj = super(BiobankOrderDao, self).insert_with_session(session, obj)
     ParticipantDao().add_missing_hpo_from_site(
         session, inserted_obj.participantId, inserted_obj.collectedSiteId)
+    self._update_history(session, obj)
+    self._update_identifier_history(session, obj)
+    self._update_sample_history(session, obj)
     return inserted_obj
 
   def _validate_model(self, session, obj):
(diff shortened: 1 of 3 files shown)


rest-api/test/unit_test/api_test/biobank_order_api_test.py
```diff
@@ -374,6 +375,16 @@ def test_insert_new_order(self):
...
 
+  def test_biobank_history_on_insert(self):
+    with self.bio_dao.session() as session:
+      self.summary_dao.insert(self.participant_summary(self.participant))
+      order_json = load_biobank_order_json(self.participant.participantId,
+                                           filename='biobank_order_2.json')
+      result = self.send_post(self.path, order_json)
+      load_biobank_order_json(self.participant.participantId, filename='biobank_order_1.json')
+      order_history = session.query(BiobankOrderHistory).first()
+      self.assertEqual(result['id'], order_history.biobankOrderId)
```
danqrodney: Also check version?

Can we also check to make sure the identifiers and samples got inserted into their history tables?

Check that this history record sticks around after an update?

rest-api/dao/biobank_order_dao.py
```diff
@@ -124,6 +124,9 @@ def insert_with_session(self, session, obj):
     inserted_obj = super(BiobankOrderDao, self).insert_with_session(session, obj)
     ParticipantDao().add_missing_hpo_from_site(
         session, inserted_obj.participantId, inserted_obj.collectedSiteId)
+    self._update_history(session, obj)
```
danqrodney: These methods get called in three different places, and always get called together. Consider combining them? (Or having a helper method that calls all three which gets invoked in the three places?)

rest-api/test/unit_test/api_test/biobank_order_api_test.py
Seeker1911: done

rest-api/dao/biobank_order_dao.py
Seeker1911: done

PR #828, opened 2018-10-31
[DC-242] Fix PM status check to also accept null

calbach:
All DB values are currently null for this column.


calbach:
FYI verified the bug and the fix manually via SQL client.

PR #829, opened 2018-11-01
Fix ETL scripts

calbach:
Currently must be run from `rest-api`. Fix `--noclone` export flags.

### DIFF
--- a/rest-api/etl/export_etl_results.sh
+++ b/rest-api/etl/export_etl_results.sh
@@ -22,7 +22,10 @@ fi
 PROJECT_AND_ACCOUNT=
 if [ "${PROJECT}" ]
 then
-  PROJECT_AND_ACCOUNT="--project ${PROJECT} --account ${ACCOUNT} --instance_name ${INSTANCE_NAME} --service_account exporter@${PROJECT}.iam.gserviceaccount.com"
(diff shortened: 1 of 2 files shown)


rest-api/etl/export_etl_results.sh
```diff
@@ -22,7 +22,10 @@ fi
 PROJECT_AND_ACCOUNT=
 if [ "${PROJECT}" ]
 then
-  PROJECT_AND_ACCOUNT="--project ${PROJECT} --account ${ACCOUNT} --instance_name ${INSTANCE_NAME} --service_account exporter@${PROJECT}.iam.gserviceaccount.com"
+  PROJECT_AND_ACCOUNT="--project ${PROJECT} --account ${ACCOUNT} --service_account exporter@${PROJECT}.iam.gserviceaccount.com"
+fi
+if [ -n "${INSTANCE_NAME}" ]; then
```
danqrodney: This is fine but you'd never use INSTANCE_NAME without also using PROJECT?

danqrodney: (So I guess this could be inside the former if statement)

calbach: Done.

What does it actually mean to omit project/account. Does that send against local?

danqrodney: Yep.

PR #830, opened 2018-11-01
Mm/external id return participant

Seeker1911:
PTSC requested that existing participants be returned in the case where they create a new participant with the new "externalId" functionality and that externalId exists already.

### DIFF
--- a/rest-api/dao/base_dao.py
+++ b/rest-api/dao/base_dao.py
@@ -331,8 +331,8 @@ class BaseDao(object):
       except IntegrityError, e:
         # SQLite and MySQL variants of the error message, respectively.
         if 'UNIQUE constraint failed' in e.message or 'Duplicate entry' in e.message:
-          if 'external_id' in e.message:
-            raise Conflict('External ID already in use, error: {}'.format(e.message))
+          # if 'external_id' in e.message:
+          #   raise Conflict('External ID already in use, error: {}'.format(e.message))
           logging.warning('Failed insert with %s: %s', tried_ids, e.message)
         else:
           raise
--- a/rest-api/dao/base_dao.py
+++ b/rest-api/dao/base_dao.py
@@ -340,7 +340,8 @@ class BaseDao(object):
     # pylint: disable=unused-argument
     # SQLite and MySQL variants of the error message, respectively.
     if 'UNIQUE constraint failed' in e.message or 'Duplicate entry' in e.message:
-      return logging.warning('Failed insert with %s: %s', tried_ids, e.message)
+      logging.warning('Failed insert with %s: %s', tried_ids, e.message)
+      return None
 
   def count(self):
     with self.session() as session:
--- a/rest-api/dao/base_dao.py
+++ b/rest-api/dao/base_dao.py
@@ -327,16 +327,21 @@ class BaseDao(object):
       try:
         with self.session() as session:
           return self.insert_with_session(session, obj)
-      except IntegrityError, e:
-        # SQLite and MySQL variants of the error message, respectively.
-        if 'UNIQUE constraint failed' in e.message or 'Duplicate entry' in e.message:
-          logging.warning('Failed insert with %s: %s', tried_ids, e.message)
+      except IntegrityError as e:
+        errors = self.handle_integrity_error(tried_ids, e, obj)
+        if errors:
+          return errors
         else:
-          raise
+          continue
     # We were unable to insert a participant (unlucky). Throw an error.
     logging.warning(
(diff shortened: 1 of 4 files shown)


rest-api/dao/base_dao.py
```diff
@@ -331,8 +330,6 @@ def _insert_with_random_id(self, obj, fields):
       except IntegrityError, e:
         # SQLite and MySQL variants of the error message, respectively.
         if 'UNIQUE constraint failed' in e.message or 'Duplicate entry' in e.message:
-          if 'external_id' in e.message:
```
danqrodney: So I would retain this logic -- this conflict can happen if two requests with the same external ID happen around the same time. But instead of raising a Conflict, I'd have the client pass in a function that can be called to determine what to do / return... and have participant_dao pass in something that does the lookup by external ID.

Seeker1911: done. (If I understood you correctly 😄 ).

rest-api/dao/base_dao.py
```diff
@@ -328,19 +327,23 @@ def _insert_with_random_id(self, obj, fields):
...
         with self.session() as session:
           return self.insert_with_session(session, obj)
-      except IntegrityError, e:
-        # SQLite and MySQL variants of the error message, respectively.
-        if 'UNIQUE constraint failed' in e.message or 'Duplicate entry' in e.message:
-          if 'external_id' in e.message:
-            raise Conflict('External ID already in use, error: {}'.format(e.message))
-          logging.warning('Failed insert with %s: %s', tried_ids, e.message)
+      except IntegrityError as e:
+        errors = self.handle_integrity_error(tried_ids, e, obj)
```
danqrodney: I wouldn't call this errors... maybe result?

rest-api/dao/participant_dao.py
```diff
@@ -339,6 +348,13 @@ def add_missing_hpo_from_site(self, session, participant_id, site_id):
     # Update the version and add history row
     self._do_update(session, participant, participant)
 
+  def handle_integrity_error(self, tried_ids, e, obj):
+    if 'external_id' in e.message:
+      existing_participant = self._check_if_external_id_exists(obj)
+      if existing_participant:
+        return existing_participant
+    return logging.warning('Failed insert with %s: %s', tried_ids, e.message)
```
danqrodney: This should be:

return super(self, ParticipantDao).handle_integrity_error(tried_ids, e, obj)

rest-api/dao/base_dao.py
Seeker1911: done

rest-api/dao/participant_dao.py
Seeker1911: ha, so I had that at first and thought I'd created some sort of circular reference or something, totally forgetting about the max_retries. Anyway, done.

rest-api/dao/base_dao.py
```diff
@@ -328,19 +327,21 @@ def _insert_with_random_id(self, obj, fields):
...
     # We were unable to insert a participant (unlucky). Throw an error.
     logging.warning(
         'Giving up after %d insert attempts, tried %s.' % (MAX_INSERT_ATTEMPTS, all_tried_ids))
     raise ServiceUnavailable('Giving up after %d insert attempts.' % MAX_INSERT_ATTEMPTS)
 
+  def handle_integrity_error(self, tried_ids, e, obj):
+    # pylint: disable=unused-argument
+    # SQLite and MySQL variants of the error message, respectively.
+    if 'UNIQUE constraint failed' in e.message or 'Duplicate entry' in e.message:
+      return logging.warning('Failed insert with %s: %s', tried_ids, e.message)
```
danqrodney: I don't think logging.warning returns anything, so technically this should work... but I'd just separate this out and say logging.warning(...) followed by return None to be sure.

rest-api/dao/participant_dao.py
```diff
@@ -72,6 +73,13 @@ def insert(self, obj):
     assert not obj.biobankId
     return self._insert_with_random_id(obj, ('participantId', 'biobankId'))
 
+  def _check_if_external_id_exists(self, obj):
+    with self.session() as session:
+      participant = session.query(Participant).filter_by(externalId=obj.externalId).first()
```
danqrodney: You can just return this. You don't need the if below.

rest-api/dao/base_dao.py
Seeker1911: Done.

rest-api/dao/participant_dao.py
Seeker1911: done

PR #834, opened 2018-11-07
DA-794 fix pm status bug

Seeker1911:
add method to check for existing non-cancelled physical measurements  and only set PM.status to cancelled if it is the only measurement.

### DIFF
--- a/rest-api/dao/physical_measurements_dao.py
+++ b/rest-api/dao/physical_measurements_dao.py
@@ -298,9 +298,9 @@ class PhysicalMeasurementsDao(UpdatableDao):
     participant_summary.lastModified = clock.CLOCK.now()
 
     if obj.status and obj.status == PhysicalMeasurementsStatus.CANCELLED:
-      if not self.existing_pm(session, participant):
+      if not self.has_uncancelled_pm(session, participant):
         # update corresponding participant summary physicalMeasurementsStatus
-        participant_summary.physicalMeasurementsStatus = obj.status
+        participant_summary.physicalMeasurementsStatus = PhysicalMeasurementsStatus.CANCELLED
 
     if participant_summary.physicalMeasurementsStatus != PhysicalMeasurementsStatus.CANCELLED:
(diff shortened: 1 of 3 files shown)


rest-api/dao/physical_measurements_dao.py
```diff
@@ -298,8 +298,9 @@ def _update_participant_summary(self, session, obj):
     participant_summary.lastModified = clock.CLOCK.now()
 
     if obj.status and obj.status == PhysicalMeasurementsStatus.CANCELLED:
-      # update corresponding participant summary physicalMeasurementsStatus
-      participant_summary.physicalMeasurementsStatus = obj.status
+      if not self.existing_pm(session, participant):
```
calbach: Shouldn't this condition be inverted? I don't follow

Seeker1911: no. its "if you do not have an existing pm that is valid, set this to cancelled"

calbach: Got it now - maybe reword as `self.has_uncancelled_pm()`

rest-api/dao/physical_measurements_dao.py
```diff
@@ -298,8 +298,9 @@ def _update_participant_summary(self, session, obj):
...
+        # update corresponding participant summary physicalMeasurementsStatus
+        participant_summary.physicalMeasurementsStatus = obj.status
```
calbach: opt: more characters, but might be slightly easier to understand if you replaced `obj.status` here with `PhysicalMeasurementsStatus.CANCELLED`

rest-api/dao/physical_measurements_dao.py
Seeker1911: done

rest-api/dao/physical_measurements_dao.py
Seeker1911: done

PR #847, opened 2018-11-21
Create tool for syncing all consents.

Seeker1911:


Seeker1911:
sorry for the confusing looking PR, this started as a cron job. But it is better ran manually as a tool.

robabram:
I realize this is a offline utility but is there be a concern about not using parameterized queries here? It looks like someone could create a sql injection attack by manipulating the source data.

rest-api/tools/ehr_upload.py
```diff
@@ -1,97 +1,134 @@
...
-        if system_call.returncode == 0:
-          print "Successfully moved folder " + google_group + '/P' + str(participant)
-        else:
-          print "There was an error moving folder " + google_group + '/P' + str(participant)
-          print "return code is : " + str(system_call.returncode)
-    else:
-      print "No participants paired with sites found for organization: " + organization
-  finally:
-    cursor.close()
+# @TODO: This spreadsheet is confidential, cant put it in version control. Need something better
```
calbach: Why not use a  flag for this ?

rest-api/tools/ehr_upload.py
```diff
@@ -1,97 +1,134 @@
...
+  hpo_data = read_hpo_report(csv_reader)
+  logging.info('Reading data complete, beginning sync...')
+  for org, data in hpo_data.items()[:2]:
+    site_paired_sql, no_paired_sql = get_sql(org)
+    logging.info('syncing participants for {}'.format(org))
+    run_sql(data['bucket'], site_paired_sql, no_paired_sql)
+
+
+def run_gsutil(gsutil):
+  gsutil = shlex.split(str(gsutil))
```
calbach: I think I've made this comment elsewhere, but surely it would be less error prone to pass an array of tokens here rather than encoding the whole thing as a string and splitting it later.

calbach: Yes. Create an array of the commandline tokens directly and skip this intermediate step (split just returns an array of tokens), e.g.


```
["gsutil", "cp", "gs://my_path", "dest"]
```

Seeker1911: shlex.split is the commonly accepted Pythonic way of passing commands to Popen().

rest-api/tools/ehr_upload.py
```diff
@@ -1,97 +1,132 @@
...
-   """.format(organization)
-  cursor = session.execute(text(sql))
-  try:
+     """.format(organization)
+
+  no_site_pairing_sql = """
+      select p.participant_id from participant p
+      left join participant_summary summary on p.participant_id = summary.participant_id
+      where p.site_id is NULL and p.organization_id in (
+      select organization_id from organization where external_id = '{}')
```
calbach: Per @robabram 's comment, I agree it would be better practice to use a SQL parameter here rather than string building. Understood that the value comes from a flag, but it looks like a safe and low effort fix

rest-api/tools/ehr_upload.py
calbach: I'm willing to be wrong, but I'm going to need a really good citation to believe that assertion.

In cases where you are starting with an arbitrary input string - sure; that's not what's going on here though. You've built all of the input parameters in code.

(this isn't blocking the review, mostly a style comment)

Seeker1911: https://docs.python.org/2.7/library/shlex.html#shlex.split
https://www.kevinlondon.com/2015/07/26/dangerous-python-functions.html

calbach: I don't see any direct relevance to this situation in either of those articles. The Popen docs mention shlex but mainly as a note that it can help to properly tokenize inputs - the examples at the bottom all use direct arrays: https://docs.python.org/2/library/subprocess.html#replacing-older-functions-with-the-subprocess-module

Anyways, fine to leave as-is, just seems less error prone to me to write (simplified, but equivalent by my understanding):

```
x = ["a", "b", "c"]
```

rather than

```
x_str = "a b c"
x = x_str.split(" ")
```

PR #862, opened 2018-12-12
[DA-631] Update last modified data in summary when enrollement status changes

robabram:
Created new unit test to check lastModified is updated when the participant enrollment status changes.

### DIFF
--- a/rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
+++ b/rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
@@ -2,11 +2,12 @@ import datetime
 import json
 from base64 import urlsafe_b64encode, urlsafe_b64decode
 import time
+import clock
 import config
 from dao.base_dao import json_serial
 from dao.biobank_stored_sample_dao import BiobankStoredSampleDao
 from dao.participant_dao import ParticipantDao
-from dao.participant_summary_dao import ParticipantSummaryDao
+from dao.participant_summary_dao import ParticipantSummaryDao, _ENROLLMENT_STATUS_SQL
 from model.biobank_stored_sample import BiobankStoredSample
 from model.participant import Participant
 from model.participant_summary import ParticipantSummary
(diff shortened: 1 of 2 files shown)


rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
```diff
@@ -375,6 +375,56 @@ def testUpdateEnrollmentStatus(self):
...
+      participantId=1,
+      biobankId=2,
+      consentForStudyEnrollment=QuestionnaireStatus.SUBMITTED,
+      consentForElectronicHealthRecords=QuestionnaireStatus.SUBMITTED,
+      consentForElectronicHealthRecordsTime=datetime.datetime(2018, 3, 1),
+      enrollmentStatus=EnrollmentStatus.INTERESTED)
+
+    self.dao.update_enrollment_status(tmp_summary)
+    self.assertEquals(EnrollmentStatus.MEMBER, tmp_summary.enrollmentStatus)
+    self.assertNotEqual(last_modified, tmp_summary.lastModified)
```
Seeker1911: Why do we need to test that two different participants have different last_modified timestamps ?

robabram: Its testing two different code paths that can change the enrollment status that are independent of each other.

robabram: Good catch, I wasn't quite getting what you were saying.  I've changed the tmp_summary to use the last_modified value to insure that the datetime assert is valid.  I'm committing the change.

Seeker1911: I guess what's confusing me here is the comparison between 2 participants. We could use FakeClock to set an exact time for last_modified and test that it changes to the new time.
[https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/test/unit_test/api_test/participant_summary_api_test.py#L252](url)

robabram: The code path update_from_biobank_stored_samples() updates the data store, so it needs a real participant record for the test to work.  The second code path update_enrollment_status() only updates the summary object's properties.  The code is really about making sure that the summary enrollment changes for the two code paths.  I have committed a new change that only uses one participant for both code path tests.  Let me know if that is what you are looking for.

rest-api/test/unit_test/dao_test/participant_summary_dao_test.py
```diff
@@ -375,6 +375,51 @@ def testUpdateEnrollmentStatus(self):
...
+      summary.lastModified = test_dt
+      summary.consentForStudyEnrollment = QuestionnaireStatus.SUBMITTED
+      summary.consentForElectronicHealthRecords = QuestionnaireStatus.SUBMITTED
+      summary.physicalMeasurementsStatus = PhysicalMeasurementsStatus.COMPLETED
+      summary.samplesToIsolateDNA = SampleStatus.RECEIVED
+      self.dao.update(summary)
+
+    ## Test Step 1: Validate update_from_biobank_stored_samples() changes lastModified.
+    reset_summary()
+    summary = self.dao.get(participant.participantId)
```
Seeker1911: Don't need this line if your getting it on line 404

PR #866, opened 2018-12-14
DA-822 refactor syncthetic data generator to accept a csv of existing participants

Seeker1911:
In order to help PTSC backfill a bunch (90k) test participants with biospecimen orders and physical measurements. We want to avoid the random percentages of withdrawn participants and random null measurements or biospecimen orders here.

### DIFF
--- a/rdr_client/generate_fake_data.py
+++ b/rdr_client/generate_fake_data.py
@@ -63,11 +63,8 @@ def _read_csv_lines(filepath):
 def generate_data_from_file(client, args):
   reader = _read_csv_lines(args.create_samples_from_file)
   request_body = [reader]
-  try:
-    print('requesting pm&b for participant')
-    client.request_json('DataGen', 'PUT', request_body)
-  except:
-    pass
+  logging.info('requesting pm&b for participant')
+  client.request_json('DataGen', 'PUT', request_body)
 
 
 
--- a/rdr_client/generate_fake_data.py
+++ b/rdr_client/generate_fake_data.py
@@ -88,7 +88,8 @@ if __name__ == '__main__':
                       help='True if biobank samples should be created')
   parser.add_argument('--create_samples_from_file',
                       help='Creates PM&B for existing participants from a csv file; requires path'
-                           ' to file')
+                           ' to file. File is expected to contain a single column of ID"s with a '
+                           'leading env. identifier. i.e. P')
   rdr_client = Client(parser=parser)
   if rdr_client.args.num_participants == 0 and not rdr_client.args.create_biobank_samples and not\
     rdr_client.args.create_samples_from_file:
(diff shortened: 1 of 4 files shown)


Seeker1911:
can ignore Codacy

> Standard pseudo-random generators are not suitable for security/cryptographic purposes.

rdr_client/generate_fake_data.py
```diff
@@ -70,7 +86,13 @@ def generate_fake_data(client, args):
   parser.add_argument('--create_biobank_samples',
                       action='store_true',
                       help='True if biobank samples should be created')
+  parser.add_argument('--create_samples_from_file',
+                      help='Creates PM&B for existing participants from a csv file; requires path'
```
calbach: Please document the expected format of this CSV file

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -138,6 +127,12 @@ def __init__(self, client, use_local_files=None):
     self._setup_questionnaires()
     self._min_birth_date = self._now - datetime.timedelta(days=_MAX_PARTICIPANT_AGE * 365)
     self._max_days_for_birth_date = 365 * (_MAX_PARTICIPANT_AGE - _MIN_PARTICIPANT_AGE)
+    self._force_measurement = False
+
+    # 5% of participants withdraw from the study
```
calbach: nit: this probably shouldn't say 5% anymore, since that's just the default value now

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -709,6 +704,26 @@ def generate_participant(self, include_physical_measurements, include_biobank_or
         last_request_time = max(last_request_time, last_hpo_change_time)
       self._submit_status_changes(participant_id, last_request_time)
 
+  def add_pm_and_biospecimens_to_participants(self, participant_id_list):
+    self._force_measurement = True
```
calbach: nit: Using persistent state here doesn't seem ideal as it could affect subsequent method calls. Would be better to just accept another optional param to the relevant methods.

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -872,7 +887,8 @@ def _make_answer_map(self, california_hpo):
     return answer_map
 
   def _submit_questionnaire_responses(self, participant_id, california_hpo, start_time):
-    if random.random() <= _NO_QUESTIONNAIRES_SUBMITTED:
+    ignore_failure = self._force_measurement
```
calbach: Why ignore failures here? Comment?

rest-api/data_gen/fake_participant_generator.py
```diff
@@ -883,37 +899,46 @@ def _submit_questionnaire_responses(self, participant_id, california_hpo, start_
...
-        method='POST',
-        body=qr_json,
-        pretend_date=submission_time)
+    try:
+      self._client.request_json(
+          _questionnaire_response_url(participant_id),
+          method='POST',
+          body=qr_json,
+          pretend_date=submission_time)
+    except RuntimeError:
```
calbach: Seems like this should probably log, even if you're not raising it

PR #867, opened 2018-12-18
[DA-623] Insure questionnaire response answer link ids are valid for questionnaire.

robabram:
_Note: This change closes a loophole that was used by many unit tests, it is unknown if Health Pro or PTSC are knowingly or inadvertently using this loophole._ 

Before this change extra/invalid answers added to a questionnaire response were silently ignored.  With this change, every questionnaire response must use either a full set or a sub set of answer link ids found in the questionnaire. No extra or invalid answer link ids are allowed.

Many of the unit tests took advantage of this loophole to use mixed questionnaire and response json data to test for specific answers within the response.  Closing this loophole required refactoring many unit tests and creating new questionnaire and response json data to test with.

### DIFF
--- a/rest-api/dao/questionnaire_response_dao.py
+++ b/rest-api/dao/questionnaire_response_dao.py
@@ -91,28 +91,24 @@ class QuestionnaireResponseDao(BaseDao):
     :param resource: A group section of the response json.
     :param link_ids: List of link ids to validate against.
     """
-    # look for child group sections and call ourselves
+    # note: resource can be either a dict or a list.
+    # if this is a dict and 'group' is found, always call ourselves.
     if 'group' in resource:
       self._validate_link_ids_from_resource_json_group(resource['group'], link_ids)
 
-    if type(resource) is list:
+    if 'question' not in resource and type(resource) is list:
       for item in resource:
         self._validate_link_ids_from_resource_json_group(item, link_ids)
 
-    # look for question section:
+    # once we have a question section, iterate through list of answers.
     if 'question' in resource:
+      for section in resource['question']:
 
-      # if question is a list, then loop through the list
-      if type(resource['question']) is list:
-        self._validate_link_ids_from_resource_json_group(resource['question'], link_ids)
-
-      # if question is not a list, check linkId directly
-      for question in resource['question']:
-        # Do not raise exception when link id is 'ignoreThis' for unit testing.
-        if 'linkId' in question and question['linkId'].lower() != 'ignorethis' \
-              and question['linkId'] not in link_ids:
+        # Do not raise exception when link id is 'ignoreThis' for unit tests.
+        if 'linkId' in section and section['linkId'].lower() != 'ignorethis' \
+              and section['linkId'] not in link_ids:
           raise BadRequest(
-            'Questionnaire response contains invalid link ID %s.' % question['linkId'])
+            'Questionnaire response contains invalid link ID %s.' % resource['linkId'])
 
   def insert_with_session(self, session, questionnaire_response):
 
--- a/rest-api/dao/questionnaire_response_dao.py
+++ b/rest-api/dao/questionnaire_response_dao.py
@@ -86,7 +86,8 @@ class QuestionnaireResponseDao(BaseDao):
   def _validate_link_ids_from_resource_json_group(self, resource, link_ids):
     """
     Look for question sections and validate the linkid in each answer. If there is a response
-    answer link id that does not exist in the questionnaire, then raise an exception.
+    answer link id that does not exist in the questionnaire, then log a message. In
+    the future this may be changed to raising an exception.
     This is a recursive function because answer groups can be nested.
     :param resource: A group section of the response json.
     :param link_ids: List of link ids to validate against.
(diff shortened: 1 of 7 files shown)


rest-api/dao/questionnaire_response_dao.py
```diff
@@ -83,15 +83,51 @@ def _validate_model(self, session, obj):  # pylint: disable=unused-argument
...
+        self._validate_link_ids_from_resource_json_group(item, link_ids)
+
+    # look for question section:
+    if 'question' in resource:
+
+      # if question is a list, then loop through the list
+      if type(resource['question']) is list:
+        self._validate_link_ids_from_resource_json_group(resource['question'], link_ids)
+
+      # if question is not a list, check linkId directly
```
calbach: Not following here, this claims question is not a list (but we didn't return in the above if statement L107, so this may not be true).

Then we iterate through the question anyways; are we assuming it's a dict instead? Should probably make this check explicit.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -83,15 +83,51 @@ def _validate_model(self, session, obj):  # pylint: disable=unused-argument
...
+      self._validate_link_ids_from_resource_json_group(resource['group'], link_ids)
+
+    if type(resource) is list:
+      for item in resource:
+        self._validate_link_ids_from_resource_json_group(item, link_ids)
+
+    # look for question section:
+    if 'question' in resource:
+
+      # if question is a list, then loop through the list
```
calbach: nit: I like the documentation in this PR, the codebase could definitely use more. IMO some of these comments (like this one) could probably be removed though. In my experience inline comments tend to be useful when they target the "why" or the high level flow, but become less useful (or harmful) as they approach a restatement of the code (those generally lead to bloat and things going out of sync). I think this one is in the latter category.

rest-api/test/unit_test/dao_test/questionnaire_response_dao_test.py
```diff
@@ -656,8 +656,17 @@ def _get_questionnaire_response_with_consents(self, *consent_paths):
         self.LN_QUESTION.linkId,
         self.EMAIL_QUESTION.linkId,
         consent_paths)
+
+    # We need to remove the unused answers in the resource so we don't trigger an unused
+    # link id exception.
+    res_len = len(resource['group']['question']) - 1
+    for idx in range(res_len, -1, -1):
+      if resource['group']['question'][idx]['linkId'].isdigit() is True:
```
calbach: nit: drop the `is True`?

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -106,8 +142,16 @@ def insert_with_session(self, session, questionnaire_response):
...
 
+    # Gather the question ids and records that match the questions in the response
     question_ids = [answer.questionId for answer in questionnaire_response.answers]
     questions = QuestionnaireQuestionDao().get_all_with_session(session, question_ids)
+
+    # DA-623: raise error when response link ids do not match our question link ids.
+    # Gather the valid link ids for this question
+    link_ids = [question.linkId for question in questions]
+    # look through the response and verify each link id is valid for each question.
+    self._validate_link_ids_from_resource_json_group(resource_json, link_ids)
```
calbach: How confident are we that this won't immediately start firing for all incoming questionnaire's? Or would incoming data in this format currently cause major data issues?

I'm not sure off-hand whether we might expect to find other values in here. If there's a risk here, consider instead pushing this initially with just a try/catch/log to monitor.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -83,15 +83,51 @@ def _validate_model(self, session, obj):  # pylint: disable=unused-argument
...
+    if 'question' in resource:
+
+      # if question is a list, then loop through the list
+      if type(resource['question']) is list:
+        self._validate_link_ids_from_resource_json_group(resource['question'], link_ids)
+
+      # if question is not a list, check linkId directly
+      for question in resource['question']:
+        # Do not raise exception when link id is 'ignoreThis' for unit testing.
+        if 'linkId' in question and question['linkId'].lower() != 'ignorethis' \
```
calbach: optional: test-only sentinel values in prod code should definitely be avoided, though understood that changing this might be a massive refactoring task we don't want to take on currently. Consider whether there's another straight forward way to achieve this.

wangy70:
As you said "it is unknown if Health Pro or PTSC are knowingly or inadvertently using this loophole", I think we need to synch with them about this change to make sure it will not break anything.

rest-api/dao/questionnaire_response_dao.py
robabram: I ran tests with recent questionnaires/response records in stable and all passed this validation check. This certainly restricts behavior that was taken advantage of in the unit tests. We are checking with PTSC to get their feedback before merging.  It may be the final decision to log for a period of time before actually enforcing the validation.

wangy70: cool 👍

rest-api/dao/questionnaire_response_dao.py
robabram: That one check fixed a lot of unit test breakage. The primary issue here is the unit test json data files for questionnaire/responses look (to me) like they were from early iterations of the module.  They have little resemblance to the format of the recent questionnaire/response json data in stable or production.  Refactoring all those data files and the unit tests would be a good sized task, probably suitable for a separate ticket.

rest-api/test/unit_test/dao_test/questionnaire_response_dao_test.py
robabram: In python database applications I like to be explicit when test boolean values since its possible that a boolean can be (None, True or False) depending on the underling data.  I get carried away with that check sometimes.

robabram:
On Tuesday, the HealthPro team indicated that this change will not affect them.  We need to get PTSC's feedback before merging.  That might have to wait until after the holidays.

wangy70:
Since you already test with recent questionnaires/response records, I think it's enough to move forward,  it's the best way to verify that.

rest-api/dao/questionnaire_response_dao.py
calbach: Sorry - I still don't follow. The comment says "question is not a list", you then proceed to iterate through the question.

The comment also implies exclusivity with line 107 which isn't true because you don't return at L107.

PR #871, opened 2018-12-31
DA-601 - Ensure biobank id exists before importing sample

robabram:
Biobank samples import pipeline could handle bad biobank IDs more gracefully


rest-api/offline/biobank_samples_pipeline.py
```diff
@@ -145,15 +147,25 @@ def _upsert_samples_from_csv(csv_reader):
...
-          written += samples_dao.upsert_all(samples)
-          samples = []
-    if samples:
-      written += samples_dao.upsert_all(samples)
+    with ParticipantDao().session() as session:
+
+      for row in csv_reader:
+        sample = _create_sample_from_row(row, biobank_id_prefix)
+        # DA-601 - Ensure biobank_id exists before accepting a sample record.
+        if session.query(Participant).filter(Participant.biobankId == sample.biobankId).count() < 1:
```
calbach: Arguably it might be nice if these showed up in the missing report, but this is definitely an improvement over crashing the pipeline.

PR #874, opened 2019-01-03
Testing changes to data generator

Seeker1911:


rdr_client/generate_fake_data.py
```diff
@@ -56,15 +56,14 @@ def generate_fake_data(client, args):
 def _read_csv_lines(filepath):
   with open(filepath, 'r') as f:
     reader = csv.reader(f)
-    reader.next()
-    return [line[0].strip() for line in reader]
+    return [line[0].strip() for line in reader][:10]
```
calbach: Not sure if you're planning to merge this PR as-is, but this line should be reverted before you do.

PR #876, opened 2019-01-04
[DA-814] Implement Sample Disposal Status from nightly report

robabram:


rest-api/participant_enums.py
```diff
@@ -56,6 +56,53 @@ class SampleStatus(messages.Enum):
...
+  """
+  Return the SampleStatus enum value for the given status from Mayo
+  :param status: a sample status value
+  :return: SampleStatus enum value
+  """
+  if status is None:
+    return SampleStatus.UNSET
+
+  # Set Received if we have an empty status value, the sample has not been disposed of yet.
+  if not status:
```
Seeker1911: The only time I could think this line would be True and the one above is not, is if status == False. Which it shouldn't be right ?

robabram: `if not status` will also return true if status is an empty string.  In the biobank nightly file, the disposed status is empty if the sample is still active.

PR #896, opened 2019-01-22
[DA-862] migrate m1api to m2pai for enrollment status

wangy70:
migrate m1api to m2pai for enrollment status


rest-api/dao/metrics_cache_dao.py
```diff
@@ -0,0 +1,106 @@
...
+        .order_by(MetricsEnrollmentStatusCache.hpoId).all()
+
+  def delete_old_records(self, n_days_ago=7):
+    with self.session() as session:
+      last_inserted_record = self.get_serving_version_with_session(session)
+      if last_inserted_record is not None:
+        last_date_inserted = last_inserted_record.dateInserted
+        seven_days_ago = last_date_inserted - datetime.timedelta(days=n_days_ago)
+        delete_sql = """
+          delete from metrics_enrollment_status_cache where date_inserted < :seven_days_ago
```
Seeker1911: Should this be greater than?

rest-api/dao/metrics_cache_dao.py
```diff
@@ -0,0 +1,106 @@
...
+        .order_by(MetricsGenderCache.hpoId).all()
+
+  def delete_old_records(self, n_days_ago=7):
+    with self.session() as session:
+      last_inserted_record = self.get_serving_version_with_session(session)
+      if last_inserted_record is not None:
+        last_date_inserted = last_inserted_record.dateInserted
+        seven_days_ago = last_date_inserted - datetime.timedelta(days=n_days_ago)
+        delete_sql = """
+          delete from metrics_gender_cache where date_inserted < :seven_days_ago
```
Seeker1911: Same as above?

rest-api/dao/metrics_cache_dao.py
wangy70: No, it's less than. This SQL is used to delete the old records before seven days timestamp to keep the table small.

Seeker1911: Ok it just reads confusing. “Less than seven days ago “ sounds earlier than.

PR #899, opened 2019-01-25
DA-843 Require prefix for biobank ID.

Seeker1911:


Seeker1911:
Whitespace fixes and (the pertinent bit here) require an environment prefix when passing in biobank id parameter in participant summary.

rest-api/dao/participant_summary_dao.py
```diff
@@ -315,6 +315,9 @@ def _add_order_by(self, query, order_by, field_names, fields):
 
   def make_query_filter(self, field_name, value):
     """Handle HPO and code values when parsing filter values."""
+    from model.config_utils import from_client_biobank_id
```
calbach: Is this a datastore dependency? Is that why you aren't importing normally? If so, please just add a comment here.

rest-api/dao/participant_summary_dao.py
```diff
@@ -315,6 +315,9 @@ def _add_order_by(self, query, order_by, field_names, fields):
...
+    if field_name == 'biobankId':
+      value = from_client_biobank_id(value)
```
calbach: This is a breaking change - are you confident that no one was using this capability up until now? If you're not sure - a safer thing to do here would be to try/catch this and just log a warning if you see it (continue with the previous behavior if you see an exception).

rest-api/dao/participant_summary_dao.py
Seeker1911: Nope, I was experimenting and forgot to move it. It's moved now 😄 
Also, shouldn't our linters catch that ?

rest-api/dao/participant_summary_dao.py
Seeker1911: I believe it's not in use, HPRO pointed out the error and I've asked PTSC if they are passing in BiobankId anywhere. That being said, I think best practice would be to "deprecate it" for a while and make it a hard error in the near future. I'll make the todo for that and switch it. Thanks.

PR #901, opened 2019-01-25
[DA-865] Update stable participants

calbach:


calbach:
FYI I also put some new bash into the playbook to make this process a bit smoother, please try it out and let me know if it works for you next time you do this process.

PR #905, opened 2019-01-29
[DA-875] 1. migrate m1api to m2api for total; 2. support filter by awardee; 3.…

wangy70:
1. migrate m1api to m2api for total; 
2. support filter by awardee; 
3. change day range limit to 600 for history=TRUE

### DIFF
--- a/rest-api/api/participant_counts_over_time_api.py
+++ b/rest-api/api/participant_counts_over_time_api.py
@@ -13,7 +13,8 @@ from participant_enums import EnrollmentStatus
 from participant_enums import Stratifications
 
 DATE_FORMAT = '%Y-%m-%d'
-DAYS_LIMIT = 100  # provisional, per design doc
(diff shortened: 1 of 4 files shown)


rest-api/api/participant_counts_over_time_api.py
```diff
@@ -89,9 +90,13 @@ def validate_params(self, start_date_str, end_date_str, stratification_str,
     except ValueError:
       raise BadRequest('Invalid end date: %s' % end_date_str)
     date_diff = abs((end_date - start_date).days)
-    if date_diff > DAYS_LIMIT:
-      raise BadRequest('Difference between start date and end date ' \
-                       'should not be greater than %s days' % DAYS_LIMIT)
+    if history != 'TRUE' and date_diff > DAYS_LIMIT_FOR_REALTIME_DATA:
```
Seeker1911: are we assuming true will always be upper case ?

PR #912, opened 2019-01-31
[DA-870] Wangy70/migrate m1api to m2api for race

wangy70:
migrate m1api to m2api for race; 
tune sql performance;


rest-api/dao/metrics_cache_dao.py
```diff
@@ -311,50 +357,287 @@ def to_client_json(self, result_set):
...
+                     calendar,
+                     hpo
+                WHERE p.hpo_id = hpo.hpo_id
+                  AND calendar.day >= :start_date
+                  AND calendar.day <= :end_date
+                  AND calendar.day >= Date(p.sign_up_time)
+              ) y
+              GROUP BY day, hpo_id, name
+              ;
+        """.format(self.race_code_dict['Race_WhatRaceEthnicity'],
```
Seeker1911: This is not sql injection safe, can you parameterize it like we do in other places ?

wangy70: Done

PR #919, opened 2019-02-05
[DA-859] Encrypt API sync tokens

robabram:
Requires adding the 'pycrypto' python library which has a 'Public Domain' license.

Jira ticket for approving new python package license [PD-3101](https://precisionmedicineinitiative.atlassian.net/browse/PD-3101)

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -21,6 +21,19 @@ activate_local_venv
 
 git submodule update --init
 
+GCLOUD=`which gcloud`
+GCP_BIN=`dirname $GCLOUD`
+SDK_ROOT=`dirname $GCP_BIN`
+echo "Cloud SDK Root: $SDK_ROOT"
+$GCLOUD components install app-engine-python app-engine-python-extras cloud-datastore-emulator
+
+GCP_SDK_ROOT=$SDK_ROOT/platform/google_appengine
+GCP_SDK_YAML=`find $GCP_SDK_ROOT -type d | grep "yaml-.*[0-9]$"`
+echo "Cloud SDK Yaml: $GCP_SDK_YAML"
+
+export sdk_dir=$GCP_SDK_ROOT
+export PYTHONPATH="$PYTHONPATH:$GCP_SDK_ROOT:$GCP_SDK_ROOT/lib:$GCP_SDK_YAML"
+
 safety check  # checks current (API) venv
 
 # Make sure JSON files are well-formed (but don't bother printing them).
--- a/rest-api/dao/base_dao.py
+++ b/rest-api/dao/base_dao.py
@@ -1,9 +1,11 @@
-from base64 import urlsafe_b64decode, urlsafe_b64encode
 import collections
 import json
 import logging
 import datetime
 import random
+from Crypto.Cipher import AES
+import config
+import urllib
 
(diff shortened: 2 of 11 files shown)


rest-api/tools/install_config.py
```diff
@@ -56,9 +58,15 @@ def main(args):
...
     comparable_file = _json_to_sorted_string(combined_config)
+    # Don't compare 'aes_key' values because it is always changed when updating.
+    comparable_file = re.sub(r'\n.*?"aes_key".*?\n', '\n', comparable_file)
+    formatted_server_config = re.sub(r'\n.*?"aes_key".*?\n', '\n', formatted_server_config)
     configs_match = _compare_configs(comparable_file, formatted_server_config, args.config_output)
 
     if not configs_match and args.update:
+      # replace 'aes_key' with a new random 16-digit string.
+      new_key = ''.join(random.choice(string.ascii_letters + string.digits) for _ in range(16))
+      combined_config['aes_key'] = new_key
```
calbach: Isn't it going to be an issue when pagination suddenly breaks on every server release?

robabram: I think I have an idea on how to change the key every 3 or 6 months without putting the key in the git project.

rest-api/dao/base_dao.py
```diff
@@ -268,9 +275,12 @@ def _add_pagination_filter(self, query, query_def, fields, first_descending):
     return query.filter(or_(*or_clauses))
 
   def _decode_token(self, query_def, fields):
-    pagination_token = query_def.pagination_token
+    pagination_token = urllib.unquote(query_def.pagination_token.encode("ascii"))
```
calbach: To avoid breaking existing users, this should ideally continue to accept both old-style and new-style tokens initially, with a TODO to remove old-style token handling in the next release.

ci/test.sh
```diff
@@ -21,6 +21,19 @@ activate_local_venv
...
 git submodule update --init
 
+GCLOUD=`which gcloud`
+GCP_BIN=`dirname $GCLOUD`
+SDK_ROOT=`dirname $GCP_BIN`
+echo "Cloud SDK Root: $SDK_ROOT"
+$GCLOUD components install app-engine-python app-engine-python-extras cloud-datastore-emulator
+
+GCP_SDK_ROOT=$SDK_ROOT/platform/google_appengine
+GCP_SDK_YAML=`find $GCP_SDK_ROOT -type d | grep "yaml-.*[0-9]$"`
```
calbach: I'm lost, what is the Cloud SDK yaml and why are we adding it to the path?

robabram: Blame CircleCI and Google.  Importing config caused an ndb import which then caused a yaml import. For whatever reason the Google SDK couldn't import the yaml library in the SDK without me adding it to the path.

rest-api/dao/base_dao.py
robabram: If an old style token comes in, I can decode it and give them an encrypted token in the reply.  That way nothing breaks and they are moved over to an encrypted token.  Will that work?

calbach: Yep, exactly

PR #920, opened 2019-02-05
[DA-870] fix race calculate algorithm

wangy70:
fix race calculate algorithm

### DIFF
--- a/rest-api/dao/metrics_cache_dao.py
+++ b/rest-api/dao/metrics_cache_dao.py
@@ -241,18 +241,34 @@ class MetricsGenderCacheDao(BaseDao):
     return client_json
 
   def get_metrics_cache_sql(self):
+
+    gender_code_dict = {
+      'GenderIdentity_Woman': 354,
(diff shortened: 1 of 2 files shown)


rest-api/dao/metrics_cache_dao.py
```diff
@@ -242,15 +243,16 @@ def to_client_json(self, result_set):
   def get_metrics_cache_sql(self):
     sql = """insert into metrics_gender_cache """
     gender_names = ['UNSET', 'Woman', 'Man', 'Transgender', 'PMI_Skip', 'Non-Binary',
-                    'Other/Additional Options']
+                    'Other/Additional Options', 'Prefer not to say']
     gender_conditions = [
-      ' (ps.gender_identity_id IS NULL OR ps.gender_identity_id=924) ',
```
Seeker1911: I'm not sure that we can expect id's to never change.

wangy70: done

PR #924, opened 2019-02-07
Drop primary key before dropping column in primary key

robabram:
Dropping a column used in a primary key is not allowed in newer versions on MySql. This is a change in behavior from 5.7.  Dropping the primary key first allows the drop column to succeed.


Seeker1911:
what do you mean by change in behavior from 5.7 ? This is 5.7

robabram:
5.7 lets you drop a column in a primary key with out stopping you.  Newer versions throw an error when dropping a column in a primary key.  This is just future proofing the migration.

rennat:
This looks correct to me. The previous almebic migration created a primary key including the field `race_name` and this PR resets the primary key to match the `metrics_cache_dao` definition

PR #925, opened 2019-02-08
[DA-858] cron job to sync site buckets consent files

rennat:
## TODOs:

- ~~Currently pulls source sheet identifier from environment, ideal solution is to put value in config (load secure config from bucket when installing config)~~
- end to end testing (still getting set up for this)


rest-api/offline/sync_consent_files.py
```diff
@@ -0,0 +1,151 @@
...
+      yield ParticipantData(*row)
+
+
+def cloudstorage_copy_objects(source, destination):
+  """Copies all objects matching the source to the destination
+
+  Both source and destination use the following format: /bucket/prefix/
+  """
+  for file_stat in cloudstorage.listbucket(source):
+    cloudstorage.copy2(
```
Seeker1911: My only concern is we’ll always copy all files. At the moment it’s about 800,000 I believe.

rest-api/offline/sync_consent_files.py
```diff
@@ -0,0 +1,151 @@
...
+    org_data.aggregate_id,
+    (
+      org_data.bucket_name
+      if org_data.bucket_name
+      else org_data_map[org_data.aggregate_id].bucket_name
+    )
+  )
+
+
+PARTICIPANT_DATA_SQL = sqlalchemy.text("""
```
Seeker1911: We can add a check for is_ghost_id from participant table (it’s brand new). And can also check for the test email pattern (@example.com) sorry I didn’t have that to begin with.

rest-api/offline/sync_consent_files.py
rennat: The newer commit now gets and checks the destination file's etag (md5 hash) to determine if it should copy.
This will save some bandwidth for already existing files but this still has 1 extra api call to get the `GCSFileStat` for each potential destination bucket path.

rest-api/offline/sync_consent_files.py
rennat: This is applied in a newer commit. The `not like` `WHERE` clause will likely have an impact on this participant query speed.

PR #928, opened 2019-02-11
[DA-889] Whitelist IPs in STABLE for Coalfire

rennat:
Need to confirm this is the correct service account.


rennat:
@Seeker1911 confirmed this is the correct Service Account

PR #929, opened 2019-02-11
[DA-873] participant counts over time EHR Consent date filtering issue

rennat:
Add initial tests to highlight the issue.
I'm committing these failing tests only at first for discussion.


rennat:
This is ready for review now.

The Codacy check does not like the way the SQL query was put together. I modified the query itself but not the way we built it.

We can do a deeper refactor to do away with python string formatting in the SQL query building but I'd suggest this larger refactor belongs in a separate pull request.

wangy70:
There is no SQL injection risk for this one, just ignore the Codacy error.

PR #932, opened 2019-02-12
[DA-668] ETL scripts: add cli flags to set paths, removed hardcoded paths

robabram:
Added '--local' flag to mysqlimport statements to fix error accessing local files.

### DIFF
--- a/rest-api/etl/setup_database_for_etl.sh
+++ b/rest-api/etl/setup_database_for_etl.sh
@@ -6,7 +6,7 @@
 # If the databases already exist, drop them prior to running this.
 # Note: this takes a while. Go get some coffee while it's running!
 
-USAGE="tools/setup_database_for_etl.sh --project <PROJECT> --account <ACCOUNT> --src-folder <FOLDER> "
+USAGE="setup_database_for_etl.sh --project <PROJECT> --account <ACCOUNT> --src-folder <FOLDER> "
 while true; do
   case "$1" in
     --account) ACCOUNT=$2; shift 2;;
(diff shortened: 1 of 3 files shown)


robabram:
I hadn't used them until this week. I saw a ticket pending to remove the hard coded paths, so I updated them.

rest-api/etl/setup_database_for_etl.sh
```diff
@@ -6,29 +6,40 @@
 # If the databases already exist, drop them prior to running this.
 # Note: this takes a while. Go get some coffee while it's running!
 
-USAGE="tools/setup_database_for_etl.sh --project <PROJECT> --account <ACCOUNT>"
```
Seeker1911: All tools assume they're being ran from rest-api/ 
For instance, this one sources `tools/auth_setup.sh` on L43 which will fail.
I'm all for the shortest distance being a straight line, but not sure we should change established patterns, this one would require quite a bit of work to establish across the board. I recommend staying with the pattern on this.

rest-api/etl/setup_local_database_for_etl.sh
```diff
@@ -23,27 +23,48 @@ DB_CONNECTION_NAME=
 CSV_DIR=/var/lib/mysql-files/rdr-csv
 OUTPUT_DIR=/tmp/rdr-sql-dump
 
-USAGE="tools/setup_local_database_for_etl.sh --account <ACCOUNT> [--generate_sql_dump] [--db_user <ROOT_DB_USER>] [--nopassword]"
```
Seeker1911: same as above.

rest-api/etl/setup_database_for_etl.sh
robabram: No problem.  I've updated the usages to show the correct 'etl' directory.

PR #935, opened 2019-02-14
[DA-912] Add capability for config mixins from GCS

rennat:
updates `tools/install_config.sh` to include config mixins from GCS.

- Syncs a hard-coded google cloud storage bucket's contents to a local folder that is ignored by git
- Includes any `.json` files from that folder when building the config

This allows some values to be included in the config on a deployment while not being present in the git repository for security.

### DIFF
--- a/rest-api/tools/install_config.py
+++ b/rest-api/tools/install_config.py
@@ -5,15 +5,20 @@ See "Config Updates" in the README for usage and permissions details.
 
 import copy
 import difflib
+import errno
 import httplib
 import json
 import logging
+import os
 import re
+import subprocess
 
 from client import Client, HttpException
 from main_util import get_parser, configure_logging
 
 BASE_CONFIG_FILE = 'config/base_config.json'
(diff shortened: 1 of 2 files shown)


rest-api/tools/install_config.py
```diff
@@ -5,15 +5,20 @@
...
 import logging
+import os
 import re
+import subprocess
 
 from client import Client, HttpException
 from main_util import get_parser, configure_logging
 
 BASE_CONFIG_FILE = 'config/base_config.json'
+SEQUESTERED_CONFIG_GCS_PATH = 'gs://all-of-us-rdr-sequestered-config-test/'
```
rennat: test bucket name is currently hard-coded here

rennat:
This would be better to do with `cloudstorage.open` calls but I was unable to get the `cloudstorage` script working in a local script. Is there a trick to this or is `cloudstorage` reserved for use from within appengine?

robabram:
The 'offline/metrics_pipeline.py' is using the cloud storage python lib.  Maybe looking there will help you.

PR #936, opened 2019-02-14
[DA-859] Refactor participant summary token sync

robabram:
Added an API parameter to disable 60 second backfill buffer.
Added new API 'ParticipantSummary/Modified' to allow grabbing a full table snapshot of what records have changed and what records are new.


Seeker1911:
I thought the agreed upon new process would be to only do the 60 sec. buffer if the link returned a type `sync` and not for a type `next`.

robabram:
This makes the 60 second backfill optional.  The default behavior is maintained if the '_backfill' parameter is never passed. '_backfill' always defaults to true if not passed.

PR #940, opened 2019-02-18
[DA-863] Cron job to skew last modified duplicates

robabram:
When there are two many duplicate last modified values in the participant summary, the cron job will skew the microseconds of the last modified values to eliminate duplicates. 

This changes the last_modified fields on the participant and participant_summary tables to store microseconds in the field. Although only duplicates on the participant summary will be skewed, adding microseconds to the participant table was necessary because the last modified timestamp is copied over to the participant summary record when the participant summary record is created.

Cron job has been tested in sandbox.


robabram:
The issue has been when there are duplicate last modified timestamps in the participant summary table.  No one has mentioned an issue with the participant table so I did not check for duplicates in it.  

But I did need to add microseconds to the participant table because last_modified value is copied over to the participant summary when the initial participant summary record is created.

wangy70:
Thanks for the explanation! The codes are all good.  I just saw "participant" is in your comment, so I thought it's in the same PR. 
Do you want to add it in this PR or some others later?

robabram:
I have updated the PR description to reflect that only the participant summary records are skewed when there are too many duplicate last modified dates and also that adding microseconds to the participant table was necessary.

PR #948, opened 2019-02-26
DA-940 add new test participants for stable

Seeker1911:


Seeker1911:
ok so what started out as a simple import test participants into stable ticket became something else. the package isort (which is a sub-dependency of another package we use) was updated on 25 February to version 4.3.9 This breaks our build. To fix it I explicitly added isort==4.3.8 as a dependency in the circle ci config so it wouldn't reinstall it at a higher version when it installed the sub-dependencies.

PR #953, opened 2019-03-01
DA-943 Create script to upload ehr consent for a single awardee

Seeker1911:
A ticket was raised 
[DA-842](https://precisionmedicineinitiative.atlassian.net/browse/DA-842) addressing the need to resync VA ehr consent files into aou179, partly because there's a 30 day lifecycle on google bucket files (and they forgot to get them) and partly because we stopped the cron job to sync all consents across the board. We removed this ability (syncing a single awardee) so I'm putting it back in.


Seeker1911:
Codacy can be ignored here

PR #954, opened 2019-03-01
DA-879 add disctinct count for pm&b visits

Seeker1911:
add number of distinct visits, distinct being defined as not have a physical measurement or biobank sample order on the same day in the participants time zone when giving a new physical measurement or biobank order. Originally this converted the time zone id of paired site, I ended up going with naively converting the UTC time of the most recent pm or bio created time to +5 (central time zone) because it's easier and more readable and still meets the requirements. There are a few random tests outside of participant summary because it's faster than adding new tests under biobank and measurement tests to look for something on a summary.


Seeker1911:
Also, I wanted a test like
`with faketime
       pm.order
       bio.order
make sure that's only 1 visit`
but the sql query uses CURDATE() which makes that hard, we might want a follow up ticket to Mock that or something, any ideas would be appreciated.

rest-api/dao/participant_summary_dao.py
```diff
@@ -514,6 +514,43 @@ def calculate_max_core_sample_time(self, participant_summary, field_name_prefix=
     else:
       return None
 
+  def calculate_distinct_visits(self, session, pid):
+    """ Participants may get PM or biobank samples on same day. This should be considered as
+    a single visit in terms of program payment to participant.
+    return Boolean: true if there has not been an order on same date (site timezone)."""
+    sql = text("""select CURDATE() as realdate , pm.created, bo.created,
```
rennat: We should revisit this soon to take in a date as a parameter so that we can extend the tests.

Seeker1911: agreed, or mock the query ? I'll make a ticket.

PR #958, opened 2019-03-04
[PD-3242] add IP range for awardee-tach

wangy70:
add IP range for awardee-tach


rest-api/config/config_stable.json
```diff
@@ -63,9 +63,24 @@
...
+            "208.72.50.0/24",
+            "208.72.51.0/25",
+            "208.72.51.128/26",
+            "208.72.51.192/27",
+            "208.72.51.224/28",
+            "208.72.51.240/29",
+            "208.72.51.248/30",
+            "208.72.51.252/31",
+            "208.72.51.254/32"
+          ]
```
Seeker1911: pending a question I asked on the original ticket regarding removing old IP's, you could cover this range with 
'208.72.51.254/22' That would give you 208.72.48.0 - 208.72.51.255. Then just add the 216.70.11.254

Seeker1911: the "66.189.127/27" is not a valid IP address (which I admit, I'm probably the one who put that one there 🙁  )

wangy70: Those original two IP ranges is not removed, I just reformat it :) 
I will correct the invalid IP ranges.

wangy70: 1. The two old invalid IP ranges are fixed.
2. '208.72.51.254/22' covers much more IPs than they provided, I still prefer to use an accurate list in case there may be some security concerns. 

The code is submitted, please review.

PR #961, opened 2019-03-05
[DA-938] Added DV biobank order mailing address table

robabram:


robabram:
You can ignore the build failure because of the migration Id.  Just let me know if that structure looks good or needs changes.

robabram:
Also, do we want to link the State field to the Code table or change it to be a String field?

robabram:
Not going to use a separate mailing address table.

PR #962, opened 2019-03-06
DA-921 cron job to update EHR status from Curation data

rennat:
TODO: finish pulling from BigQuery

I've run into local environment issues connecting to BigQuery. I'm working around them.

Included here is testing for everything downstream from executing the query against BigQuery.


rennat:
This is now updated to scan the curation bucket for EHR update times directly, this negates the need for an additional query to BigQuery.

PR #964, opened 2019-03-07
[DA-952] Create new table for Salivary pilot to store biobank DV orders

robabram:
New model uses standard SqlAlchemy column types to ensure sqlite unit tests continue to work.  
Update 'created' and 'modified' to auto populate in migration.
Auto create a history table for the new table. 
Auto add triggers to auto track table changes into history table.


Seeker1911:
are we doing anything about Codacy ?

robabram:
Codacy is complaining about a possible SQL injection, it can be ignored.  The code creates and removes the automatic history table for the given table name and is only used in migrations.

PR #965, opened 2019-03-08
Add missing alembic reversable handler for Stored Procedures

robabram:
Add separate alembic handler for CREATE PROCEDURE.
Clean up migration code.
Add missing unittest_schemes functions for creating Functions/Procedures in unittests.


robabram:
The Alembic user needs EXECUTE permission because all Stored Procedures and Functions are created by the Alembic user.  In the MySQL security schema we have in GCP any functions/procs are executed using the MySQL user who created it.  Because the Alembic user does not have the EXECUTE permission, we see this error when we call a function using the RDR user.

[42000][1370] execute command denied to user 'alembic'@'%' for routine 'rdr.fn_get_code_id_from_key'

From the MySQL Docs:
_The SQL SECURITY characteristic can be DEFINER or INVOKER to specify the security context; that is, whether the routine executes using the privileges of the account named in the routine DEFINER clause or the user who invokes it. This account must have permission to access the database with which the routine is associated. The default value is DEFINER. **The user who invokes the routine must have the EXECUTE privilege for it, as must the DEFINER account if the routine executes in definer security context**._ 

Since we already have received permission to give the RDR user EXECUTE permission, I don't believe we need to seek additional permission to give it to the Alembic user.

PR #967, opened 2019-03-11
[DA-925] Capture and use questionnaire authored timestamp 

robabram:
Adds a new 'authored' column to the Questionnaire Response table.
Captures the 'authored' timestamp from questionnaires, uses 'created' value if no 'authored' value is present.
Timestamps in participant summary will now use the 'authored' timestamp instead of the 'created' timestamp.
Updated fake data generator to randomly set authored time to 0 seconds 270 in the past.  1 in 10 chance of being identical as the 'created' timestamp.


rest-api/alembic/versions/648d99c5c414_add_authored_column.py
```diff
@@ -0,0 +1,59 @@
...
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('questionnaire_response', sa.Column('authored', model.utils.UTCDateTime(), nullable=True))
+    # ### end Alembic commands ###
+    op.execute('ALTER TABLE questionnaire_response CHANGE COLUMN `authored` `authored` DATETIME NULL AFTER created')
```
wangy70: Why we need to change it here?

robabram: The change is setting the new field position so it is after the 'created' column instead of at the end of the table. That way it doesn't have to be after the big 'resource' column when you select from the table.

robabram:
I have also updated the 'sp_get_questionnaire_answers' stored procedure to include the 'authored' timestamp.

This migration is interesting because it replaces the existing stored procedure using the alembic ReplacableObject methodology, outlined here: https://alembic.sqlalchemy.org/en/latest/cookbook.html#replaceable-objects.

When you rollback the migration (IE: "alembic downgrade -1", the original version of the stored procedure is put back into the db. The link might be useful reading for everyone interested.

robabram:
Changes to jira ticket include capturing language from surveys and not switching to using 'authored' timestamp to update participant summary columns.

rest-api/dao/questionnaire_response_dao.py
```diff
@@ -353,9 +355,21 @@ def from_client_json(self, resource_json, participant_id=None, client_id=None):
...
                        questionnaire.questionnaireId)
+    authored = None
+    if fhir_qr.authored and fhir_qr.authored.date:
+      authored = fhir_qr.authored.date
+
+    language = None
+    if fhir_qr.extension:
+      for ext in fhir_qr.extension:
+        if 'iso21090-ST-language' in ext.url:
+          language = ext.valueCode[:2]
```
wangy70: When I work on other tickets, I accidentally found that there are some values of `primary_language` like below:

`select distinct primary_language from participant_summary`
```
en
es
"en"
"es"
```
I doubt the double quotes are from the xml extension, but I didn't dig more yet, just FYI. I think before we know what happened, maybe we need to take care of this to make sure it will not cause any exception.

rest-api/model/questionnaire_response.py
```diff
@@ -16,6 +16,8 @@ class QuestionnaireResponse(Base):
   participantId = Column('participant_id', Integer, ForeignKey('participant.participant_id'),
                          nullable=False)
   created = Column('created', UTCDateTime, nullable=False)
+  authored = Column('authored', UTCDateTime, nullable=True)
+  language = Column('language', String(2), nullable=True)
```
wangy70: Same comments as above.

rest-api/dao/questionnaire_response_dao.py
robabram: In this case the FHIR object strips the double quotes and only the language code is stored.  I was dealing this the other day, where another field ended up wrapped with double quotes. I believe the quotes get captured when code is accessing data directly from a JSON string.  I saw this issue in a unittest recently.

wangy70: Nice

PR #969, opened 2019-03-12
[DA-932] public metrics export api for new ppi

wangy70:
1. add new metrics questionnaireOnHealthcareAccess, questionnaireOnMedicalHistory, questionnaireOnMedications, questionnaireOnFamilyHealth  for PPI 
2. improve the realtime enrollment status metric SQL
3. update readme, remove EHR_CONSENT and EHR_RATIO stratification, because these are moved to our new EHR API
4. refactor participant_counts_over_time_api
5. add logic to  metrics_cache_dao to support new PUBLIC METRIC API requirements


wangy70:
The "Possible SQL injection" can be ignored, because that variable was checked and can only be the value of the constant enum.

rest-api/dao/metrics_cache_dao.py
```diff
@@ -985,6 +985,49 @@ def to_client_json(self, result_set):
...
 
+  def to_public_metrics_client_json(self, result_set):
+    client_json = []
+    for record in result_set:
+      new_item = {
+        'date': record.date.isoformat(),
+        'metrics': {
+          'completed': {
+            'Registered': record.registered,
+            'Consent_Enrollment': record.consentEnrollment,
```
robabram: Is there a reason for returning a new column name instead of the model column name here? It seems like a standard that APIs return the model column name.

wangy70: Yes, I agree with you, that make things easy for us. I think the current situation is that all the JSON data are returned with a "_" format name, this may inherited from the old APIs. And now all the APIs are using by the dashboard frontend, it's not easy to change them unless we make two versions of them and ask healthPro to update to the new one.

wangy70:
Sorry, just pushed a uncommitted change by accident, working on the rollback.

rest-api/dao/metrics_cache_dao.py
robabram: Ok, I guess its a design decision we are stuck with.

wangy70:
Revert is done, no code is changed.

PR #971, opened 2019-03-14
DA-921 fix missed reference to earlier argument style

rennat:


rennat:
This adds testing for hpo filtering after @wangy70 found an issue.

PR #972, opened 2019-03-14
[DA-933] create endpoints for public metrics export api

wangy70:
public metrics API
create new endpoints

### DIFF
--- a/rest-api/api/participant_counts_over_time_api.py
+++ b/rest-api/api/participant_counts_over_time_api.py
@@ -56,9 +56,9 @@ class ParticipantCountsOverTimeApi(Resource):
       # Validate dates
       if not params['end_date']:
         raise BadRequest('end date should not be empty')
-      start_date = datetime.datetime.strptime('2017-01-01', DATE_FORMAT).date()
       try:
         end_date = datetime.datetime.strptime(params['end_date'], DATE_FORMAT).date()
+        start_date = end_date
       except ValueError:
         raise BadRequest('Invalid end date: %s' % params['end_date'])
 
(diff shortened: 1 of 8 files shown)


rest-api/api/participant_counts_over_time_api.py
```diff
@@ -56,9 +56,9 @@ def validate_params(self, params):
       # Validate dates
       if not params['end_date']:
         raise BadRequest('end date should not be empty')
-      start_date = datetime.datetime.strptime('2017-01-01', DATE_FORMAT).date()
       try:
         end_date = datetime.datetime.strptime(params['end_date'], DATE_FORMAT).date()
+        start_date = end_date
```
rennat: Does this cause problems if the start and end date are the same?

wangy70: The is for querying the data by a cutoff date instead of a date range. Technically, in this case the start_date should never be used, I assign a value for it just in case.

rest-api/test/unit_test/api_test/public_metrics_api_test.py
```diff
@@ -0,0 +1,923 @@
...
+    dao = MetricsEnrollmentStatusCacheDao(MetricsCacheType.PUBLIC_METRICS_EXPORT_API)
+    service.refresh_data_for_metrics_cache(dao)
+
+    qs = """
+          &stratification=ENROLLMENT_STATUS
+          &startDate=2018-01-01
+          &endDate=2018-01-08
+          """
+
+    qs = ''.join(qs.split())  # Remove all whitespace
```
rennat: There is nothing wrong with this way of cleaning whitespace out of a string. There is a more common pythonic way though:
```
qs = (
  '&stratification=ENROLLMENT_STATUS'
  '&startDate=2018-01-01'
  '&endDate=2018-01-08'
)
```

Python automatically concatenates consecutive string literals.

wangy70: Cool, thanks for the tip! I will change it to this way.

PR #973, opened 2019-03-15
[DA-955] Extend Biobank Samples Pipeline to read status updates of salivary orders

robabram:
- Created optimized SQL statement to update participant summary DV sample status fields. Add a TODO to optimize the existing SQL in the future.
- Create index on biobank_stored_sample.test field.  Should speed up pipeline a bit.
- Change 'created' and 'modified' columns to be Null so SQLAlchemy doesn't force values for them when inserting new records. 
- Fix up 'created' column so it is in the right column position and auto sets the value.


rest-api/dao/participant_summary_dao.py
```diff
@@ -403,12 +403,53 @@ def update_from_biobank_stored_samples(self, participant_id=None):
 
     sample_sql = replace_null_safe_equals(sample_sql)
     counts_sql = replace_null_safe_equals(counts_sql)
+
+
     with self.session() as session:
       session.execute(sample_sql, sample_params)
       session.execute(counts_sql, counts_params)
       session.execute(enrollment_status_sql, enrollment_status_params)
+      # TODO: Change this to the optimized sql in _update_dv_stored_samples()
```
rennat: is this a todo for another pull request later or a lingering comment?

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_test.py
```diff
@@ -33,11 +40,96 @@ def setUp(self):
...
+        ('identifiers', [BiobankOrderIdentifier(system=u'a', value=u'c')]),
+        ('samples', [BiobankOrderedSample(
+            test=u'1SAL2',
+            description=u'description',
+            processingRequired=True)]),
+        ('dvOrders', [BiobankDVOrder(
+          participantId=participantId
+        )])):
+      if k not in kwargs:
+        kwargs[k] = default_value
```
rennat: alternative dict merging option for better readability:
```
defaults = {'biobankOrderId': u'1', ...}
kwargs = dict(defaults, **kwargs)
```

rest-api/dao/participant_summary_dao.py
robabram: There is no ticket yet, but its on a lot of minds that the biobank nightly processing job is painfully inefficient and occasionally fails because of GCP timing something out. A large part of that stems from non-optimized SQL statements.

rest-api/test/unit_test/offline_test/biobank_samples_pipeline_test.py
robabram: I copied that code method block from another unittest. Your readability optimization would be a lot better, but I think we should wait until the bit Python 3 re-write to get serious about making code be more pythonic.

PR #978, opened 2019-03-18
DA-951 add MayoLink api

Seeker1911:


Seeker1911:
Tests still need to be added, and other tickets, but this should be a good base to work off of from here. Happy to have a call if anyone wants to walk through it. Sorry it's so long, I didn't have time to make it short.

Seeker1911:
Codacy error is for parsing xml. Can be ignored, we are parsing the response from mayolink api. Hopefully we can trust them.

PR #986, opened 2019-03-21
use SimpleFhirR4Reader in DvOrderDao

rennat:
also extend the reader and reader tests

### DIFF
--- a/rest-api/dao/dv_order_dao.py
+++ b/rest-api/dao/dv_order_dao.py
@@ -5,6 +5,7 @@ from api_util import format_json_code, get_code_id, format_json_enum, parse_date
 from dao.base_dao import UpdatableDao
 from dao.code_dao import CodeDao
 from dao.participant_summary_dao import ParticipantSummaryDao
+from fhir_utils import SimpleFhirR4Reader
 from model.biobank_dv_order import BiobankDVOrder
 from werkzeug.exceptions import BadRequest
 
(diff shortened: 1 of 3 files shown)


rest-api/dao/dv_order_dao.py
```diff
@@ -120,11 +127,11 @@ def from_client_json(self, resource_json, id_=None, expected_version=None,
         order.id = self.get_id(order)[0]
         order.created = self._get_created_date(participant_id, id_)
         order.version = expected_version
-        order.barcode = resource['barcode']
+        order.barcode = fhir_resource.barcode  # NOTE: not in the FHIR spec
```
rennat: This line will behave the same as before this change, but `barcode` is not present in the example or in the FHIR spec for `SupplyRequest`

rest-api/dao/dv_order_dao.py
```diff
@@ -120,11 +127,11 @@ def from_client_json(self, resource_json, id_=None, expected_version=None,
...
         # @TODO: foreign key to biobank order.biobank order id. implement in DA-953
         # order.biobankOrderId = resource['biobank_order_id']
-        order.biobankStatus = resource['status']
-        order.biobankReceived = parse_date(resource['received'])
+        order.biobankStatus = fhir_resource.status
+        order.biobankReceived = parse_date(fhir_resource.received)  # NOTE: not in the FHIR spec
```
rennat: This line will behave the same as before this change, but `received` is not present in the example or in the FHIR spec for `SupplyRequest`

Seeker1911: Thanks for doing this, barcode is internal to our table structure, we infer it from another field, same with received I believe. This will be great for our next dry run. I can tune up things like barcode based on some new request data I have from our partners. This is a huge help. 👍

PR #987, opened 2019-03-25
integrate biobank order into direct volunteer salivary orders

Seeker1911:


rest-api/api/dv_order_api.py
```diff
@@ -14,12 +16,18 @@ def __init__(self):
...
-    if not p_id:
-      raise BadRequest('Request must include participant id and must be of type integer')
+    fhir_resource = SimpleFhirR4Reader(resource)
+    if resource['resourceType'] == 'SupplyRequest':
+      p_id = from_client_participant_id(fhir_resource.contained.get(resourceType='Patient').
+                                        identifier.get(
+                                  system='http://joinallofus.org/fhir/participantId').value)
+    if resource['resourceType'] == 'SupplyDelivery':
+      # @todo: do something with delivery
+      p_id = from_client_participant_id(resource['patient']['identifier'][0]['value'])
```
rennat: we'll need to address these direct index accesses (cannot guarantee index `0` is the correct one)

rest-api/api/dv_order_api.py
```diff
@@ -14,12 +16,18 @@ def __init__(self):
...
     response = super(DvOrderApi, self).post(participant_id=p_id)
-    order_id = resource['identifier'][0]['code']
+    order_id = resource['identifier'][0]['value']
```
rennat: same as above

PR #989, opened 2019-03-25
Add option to connect to replica SQL instances.

robabram:
Change Staging replica instance name to 'rdrbackupdb1'.
Code cleanup and organization.

This also adds a new command line parameter "--enable-replica" to enable connections to the replica SQL instances for each project.

image


robabram:
After having to switch between primary and replica connections multiple times on Friday, I thought it would be nice to be able to enable replica connections automatically in the sql proxy service.

PR #990, opened 2019-03-26
DA-980 BiobankDVOrder delivery update endpoint

rennat:


rennat:
@Seeker1911 This gets the basics in place for the update `PUT` operation. This is *not* adequately tested so far.

PR #996, opened 2019-03-28
ehr metrics: unify ambiguous term `site` as organizations

rennat:
In the original work specification for the EHR metrics API, the term `site` was interpreted in different ways by different stakeholders.
After discussions, we decided that these APIs should work only on `Organization`.


rennat:
@wangy70 This change will impact your work on the public metrics api. I will look at it.

wangy70:
> @wangy70 This change will impact your work on the public metrics api. I will look at it.

wangy70:
> @wangy70 This change will impact your work on the public metrics api. I will look at it.

Thanks, let's talk about it on Monday.

rest-api/alembic/versions/ac995f0b54b8_ehr_metrics_use_organizations.py
```diff
@@ -0,0 +1,66 @@
...
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('ehr_receipt', sa.Column('organization_id', sa.Integer(), nullable=False))
+    op.drop_constraint(u'ehr_receipt_ibfk_1', 'ehr_receipt', type_='foreignkey')
```
wangy70: Do you need the columns in a specified order? If yes, we may need to specify the order by using `AFTER`.

rennat: No the column order does not matter for this table.

PR #998, opened 2019-03-29
update dv orders based on logging errors in test

Seeker1911:


Seeker1911:
After speaking with PTSC devs and looking at logs in test I've made a few updates to the api.

rest-api/api/dv_order_api.py
```diff
@@ -133,8 +144,9 @@ def _put_supply_delivery(self, resource, bo_id):
     order.shipmentEstArrival = eta.date()
     order.shipmentStatus = tracking_status_enum
 
-    self._do_update(order)
-    return self._make_response(order)
+    # self._do_update(order)
+    # return self._make_response(order)
+    return order
```
rennat: Will this change skip writing the updates?

Seeker1911: no it is called in updateable api, but I do need to remove the comments

PR #1003, opened 2019-04-03
update deploy process with Cron YAML builder

rennat:
This change adds a cron.yaml builder script with inheritance and the ability to override/remove inherited cron jobs.

This is needed so we can have mostly shared cron definitions with a few exceptions.

---

## The old system
- if project name is `all-of-us-rdr-sandbox`
    - use `cron_sandbox.yaml`
- else
   - use `cron_default.yaml`

## This system
1. Build list of cron files to combine
    - start with `cron_default.yaml`
    - add `cron_<project_shortname>.yaml` if exists
2. Iterate each job from files in order
    - identify jobs by `description` (cannot be overridden or removed if missing `description`)
    - add new jobs
    - overwrite existing jobs
    - remove existing job if new job is missing either `url` or `schedule` (the required fields)
3. Serialize compiled job list as `cron.yaml`

### DIFF
--- a/rest-api/cron_default.yaml
+++ b/rest-api/cron_default.yaml
@@ -39,13 +39,3 @@ cron:
   schedule: every day 00:10
   timezone: America/New_York
   target: offline
-- description: Sync site bucket consent files
-  url: /offline/SyncConsentFiles
-  schedule: 1 of month 00:00
-  timezone: America/New_York
-  target: offline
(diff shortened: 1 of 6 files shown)


rennat:
The Codacy check fails because we use `yaml.load` instead of `yaml.safe_load`. In this case we are only loading files that we have defined so this is acceptable.

PR #1011, opened 2019-04-09
make latest a string

Seeker1911:


Seeker1911:
test doesn't like the fact that "latest" wasn't an explicit string, even though it works on dev_appserver and master and Googles on docs show it written as a constant. 🤷‍♂️

PR #1021, opened 2019-04-11
Tn/da 1024 genomic validation

rennat:


rest-api/model/genomics.py
```diff
@@ -40,6 +40,7 @@ class GenomicSet(Base):
   # have mysql always update the modified data when the record is changed
   modified = Column('modified', DateTime, nullable=True)
 
+  version = Column('version', Integer, nullable=False, default=1)
```
rennat: Before we can merge this pull request we need to determine if this is necessary. If so then I will add an alembic migration for it.

If I remove the `version` property I get:
```Traceback (most recent call last):
  File "/Users/jonathannetterville/vumc/raw-data-repository/rest-api/test/unit_test/genomic_test/genomic_validation_test.py", line 141, in test_withdrawn
    validate_and_update_genomic_set_by_id(genomic_set.id)
  File "/Users/jonathannetterville/vumc/raw-data-repository/rest-api/genomic/validation.py", line 42, in validate_and_update_genomic_set_by_id
    dao.member_dao.update(member)
  File "/Users/jonathannetterville/vumc/raw-data-repository/rest-api/dao/base_dao.py", line 438, in update
    return self.update_with_session(session, obj)
  File "/Users/jonathannetterville/vumc/raw-data-repository/rest-api/dao/base_dao.py", line 430, in update_with_session
    self._validate_update(session, obj, existing_obj)
  File "/Users/jonathannetterville/vumc/raw-data-repository/rest-api/dao/base_dao.py", line 413, in _validate_update
    if existing_obj.version != obj.version:
AttributeError: 'GenomicSetMember' object has no attribute 'version'
```

rest-api/genomic/validation.py
```diff
@@ -0,0 +1,81 @@
...
+  if row.sex_at_birth not in GENOMIC_VALID_SEX_AT_BIRTH_VALUES:
+    return GenomicValidationStatus.INVALID_SEX_AT_BIRTH
+  if row.birth_date > dob_cutoff:
+    return GenomicValidationStatus.INVALID_AGE
+  if not all(map(
+    functools.partial(operator.contains, GENOMIC_VALID_SAMPLE_STATUSES),
+    [row.sample_status_1ED04, row.sample_status_1SAL2]
+  )):
+    return GenomicValidationStatus.INVALID_BIOBANK_ORDER
+  if row.ny_flag and not row.zip_code:
```
rennat: @wangy70 is this the intended validation behavior?

wangy70: Just talked with Katie about the zip code validation, based on the way of the current requirement doc, we need to check every participant no matter ny_flag yes or no. And Katie will double check with them.

rest-api/model/genomics.py
wangy70: I'm not sure about this part, any thought? @robabram

rest-api/dao/genomics_dao.py
```diff
@@ -38,3 +114,33 @@ def upsert(session):
...
+    )
+    parameter_sets = [
+      {
+        'member_id': member_id,
+        'status': int(status),
+        'time': now if status == GenomicValidationStatus.VALID else None,
+      }
+      for member_id, status in member_id_status_pair_iterable
+    ]
+    session.execute(query, parameter_sets)
```
rennat: This may need to be re-worked to update the `version` column as well but I'm not sure I can make this error-out if the version number is not the expected version.
@robabram thoughts?

rest-api/model/genomics.py
rennat: I talked with @robabram about this and we agree that there is no concurrent write risk with the intended use case of these models so we can safely override `def _validate_update()` (in the same way that `ParticipantSummaryDao` does) to not require a `version` field.

rest-api/dao/genomics_dao.py
rennat: After talking with @robabram about the `version` field. We arrived at not needing it for this model which means we can safely have this bulk update.

rest-api/dao/base_dao.py
```diff
@@ -410,7 +417,7 @@ def _validate_update(self, session, obj, existing_obj):
     """
     if not existing_obj:
       raise NotFound('%s with id %s does not exist' % (self.model_type.__name__, id))
-    if existing_obj.version != obj.version:
+    if self.validate_version_match and existing_obj.version != obj.version:
```
rennat: @robabram @wangy70 this is the `UpdatableDao`'s `version` behavior change we have discussed so far.

PR #1023, opened 2019-04-15
fix for distinct visit counts bug

Seeker1911:


rest-api/dao/participant_summary_dao.py
```diff
@@ -555,42 +556,33 @@ def calculate_max_core_sample_time(self, participant_summary, field_name_prefix=
...
+      for measurement in existing_measurements:
+        if measurement.created.date() == created.date() and measurement.physicalMeasurementsId\
+          != id_:
+          day_has_measurement = True
+    is_distinct_visit = not (day_has_order or day_has_measurement)
+    return is_distinct_visit
+    # if day_has_order or day_has_measurement:
+    #   # not a distinct visit.
+    #   return False
+    # return True
```
rennat: let's remove these commented lines

rest-api/dao/biobank_order_dao.py
```diff
@@ -174,6 +174,13 @@ def get_with_children(self, obj_id):
     with self.session() as session:
       return self.get_with_children_in_session(session, obj_id)
 
+  def get_biobank_order_for_participant(self, pid):
```
rennat: can we rename this to the plural `..._orders_...`

Seeker1911: done

rest-api/dao/participant_summary_dao.py
Seeker1911: done

PR #1024, opened 2019-04-15
[DA-1023] Wangy70/genomic set input file handler

wangy70:
Implement genomic set input file handler:
1. read_genomic_set_from_bucket
2. create_genomic_set_status_result_file


wangy70:
All the `format` params are from our enum, the SQL injection warning can be ignored.

wangy70:
I changed it to query params to avoid SQL injection warnings. For this case, I can't use the alchemy methods, because the `SqlExporter` need a SQL string as a parameter.

PR #1030, opened 2019-04-19
Mm/add biobank address

Seeker1911:
adding logic for biobank address. Which comes in as a normal address on one of the put requests for supply delivery. So here I compare the order address to the known biobank address for the logic flow.


Seeker1911:
also changed alembic env file to pick up type changes in models.

rest-api/dao/dv_order_dao.py
```diff
@@ -24,6 +24,11 @@ class DvOrderDao(UpdatableDao):
   def __init__(self):
     self.code_dao = CodeDao()
     super(DvOrderDao, self).__init__(BiobankDVOrder)
+    self.biobank_address = {'city': "Rochester", 'state': "MN",
```
rennat: This looks like something used in testing that was left in.

robabram: I believe that is the real address of the Biobank where samples are shipped to.

Seeker1911: correct, it was intended to be left there.

PR #1033, opened 2019-04-22
[DA-1034] Added handling for unhandled exception errors.

robabram:
Removed value setting for modified and created fields, these are no longer necessary after latest updates to the table.


rest-api/api/dv_order_api.py
```diff
@@ -96,13 +113,19 @@ def put(self, bo_id):  # pylint: disable=unused-argument
...
+    # handle invalid FHIR documents
+    try:
+      fhir_resource = SimpleFhirR4Reader(resource)
+      barcode_url = None
+      if fhir_resource.extension.get(url=VIBRENT_FULFILLMENT_URL).valueString == 'shipped':
+        barcode_url = fhir_resource.extension.get(url=VIBRENT_BARCODE_URL).url
+      pid = fhir_resource.contained.get(
+        resourceType='Patient').identifier.get(system=VIBRENT_FHIR_URL + 'participantId')
+      p_id = from_client_participant_id(pid.value)
+    except AttributeError:
```
Seeker1911: Not sure we need this one, SimpleFhirR4Reader returns very explicit error messages that would be more helpful.

rest-api/api/dv_order_api.py
```diff
@@ -125,15 +148,21 @@ def _put_supply_request(self, resource, bo_id):
...
+      fhir = SimpleFhirR4Reader(resource)
+      participant_id = fhir.patient.identifier.value
+      p_id = from_client_participant_id(participant_id)
+      update_time = dateutil.parser.parse(fhir.occurrenceDateTime)
+      carrier_name = fhir.extension.get(url=VIBRENT_FHIR_URL + 'carrier').valueString
+      eta = dateutil.parser.parse(fhir.extension.get(
+        url=VIBRENT_FHIR_URL + "expected-delivery-date").valueDateTime)
+      tracking_status = fhir.extension.get(
+        url=VIBRENT_FHIR_URL + 'tracking-status').valueString
+    except AttributeError:
```
Seeker1911: same as above

rest-api/api/dv_order_api.py
robabram: I was testing with limited parameters and an un-handled attribute error was being thrown.  This is really to catch someone sending empty or nearly empty json.

PR #1038, opened 2019-04-24
DA-1080 Fix delivery bug and refactor supplydelivery request

Seeker1911:


rest-api/dao/dv_order_dao.py
```diff
@@ -196,6 +191,8 @@ def from_client_json(self, resource_json, id_=None, expected_version=None,
         url=VIBRENT_ORDER_URL).valueString
       if id_ is None:
         order.version = 1
+        order.modified = clock.CLOCK.now()
+        order.created = clock.CLOCK.now()
```
robabram: After my auto populate fix PR is merged, specifically setting 'created' and 'modified' should not be needed.

Seeker1911: removed

rest-api/dao/dv_order_dao.py
```diff
@@ -138,31 +137,26 @@ def from_client_json(self, resource_json, id_=None, expected_version=None,
...
+        existing_obj.streetAddress1 = order_address.line[0]
+        existing_obj.zipCode = order_address.postalCode
+
+        if len(order_address._obj['line'][0]) > 1:
+          try:
+            existing_obj.streetAddress2 = order_address._obj['line'][1]
+          except IndexError:
+            pass
+
+      elif order_address['use'] == 'work':
```
rennat: on line 140 this is referenced as `order_address._obj['use']`. The result should be the same either way but maybe we should query this value out of the fhir object once and then use it in comparisons.

Seeker1911:
ok, for real this time, not sure why I didn't test changes (I assume).

PR #1041, opened 2019-04-24
[DA-1017] add enrollment status stratification to public metric api

wangy70:
add enrollment status stratification to public metric api


wangy70:
No SQL injection risk, the warning can be ignored.

PR #1045, opened 2019-04-26
[DA-983] Modify fake participant generator

robabram:
Because the ticket has so many specifications to meet, I created a new participant generator program to allow creating spec test participants.

This PR does the following.
- Breaks the old participant generator code into multiple generators that can be used independently.
- Creates a new client cli program to execute multiple tools 'rest-api/client_cli/rdr.py', see README.md.
- Added a tool (verify) that can verify that the local environment is good for calling APIs in the GCP projects.
- Added a tool (random-data) that replicates the old participant client tool to verify that the new tool chain works correctly.
- Added a tool (spec-data) which can take a spreadsheet and create specialized test participants.
- Pulls in multiple 'fhirclient' libraries into 'rest-api/lib_fhir'.
- Finished 'services/gcp_utils.py' so every bash script that calls 'gcloud' or 'gsutil' can be replaced with a pure python program.
- Adds the 'requests' and 'requests-toolbelt' libraries to enable using Requests as an option.


robabram:
The codacy review can be ignored.  The issues are mainly from the re-hashing of the participant generator code and won't be uploaded to app engine.

robabram:
Split into 3 new PRs, closing this one.

PR #1054, opened 2019-04-30
Add multiple fhirclient library versions to project

robabram:


robabram:
Codacy issues are `assert` calls in the fhirclient library and also complaining about the bash lines gathering files for linting.

Seeker1911:
I've removed `assert` condition from Codacy, the issue that it would cause doesn't apply to RDR anyway.

Seeker1911:
would it possible to add the fhir client's as git sub-modules ?

robabram:
I hadn't thought about sub-modules, but it should be fine.

robabram:
Its not possible to add Git Tag branches as sub-modules.  This is the best way to add multiple versions of the fhirclient library.

PR #1058, opened 2019-05-06
[DA-944] refactored primary etl script

robabram:
Changed back from View to a INSERT INTO SELECT, because filtering took way too long in that statement.  Instead, I added multiple steps after creating the src_clean table to mark which records should be filtered.  This keeps the execution speed of the script about the same as before filtering.


Seeker1911:
could Codacy be right here ? I don't see anything wrong.

robabram:
I don't think Codacy is using the correct SQL dialect to analyze the file.  The line "ALTER TABLE cdm.src_clean ADD KEY (participant_id);" has been there since before I made any changes to the ETL script.

PR #1059, opened 2019-05-06
[DA-1111] Update IP addresses for TACH

robabram:


rest-api/config/config_prod.json
```diff
@@ -80,8 +80,10 @@
...
         "ip4": [
-          "66.189.127/27",
-          "68.184.42.162/27"
+          "161.249.83.116",
+          "161.249.147.132",
+          "161.249.241.64/26",
+          "161.249.150.128/26"
         ]
       },
       "description": "Using subnet mask these ranges cover 66.189.127.0-32 and 68.184.42.162.160-191"
```
Seeker1911: please update descriptions. Also, you may want to double check with Celecia that these IP's would apply to prod and stable. Usually different ranges are used for each env. This may only be a prod change (assuming from the urgency of the ticket)

rest-api/config/config_stable.json
```diff
@@ -60,24 +60,10 @@
...
-            "208.72.49.0/24",
-            "208.72.50.0/24",
-            "208.72.51.0/24"
+            "161.249.83.116",
+            "161.249.147.132",
+            "161.249.241.64/26",
+            "161.249.150.128/26"
           ]
         },
         "description": "Using subnet mask these ranges cover 66.189.127.0-32, 68.184.42.162.160-191, 216.70.11.254 and 208.72.48.248 – 208.72.51.254"
```
Seeker1911: description

PR #1074, opened 2019-05-16
DA-1097 refactor number of distinct visit counting 

Seeker1911:
covering more cases. i.e. amending orders, cancelling, etc. Added specific test case examples from the PRD.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -47,7 +47,7 @@ load-plugins=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-#disable=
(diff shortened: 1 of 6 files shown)


pylintrc
```diff
@@ -47,7 +47,7 @@ load-plugins=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-#disable=
+disable=C0111
```
Seeker1911: oh yeah, also got tired of the "missing docstring" message in my linter. If anyone is unhappy with this decision I"m happy to remove this (and keep it in my global pylintrc).

PR #1075, opened 2019-05-17
DA-964 refactor ppi participant views for qa table exporter

Seeker1911:
added filtering of `is_ghost_id is not true`

### DIFF
--- a/rest-api/alembic/versions/71614b7c6482_filter_ghosts_from_views.py
+++ b/rest-api/alembic/versions/71614b7c6482_filter_ghosts_from_views.py
@@ -1,0 +1,505 @@
+"""filter ghosts from views
+
+Revision ID: 71614b7c6482
+Revises: 3c91f6fd0585
+Create Date: 2019-05-17 09:15:12.549159
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import model.utils
+
+from dao.alembic_utils import ReplaceableObject
+
+from participant_enums import PhysicalMeasurementsStatus, QuestionnaireStatus, OrderStatus
+from participant_enums import WithdrawalStatus, WithdrawalReason, SuspensionStatus, QuestionnaireDefinitionStatus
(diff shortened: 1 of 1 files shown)


rest-api/alembic/versions/71614b7c6482_filter_ghosts_from_views.py
```diff
@@ -0,0 +1,505 @@
...
+     LEFT OUTER JOIN code state_code ON ps.state_id = state_code.code_id
+     LEFT OUTER JOIN code recontact_method_code ON ps.recontact_method_id = recontact_method_code.code_id
+     LEFT OUTER JOIN code language_code ON ps.language_id = language_code.code_id
+     LEFT OUTER JOIN code gender_code ON ps.gender_identity_id = gender_code.code_id
+     LEFT OUTER JOIN code sex_code ON ps.sex_id = sex_code.code_id
+     LEFT OUTER JOIN code sexual_orientation_code ON ps.sexual_orientation_id = sexual_orientation_code.code_id
+     LEFT OUTER JOIN code education_code ON ps.education_id = education_code.code_id
+     LEFT OUTER JOIN code income_code ON ps.income_id = income_code.code_id
+     WHERE p.withdrawal_status = 1 AND # NOT_WITHDRAWN
+           (ps.email IS NULL OR ps.email NOT LIKE '%@example.com') AND
```
wangy70: Does comment allowed in """ """ ?

rest-api/alembic/versions/71614b7c6482_filter_ghosts_from_views.py
```diff
@@ -0,0 +1,505 @@
...
```
wangy70: same as above

rest-api/alembic/versions/71614b7c6482_filter_ghosts_from_views.py
Seeker1911: yes if it's raw sql it's fine and distinguishes between any `'` and `"` quotes that may be in the query.

PR #1080, opened 2019-05-22
[DA-1131] add columns to genomic_set_member_history table

wangy70:
add columns to genomic_set_member_history table


rennat:
This migration looks good

PR #1085, opened 2019-05-24
[DA-1177] genomic manifest file second modification

wangy70:
genomic manifest file second modification:
1. leave 'value' column blank, biobank will fill it
2. add prefix 'T' for biobank_id
3. ingest client id back to db


rest-api/genomic/genomic_biobank_menifest_handler.py
```diff
@@ -24,6 +24,8 @@
 # sample suffix: -v12019-04-05-00-30-10.csv
 _RESULT_CSV_FILE_SUFFIX_LENGTH = 26
 
+BIOBANK_ID_PREFIX = 'T'
```
rennat: Is this a universal constant or does it need to be put into the application config?

wangy70: Not sure about this, I got this request for the genomic manifest file.

wangy70: I can talk with Michael next week to double check, if yes, I can submit a new change for it.

rennat: if this will need to be a different prefix on different environments (similar to participant prefix) then we will move it to an application config parameter. Until it becomes an issue, this is fine.

PR #1087, opened 2019-05-29
[PD-2165] add ip to whitelist for service account

wangy70:
add ip to whitelist for service account

### DIFF
--- a/rest-api/config/config_prod.json
+++ b/rest-api/config/config_prod.json
@@ -124,6 +124,19 @@
       "whitelisted_ip_ranges" : {
         "ip4": ["160.129.251.78"]
       }
+    },
+    "awardee-newyork@all-of-us-ops-data-api-prod.iam.gserviceaccount.com": {
+      "roles": ["awardee_sa"],
+      "awardee": "newyork",
+      "whitelisted_ip_ranges" : {
+        "ip4": [
+          "157.139.237.172",
+          "157.139.9.22",
+          "140.251.21.110",
+          "157.139.9.228",
+          "157.139.21.56"
+        ]
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


rest-api/config/config_prod.json
```diff
@@ -124,6 +124,19 @@
       "whitelisted_ip_ranges" : {
         "ip4": ["160.129.251.78"]
       }
+    },
+    "awardee-newyork@all-of-us-ops-data-api-prod.iam.gserviceaccount.com": {
+      "roles": ["awardee_sa"],
+      "awardee": "newyork",
```
Seeker1911: I believe this should be columbia.

rest-api/config/config_stable.json
```diff
@@ -97,6 +97,13 @@
           "ip4": ["156.145.115.47", "156.145.115.138", "156.145.115.188", "157.139.21.56", "157.139.237.172", "157.139.9.228"]
         }
     },
+    "awardee-newyork@all-of-us-ops-data-api-stable.iam.gserviceaccount.com": {
+        "roles": ["awardee_sa"],
+        "awardee": "newyork",
```
Seeker1911: same.

rest-api/config/config_prod.json
wangy70: Are you sure? I saw there is another account for columbia is "awardee-columbia@all-of-us-ops-data-api-stable.iam.gserviceaccount.com". And in the ticket PD-2165, Celecia said it's "newyork".

Seeker1911: we use column A of this sheet https://docs.google.com/spreadsheets/d/1CcIGRV0Bd6BIz7PeuvrV6QDGDRQkp83CJUkWAl-fG58/edit#gid=1076878570
There is no "new york" awardee ID. 
We have a columbia because another organization under the awardee has already signed up.
We can have duplicates here, the `awardee` key is used to ensure the ops-data api doesn't search for other awardees.

wangy70: done

rest-api/config/config_stable.json
wangy70: done

PR #1088, opened 2019-05-29
update_ehr_status cron job simplified dataset from bigquery

rennat:


rest-api/cloud_utils/bigquery.py
```diff
@@ -1,100 +1,185 @@
...
+import re
 
+import pytz
 from googleapiclient.discovery import build
 from google.appengine.api import app_identity
 
 
-SOCKET_TIMEOUT = 600000
-BQ_DEFAULT_RETRY_COUNT = 10
+class BigQueryJob(object):
```
rennat: This `BigQueryJob` class will get merged into Robert's work with the bigquery utils when they are merged in.

PR #1091, opened 2019-05-31
[DA-1185] Build BQ tools and schema support

robabram:
This PR creates a new directory named `tools_libs` under `rest-api/tools/`. This will be the place for cli tools written using the new python GCP tooling. 

I have also added Bash Completion support for the new tools, see `rest-api/tools/tool_libs/tools.bash`.  You can use it by running: `. tools/tool_libs/tools.bash`.

You can run the new tools by using `python -m tools --help`.

### DIFF
--- a/rest-api/model/bq_base.py
+++ b/rest-api/model/bq_base.py
@@ -1,0 +1,331 @@
+#
+# BigQuery table schema support
+#
+import collections
+from exceptions import Exception
+import importlib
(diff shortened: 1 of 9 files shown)


rennat:
This is overall good. I would like to see some pure python unit tests for the bigquery utility classes. This will demonstrate their usage and give us some confidence that things are behaving as expected.

PR #1100, opened 2019-06-05
[DA-1150] add patientStatus to ParticipantSummary API

rennat:
# Changes

## `ParticipantSummary.patientStatus` relationship
This `relationship` was added on the model for use in queries.

## `BaseApi.query` & `query.FieldFilter` refactoring
The core logic has not changed but has been rearranged to allow subclassing `FieldFilters` for custom filter implementation.

## `BaseDao` & `query.Query` new `options` argument
`query.Query` objects can now specify `options` to be passed into a `sqlalchemy.query`. This allows for lazy/eager loading of children as needed when building a query.

# Consequences

## Performance
- `ParticipantSummaryApi` now eagerly loads children for `patientStatus` using `subqueryload()`
- `ParticipantSummaryApi` queries that target patientStatus will use subquery (`relationship.any()`)

These subqueries seemed like a reasonable start but we need to evaluate performance and determine if we should be joining instead.


rest-api/api/participant_summary_api.py
```diff
@@ -10,7 +10,10 @@
 
 class ParticipantSummaryApi(BaseApi):
   def __init__(self):
-    super(ParticipantSummaryApi, self).__init__(ParticipantSummaryDao())
+    super(ParticipantSummaryApi, self).__init__(
+      ParticipantSummaryDao(),
+      get_returns_children=True
```
robabram: I have a performance concern when users use this API for sync'ing records.  I think we should put this under a new temporary testing API, excluded in Prod, to compare various usage scenarios vs the current API.

rennat:
After discussing this, we intend to merge this branch as-is. 
*Before* we merge this branch, we will create and use an aggregate profiling client program to be able to compare the performance before and after this code goes into test.

PR #1109, opened 2019-06-07
[DA-1199] Fix POST request response, add history table

robabram:
Change POST method to return an object instead of list.
Add new style history table to patient status.


rest-api/test/unit_test/api_test/patient_status_api_test.py
```diff
@@ -63,16 +63,22 @@ def test_patient_status_udpated(self):
...
-    self.send_put(url, data, expected_status=httplib.OK)
+    resp = self.send_put(url, data, expected_status=httplib.OK)
+    self.assertDictContainsSubset(data, resp)
 
     # Get record and test that our test_data dict is in the resp.response dict.
     resp = self.send_get(url)
     self.assertDictContainsSubset(data, resp)
 
-  # TODO: When new style history tables and triggers have been added to unit tests, test history URL.
+    # TODO: When new style history tables and triggers have been added to unit tests, test history URL.
```
robabram: Its still not possible to test against the new style history tables. I'm creating a ticket for that.

PR #1114, opened 2019-06-19
[DA-1220] make ehr update data sources configurable

rennat:
The existing bigquery table names were hard-coded. A view needed to change immediately after the initial release so this work allows changing the data view sources via the configuration.

### DIFF
--- a/rest-api/test/unit_test/offline_test/update_ehr_status_test.py
+++ b/rest-api/test/unit_test/offline_test/update_ehr_status_test.py
@@ -5,6 +5,7 @@ import mock
 import pytz
 
 from clock import FakeClock
+import config
 from dao.ehr_dao import EhrReceiptDao
 from dao.hpo_dao import HPODao
 from dao.organization_dao import OrganizationDao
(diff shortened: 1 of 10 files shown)


rennat:
The codacy complaints here are acceptable because this string formatting of a SQL-like object is using only configuration parameters.

rest-api/test/unit_test/offline_test/update_ehr_status_test.py
```diff
@@ -77,6 +145,24 @@ def _make_participant(self, hpo, org, int_id):
...
 
+  @mock.patch('offline.update_ehr_status.make_update_organizations_job')
+  @mock.patch('offline.update_ehr_status.make_update_participant_summaries_job')
+  def test_skips_when_no_job(self, mock_summary_job, mock_organization_job):
+    mock_summary_job.return_value = None
+    mock_organization_job.return_value = None
+    with FakeClock(datetime.datetime(2019, 1, 1)):
+      offline.update_ehr_status.update_ehr_status()
+
+    summary = self.summary_dao.get(11)
```
rennat: I'm submitting another commit to improve this test case

PR #1123, opened 2019-06-24
move mayo api template to module and remove os package

Seeker1911:


Seeker1911:
we can ignrore the codacy error for etree (when parsing our own xml in the module)

PR #1126, opened 2019-06-25
v1-50-v5 HOTFIX: EHR Metrics API and Update cron job fixes

rennat:
This changeset does two things:

1. improve robustness when parsing data from biquery (handle new edge cases)
2. adjust filter logic for two queries to make the counts identical


rennat:
Codacy checks are not running for this hotfix branch

PR #1133, opened 2019-06-26
[DA-1187] Build participant JSON data for transfer to BigQuery

robabram:
NOTE: Includes adding `Enum34` python package.  Remember to run `tools/setup_env.sh` after merge to master.
Includes misc improvments for new tool utilities.
`version for tanner` was a commit for Tanner to review an issue and is fixed in the next commit.


rest-api/model/bq_base.py
```diff
@@ -309,22 +338,135 @@ class BQView(object):
   sql = None
 
 
+# class BQSession(object):
+#
```
Seeker1911: did you mean to remove these ?

rest-api/model/bq_base.py
```diff
@@ -309,22 +338,135 @@ class BQView(object):
...
+    :return: value
+    """
+    return getattr(self, item)
+
+  def update_values(self, data):
+    """
+    Update this object with the given dict values, validate against BQSchema if available.
+    :param data: dict of data values to add/update.
+    """
+    def update(d, u, s):
```
Seeker1911: what does d/u/s represent ?

rest-api/model/bq_base.py
robabram: Not at this time.  I want to create a testing API that replicates the current PS API, so BQSession might be used to combine Tanner's BQ query job code with the schema code for that testing API.

rest-api/model/bq_base.py
robabram: d=dict to update
u=dict to update from
s=schema

PR #1135, opened 2019-06-27
[DA-1188] cron jobs for BigQuery

robabram:
Contains two cron jobs.
1. A manual only job that can rebuild all participant summary records in the `bigquery_sync` table and force them to be synchronized to BigQuery.
2. An automatic job that sync's as many records from the `bigquery_sync` table it can in 90 seconds.


robabram:
The Codacy issue can be ignored, the sql does not take user input.

PR #1141, opened 2019-07-01
[DA-1203] modify withdrawal report

wangy70:
Modify Withdrawal Report to only send Biobank IDs of participants that had samples collected


wangy70:
You can ignore the SQL injection alert, it's a static string.

PR #1160, opened 2019-07-12
[Legacy] fix eager loading of patient status on participant summary

rennat:


Seeker1911:
should we just close this ?

PR #1161, opened 2019-07-12
[DA-1236] Change Cron job to utilize task queues to re-gen BQ data

robabram:


rest-api/offline/main.py
```diff
@@ -190,19 +190,20 @@ def genomic_pipeline():
 
 @app_util.auth_required_cron
 @_alert_on_exceptions
-def bigquery_rebuild():
-  # this should always be a manually run job, but we have to schedule it.
+def bigquery_rebuild_cron():
+  """ this should always be a manually run job, but we have to schedule it at least once a year. """
   now = datetime.utcnow()
```
Seeker1911: is this imported somewhere ?

robabram: It is used lower in the same file in the offline_app.add_url_rule() call to define the cron job call function.

PR #1165, opened 2019-07-15
DA-1235 fix stable test participant pairing

Seeker1911:
We were passing in an unnecessary argument. num_participants (which was unused anyway) was overwriting the keyword param reader in import_participants.py Line 145.


Seeker1911:
creating the tools testing client grew way beyond the scope of this ticket. I removed the commit and I'll merge this fix in. We can work on the new tool seperately.

PR #1171, opened 2019-07-19
DA-1194 download csv files from buckets

Seeker1911:


Seeker1911:
#ignore_codacy 😃
using os.system from known/good/controlled Google bucket

rest-api/tools/download_data_files.py
```diff
@@ -0,0 +1,24 @@
+import os
+""" Downloads csv files from Google buckets and places in data/ dir (gitignored).
+"""
+
+# @todo: change this to the new proper bucket once it's made
+SOURCE_BUCKET = 'all-of-us-rdr-sequestered-config-test/'
```
wangy70: Do we need to keep `export_organizations.sh` before we have the proper bucket name?

Seeker1911: good point. I put it back.

PR #1176, opened 2019-07-26
addding bucket name to download tool

Seeker1911:
added the real bucket name for the `export_organizations.sh` replacement tool.


Seeker1911:
we should wait to remove the old tool until we've confirmed the business has switched over to using the buckets.

PR #1181, opened 2019-07-31
[DA-1251] Create table to capture raw API request payload 

robabram:


rest-api/api/base_api.py
```diff
@@ -1,19 +1,38 @@
...
 from google.appengine.ext import deferred
-
+from sqlalchemy import inspect
+from sqlalchemy.exc import NoInspectionAvailable
 
 DEFAULT_MAX_RESULTS = 100
 MAX_MAX_RESULTS = 10000
 
 
+def _deferred_save_raw_request(log):
```
Seeker1911: Seems like putting this in the base_dao would be the more natural place to put this instead of importing.

rest-api/api/base_api.py
```diff
@@ -61,6 +116,7 @@ def get(self, id_=None, participant_id=None):
       if participant_id != obj.participantId:
         raise NotFound("%s with ID %s is not for participant with ID %s" %
                        (self.dao.model_type.__name__, id_, participant_id))
+    self._save_raw_request(obj)
```
Seeker1911: Could we just put this call in _make_response once ? Seems like all the verbs call it.

robabram: Its possible that not all API's call `super()` in their `_make_response` override.  This way makes sure `_save_raw_request()` is always called, and is future proof even if all child implementations call the `super()` method now.

rest-api/api/base_api.py
```diff
@@ -44,6 +63,42 @@ def _get_request_arg_bool(self, key, default=False):
...
+  def _save_raw_request(self, obj):
+    """ Create deferred task to save the request payload and possibly link it to a table record """
+    log = RequestsLog()
+
+    log.endpoint = request.endpoint
+    log.method = request.method
+    log.url = request.url
+    log.resource = request.data
+    log.version = int(request.url.split('/')[4][1:])
+
```
wangy70: I'm a little confuse about this `log.version`,  do we always have this version value in the url? Do we need to catch the exception here?

rest-api/api/base_api.py
robabram: I've moved the function to the base_dao.py file.

rest-api/api/base_api.py
robabram: As long as we use following path format `https://xxxx/rdr/v1/xxxx` , it will return 1 or what ever integer is the version number.

wangy70: Oh, I forgot the prefix. Thanks for the explanation.

PR #1196, opened 2019-08-09
start moving cloud storage functions

Seeker1911:


Seeker1911:
There doesn't seem to be a way to emulate the storage api on localhost for testing. But this can at least start us in the direction of running the app.

rdr_service/offline/bigquery_sync.py
```diff
@@ -5,7 +5,7 @@
 from datetime import datetime
 
 from rdr_service.config import GAE_PROJECT
-from google.appengine.api import taskqueue
+#from google.appengine.api import taskqueue @todo
```
rennat: I added a placeholder for this in #1197
you can drop the changes in this file from this PR

Seeker1911: done

PR #1215, opened 2019-08-16
Fixed up database connection to use a dedicated unittest connection

robabram:
Dropped using a connection string from the os environment, now uses a dedicated unittest connection string in the local connection config file (.configs/db_config.json).

ALSO remember to call super() for BaseTestCase.setUp().

### DIFF
--- a/tests/helpers/unittest_base.py
+++ b/tests/helpers/unittest_base.py
@@ -1,3 +1,4 @@
+import os
 import faker
 import unittest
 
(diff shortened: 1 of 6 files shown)


tests/helpers/unittest_base.py
```diff
@@ -13,11 +15,21 @@
...
+    def __init__(self, *args, **kwargs):
+        super(BaseTestCase, self).__init__(*args, **kwargs)
+        # make sure this gets called before anything else.
         reset_mysql_instance()
+        self.fake = faker.Faker()
+
+    def setUp(self) -> None:
+        super(BaseTestCase, self).setUp()
+
+        reset_mysql_instance()
```
Seeker1911: I don't think we want this in __init__ and setup do we ?

robabram: Part of the problem on Friday was the `setUp()` method test_site_dao.py didn't call `reset_mysql_instance()`. We don't really need to call `reset_mysql_instance()` in the init method, we just need to set the `UNITTEST_FLAG` environment variable in the init method.  I've added a new commit that fixes that.

PR #1219, opened 2019-08-20
Fixed missing barcode in API call.

robabram:
Removed fields that PTSC removed from codebook.


rest-api/dao/bigquery_sync_dao.py
```diff
@@ -207,14 +207,10 @@ def _prep_consentpii_answers(self, p_id, session):
       'middle_name': qnan.PIIName_Middle,
```
Seeker1911: remove BQ file ?

rest-api/model/bq_participant_summary.py
```diff
@@ -173,12 +173,8 @@ class BQParticipantSummarySchema(BQSchema):
   distinct_visits = BQField('distinct_visits', BQFieldTypeEnum.INTEGER, BQFieldModeEnum.NULLABLE)
```
Seeker1911: remove bq file ?

PR #1268, opened 2019-09-09
minor updates to skipped tests

Seeker1911:


rdr_service/storage.py
```diff
@@ -3,11 +3,13 @@
...
 import datetime
 import hashlib
+import pathlib
 
 from contextlib import ContextDecorator
 from tempfile import mkstemp
 from abc import ABC, abstractmethod
 from google.cloud import storage
+from google.cloud import storage
 from google.cloud.storage import Blob
```
wangy70: dup import

tests/helpers/unittest_base.py
```diff
@@ -139,21 +139,26 @@ def setUp(self, with_data=True, with_consent_codes=False) -> None:
...
+    def setup_storage(self):
         temp_folder_path = mkdtemp()
+        self.addCleanup(shutil.rmtree, temp_folder_path)
         os.environ['RDR_STORAGE_ROOT'] = temp_folder_path
 
-    @staticmethod
-    def setup_config():
-        if os.environ.get('UNITTEST_CONFIG_FLAG'):
-            return
-        data = read_dev_config(os.path.join(os.path.dirname(__file__), "../../rdr_service/config/base_config.json"),
```
wangy70:
```
if os.environ.get('UNITTEST_CONFIG_FLAG'):
            return
```
This will avoid we setup the config for every test case, are you sure you want to remove it?

rdr_service/storage.py
```diff
@@ -68,7 +74,12 @@ def _get_blob_name_from_local_path(self, local_path):
         return blob_name
 
     def open(self, path, mode):
-        return open(self._get_local_path(path), mode)
+        local_path = self._get_local_path(path)
+        directory = os.path.dirname(local_path)
+        if directory:
+            pathlib.Path(directory).mkdir(parents=True, exist_ok=True)
+
```
wangy70: Why do we create the directory before we open a file in it?

rdr_service/storage.py
Seeker1911: done

tests/helpers/unittest_base.py
Seeker1911: yes so we can keep or overwrite configs

rdr_service/storage.py
Seeker1911: to ensure parents exist or are created. Some tests pass in a hierarchy and this goes to temp.

tests/helpers/unittest_base.py
rennat: Yes we want the configs to be clean for each test.
This does have the added overhead of interacting with the filesystem a lot. If we want to speed this up then we should look into an alternate in-memory `ConfigProvider`

rdr_service/storage.py
wangy70: @rennat  I don't think we need create the directory when we open a file. If the file is not there, even you create the directory, the file still not there. I didn't get the intention of this part. What do you think?

rennat: You are right that we do not need to when opening for reading. However, we do need to create the directory when opening for writing.
We use the filesystem storage provider only for local development so this shortcut seemed reasonable. To get a `complete` solution for this we would need a file-wrapper that only creates the directories when `write` is called. This seemed like a lot of extra work only to fix a minor inconvenience for local development.

summary: yes it is a problem but it has minimal impact and it is not worth the time to fix it properly right now

wangy70: OK, I got you. When I do my test cases, I manually create the directories before I open them, so that's why I didn't need this part.

PR #1274, opened 2019-09-10
Migrate BigQuery hotfix

robabram:
Updates to tests and mysql unittest helper.


rdr_service/api/dv_order_api.py
```diff
@@ -114,9 +183,9 @@ def _put_supply_request(self, resource, bo_id):
             )
             p_id = from_client_participant_id(pid.value)
         except AttributeError as e:
-            raise BadRequest(e.message)
+            raise BadRequest(e)
```
Seeker1911: I think we need str(e)

Seeker1911:
```suggestion
            raise BadRequest(str(e))
```

rdr_service/api/dv_order_api.py
```diff
@@ -114,9 +183,9 @@ def _put_supply_request(self, resource, bo_id):
...
         except Exception as e:
-            raise BadRequest(e.message)
+            raise BadRequest(e)
```
Seeker1911:
```suggestion
            raise BadRequest(str(e))
```

rdr_service/api/dv_order_api.py
```diff
@@ -125,47 +194,65 @@ def _put_supply_request(self, resource, bo_id):
...
-            tracking_status = fhir.extension.get(url=VIBRENT_FHIR_URL + "tracking-status").valueString
-            tracking_id = fhir.identifier.get(system=VIBRENT_FHIR_URL + "trackingId").value
+            tracking_status = fhir.extension.get(
+                url=VIBRENT_FHIR_URL + "tracking-status"
+            ).valueString
+            if tracking_status:
+                tracking_status = tracking_status.lower()
         except AttributeError as e:
-            raise BadRequest(e.message)
+            raise BadRequest(e)
```
Seeker1911:
```suggestion
            raise BadRequest(str(e))
```

rdr_service/api/dv_order_api.py
```diff
@@ -125,47 +194,65 @@ def _put_supply_request(self, resource, bo_id):
...
         except Exception as e:
-            raise BadRequest(e.message)
+            raise BadRequest(e)
```
Seeker1911:
```suggestion
            raise BadRequest(str(e))
```

tests/test-data/dv_order_api_post_supply_delivery.json
```diff
@@ -51,7 +51,7 @@
   "extension": [
     {
       "url": "http://joinallofus.org/fhir/tracking-status",
-      "valueString": "enroute"
+      "valueString": "in_transit"
```
Seeker1911: perhaps we should make this upper case to ensure we're catching it correctly everywhere.

robabram: PTSC is sending it to us lowercase, so I think we should probably keep it lowercase here.

PR #1285, opened 2019-09-13
add ops data api readme

Seeker1911:


Seeker1911:
apologize for having the json files in this PR too. However that has already been approved so it's a no-op in this case. The README is because the root directory ops data readme link was socialized, so we need to keep it there.

PR #1309, opened 2019-09-25
Misc testing and fixing items.

robabram:


rdr_service/storage.py
```diff
@@ -170,6 +170,13 @@ def __init__(self, provider, blob):
         self.temp_file = None
         self.temp_file_path = None
 
+    def __iter__(self):
+        # TODO: Download file now and
```
Seeker1911: and then ?

robabram: Removed todo.

PR #1314, opened 2019-09-27
[DA-1298] Hierarchy content implementation

wangy70:
Hierarchy content implementation

### DIFF
--- a/rest-api/dao/organization_hierarchy_sync_dao.py
+++ b/rest-api/dao/organization_hierarchy_sync_dao.py
@@ -26,7 +26,7 @@ class OrganizationHierarchySyncDao(BaseDao):
     self.organization_dao = OrganizationDao()
     self.site_dao = SiteDao()
 
-  def from_client_json(self, resource_json, id_=None, expected_version=None, client_id=None):
+  def from_client_json(self, resource_json, id_=None, expected_version=None, client_id=None):  # pylint: disable=unused-argument
     try:
       fhir_org = lib_fhir.fhirclient_3_0_0.models.organization.Organization(resource_json)
     except FHIRValidationError:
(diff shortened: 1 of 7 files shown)


rest-api/dao/organization_hierarchy_sync_dao.py
```diff
@@ -0,0 +1,385 @@
...
+    obj_type = self._get_type(hierarchy_org_obj)
+
+    operation_funcs = {
+      'AWARDEE': self._update_awardee,
+      'ORGANIZATION': self._update_organization,
+      'SITE': self._update_site
+    }
+
+    if obj_type not in operation_funcs:
+      raise BadRequest('No awardee-type info found in payload data.')
```
Seeker1911: I think this wording could become confusing, maybe something without explicitly mentioning one of (awardee, organization, site)

rest-api/dao/organization_hierarchy_sync_dao.py
```diff
@@ -0,0 +1,385 @@
...
+from model.organization import Organization
+from model.site import Site
+from model.site_enums import ObsoleteStatus, SiteStatus, EnrollingStatus, DigitalSchedulingStatus
+from dao.hpo_dao import HPODao
+from participant_enums import OrganizationType
+from dao.organization_dao import OrganizationDao
+from dao.site_dao import SiteDao
+from dateutil.parser import parse
+
+_FHIR_SYSTEM_PREFIX = 'http://all-of-us.org/fhir/sites/'
```
Seeker1911: We have a similar structure in api_util.py. We might consider consolidating this constant there.

wangy70: done

rest-api/dao/organization_hierarchy_sync_dao.py
wangy70: 'awardee-type' is the field name in their payload.

PR #1315, opened 2019-09-30
Mm/logging

Seeker1911:
add logging middleware and a first attempt at auto deploying to sandbox for python 3

### DIFF
--- a/rdr_service/storage.py
+++ b/rdr_service/storage.py
@@ -1,3 +1,4 @@
+import io
 import os
 import glob
 import shutil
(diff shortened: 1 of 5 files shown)


Seeker1911:
I disagree whole heartedly disagree with Codacy's choice of algorithms here.

rdr_service/storage.py
```diff
@@ -225,6 +225,36 @@ def __enter__(self):
     def __exit__(self, *exc):
         self.close()
         return False
+        
+    def __iter__(self):
```
robabram: This function conflicts with the same function definition at line 177.  The one at line 177 probably needs to be removed.

robabram: Never mind, saw you removed it in a later commit.

PR #1323, opened 2019-10-02
Py3 BigQuery Fixes for Sync process and Import Organizations

robabram:
Fixes a minor BigQuery sync issue.
Fixes the `localhost` BigQuery sync issue after running `import_organizations.sh`.
Use the DB connection string created by the cli tools when they are run.  Otherwise we only connect to our localhost DB when running cli tools.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -31,10 +31,10 @@ cd rdr_service
 
 # Shouldn't need this anymore
 # ./tools/install_config.sh --config=config/config_dev.json --update
-./tools/setup_local_database.sh --db_user root --db_name rdr
+# ./tools/setup_local_database.sh --db_user root --db_name rdr
 
 cd $PROJ_DIR
 
 ./ci/check_licenses.sh
 
-python -m unittest discover -v -s tests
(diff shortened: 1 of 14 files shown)


Seeker1911:
looks good so far.

ci/test.sh
```diff
@@ -31,10 +31,10 @@ cd rdr_service
 
 # Shouldn't need this anymore
 # ./tools/install_config.sh --config=config/config_dev.json --update
-./tools/setup_local_database.sh --db_user root --db_name rdr
+# ./tools/setup_local_database.sh --db_user root --db_name rdr
```
Seeker1911: dont we still need this ?

robabram: If the unit tests all use the 9306 db instance, we don't.  I believe commenting it out will let us catch tests that try to use the 3306 db instance.  Using the 3306 db instance would indicate that something is wrong that needs correcting.

Seeker1911: ok, isnt this how CI runs its tests though ?

robabram: It runs them the same way as we do, it spins up a db instance on port 9306.

PR #1325, opened 2019-10-02
[DA-1316] Bug Fix for Salivary Expected Delivery Date

j-kanuch:
Changed the condition to evaluate to True if there is an available 'expected-delivery-date' URL.
Also expanded test_dv_order_dao.py to test for the expected delivery date and other expected data.

### DIFF
--- a/tests/dao_tests/test_dv_order_dao.py
+++ b/tests/dao_tests/test_dv_order_dao.py
@@ -86,7 +86,6 @@ class DvOrderDaoTestBase(BaseTestCase):
 
     def test_from_client_json(self):
         # pylint: disable=unused-variable
-        
         req_payload = self.send_post(
             "SupplyRequest",
             request_data=self.post_request,
--- a/tests/dao_tests/test_dv_order_dao.py
+++ b/tests/dao_tests/test_dv_order_dao.py
@@ -111,7 +111,8 @@ class DvOrderDaoTestBase(BaseTestCase):
 
             # return a BiobankDVOrder object from database
             with self.dao.session() as session:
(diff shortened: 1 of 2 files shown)


tests/dao_tests/test_dv_order_dao.py
```diff
@@ -83,6 +84,39 @@ def test_enumerate_tracking_status(self):
         )
         self.assertEqual(status, OrderShipmentTrackingStatus.IN_TRANSIT)
 
+    def test_from_client_json(self):
+        # pylint: disable=unused-variable
+        req_payload = self.send_post(
```
Seeker1911: if you're not going to use the variables, why assign them at all?

tests/dao_tests/test_dv_order_dao.py
```diff
@@ -83,6 +84,39 @@ def test_enumerate_tracking_status(self):
...
+            request_data=self.post_request,
+            expected_status=http.client.CREATED,
+        )
+        deliv_payload = self.send_post(
+            "SupplyDelivery",
+            request_data=self.post_delivery,
+            expected_status=http.client.CREATED,
+        )
+
+        existing_obj = self.dao.from_client_json(self.post_delivery, participant_id=self.participant.participantId)
```
Seeker1911: The typical pattern is to use something like `existing_obj` when we fetch a record from the DB and are comparing it to a new obj. This name seems a little misleading.

Seeker1911:
You should check the database values to test that it is inserted as expected and if a change is made that it is also reflected in the record.

PR #1335, opened 2019-10-07
DA-1273 update docs

Seeker1911:


Seeker1911:
Put ops data api back in root as the url has been shared with partners across the AOU org. Theyve expressed the desire to put it back as we have in master already.

PR #1341, opened 2019-10-08
Implemented migration to fix history table

j-kanuch:


Seeker1911:
we'll do this PR in tandem with the code change PR so theres not a time gap where these tables are essentially broken.

rdr_service/alembic/versions/15e05f673f5f_altering_biobank_dv_order_history.py
```diff
@@ -0,0 +1,116 @@
...
+    # Procedure to remove the biobank_reference column and move data to correct columns of the history table
+    # Make a bkup_ table
+    op.execute(
+        "CREATE TABLE bkup_biobank_dv_order_history AS SELECT * FROM biobank_dv_order_history;"
+    )
+
+    # PREP THE tmp_ TABLE
+    # create the tmp_ table
+    op.execute(
+        "CREATE TABLE tmp_biobank_dv_order_history AS select * from biobank_dv_order_history;"
```
robabram: You could add `limit 1` to this statement since you are truncating the new temp table in the next step.

rdr_service/alembic/versions/15e05f673f5f_altering_biobank_dv_order_history.py
```diff
@@ -94,6 +94,8 @@ def upgrade_rdr():
     # Clean out the tmp_ table
     op.execute("DROP TABLE tmp_biobank_dv_order_history;")
 
+    # Removing the bkup_ table
+    op.execute("DROP TABLE bkup_biobank_dv_order_history;")
```
robabram: I think this should be commented out, this should be a manual step.  We want to keep the backup table until we have verified everything is correct.  Once verified, we should manually drop the backup table.

j-kanuch: Sounds good. I commented it out.

PR #1344, opened 2019-10-11
Modified biobankStatus to source from Mayolink's API

j-kanuch:
Added `biobankStatus` to API test. Removed from DAO test. Writing 'status' field from Mayolink's response to `biobankStatus`.

### DIFF
--- a/rdr_service/dao/dv_order_dao.py
+++ b/rdr_service/dao/dv_order_dao.py
@@ -27,7 +27,6 @@ from rdr_service.model.config_utils import to_client_biobank_id
 from rdr_service.model.utils import to_client_participant_id
 from rdr_service.participant_enums import BiobankOrderStatus, OrderShipmentStatus, OrderShipmentTrackingStatus
 
-
 class DvOrderDao(UpdatableDao):
     def __init__(self):
         self.code_dao = CodeDao()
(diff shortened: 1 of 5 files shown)


tests/dao_tests/test_dv_order_dao.py
```diff
@@ -78,8 +77,6 @@ def test_insert_biobank_order(self):
         self.assertEqual(put_response["version"], 4)
         self.assertEqual(put_response["meta"]["versionId"].strip("W/"), '"4"')
         self.assertEqual(put_response["barcode"], "SABR90160121INA")
-        # self.assertEqual(put_response["biobankOrderId"], "12345")
-        self.assertEqual(put_response["biobankStatus"], "Delivered")
         self.assertEqual(put_response["order_id"], 999999)
```
Seeker1911: You should add an assert from a GET request showing it is inserted and fetchable from the DB.

rdr_service/dao/dv_order_dao.py
```diff
@@ -138,6 +137,13 @@ def from_client_json(
             if not existing_obj:
                 raise NotFound("existing order record not found")
 
+            # handling of biobankStatus from Mayolink API
+            try:
+                existing_obj.biobankStatus = resource_json['biobankStatus']
```
Seeker1911: resource json here would be the payload from the request which will not contain `biobankStatus`.

j-kanuch: Shouldn't `biobankStatus` exist in a PUT request here, set in `to_client_json`? The `model` parameter would be the response from MayoLink, and `for_update` is `True`.
Otherwise you are right, it would not exist, hence the catching on `KeyError` below.

Seeker1911: If biobank status is the status of the biobank order from the mayolink api then PTSC should not be updating it.

PR #1345, opened 2019-10-14
revert_senamtic_version_column

wangy70:
revert the table change of questionnaire semantic version


rest-api/alembic/versions/a53100879199_add_senamtic_version_column_to_.py
```diff
@@ -0,0 +1,59 @@
+"""add senamtic_version column to questionnaire
```
Seeker1911: what does this file do?

wangy70: This is the original alembic file which was deleted when that PR was reverted. But we still need this file as a place holder because the alembic migration need this one in it's own version system.

PR #1356, opened 2019-10-17
tune_metrics_v2_cron_job

wangy70:
The idea is to create a temp table at the very beginning to store all the participant and participant summary data.
And then run the cron job against the temp table, so it will not lock the participant and participant summary tables.

### DIFF
--- a/rest-api/dao/metrics_cache_dao.py
+++ b/rest-api/dao/metrics_cache_dao.py
@@ -37,16 +37,8 @@ class MetricsCacheJobStatusDao(UpdatableDao):
       index_name_list = list(set(index_name_list))
 
       for index_name in index_name_list:
-        if index_name not in ['PRIMARY', 'participant_summary_hpo']:
+        if index_name != 'PRIMARY':
           session.execute('ALTER TABLE  metrics_tmp_participant DROP INDEX  {}'.format(index_name))
-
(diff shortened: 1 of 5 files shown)


rest-api/dao/metrics_cache_dao.py
```diff
@@ -19,6 +20,66 @@
...
+        if index_name not in ['PRIMARY', 'participant_summary_hpo']:
+          session.execute('ALTER TABLE  metrics_tmp_participant DROP INDEX  {}'.format(index_name))
+
+      session.execute('CREATE INDEX idx_sign_up_time ON metrics_tmp_participant (sign_up_time)')
+      session.execute('CREATE INDEX idx_consent_time ON metrics_tmp_participant '
+                      '(consent_for_study_enrollment_time)')
+      session.execute('CREATE INDEX idx_member_time ON metrics_tmp_participant '
+                      '(enrollment_status_member_time)')
+      session.execute('CREATE INDEX idx_sample_time ON metrics_tmp_participant '
+                      '(enrollment_status_core_stored_sample_time)')
```
robabram: Creating the indexes on the table before inserting the records will slow down the inserts.  Can the index creation lines be moved to after the data has been inserted?

wangy70: done

PR #1380, opened 2019-10-29
[DA-1350] Populate biobank_order.finalized_time from biobank_ordered_sample

j-kanuch:


rest-api/dao/biobank_order_dao.py
```diff
@@ -352,6 +352,8 @@ def from_client_json(self, resource_json, id_=None, expected_version=None,
         resource.processed_info)
     order.finalizedUsername, order.finalizedSiteId = self._parse_handling_info(
         resource.finalized_info)
+    # order.finalizedTime uses the time from biobank_ordered_sample.finalized
+    order.finalizedTime = parse_date(self.get_random_sample_finalized_time(resource).isostring)
```
robabram: There is potential for get_random_sample_finalized_time() to return a None value which would cause an exception here.

j-kanuch: I'll catch the AttributeError just in case. In regards to the return value, @robabram  is correct, I don't need parse_date. I just needed to return the `date` property of the FHIRDate object, and then remove the offset.

PR #1389, opened 2019-10-30
[DA-1350] Populate biobank_order.finalized_time from biobank_ordered_sample

j-kanuch:


tests/dao_tests/test_biobank_order_dao.py
```diff
@@ -114,6 +116,13 @@ def test_from_json(self):
         self.assertEqual(2, order.finalizedSiteId)
         self.assertEqual("bob@pmi-ops.org", order.finalizedUsername)
 
+        # testing finalized_time
+        samples_finalized_time = None
+        for sample in order_json['samples']:
+            samples_finalized_time = parse_date(sample['finalized'])
+            break
```
wangy70: Do we need a loop here?
Seems it's same as: 
`samples_finalized_time = parse_date(order_json['samples'][0]['finalized'])`

j-kanuch: I agree, and was on the fence about doing it that way. I'm using a loop for consistency because the code that is being tested uses a loop as well. From the biobank_order_dao.py:
```
  def get_random_sample_finalized_time(self, obj):
    """all samples are set to same finalized time in an order, we only need one."""
    for sample in obj.samples:
      if sample.finalized is not None:
        return sample.finalized
```

I can update the test since it seems better to not loop in this instance.

PR #1405, opened 2019-11-11
[ROC-228] Added whitelisted IP for awardee-tach on stable.

j-kanuch:
This PR adds a whitelisted IP for the awardee-tach service account on `stable`.


Seeker1911:
I thought this IP would be added to prod as well

j-kanuch:
> I thought this IP would be added to prod as well

It already was added to the `prod` and `stable` configs in a previous oncall request. At some point it was removed from `stable`. This new request was to add it to both `prod` and `stable`, but only `stable` needs to be updated.

PR #1408, opened 2019-11-12
Mm/participant origination

Seeker1911:


rdr_service/dao/participant_dao.py
```diff
@@ -134,11 +134,12 @@ def _validate_update(self, session, obj, existing_obj):
             and obj.withdrawalReasonJustification is None
         ):
             raise BadRequest("missing withdrawalReasonJustification in update")
-        if get_oauth_id() != existing_obj.participantOrigination:
-            logging.warning(f"{get_oauth_id()} tried to modify participant from \
-                    {existing_obj.participantOrigination}")
-            raise BadRequest(f"{get_oauth_id()} not able to update participant from source \
-                    {existing_obj.participantOrigination}")
+        if existing_obj:
+            if get_oauth_id() != existing_obj.participantOrigination:
```
robabram: Doesn't get_oauth_id() return an email address?  How will this match up to the participant origination value?

robabram: Nevermind, I saw that you changed that to a call to lookup origination id from email.

Seeker1911: that looks like an old version

PR #1446, opened 2019-12-02
[DA-1385] Ignoring non-gc metrics files in bucket; updated tests.

j-kanuch:
- Added filter when generating file list to only look for 'datamanifest' files.
- Needed to rename the test files to include the above so they are picked up. 
- Added a new test file `test_empty_wells.csv` that has the fake data currently in the `stable` bucket for unit testing.


rdr_service/genomic/genomic_job_controller.py
```diff
@@ -112,7 +112,8 @@ def _get_uningested_file_names_from_bucket(self, bucket_name):
         """
         files = list_blobs('/' + bucket_name)
         files = [s.name for s in files
-                 if self.archive_folder_name not in s.name.lower()]
+                 if self.archive_folder_name not in s.name.lower()
+                 if 'datamanifest' in s.name.lower()]
```
wangy70, line 116: Is this the same as `if self.archive_folder_name not in s.name.lower() and 'datamanifest' in s.name.lower()` ?

j-kanuch: Yes, the code does the same thing, actually, it is not more efficient, because Python does lazy evaluation. But same code.

wangy70 (approved):

PR #1475, opened 2019-12-06
refactor dv order barcode status

Seeker1911:


j-kanuch (approved):

Seeker1911:
according to PTC docs, they only send barcode when status is shipped.

PR #1488, opened 2019-12-10
From scratch CircleCI configuration script.

robabram:
Update script version to 2.1 from 2.0.
Moved old/unused scripts to `old_scripts` folder.
Created new bash scripts for CircleCI Steps.
Does not include Jira integration yet.
Updated OpenSLL command to work with newer version of OpenSSL.  Someone needs to research what current OpenSSL coder/decoder/hashing settings we should use.  Our current settings end up using MD5 hashing which is depreciated.

The `test_gc_validation_metrics_end_to_end` still intermittently fails.


Seeker1911:
wow, a total rewrite and we are better for it. 👍🏻
I have two small concerns. I think we should comment out or remove the reference to master branch until we reintegrate it. 
Is there a plan on replacing the release_notes script ?

Seeker1911 (approved):

PR #1515, opened 2020-01-02
DA-1268 update appengine tool to restart instances

Seeker1911:
This update adds the restart instances functionality when deploying or when updating configs.
Since deploy also updates configs, I've passed in a new flag to prevent deployment from restarting instances when it calls the config update. Also added a function in app engine manager to allow calling a single gcp_util function (mostly for testing).

### DIFF
--- a/rdr_service/services/gcp_utils.py
+++ b/rdr_service/services/gcp_utils.py
@@ -821,16 +821,15 @@ def gcp_deploy_app(project, config_files: list, version: str = None, promote: bo
 
     return True
 
-def gcp_restart_instances(project=None):
+def gcp_restart_instances(self):
     """
     Restart running instances of an App Engine environment.
     :return: True if successful, else False.
     """
 
-    if not project:
-        project = gcp_get_current_project()
+    project = self.gcp_env.project if self.gcp_env.project else gcp_get_current_project()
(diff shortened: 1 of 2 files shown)


rdr_service/services/gcp_utils.py
```diff
@@ -821,16 +821,15 @@ def gcp_deploy_app(project, config_files: list, version: str = None, promote: bo
 
     return True
 
-def gcp_restart_instances(project=None):
+def gcp_restart_instances(self):
```
robabram: Changing this to require an object with a `gcp_env` property makes this less flexible and isn't consistent with the other functions in this file.  Also, does this function restart instances for all services or only for the default service?  Can we use this function to only restart the cron service instances?

Seeker1911: I wouldn't say its required, if self does not have `gcp_env` it calls the `get_current_project` function. But I see your point and I'm happy to change it to better reflect other functions in the module. As it is this function only restarts default, I didn't see a reason to allow other services but I can add that if you'd like.

robabram: It wouldn't hurt to allow it to restart other services besides the default service.

robabram (approved):

PR #1518, opened 2020-01-09
[PDR-29] Fix Enrollment Status numbers

robabram:
Wrote new PDR tool to update BigQuery data for small sets of participants.
Updated Enrollment Status logic.
Add BQ processing to Biobank Reconciliation process.
Capture newly created participants.
Added Enrollment Status Unittests.

### DIFF
--- a/rdr_service/participant_enums.py
+++ b/rdr_service/participant_enums.py
@@ -78,6 +78,7 @@ class QuestionnaireDefinitionStatus(messages.Enum):
 class EnrollmentStatus(messages.Enum):
     """A status reflecting how fully enrolled a participant is"""
 
+    # REGISTERED = 0  # Adding this value breaks metric counts over time.
     INTERESTED = 1
     MEMBER = 2
     FULL_PARTICIPANT = 3
--- a/rdr_service/participant_enums.py
+++ b/rdr_service/participant_enums.py
@@ -78,7 +78,7 @@ class QuestionnaireDefinitionStatus(messages.Enum):
 class EnrollmentStatus(messages.Enum):
     """A status reflecting how fully enrolled a participant is"""
 
-    # REGISTERED = 0  # Adding this value breaks metric counts over time.
+    # REGISTERED should not be added here as doing so will break metric counts over time.
     INTERESTED = 1
     MEMBER = 2
     FULL_PARTICIPANT = 3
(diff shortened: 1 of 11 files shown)


rdr_service/participant_enums.py
```diff
@@ -78,6 +78,7 @@ class QuestionnaireDefinitionStatus(messages.Enum):
 class EnrollmentStatus(messages.Enum):
     """A status reflecting how fully enrolled a participant is"""
 
+    # REGISTERED = 0  # Adding this value breaks metric counts over time.
     INTERESTED = 1
```
wangy70: This is the old version of enrollmentStatus, there is no `REGISTERED` status, I think we can remove this comment line.

rdr_service/tools/tool_libs/pdr_update_tool.py
```diff
@@ -0,0 +1,174 @@
...
+        if self.args.pid:
+            if self.update_single_pid(self.args.pid) == 0:
+                _logger.info(f'Participant {self.args.pid} updated.')
+            else:
+                _logger.error(f'Participant ID {self.args.pid} not found.')
+
+        return 1
+
+
+def run():
```
Seeker1911, line 135: I think having a run method and a run function could be potentially confusing. Perhaps we could call this `main`

rdr_service/participant_enums.py
robabram: I want to leave it in as a reminder, when I put it in it broke some metrics dao code that looks up the enum value by array position.

rdr_service/tools/tool_libs/pdr_update_tool.py
robabram: If you want to change it, I don't mind.  We should probably do it under a different ticket and change it in all the new tools at the same time.  Also `setup.py` would need to be changed to call the main function instead of run.

Seeker1911: Ok, we’ll call that a todo 👍

Seeker1911 (approved):

rdr_service/participant_enums.py
wangy70: I'm so confused about the comment, maybe at least we can change it to something else. 
You add a new status which is not belong to there and break something, that's the expected result. My point is if it is left there, it looks like `REGISTERED` is one of the valid status and we commented it out just because it will break something.  But the truth is `REGISTERED` is not a valid status for the old version of enrollmentStatus at all.

wangy70 (approved):

rdr_service/participant_enums.py
robabram: Made comment change and commited it.

PR #1542, opened 2020-01-24
[DA-1427] CVL Broad WGS Manifest Workflow

j-kanuch:
This PR implements the workflow for generating the CVL WGS manifest. Two new genomic components were created: `ManifestDefinitionProvider` and `ManifestCompiler`. 
A new branch of the genomic pipeline was created to build the CVL WGS manifest.

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -656,10 +656,7 @@ class GenomicBiobankSamplesCoupler:
 class ManifestDefinitionProvider:
     """
     Helper class to produce the definitions for each manifest
-    Separated here so that all manifest definitions are in one code block
-    (unlike, for example, the biobank samples reports)
     """
-
     # Metadata for the various manifests
     ManifestDef = namedtuple('ManifestDef', ["job_run_field",
                                              "source_data",
@@ -680,7 +677,7 @@ class ManifestDefinitionProvider:
         Creates the manifest definitions to use when generating the manifest
         based on manifest type
         """
-        # TODO: Put destination buckets in config when naming convention defined
+        # Set each Manifes Definition as an instance of ManifestDef()
         self.MANIFEST_DEFINITIONS[GenomicManifestTypes.DRC_CVL_WGS] = self.ManifestDef(
             job_run_field='cvlManifestWgsJobRunId',
             source_data=self._get_source_data_query(GenomicManifestTypes.DRC_CVL_WGS),
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -9,7 +9,7 @@ import re
 import datetime
 from collections import deque, namedtuple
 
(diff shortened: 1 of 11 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -646,3 +651,159 @@ def _get_sex_at_birth(self, participant_id):
                           .first()
         return self._SEX_AT_BIRTH_CODES.get(
                  result[0].lower().split('_')[-1], 'NA') if result else 'NA'
+
+
+class ManifestDefinitionProvider:
```
Seeker1911, line 656: You missed an opportunity here: `ManifestDestinyProvider`

rdr_service/genomic/genomic_job_components.py
```diff
@@ -646,3 +651,159 @@ def _get_sex_at_birth(self, participant_id):
...
+
+        self.MANIFEST_DEFINITIONS = dict()
+        self._setup_manifest_definitions()
+
+    def _setup_manifest_definitions(self):
+        """
+        Creates the manifest definitions to use when generating the manifest
+        based on manifest type
+        """
+        # Set each Manifes Definition as an instance of ManifestDef()
```
Seeker1911, line 680:
```suggestion
        # Set each Manifest Definition as an instance of ManifestDef()
```

rdr_service/genomic/genomic_job_components.py
```diff
@@ -646,3 +651,159 @@ def _get_sex_at_birth(self, participant_id):
...
+        Main execution method for ManifestCompiler
+        :return: result code
+        """
+        self.manifest_def = self.def_provider.get_def(manifest_type)
+        source_data = self._pull_source_data()
+        if source_data:
+            self.output_file_name = self.manifest_def.output_filename
+            logging.info(
+                f'Preparing manifest of type {manifest_type}...'
+                f'{self.manifest_def.destination_bucket}/{self.manifest_def.output_filename}'
```
Seeker1911, line 769: We've never defined the logging formatter to use f strings. It will use % behind the scenes. 
See [python formatting](https://docs.python.org/3/library/logging.html#logging.Formatter) for info.  
Until we do, we should use the % style formatting.

Seeker1911 (approved):

rdr_service/genomic/genomic_job_components.py
j-kanuch: I'll update this in the next PR I have for the CVL Array manifests, probably on Monday.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Google AppEngine 2 logs don't like %-style though, so I updated most of the messages back in https://github.com/all-of-us/raw-data-repository/commit/8ab024f986621eef4dd5a4937505e60dc681c7f4.  I think the default formatter is converted by AppEngine Logging behind the scenes.

PR #1555, opened 2020-01-31
[DA-1428] Create Color Array Manifest

j-kanuch:
PR Implements the CVL Array Manifest workflow as part of the WGS manifest workflow and the unit tests to support it.  There were some minor modifications to the entire workflow as more information came to light in regard of upstream requirements.


rdr_service/offline/genomic_pipeline.py
```diff
@@ -99,4 +99,9 @@ def create_cvl_manifests():
     Entrypoint for CVL Manifest workflow
     Sources from list of biobank_ids in CVL reconciliation report
     """
-    pass
+    # So Far only WGS manifest requirements are really defined
+    job_id = GenomicJob.CREATE_MANIFEST_CVL_WGS
+    run_controller = genomic_job_controller.GenomicJobController(
+        job_id)
+    result = run_controller.run_cvl_wgs_manifest()
```
robabram: This function can return either a result or an error, why are you not checking for an error here?

robabram (approved):

PR #1572, opened 2020-02-07
Initial spec generator unittest integration

robabram:
This is a start on integrating the spec generators into the unittests.


rdr_service/data_gen/generators/hpo.py
```diff
@@ -157,8 +161,8 @@ def get_site(self, site_id):
 
         for site in self._hpo_sites:
             if (
-                site_id == site["Site ID / Google Group"]
-                or "hpo-site-{0}".format(site_id) == site["Site ID / Google Group"]
+                site_id.lower() == site["Site ID / Google Group"].lower()
```
Seeker1911, line 167: we don't need to call `lower()` twice on these lines.

tests/helpers/mysql_helper.py
```diff
@@ -136,6 +140,7 @@ def _initialize_database(with_data=True, with_consent_codes=False):
     config.override_setting('unittest_db_connection_string', db_conn_str)
 
     if with_data:
+        # _setup_hpos()
```
Seeker1911: why

Seeker1911 (approved):

PR #1574, opened 2020-02-07
[DA-1452] Refactor Biobank Return Manifest workflow for non-analyst process

j-kanuch:
This PR refactors the Genomic file ingestion process to be more generic.  This implements the ingestion of the Biobank Return Manifest and other files into the ongoing (non-analyst) New Participant Workflow.


rdr_service/model/genomics.py
```diff
@@ -84,8 +84,7 @@ class GenomicSetMember(Base):
 
     validatedTime = Column("validated_time", DateTime, nullable=True)
 
-    sampleId = Column('sample_id', String(80),
-                      ForeignKey("biobank_stored_sample.biobank_stored_sample_id"), nullable=True)
+    sampleId = Column('sample_id', String(80), nullable=True)
```
wangy70, line 87: The model is changed, but I didn't find the db migration file in this PR. Does this change already applied in DB?

rdr_service/model/genomics.py
```diff
@@ -185,8 +184,7 @@ class GenomicGCValidationMetrics(Base):
 
     biobankId = Column('biobank_id', String(80), nullable=False)
     # Ingested Data
-    sampleId = Column('sample_id', String(80),
-                      ForeignKey("biobank_stored_sample.biobank_stored_sample_id"), nullable=True)
+    sampleId = Column('sample_id', String(80), nullable=True)
```
wangy70, line 187: Same as above

rdr_service/model/genomics.py
j-kanuch: Actually, I made the model change in my last PR but then it was decided that we hold on changing this.  I deleted the migration file as we're not needing the FK on this field. But I didn't go back and change this file. The DB doesn't have the foreign key, and shouldn't as of now, so this is really just fixing the change that I had in my last PR.

rdr_service/model/genomics.py
j-kanuch: Same as above too.

wangy70 (approved):

PR #1575, opened 2020-02-10
check enrollment status cron

Seeker1911:


rdr_service/offline/enrollment_check.py
```diff
@@ -0,0 +1,95 @@
...
+        if create_ticket:
+            jira_handler = JiraTicketHandler()
+            jira_handler.create_ticket('Participant Status offline job', str(results).strip('[]'), issue_type='Task',
+                                       board_id='ROC')
+        return False
+    else:
+        return True
+
+
+def get_last_30_days():
```
robabram: Why just 30 days?

Seeker1911: it runs once a month. The actual number is 31 days. Looks like I forgot to change the name of the function.

Seeker1911: renamed

robabram (approved):

PR #1581, opened 2020-02-13
[DA-1466] Implement Genomic Centers Manifest into Workflow

j-kanuch:
This PR adds the GC manifest workflow to the ongoing genomic pipeline. This "ingestion" is currently more akin to a "reconciliation" as no new data is being ingested from the GC manifest files; we are only checking existence of a sample_id and then writing the genomic_job_run_id to the genomic_set_member. A migration is included to add a column for `reconcileGCManifestJobRunId`. The other reconciliation columns were renamed to be more explicit.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -204,6 +218,23 @@ def _ingest_bb_return_manifest(self):
         except RuntimeError:
             return GenomicSubProcessResult.ERROR
 
+    def _ingest_gc_manifest(self, data):
+        """
+        Updates the GenomicSetMember with GC Manifest data (just manifest
```
Seeker1911:
```suggestion
        Updates the GenomicSetMember with GC Manifest data (just manifest)
```

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -133,11 +135,29 @@ def run_biobank_return_manifest_workflow(self):
...
                                             sub_folder=self.sub_folder_name)
         try:
-            logging.info('Running Biobank Return Manifest Workflow.')
             self.job_result = self.ingester.generate_file_queue_and_do_ingestion()
         except RuntimeError:
             self.job_result = GenomicSubProcessResult.ERROR
 
+    def run_genomic_centers_manifest_workflow(self):
+        """
+        Uses ingester to ingest Genomic Manifest files.
```
Seeker1911:
```suggestion
        Uses GenomicFileIngester to ingest Genomic Manifest files.
```

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -133,11 +135,29 @@ def run_biobank_return_manifest_workflow(self):
...
+        Currently leaves files in bucket.
```
Seeker1911: If this is likely to change, I'd fear this comment becoming stale and not getting updated.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Updated this with the most-recent commit.

rdr_service/genomic/genomic_job_controller.py
j-kanuch: Updated this with the most-recent commit.

rdr_service/genomic/genomic_job_controller.py
j-kanuch: I removed the comment since it isn't really necessary and can become stale.

Seeker1911 (approved):

PR #1594, opened 2020-02-20
BigQuery: Ensure we are getting all survey questions.

robabram:
Allow getting the config from a specific project id.
The `sp_get_code_module_items` stored proc needed another level to capture all questions.


rdr_service/alembic/versions/7ab9205d1bc6_function_fix.py
```diff
@@ -0,0 +1,95 @@
...
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    op.replace_sp(sp_get_code_module_items, replaces="a43f72b7c848.sp_get_code_module_items")
+
+def downgrade_rdr():
+    op.replace_sp(sp_get_code_module_items, replace_with="a43f72b7c848.sp_get_code_module_items")
```
Seeker1911, line 82: downgrade with the same migration ?

robabram: In the downgrade, the operative is `replace_with`, so it is replacing the current stored procedure with the previous stored procedure.

Seeker1911: 👍🏻

Seeker1911 (approved):

PR #1648, opened 2020-03-17
[ROC-366] Added filter for removing DV test participants

j-kanuch:
This PR adds a new column to biobank_dv_order: is_test_sample. The Biobank reconciliation reports were updated to filter out test participants using this field in the report SQL. The Missing Salivary report test was updated to test this filter worked correctly.


Seeker1911 (approved):
Perhaps I'm being picky but "test sample" might not be the best word to use. All samples are tests and here we mean it to reflect a fake sample. This could be confusing. Approved either way.

PR #1651, opened 2020-03-19
DA-1515 update gror not sure values

Seeker1911:
Refactored genomic consent to accept an "unsure" response. 
Updated tests to match.
Removed some of the legacy PyCharm tools.


wangy70:
Are you sure you want to submit all your IDE runConfigurations files?

Seeker1911:
The `.idea` directory is already shared in our project.
most of these are just deleting the old version of the tools we use.

wangy70:
It seems you remove all the content of those files. Why?

wangy70:
I think some of them are still in use, like Unittests__Partial.xml , I use it for unit test.

Seeker1911:
I had two unittests_partial. One that works and one that assumes the old directory structure. Are you sure you use that particular one ?

wangy70:
Yes, I double checked, I use Unittests__Partial.xml, and I believe Robert also use this one.

wangy70 (approved):

PR #1688, opened 2020-04-22
[ROC-447] HOTFIX: Implement new GROR module and answer codes.

j-kanuch:
Changed test-data json for gror consent questionnaire responses. Modified module name and answer codes for GROR.


j-kanuch:
Don't want to merge to devel.

PR #1700, opened 2020-04-28
[DA-1514] Genomic Alerts for Missing Data Files

j-kanuch:
This PR implements a Jira alerting system for the genomic data manifest workflow. Files listed in the data manifest but missing from the genotyping data bucket will be reported in a Jira ticket to the ROC board in the current sprint.


rdr_service/services/jira_utils.py
```diff
@@ -34,15 +37,34 @@ def __init__(self):
 
         self._connect_to_jira()
 
+    def set_jira_credentials_from_config(self):
+        """
+        Sets the Jira API username and password
+        based on whether set in environment or in config
+        :return:
+        """
```
robabram: I thought we talked about just putting the JIRA credentials in the App Configs instead of reading a separate file.

j-kanuch: Yes, we did. I was running into issues with getting the credentials during testing without putting them in the .current_config.json or config_dev.json. 
This would have exposed them to the repo since those files are still public.

robabram: The trick to that is to read them from the local environment variables before reading from the config.  So you would check `os.environ` to see if the JIRA credentials are there and if so use them for local testing.

j-kanuch: I've implemented the update to use the app configs.

wangy70 (approved):

PR #1710, opened 2020-05-04
[DA-1459] Endpoint for creating/updating individual biobank specimen

kenny-skaggs:
This sets up the API endpoint for creating and updating individual biobank specimen. There were some fields on the model that have slightly different names in the API documentation.

I also did some things that I'd like another opinion on with looking up existing items. Our primary key for the specimen is the combination of orderId and our own id field. The rlimsId is what we receive from the client as the unique id. So I use that to retrieve our id for existing objects.


kenny-skaggs:
I realize I should maybe have clarified a bit more. This doesn't complete the ticket, but is work-in-progress toward it. From my understanding I'll still need the migration endpoint for finishing out the ticket.

rdr_service/api/biobank_specimen_api.py
```diff
@@ -0,0 +1,36 @@
...
+from werkzeug.exceptions import BadRequest
+
+
+class BiobankSpecimenApi(UpdatableApi):
+    def __init__(self):
+        super().__init__(BiobankSpecimenDao(), get_returns_children=True)
+
+    @auth_required(HEALTHPRO)
+    def post(self, p_id):
+        return super().post(participant_id=p_id)
```
wangy70: Do we need POST here? It seems we only support GET and PUT in rdr_service/main.py for BiobankSpecimenApi.

rdr_service/api/biobank_specimen_api.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @auth_required(HEALTHPRO)
+    def put(self, *args, **kwargs):  # pylint: disable=unused-argument
+        resource = request.get_json(force=True)
+
+        for required_field in ['rlimsID', 'orderID', 'testcode', 'participantID']:
+            if required_field not in resource:
+                raise BadRequest("Missing field: %s" % required_field)
+
+        participant_id = from_client_participant_id(resource['participantID'])
```
wangy70: I just double confirmed one thing, for the participant ID in the payload, it's biobank's participant id, in our system, we call it biobankID, it's one on one mapping with our participantID.
Here is the RLIMS field mapping: https://docs.google.com/spreadsheets/d/1oBwWN5It5SESITmuIVrL-YjzorVDRyD8/edit#gid=804532217

And for this biobankID, in RDR it's 9 digits, but outside RDR, we add a prefix for it. 
Here is the detail: https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/model/config_utils.py#L9

Sorry that I jus realize this when I read your code.

rdr_service/api/biobank_specimen_api.py
kenny-skaggs: You're right, it looks like the design document only calls for PUT as well. I'll remove it.

kenny-skaggs: Should I also remove the GET method from this and main.py? The documentation only specifies PUT.

rdr_service/dao/biobank_specimen_dao.py
```diff
@@ -0,0 +1,98 @@
...
+
+        return result
+
+    #pylint: disable=unused-argument
+    def from_client_json(self, resource, id_=None, expected_version=None, participant_id=None, client_id=None):
+        order = BiobankSpecimen(rlimsId=resource['rlimsID'], orderId=resource['orderID'], testCode=resource['testcode'],
+                                biobankId=from_client_biobank_id(resource['participantID']))
+
+        if not self.exists(resource):
+            order.created = clock.CLOCK.now()
```
wangy70: We have a model event listener to help you handle the `created` and `modified` fields. 
Here is an example: 
https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/model/biobank_order.py#L302

Your code is also good, just for your reference :)

rdr_service/dao/biobank_specimen_dao.py
```diff
@@ -0,0 +1,98 @@
...
+            for status_field_name, parser in [('status', None),
+                                              ('freezeThawCount', None),
+                                              ('location', None),
+                                              ('quantity', None),
+                                              ('quantityUnits', None),
+                                              ('deviations', None),
+                                              ('processingCompleteDate', parse_date)]:
+                self.map_optional_json_field_to_object(resource['status'], order, status_field_name, parser=parser)
+
+        order.version = 1
```
wangy70, line 70: Do we have `version` for this model?

rdr_service/dao/biobank_specimen_dao.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if 'status' in resource:
+            for status_field_name, parser in [('status', None),
+                                              ('freezeThawCount', None),
+                                              ('location', None),
+                                              ('quantity', None),
+                                              ('quantityUnits', None),
+                                              ('deviations', None),
+                                              ('processingCompleteDate', parse_date)]:
+                self.map_optional_json_field_to_object(resource['status'], order, status_field_name, parser=parser)
```
wangy70, line 68: It seems you only handled part of the payload. In the swagger API description, the full payload is something like below:
```
{
  "rlimsID": "string",
  "participantID": "string",
  "orderID": "string",
  "testcode": "string",
  "repositoryID": "string",
  "studyID": "string",
  "cohortID": "string",
  "sampleType": "string",
  "status": {
    "status": "string",
    "freezeThawCount": 0,
    "location": "string",
    "quantity": "string",
    "quantityUnits": "string",
    "processingCompleteDate": "2020-05-05T19:40:28.534Z",
    "deviations": "string"
  },
  "disposalStatus": {
    "reason": "string",
    "disposalDate": "2020-05-05T19:40:28.534Z"
  },
  "collectionDate": "2020-05-05T19:40:28.534Z",
  "confirmationDate": "2020-05-05T19:40:28.534Z",
  "attributes": [
    {
      "name": "string",
      "value": "string"
    }
  ],
  "aliquots": [
    {
      "rlimsID": "string",
      "sampleType": "string",
      "status": {
        "status": "string",
        "freezeThawCount": 0,
        "location": "string",
        "quantity": "string",
        "quantityUnits": "string",
        "processingCompleteDate": "2020-05-05T19:40:28.534Z",
        "deviations": "string"
      },
      "disposalStatus": {
        "reason": "string",
        "disposalDate": "2020-05-05T19:40:28.534Z"
      },
      "childPlanService": "string",
      "initialTreatment": "string",
      "containerTypeID": "string",
      "datasets": [
        {
          "rlimsID": "string",
          "name": "string",
          "status": "string",
          "datasetItems": [
            {
              "paramID": "string",
              "displayValue": "string",
              "displayUnits": "string"
            }
          ]
        }
      ],
      "aliquots": [
        null
      ]
    }
  ]
}
```

tests/api_tests/test_biobank_specimen_api.py
```diff
@@ -0,0 +1,162 @@
...
+            'orderID': self.bio_order.biobankOrderId,
+            'participantID': config_utils.to_client_biobank_id(self.participant.biobankId),
+            'testcode': 'test 1234567'
+        }
+        rlims_id = payload['rlimsID']
+        initial_result = self.send_put(f"Biobank/specimens/{rlims_id}", request_data=payload)
+
+        new_payload = payload
+        new_payload['testcode'] = 'updated testcode'
+        self.send_put(f"Biobank/specimens/{rlims_id}", request_data=new_payload, headers={"if-match": 'W/"1"'})
```
wangy70: We don't need "if-match" header for this API.

wangy70 (commented):
I like your code style, the code looks very comfortable. 
There are some requirement things need to re-check, please see my comments for detail.

wangy70:
BTW, the combined primary key id+orderID is not as design, I'm not sure what's the purpose for it. If you think it's unnecessary, I'm good to just use id :)

rdr_service/dao/biobank_specimen_dao.py
kenny-skaggs: Good idea, I like the event listener approach. I'll make the change now.

rdr_service/dao/biobank_specimen_dao.py
kenny-skaggs: We don't. I'll remove this.

kenny-skaggs: So I'm finding that removing this causes issues in BaseApi's _make_response method https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/api/base_api.py#L301. That method attempts to access a version field on the object. Is it better to leave it there, or to set up a system of configuring the BaseApi to ignore version for specimen endpoints?

tests/api_tests/test_biobank_specimen_api.py
kenny-skaggs: Ah, sorry I missed this. Fixing now.

rdr_service/dao/biobank_specimen_dao.py
kenny-skaggs: Good point. I understood them to be part of the implementation for the other endpoints. For example, I thought implementation of handling disposalStatus would be part of the 4th endpoint listed in the tech design document.

kenny-skaggs:
I'm a fan of just using the ID. It'll at least make the tests a little cleaner.

wangy70 (approved):
Approve this one to avoid the PR getting too big. A new PR will track other issues.

PR #1765, opened 2020-05-29
[DA-1440] Drop deprecated 'old_resource' column from PM table

pegbertsch:
Remove remaining references to the old column

Alembic also picked up an unrelated mod involving a previously nullable column in another table, which was deemed safe to commit


kenny-skaggs (approved):

pegbertsch:
Leaving this PR open until next week, need to discuss w/team and consider timing for merge/inclusion in a release since this database change can cause downtime for end users.

robabram (approved):

pegbertsch:
Merging this PR.  Profiling manual deploy time on test wasn't too helpful, not enough records in the PM table (vs. count on Prod).   Will still need to do final deploy to prod in off hours in case it takes additional time and blocks access to the table.

PR #1766, opened 2020-05-29
Adding codeowners file

kenny-skaggs:
According to https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners I _think_ this will automatically select us all as reviewers for new PRs. I don't know if those extra clicks were getting to anyone else as much, but I thought this would be fun and convenient.


rdr_service/offline/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
```
robabram: Why in put the file in 'rdr_service/offline'?

kenny-skaggs: Oops, meant to put it at the root. Moving it now.

kenny-skaggs: Ok, should be good now.

robabram (approved):

PR #1770, opened 2020-06-02
[DA-1549] Genomic Outreach API; GEM Report States

j-kanuch:
This PR implements the Genomic Outreach API to serve PTSC with GEM report state resources. This includes look-ups by date-range and participant_id. This PR also includes refactoring of the GEM A1, A2, and A3 manifests to use the new FSM created for the genomic system instead of relying on genomic set members' job_run_ids to determine report state.

### DIFF
--- a/rdr_service/api/genomic_api.py
+++ b/rdr_service/api/genomic_api.py
@@ -1,5 +1,9 @@
+from dateutil import parser
+
+from flask import request
 from werkzeug.exceptions import NotFound, BadRequest
 
+from rdr_service import clock
 from rdr_service.api.base_api import BaseApi
 from rdr_service.api_util import GEM, RDR_AND_PTC
 from rdr_service.app_util import auth_required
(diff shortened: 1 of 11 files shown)


rdr_service/dao/genomics_dao.py
```diff
@@ -411,6 +412,60 @@ def get_null_field_members(self, field):
...
+                    GenomicWorkflowState.A2
+                )) &
+                (
+                    (ParticipantSummary.consentForGenomicsROR != QuestionnaireStatus.SUBMITTED) |
+                    (ParticipantSummary.consentForStudyEnrollment != QuestionnaireStatus.SUBMITTED)
+                ) &
+                (
+                    (ParticipantSummary.consentForGenomicsRORAuthored > _date) |
+                    (ParticipantSummary.consentForStudyEnrollmentAuthored > _date)
+                )
```
kenny-skaggs, line 440: Should there be a logical link between checking the status and authored time for that status? This seems like it could find a participant that has a no consent status for GROR authored before the date, but has made a change to their 'study for enrollment' status after the date.

Same sort of concept down in the reconsented method.

rdr_service/api/genomic_api.py
```diff
@@ -29,3 +29,20 @@ def get(self, mode=None, p_id=None):
...
+    def get(self, mode=None):
+        if mode not in ('GEM', 'RHP'):
+            raise BadRequest("GenomicOutreach Mode required to be \"GEM\" or \"RHP\".")
+
+        if mode == "GEM":
+            report_states_result = self.dao.date_lookup(mode)
+
+            return self._make_response(report_states_result)
+
+        return BadRequest
```
robabram: Ultra compact: "return self._make_response(self.dao.date_lookup(mode)) if mode == "GEM" else BadRequest"

robabram (approved):

rdr_service/dao/genomics_dao.py
j-kanuch: Yes, I see what you mean. It should be (pseudo-code) `(ror_consent != 1 AND ror_authored > date) OR (general_consent != 1 AND consent_authored > date)`.

rdr_service/api/genomic_api.py
j-kanuch: Yes, I'll compact this, thanks.

j-kanuch: I just realized I had moved this into another method.

rdr_service/dao/genomics_dao.py
kenny-skaggs: Looks good 👍

PR #1773, opened 2020-06-03
[DA-1597] Get the correct answer for each participant consent

robabram:


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -582,9 +577,12 @@ def get_module_answers(ro_dao, module, p_id):
             if not results:
                 return None
 
+            # Match the specific questionnaire response id otherwise return answers for the most recent response.
             for row in results:
-                data = ro_dao.to_dict(row, result_proxy=results)
+                if qr_id and row.questionnaire_response_id != qr_id:
+                    continue
 
+                data = ro_dao.to_dict(row, result_proxy=results)
```
kenny-skaggs, line 585: This looks for the first row to be the latest response, and the SQL orders by created time. Is it possible that PTSC could send us responses out of order in some way (like collecting a bunch of them and sending them all at once and not in the order they received them). I'm wondering if there's some strange way that the order of the created time doesn't match with the order of authored time. That scenario is probably really out there, but it came to mind.

robabram: PTSC can delay sending us survey responses, but as I understand it they never send us surveys out of order.  So I believe it is safe to order by `created`.  Additionally we need to use `created` which is in an index and `authored` is not.

kenny-skaggs: 👍 I'm all for the indexed column then

kenny-skaggs (approved):

PR #1777, opened 2020-06-05
[DA-1596] Cohort 2 AW0 Manifest

j-kanuch:
This PR implements the AW0 manifest for Cohort 2. The new consent question code is outstanding and will need to be added when RDR receives this.
Some additional refactoring was done to split the Cohort 3 and Cohort 2 processes.

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -1072,7 +1072,7 @@ class GenomicBiobankSamplesCoupler:
         :param: from_date
         :return: list of tuples (bid, pid, biobank_identifier.value, collected_site_id)
         """
-        # TODO: add Genomic RoR Consent when that Code is added
+
         _new_samples_sql = """
         SELECT DISTINCT
           ss.biobank_id,
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -41,7 +41,7 @@ from rdr_service.participant_enums import (
     GenomicSetMemberStatus,
     SuspensionStatus,
     GenomicWorkflowState,
-)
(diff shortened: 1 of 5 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -1095,22 +1107,99 @@ def _get_new_biobank_samples(self, from_date):
...
+        :return:
+        """
+
+        _c2_samples_sql = """
+                SELECT DISTINCT
+                  ss.biobank_id,
+                  p.participant_id,
+                  o.biobank_order_id,
+                  o.collected_site_id,
+                  ss.biobank_stored_sample_id,
```
robabram, line 1158: Why did you go with DISTINCT instead of using these fields in a GROUP BY clause?  How well does this query perform?

j-kanuch: In this case, I only needed unique rows, I wasn't aggregating the results, so I decided to use DISTINCT. For this query, DISTINCT slightly outperforms GROUP BY. I did some comparison testing with the large dataset on PROD and DISTINCT seemed to beat GROUP BY on average between 15-20%.
It took around 4-5 seconds for DISTINCT and 6-8 seconds for  GROUP BY. The daily run will have less data than the thousands of records currently on PROD, as the date-limit will be only 1-day, so the difference may be trivial in the long run.

robabram: Sounds good, thanks for the info.

robabram (approved):

rdr_service/genomic/genomic_job_components.py
j-kanuch: No problem, thanks for keeping me honest!

PR #1780, opened 2020-06-09
[ROC-512] Deactivating previous street address 2

kenny-skaggs:
We're clearing line 2 of the street address when receiving an updated line 1. But the previous response for street address line 2 remains active. Previous questionnaire responses are deactivated when a new response to the same question is received. This sets the code up to deactivate previous responses for line 2 of the street address if we receive a response for line 1 or 2.

The solution I went with is to load the line 2 question if we're loading the line 1 question too. That way, later on in the code when it's loading current answers, the current answer for line 2 is loaded as well and gets an end date set on it.

### DIFF
--- a/tests/api_tests/test_participant_api.py
+++ b/tests/api_tests/test_participant_api.py
@@ -8,7 +8,10 @@ from rdr_service.code_constants import PPI_SYSTEM, RACE_WHITE_CODE, PMI_SKIP_COD
 from rdr_service.concepts import Concept
 from rdr_service.dao.hpo_dao import HPODao
 from rdr_service.dao.biobank_order_dao import BiobankOrderDao
+from rdr_service.model.code import Code
 from rdr_service.model.hpo import HPO
+from rdr_service.model.questionnaire import QuestionnaireQuestion
+from rdr_service.model.questionnaire_response import QuestionnaireResponseAnswer
 from rdr_service.participant_enums import (
     OrganizationType,
     SuspensionStatus,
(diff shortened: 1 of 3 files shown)


tests/api_tests/test_participant_api.py
```diff
@@ -525,6 +528,16 @@ def test_street_address_two_clears_on_no_answer(self):
         participant_summary = self.send_get("Participant/%s/Summary" % participant_id)
         self.assertNotIn("streetAddress2", participant_summary)
 
+        # Make sure the street address 2 answer is set inactive too
+        street_address_2_active_answer = self.session.query(QuestionnaireResponseAnswer)\
+                                             .join(QuestionnaireQuestion)\
+                                             .join(Code, Code.codeId == QuestionnaireQuestion.codeId)\
+                                             .filter(Code.value == 'PIIAddress_StreetAddress2',
+                                                     QuestionnaireResponseAnswer.endTime.is_(None))\
+                                             .one_or_none()
```
pegbertsch, line 537: If you have added _imply_street_address_2_from_street_address_1 logic and submitted any response with a street_address_1 in the test data, is matching one_or_none() on the query too broad now?   Should it always match exactly one active street_address_2 answer now?

kenny-skaggs: The implication logic is to load any previous street address 2 responses and deactivate them if given a response to street address 1. So at this point in the test I'm expecting to not have any active street address 2 responses in the database. I think I could narrow it down by using the query's `exists` method and asserting that is false if that would be clearer.

pegbertsch: Exiting test case logic is fine.  I think I misunderstood what that "imply" step does in appending the extra street address 2 code id (i.e., won't matter/won't find anything if it was never answered before vs. me thinking it could cause a "None" answer to be implied)

pegbertsch (approved):

PR #1781, opened 2020-06-09
[ROC-520] mayolinkapi_state_length_limit

wangy70:


rdr_service/dao/dv_order_dao.py
```diff
@@ -100,7 +100,7 @@ def _filter_order_fields(self, resource, pid):
                     "address1": summary.streetAddress,
                     "address2": summary.streetAddress2,
                     "city": summary.city,
-                    "state": code_dict["state"],
+                    "state": code_dict["state"][-2:] if code_dict["state"] else code_dict["state"],
```
kenny-skaggs: `format_json_code` sets state as the PIIState code if there is one, but it could also be set as "UNSET" or "UNMAPPED". Should there be any logic for those scenarios, like maybe not sending the state field if it's not set or can't be mapped?

wangy70: Good catch! I will communicate with MayoLink team about these two  "UNSET" or "UNMAPPED". Maybe we can just send empty string to them.

wangy70: @kenny-skaggs  Done

kenny-skaggs: 👍

kenny-skaggs (approved):

PR #1785, opened 2020-06-12
[DA-1570] Zipping VA files in consent sync cron job

kenny-skaggs:
These changes are to zip VA consent files in the consent sync cron job. The tool had that functionality already, so I moved that over to the cron script file and have them both using it now. There's also a bunch of refactoring to have them share more functionality. And this schedules the VA consent sync for half an hour after the regular sync.

One big change of note is that I've removed the use of CloudTasks from the cron script. I needed to create the zip files after downloading the consent files and there didn't seem to be a way of having that run after the tasks were complete.


rdr_service/offline/sync_consent_files.py
```diff
@@ -15,13 +15,14 @@
...
 
 SOURCE_BUCKET = {
     "vibrent": "ptc-uploads-all-of-us-rdr-prod",
     "careevolution": "ce-uploads-all-of-us-rdr-prod"
 }
 DEFAULT_GOOGLE_GROUP = "no-site-assigned"
-TEMP_CONSENTS_PATH = "./temp_consents"
+TEMP_CONSENTS_PATH = "/tmp/temp_consents"
 
 ParticipantData = collections.namedtuple("ParticipantData", ("participant_id", "origin_id", "google_group", "org_id"))
```
robabram: We probably should not hard code the temp directory path here.  You can use the tempfile.gettempdir() function to get the current temp directory.  https://docs.python.org/3.7/library/tempfile.html#tempfile.gettempdir

rdr_service/offline/sync_consent_files.py
```diff
@@ -72,6 +73,7 @@ def copy_file(source, destination, participant_id, dry_run=True, zip_files=False
 
 def archive_and_upload_consents(dry_run=True):
     logging.info("zipping and uploading consent files...")
+    storage_provider = GoogleCloudStorageProvider()
     for bucket_dir in _directories_in(TEMP_CONSENTS_PATH):
         for org_dir in _directories_in(bucket_dir):
```
robabram: Can you use the `get_storage_provider()` function in `rdr_service/storage.py` here, as this method can correctly get either the local or cloud storage provider depending on where the code is being run.

kenny-skaggs: I may be misunderstanding, but I don't think I would want it to use the LocalStorageProvider whether we're running the tool on our machines or if it's running in the cloud as part of the cron job. In either case I want to access the GoogleCloudStorageProvider's implementation of `upload_from_file`.

rdr_service/offline/sync_consent_files.py
kenny-skaggs: Updating now. Thanks, I didn't know that existed.

rdr_service/offline/sync_consent_files.py
robabram: Ok.  I just wanted to make sure you were aware of the `get_storage_provider()` function.

robabram (approved):

PR #1788, opened 2020-06-16
[ROC-527] Creating tool for pulling cope survey result counts

kenny-skaggs:
CE is asking for answer counts for a specific set of question codes. They'd like to see the total number of answers for each of the questions and the counts for each option on that question. There's an example on the ticket of the type of output they'd like to see that could clarify what they're looking for.

I didn't follow their output very closely, but I've emailed Kyle the results of the tool and he's happy with the output structure.

We'll need to use this at the end of the month to pull the June COPE survey data, and likely again at the end of July for that month's survey.

### DIFF
--- a/rdr_service/tools/tool_libs/cope_answers.py
+++ b/rdr_service/tools/tool_libs/cope_answers.py
@@ -6,12 +6,10 @@
 import argparse
 import json
 import logging
-import random
 import sys
 
 from rdr_service.dao import database_factory
-from rdr_service.services.gcp_utils import gcp_format_sql_instance, gcp_make_auth_header
-from rdr_service.services.system_utils import make_api_request, setup_logging, setup_i18n
+from rdr_service.services.system_utils import setup_logging, setup_i18n
 from rdr_service.tools.tool_libs import GCPProcessContext
 
 _logger = logging.getLogger("rdr_logger")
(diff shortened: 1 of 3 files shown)


rdr_service/tools/tool_libs/cope_answers.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        if resp_code != 200:
+            _logger.error(resp_data)
+            _logger.error("failed to retrieve config, aborting.")
+            return 1
+
+        passwd = resp_data["rdr_db_password"]
+        if not passwd:
+            _logger.error("failed to retrieve database user password from config.")
+            return 1
```
robabram: This can be simplified to `passwd = self.gcp_env.get_app_db_config()['rdr_db_password']`.  I also don't see where passwd is used.

rdr_service/tools/tool_libs/cope_answers.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # connect a sql proxy to the current project
+        _logger.info("starting google sql proxy...")
+        port = random.randint(10000, 65535)
+        instances = gcp_format_sql_instance(self.gcp_env.project, port=port)
```
robabram: These are unnecessary, the `activate_sql_proxy` will do the exact same steps if the `instance` and `port` arguments are not passed to the function.

rdr_service/tools/tool_libs/cope_answers.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @staticmethod
+    def _new_question_tracker():
+        return {
+            'total_answers': 0,
+            'answers': {}
+        }
+
+    def run(self):
+        self._setup_database_connection()
```
robabram: The `self._setup_database_connection()` function returns a value if there is an error.  Its not checked here and the code continues executing when there is an error.

rdr_service/tools/tool_libs/cope_answers.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # connect a sql proxy to the current project
+        _logger.info("starting google sql proxy...")
+        port = random.randint(10000, 65535)
+        instances = gcp_format_sql_instance(self.gcp_env.project, port=port)
+        proxy_pid = self.gcp_env.activate_sql_proxy(instance=instances, port=port)
+        if not proxy_pid:
+            _logger.error("activating google sql proxy failed.")
+            return 1
+
```
robabram: There is no return value here for a successful connection.

tests/helpers/unittest_base.py
```diff
@@ -220,6 +223,11 @@ def unique_participant_biobank_id(self):
         self._next_unique_participant_biobank_id += 1
         return next_biobank_id
 
+    def unique_questionnaire_response_id(self):
```
robabram, line 296: There is the `QuestionnaireTestMixin` class that encapsulates Questionnaire functions for unittests, why did you not put all these Questionnaire functions in that class?

kenny-skaggs:
@robabram, thanks for pointing out the issues with the database connection setup. I've cleaned it up with the things you've pointed out.

I wasn't aware of the QuestionnaireTestMixin. I can move the questionnaire methods there, but I've been thinking that their functionality fits better with some of the other sort of factory or helper methods that create objects in the database for getting the test data setup. I've been imagining that I would move all those functions into a helper class that the UnitTest classes would use rather than have as part of their definition. I could do that now too, but I didn't want to flood a PR with that refactor and bury changes I made for a ticket. Let me know what you think would be best.

robabram:
I think it would be fine for now if you moved those functions to the QuestionnaireTestMixin and did a refactor sometime in the future.

kenny-skaggs:
Ok, sounds good. Should be ready @robabram

robabram (approved):

PR #1789, opened 2020-06-16
[DA-1598] support address for quest biobank order

wangy70:


rdr_service/dao/biobank_order_dao.py
```diff
@@ -617,6 +656,20 @@ def _make_mayolink_order(self, participant_id, resource):
...
+    def _add_quest_site_address(cls, order, resource):
+        if resource.collected_info and resource.collected_info.address:
+            address = BiobankQuestOrderSiteAddress(city=resource.collected_info.address.city,
+                                                   state=resource.collected_info.address.state,
+                                                   zipCode=resource.collected_info.address.postalCode,
+                                                   address1=resource.collected_info.address.line[0]
+                                                   if resource.collected_info.address.line else None,
+                                                   address2=resource.collected_info.address.line[1]
+                                                   if len(resource.collected_info.address.line) > 1 else None)
+            order.questSiteAddress = address
```
kenny-skaggs, line 669: Would this create a new address record for each order? If so, should we potentially save some storage space by changing the relationship to be many orders to one address (rather than have a column for the order id on the address) and perform a search to see if we are already aware of a site's address?

wangy70: This site address is different from the Site table we already have. For quest biobank orders, we already have the relationship between biobank order table and site table, but this site(in site table) is just a place holder. Quest has hundreds of real sites for the collection work, we don't have those site information, so they just pass the address when they create the biobank orders. Actually we don't have an object oriented "site" for this scenario, when they update order, the can update this address.

kenny-skaggs: I think I follow what you mean, but I'm still wondering if we could create a single record in the BiobankQuestOrderSiteAddress table for two orders that share the same address. So if two orders have the same site address I think right now we'd have two records in the BiobankQuestOrderSiteAddress table, one for each of them. In that case the only difference between the records would be the Quest order they're tied to. Would multiple orders come from the same site?

wangy70: We have limited information about the address now. Currently we get the address from the payload as string, there is no site id or site name bind with the order, so we don't know the mapping between order and site, we only get the address for state, city, zip code and street as string, so for this scenario I prefer to treat this address as the attributes of an order. In this way, when an order create/update request comes, we don't need to search all the address fields from the existing address records, it's easier for the code logic. If you concern about the storage space, I think the difference maybe only 1 or 2 MB for a year's data.

Anyway, I think both ways are good for this ticket, they are only different designs from different perspective.

kenny-skaggs: Ok, sounds good to me.

kenny-skaggs (approved):

PR #1799, opened 2020-06-19
[DA-1544] Add building resource records to BigQuery tasks.

robabram:
Restructure BigQuery tasks to run under Resource service.

### DIFF
--- a/rdr_service/resource/generators/_base.py
+++ b/rdr_service/resource/generators/_base.py
@@ -4,10 +4,12 @@
 #
 import hashlib
 import json
+import logging
 import re
 
 from marshmallow_jsonschema import JSONSchema
 from sqlalchemy import and_
+from sqlalchemy.exc import IntegrityError
 
 from rdr_service.dao.resource_dao import ResourceDataDao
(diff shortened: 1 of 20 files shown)


rdr_service/resource/generators/_base.py
```diff
@@ -61,7 +63,11 @@ def _get_or_create_type_record(self, dao, schema_meta):
             rec.typeUID = type_uid
 
             session.add(rec)
-            session.commit()
+            try:
+                session.commit()
+            except IntegrityError:
+                # Record already exists, parallel tasks can cause this.  We should not see these often.
+                logging.warning(f'Resource type record already exists for {rec.typeName}.')
```
kenny-skaggs: What happens to the resource when the type already exists? Should this attempt to return it?

robabram: This is to handle an edge case where the Type record does not exists but two parallel processes are trying to create it at the same time, which happened during testing on Sandbox.  This does need fixing to try retrieving the record in the Try block.

kenny-skaggs (approved):

PR #1803, opened 2020-06-23
Biobank Support: Updating biobank specimen migration endpoint response

kenny-skaggs:
Lary and I talked a bit in the biobank channel and decided to make the migration endpoint's response a little easier to parse and work with. This essentially takes the summary and error strings and converts them to JSON objects. I've also added logging for the endpoint so we can see error details for individual specimen and help track down problems a little quicker.

### DIFF
--- a/rdr_service/api/biobank_specimen_api.py
+++ b/rdr_service/api/biobank_specimen_api.py
@@ -1,4 +1,5 @@
 from flask import request
+import logging
 
 from rdr_service.api.base_api import UpdatableApi, log_api_request
 from rdr_service.api_util import BIOBANK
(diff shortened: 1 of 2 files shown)


pegbertsch (commented):
Is there any additional communication of this change that has to occur before it can be released, since it's changing response content?  Or doc updates?

kenny-skaggs:
The response wasn't originally documented so this wouldn't be changing it away from anything anyone's relying on. I'll go ahead and update the technical design with information for the new response so we have it recorded somewhere.

pegbertsch (approved):

PR #1806, opened 2020-06-25
[DA-1623] Creating date range intersection utility class

kenny-skaggs:
These changes are to help with changing how we find enrollment status. BigQuery calculates the participant's enrollment status by looking at the current state of their data. To make sure they stay keep the Core enrollment status if they've ever had it, the BigQuery calculation needs to determine if they've ever had it. This creates a new class to help with that calculation.

The DateCollection class receives start and stop dates in order to set up date ranges that signify when the consents are active. My plan is to track all the date ranges that a particular criteria (ehr consent, consent for study, physical measurements, ...) and get the intersection of them all to find where a participant had everything they needed to be a Core participant.

### DIFF
--- a/rdr_service/app_util.py
+++ b/rdr_service/app_util.py
@@ -334,3 +334,94 @@ def datetime_as_naive_utc(value):
 def is_care_evo_and_not_prod():
     return GAE_PROJECT != "all-of-us-rdr-prod" and get_account_origin_id() == "careevolution"
 
+
+class DateRange:
+    def __init__(self, start, end):
+        self.start = start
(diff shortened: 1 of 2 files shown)


robabram:
Kenny, I see where you are going with this but I have a concern about rebuild performance.  We will want to make sure we only call this for participants where it is needed and not for all participants.

robabram (approved):

kenny-skaggs:
@robabram, agreed. I'll set it up so that we only check the date overlaps if there's a possibility of it.

PR #1813, opened 2020-06-26
[DA-1595] Collection Tube ID

j-kanuch:
This PR refactors how `sampleId` is used in the genomic system. Sample ID downstream of the AW0 manifest represents a child of the original sample ID. The original sample is being represented by `collectionTubeId`. This is ingested in the AW1 manifest. Additionally, some lookups needed to be changed, as well as test data and the test filenames to distinguish `sampleId` from `collectionTubeId`.


rdr_service/dao/genomics_dao.py
```diff
@@ -342,6 +343,21 @@ def get_member_from_sample_id(self, sample_id, genome_type):
...
+        Needs a genome type
+        :param genome_type: aou_wgs, aou_array, aou_cvl
+        :param tube_id:
+        :return: a GenomicSetMember object
+        """
+        with self.session() as session:
+            member = session.query(GenomicSetMember).filter(
+                GenomicSetMember.collectionTubeId == tube_id,
+                GenomicSetMember.genomeType == genome_type
+            ).first()
```
pegbertsch, line 358: Is this always a "none or 1 result" expectation on the query?  Any reason to add a defensive check for multiple results?

j-kanuch: It should only ever be 1 or none. SqlAlchemy .first() doesn't raise an exception if None. SqlAlchemy .one() will raise an exception if no results are found, which I wasn't handling up the call stack. That may be something to add in a future PR.

pegbertsch (approved):

PR #1814, opened 2020-06-26
[DA-1623] Changes to enrollment status calculation

kenny-skaggs:
The goal of these changes is to change how we find a participant's enrollment status in two ways. The first way is that cohort 3 participants are required to have completed GROR consent. The second way is that if a participant ever was a Core participant, they shouldn't lose that status for any reason.

Three main areas had to be updated for this:
1. RDR's participant summary updates when we receive questionnaire responses. I was able to get a check there to see if they already have Core status. If they do then we don't change anything. I'll be working on a SQL query to retroactively make sure no participants have lost core status that shouldn't have.
2. The BigQuery code for calculating a participant's enrollment status.
3. And Resource's ParticipantGenerator that will be replacing the BigQuery code.

The ParticipantGenerator and BigQuery code are mostly the same (minus some naming in the data dictionaries). So those changes and their tests will be very similar if not duplicated.

Other miscellaneous changes:
- I've updated the DateCollection utility a bit to account for the possibility that we could get multiple YES responses for consents in a row.


wangy70:
I just reviewed `rdr_service/dao/participant_summary_dao.py`, it looks good. I'm not familiar with the BigQuery part, I think @robabram is the right person for that part.

rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -419,77 +430,164 @@ def _calculate_enrollment_status(self, p_id, ro_session, ro_summary):
...
+
         # check physical measurements
+        physical_measurements_date = datetime.datetime.min
         if 'pm' in ro_summary:
             for pm in ro_summary['pm']:
                 if pm['pm_status_id'] == int(PhysicalMeasurementsStatus.COMPLETED) or \
-                    (pm['pm_finalized'] and pm['pm_status_id'] != int(PhysicalMeasurementsStatus.CANCELLED)):
+                        (pm['pm_finalized'] and pm['pm_status_id'] != int(PhysicalMeasurementsStatus.CANCELLED)):
                     pm_complete = True
+                    physical_measurements_date = min(physical_measurements_date, pm['pm_finalized'])
```
pegbertsch, line 479: Per offline conversation:  Seems this will never result in an actual date from the participant history, physical_measurements_date will always retain the original datetime.datetime.min assigned value from line 473?   Is that the desired logic here?

kenny-skaggs: Nope, updating it now to find the earliest physical measurements time.

pegbertsch (approved):
Approved.

PR #1818, opened 2020-07-01
[DA-1546] Implement a subset of the Resource APIs

robabram:
I also consolidated a couple of duplicate functions from the 'main.py' files into 'services/flask.py'.


rdr_service/api/resource_api.py
```diff
@@ -0,0 +1,290 @@
...
+            for _id in [resource.pk_id, resource.pk_alt_id]:
+                query = session.query(ResourceData.id, ResourceData.created, ResourceData.modified,
+                                      ResourceData.resource)
+                if isinstance(_id, int):
+                    query = query.filter(ResourceData.resourcePKID == _id)
+                elif isinstance(_id, str):
+                    query = query.filter(ResourceData.resourcePKAltID == _id)
+                else:
+                    raise ValueError(f'Invalid resource id: {_id}.')
+                # sql = self.dao.query_to_text(query)
```
kenny-skaggs, line 157: Was this left over from testing? There are a few more `query_to_text` calls throughout the changes.

rdr_service/resource/generators/code.py
```diff
@@ -29,7 +28,6 @@ def make_resource(self, _pk):
             return generators.ResourceRecordSet(schemas.CodeSchema, data)
 
 
-@task_auth_required
```
kenny-skaggs, line 32: Is this an endpoint? If so, is it protected another way?

rdr_service/api/resource_api.py
robabram: I like to leave those commented out lines there for testing in the future.

rdr_service/resource/generators/code.py
robabram: This is showing the line is being removed from the file.

kenny-skaggs: Right, sorry I wasn't very clear. I was referring to the `rebuild_codebook_resources_task` method below as the potential endpoint. By removing the task_auth_required decorator, does that expose it a potentially unwanted way?

robabram: If you follow the usages for the `rebuild_codebook_resources_task` function you'll see that those functions have either the task API or cron API decorators.  Really, I think I copied that decorator there and accidentally let it slip into a commit.

kenny-skaggs: Makes sense 👍

kenny-skaggs (approved):

PR #1822, opened 2020-07-02
Fixing exception handling in biobank migration endpoint

kenny-skaggs:
I had built out some error handling in the biobank migration endpoint but recently found that there was an issue with how it works. My goal was to allow for errors on individual specimen but continue importing what the API could from a request (rather than crashing and taking no specimen that would have been valid). To account for any errors that could come up when saving things to the database I now commit the session at the end of processing each specimen and roll it back if any exceptions occur.


pegbertsch (commented):
Reviewed and found no issues, but would prefer 2nd reviewer to sign off on introduction of interim commits/rollbacks.

j-kanuch (approved):

PR #1823, opened 2020-07-02
[DA-1646] Add IGNORE state to lookups and Genomic utility tool

j-kanuch:
This PR implements a new genomic IGNORE state. The queries were updated to exclude `genomic_set_member` records that have the IGNORE state. Unit tests were updated. Other states were added to unit tests since the state was added to the lookup queries. The genomic utility tool was also updated to allow for setting the state to IGNORE for specified `genomic_set_member` records.


pegbertsch (approved):
My knowledge of the business logic for this workflow is still limited, but saw no issues.

PR #1825, opened 2020-07-02
[ROC-553] Fix PPI check tool

robabram:
Fix code that downloads CSV files from Google Docs.  Switched to using 'urllib3' and added a fallback to call 'curl'.

### DIFF
--- a/.idea/runConfigurations/Client__PPI_Check.xml
+++ b/.idea/runConfigurations/Client__PPI_Check.xml
@@ -13,7 +13,7 @@
     <option name="ADD_SOURCE_ROOTS" value="true" />
     <EXTENSION ID="PythonCoverageRunConfigurationExtension" runner="coverage.py" />
     <option name="SCRIPT_NAME" value="client" />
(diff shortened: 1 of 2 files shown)


pegbertsch (commented):
Is there a root cause for what changed on google side that required these changes?   

Are there any security considerations given the CERT_NONE (urllib3) and --insecure (curl) options used?

pegbertsch (approved):
See comment on security question given the nature of the download options (may be non-issue), but otherwise approved.

robabram:
It appears that Google cleaned up or enabled stricter validation URL arguments for Google docs.  There were a couple of issues with the URL building in the code. Additionally, Google started sending a HTTP redirect to a new URL for the download.  The python Requests library was doing something wrong when handling the redirect and would fail.  Switching to URLLIB3 fixed the redirect problem.  

As for your cert question; there are no cert bundles on the App Engine instances to verify against, so all SSL cert checking fails even though the Google web server certs are valid.

PR #1826, opened 2020-07-06
[DA-1628] Finding files by name for samples import

kenny-skaggs:
The samples import and reconciliation report cron jobs attempt to find the latest samples inventory CSV by looking at a set of files in the bucket and then trying to use the newest one. This restricts that set of files down to ones that match a specific name.

While working on this I found that if we try to log anything that's not a string then we don't get the GAE logs at all. I don't think this was contributing to anything we've seen so far, but I thought adding some extra processing for that in the gcp_logger could help prevent that issue in the future.

### DIFF
--- a/rdr_service/config.py
+++ b/rdr_service/config.py
@@ -85,8 +85,8 @@ EHR_STATUS_BIGQUERY_VIEW_PARTICIPANT = "ehr_status_bigquery_view_participant"
 EHR_STATUS_BIGQUERY_VIEW_ORGANIZATION = "ehr_status_bigquery_view_organization"
 HPO_REPORT_CONFIG_MIXIN_PATH = "hpo_report_config_mixin_path"
 LOCALHOST_DEFAULT_BUCKET_NAME = 'local_bucket'
-BIOBANK_SAMPLES_INVENTORY_FILE_PATTERN = 'biobank_samples_inventory_file_pattern'
-BIOBANK_SAMPLES_INVENTORY_MANIFEST_FILE_PATTERN = 'biobank_samples_inventory_manifest_file_pattern'
+BIOBANK_SAMPLES_DAILY_INVENTORY_FILE_PATTERN = 'biobank_samples_daily_inventory_file_pattern'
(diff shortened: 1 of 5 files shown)


rdr_service/config.py
```diff
@@ -85,6 +85,8 @@
 EHR_STATUS_BIGQUERY_VIEW_ORGANIZATION = "ehr_status_bigquery_view_organization"
 HPO_REPORT_CONFIG_MIXIN_PATH = "hpo_report_config_mixin_path"
 LOCALHOST_DEFAULT_BUCKET_NAME = 'local_bucket'
+BIOBANK_SAMPLES_INVENTORY_FILE_PATTERN = 'biobank_samples_inventory_file_pattern'
+BIOBANK_SAMPLES_INVENTORY_MANIFEST_FILE_PATTERN = 'biobank_samples_inventory_manifest_file_pattern'
```
pegbertsch: Would it help if these had more obvious names that reflect the daily vs. monthly inventory report purpose?   E.g., BIOBANK_SAMPLES_DAILY_INVENTORY_FILE_PATTERN / MONTLY_INVENTORY_MANIFEST_FILE_PATTERN?

rdr_service/offline/biobank_samples_pipeline.py
```diff
@@ -145,7 +146,7 @@ def get_last_biobank_sample_file_info(monthly=False):
     if monthly:
```
pegbertsch, line 147: Do we still need this logic if we're screening on specific file name formats now?  Or is it still more efficient to also qualify the bucket/path name further?

pegbertsch (approved):
Minor/optional suggestions but no issues

rdr_service/config.py
kenny-skaggs: Good idea, updating it now.

rdr_service/offline/biobank_samples_pipeline.py
kenny-skaggs: Good catch. I don't think we'd need it, but it could help with narrowing things down when searching through names. I don't see any immediate need to take it out.

pegbertsch (approved):

PR #1830, opened 2020-07-07
[DA-1607] DV salivary orders: creating barcode BiobankOrderIdentifier and finalized BiobankOrder

kenny-skaggs:
The ticket mentions saliva pilot instead of all saliva dv orders, but based on some tickets I've been digging through the terms seem to be interchangeable.

These changes work with BiobankOrders created when we receive DV orders. Since the RDR is creating biobank orders internally at the same time they're sent to MayoLink this sets the finalized time for the orders. And (per the ticket) this also creates a BiobankOrderIdentifier for the order that has the MayoLink barcode as a value and pmi-ops as the system.

### DIFF
--- a/rdr_service/dao/dv_order_dao.py
+++ b/rdr_service/dao/dv_order_dao.py
@@ -28,6 +28,7 @@ from rdr_service.model.biobank_order import BiobankOrder, BiobankOrderIdentifier
     MayolinkCreateOrderHistory
 from rdr_service.model.config_utils import to_client_biobank_id
 from rdr_service.model.utils import to_client_participant_id
+from rdr_service.offline.biobank_samples_pipeline import _PMI_OPS_SYSTEM
 from rdr_service.participant_enums import BiobankOrderStatus, OrderShipmentStatus, OrderShipmentTrackingStatus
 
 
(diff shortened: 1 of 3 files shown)


rdr_service/dao/dv_order_dao.py
```diff
@@ -260,7 +260,7 @@ def insert_biobank_order(self, pid, resource):
         obj = BiobankOrder()
         obj.participantId = int(pid)
         obj.created = clock.CLOCK.now()
-        obj.created = datetime.datetime.now()
+        obj.finalizedTime = obj.created
```
pegbertsch, line 263: old code vs. new code:  Is the old code ending up with obj.created = datetime.datetime.now() a bug since it's local time?   Or is it functionality equivalent on our GCP instances to utcnow() as returned by clock.CLOCK.now()?   Wondering if previously inserted biobank order records were using an inconsistent timestamp value.

kenny-skaggs: Good catch, I wasn't aware that `CLOCK.now()` did UTC while `datetime.now()` was local. It looks like the server is set to UTC. The `created` field for request logs are saved with UTC time (latest request logs in prod currently are in the 1 o'clock time). So I found a timestamp of an existing dv order in prod and matched it with with the request log entry recorded for it and the created times are within a second of eachother.

pegbertsch (approved):

PR #1831, opened 2020-07-08
[DA-1648] Update to Cohort 2 AW0 Samples Logic

j-kanuch:
This PR changes the way samples are created in the genomic system's Cohort 2 AW0 manifest. First, a list of participants are generated that match the reconsent and validation criteria. Next, a lookup is performed on each participant to get the relevant sample that matches the following rules:
- Most-recently collected 1ED04 (blood) or 1SAL2 (saliva) sample
- If samples were collected at the same time, use the 1ED04 sample

The AW2 Array sub-folder location was also updated to match the SOP documentation.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -1219,14 +1245,99 @@ def _get_new_biobank_samples(self, from_date):
...
+                WHEN ps.withdrawal_status = :withdrawal_param THEN 1 ELSE 0
+              END as valid_withdrawal_status,
+              CASE
+                WHEN ps.suspension_status = :suspension_param THEN 1 ELSE 0
+              END as valid_suspension_status,
+              CASE
+                WHEN ps.consent_for_study_enrollment = :general_consent_param THEN 1 ELSE 0
+              END as general_consent_given,
+              CASE
+                WHEN ps.date_of_birth < DATE_SUB(now(), INTERVAL :dob_param*365 DAY) THEN 1 ELSE 0
```
kenny-skaggs: Finding the participant age this way doesn't account for leap years, `DATE_SUB(NOW(), INTERVAL :dob_param YEAR)` would.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1036,58 +1049,71 @@ def create_new_genomic_participants(self, from_date):
...
+        participants = self._get_new_c2_participants(from_date)
+        participant_matrix = self.GenomicSampleMeta(*participants)
+
+        if len(participants) > 0:
+            for i, _bid in enumerate(participant_matrix.bids):
+                sample_data = self._get_usable_sample_from_participant(pid=participant_matrix.pids[i],
+                                                                       bid=_bid)
+
+                # update the sample id and collected site
+                if sample_data != (None, None):
```
kenny-skaggs: Would this match an empty return or should there be another None?

j-kanuch: Good catch, the biobank order id was a late addition. I may change it to the following since all values are required:
`if None not in sample_data:`

rdr_service/genomic/genomic_job_components.py
j-kanuch: Good call, I will update.

kenny-skaggs (approved):

PR #1832, opened 2020-07-09
improve_error_message

wangy70:


pegbertsch (approved):

wangy70:
I will hold this one because I just realized some one may catch the old 504 error for token refresh. I saw from the log that PTSC maybe one of them.

PR #1834, opened 2020-07-10
[DA-1624] ehr_consent_expired_status

wangy70:


rdr_service/dao/participant_summary_dao.py
```diff
@@ -586,11 +588,13 @@ def calculate_enrollment_status(
...
             ):
                 return EnrollmentStatus.FULL_PARTICIPANT
-            return EnrollmentStatus.MEMBER
+            if consent_expire_status != ConsentExpireStatus.EXPIRED:
+                return EnrollmentStatus.MEMBER
         return EnrollmentStatus.INTERESTED
 
     def calculate_member_time(self, consent, participant_summary):
-        if consent and participant_summary.enrollmentStatusMemberTime is not None:
+        if (consent or participant_summary.ehrConsentExpireStatus == ConsentExpireStatus.EXPIRED) and \
```
kenny-skaggs: Before this change, any participants that explicitly refused EHR consent having having a consent would have their member time cleared. Now if a participant explicitly refuses EHR consent after they have an expiration this will still give them an enrollmentStatusMemberTime. Is that desired behavior?

kenny-skaggs (commented):
Have a specific comment in the DAO that may need to be resolved.

Additionally, this logic will need to be implemented in the BigQuery and Resource API code. But I imagine those implementations can wait for another PR due to the time sensitivity of getting this out.

rdr_service/dao/participant_summary_dao.py
wangy70: Ah, good catch! I thought even it's expired, we still need the `enrollmentStatusMemberTime` for the participant count over time, but you are right, we didn't do that for the withdrawal participants. I will change this back to keep consistency. I will also check with Katie about the requirement for the participant count over time dashboard, if they need, we can do a fix for that.

wangy70:
Thanks for the reminder, I will have a new PR for the BigQuery and Resource API code.

kenny-skaggs (approved):

PR #1836, opened 2020-07-13
fake_participant_generator_skip_when_code_not_found

wangy70:


rdr_service/data_gen/generators/questionnaires.py
```diff
@@ -691,6 +691,9 @@ def _answer_question(self, question, answer, quantity=None):
     :param quantity: models.quantity.Quantity object
     :return: QuestionnaireResponseGroupQuestion object
     """
+        if question is None:
```
pegbertsch, line 694: I was looking at how the question parameter is generated via _find_question(), e.g:
self._answer_question(self._find_question("ExtraConsent_WelcomeVideo"),...)

Would it be helpful if _find_question() did debug logging when it falls through to its "return None" case on line 871, to log the question_code string that this "code not found" message refers to?

pegbertsch (approved):
Suggestion (optional) to see if debug logging could be enhanced in cases where we skip.

rdr_service/data_gen/generators/questionnaires.py
wangy70: Good point, just added the log.

PR #1837, opened 2020-07-13
[DA-1624] bq and resource api for consent_expire

wangy70:


kenny-skaggs:
I may be misunderstanding it, but BigQuery and the ResourceAPI seem to have their own logic for finding the enrollment status. For ResourceAPI it's at https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/resource/generators/participant.py#L426 and for BigQuery it's at https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/bq_participant_summary_dao.py#L423. For both you'd have to adjust the logic for finding the enrollment status based on current data and you'd also need to do something about the backfill logic to make sure it doesn't set them to Core if they're not supposed to be (because of an expired EHR).

wangy70:
Yea, this is still in progress. I just checked with @robabram , the response value will be automatically covered by the existing code. I will take a look at the enrollment status part. 

For the backfill you mention, I didn't get it, could you explain more? Thanks!

wangy70:
Close this one because there are a lot of changes and conflicts on bq_participant_summary_dao.py, will submit a new PR for the BQ and resource API enrollment status calculation.

PR #1847, opened 2020-07-16
[DA-1655] Update W3, Genomics Cron, AW0 Filename

j-kanuch:
This PR contains the following updates for the Genomics Pipeline:
- Update columns on W3 manifest
- Removes all jobs from cron schedule and add cohort 2 AW0 workflow to run Mondays
- Update the AW0 filename for Biobank's latest request

### DIFF
--- a/rdr_service/genomic/genomic_biobank_manifest_handler.py
+++ b/rdr_service/genomic/genomic_biobank_manifest_handler.py
@@ -172,6 +172,5 @@ def _get_output_manifest_file_name(genomic_set_id, timestamp=None, cohort_id=Non
     )
     cohort = f"_{cohort_id}" if cohort_id else ""
     folder_name = config.getSetting(GENOMIC_BIOBANK_MANIFEST_FOLDER_NAME)
-    #full_name = f'{folder_name}/{_MANIFEST_FILE_NAME_PREFIX}{cohort}-{str(genomic_set_id)}-v1{now_cdt_str}.CSV'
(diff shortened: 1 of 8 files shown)


rdr_service/genomic/genomic_biobank_manifest_handler.py
```diff
@@ -170,7 +170,8 @@ def _get_output_manifest_file_name(genomic_set_id, timestamp=None, cohort_id=Non
     now_cdt_str = (
         _UTC.localize(file_timestamp).astimezone(_US_CENTRAL).replace(tzinfo=None).strftime(OUTPUT_CSV_TIME_FORMAT)
     )
-    cohort = f"-{cohort_id}" if cohort_id else ""
+    cohort = f"_{cohort_id}" if cohort_id else ""
     folder_name = config.getSetting(GENOMIC_BIOBANK_MANIFEST_FOLDER_NAME)
-    full_name = f'{folder_name}/{_MANIFEST_FILE_NAME_PREFIX}{cohort}-{str(genomic_set_id)}-v1{now_cdt_str}.CSV'
+    #full_name = f'{folder_name}/{_MANIFEST_FILE_NAME_PREFIX}{cohort}-{str(genomic_set_id)}-v1{now_cdt_str}.CSV'
```
pegbertsch: Is the format of the full_name value nailed down or are we leaving both formats in the code just in case?

pegbertsch (approved):
Question to confirm if we can remove commented out full_name format or not, but no issues.

rdr_service/genomic/genomic_biobank_manifest_handler.py
j-kanuch: Yes, this can probably be removed.

PR #1850, opened 2020-07-17
[ROC-659] Filter excluded fields when using Codebook data

robabram:
Fixes issue with PII fields being added to PDR BigQuery.


pegbertsch (commented):
I'm not familiar with how the lists of excluded fields for each _BQModuleSchema class was identified.  Is it correct that there are no excluded fields currently defined for the BQPDRCOPEMaySchema?

robabram:
The COPE survey doesn't have any PII questions in it.

pegbertsch (approved):

PR #1852, opened 2020-07-21
[DA-1664] Cohort 1 AW0 Workflow

j-kanuch:
This PR implements the Cohort 1 workflow for the AW0 manifest. Methods originally for Cohort 2 only were refactored to be reusable between both Cohorts 1 & 2. As an unrelated addition, the SendGrid sender email was updated to "no-reply@pmi-ops.org" and included in this PR as well.


rdr_service/offline/genomic_pipeline.py
```diff
@@ -58,6 +58,15 @@ def c2_participant_workflow():
         controller.run_c2_participant_workflow()
 
 
+def c1_participant_workflow():
+    """
+    Entrypoint for Cohort 1 Participant Workflow,
+    Sources from Cohort 1 participants that have reconsented.
+    """
+    with GenomicJobController(GenomicJob.C2_PARTICIPANT_WORKFLOW) as controller:
```
kenny-skaggs: Should this be initialized with the C1 workflow id?

j-kanuch: Good catch, this was a copy/paste error. I've updated the code.

kenny-skaggs (approved):

PR #1856, opened 2020-07-22
[DA-1670] Updated tool to create Cohort 1 AW0 manifests.

j-kanuch:
This PR extends the genomic utility tool to run the AW0 manifest for Cohort 1 locally.

### DIFF
--- a/rdr_service/tools/tool_libs/genomic_utils.py
+++ b/rdr_service/tools/tool_libs/genomic_utils.py
@@ -244,8 +244,12 @@ class GenerateManifestClass(GenomicManifestBase):
                 return 1
 
             if int(self.args.cohort) == 2:
-                _logger.info('Running c2 workflow')
+                _logger.info('Running Cohort 2 workflow')
                 return self.generate_local_c2_manifest()
+
+            if int(self.args.cohort) == 1:
+                _logger.info('Running Cohort 1 workflow')
+                return self.generate_local_c1_manifest()
 
     def generate_local_c2_manifest(self):
         """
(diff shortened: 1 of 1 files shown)


rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -304,6 +307,62 @@ def export_c2_manifest_to_local_file(self, set_id):
...
+    def generate_local_c1_manifest(self):
+        """
+        Creates a new C1 Manifest locally
+        :return:
+        """
+
+        last_run_time = self.dao.get_last_successful_runtime(GenomicJob.C1_PARTICIPANT_WORKFLOW)
+
+        if last_run_time is None:
+            last_run_time = datetime.datetime(2020, 7, 27, 0, 0, 0, 0)
```
pegbertsch, line 319: Since last_run_time will be used to find/filter for new C1 participants (line 324), just want to confirm we really only care about the M/D/Y here and there are no timezone offset issues (e.g, UTC vs. "midnight Eastern" regarding when the PTSC release happens), like we had with the cutoff datetime for cohort assignments.

j-kanuch: In actuality, we want to pick up any participant that has answered the reconsent-review question yes. This default time is more to limit the query than to actually filter out any participants before this time. Once the first run happens, `last_run_time` is set with the UTC time from the `start_time` of that first run. This will pick up the participants that have answered the reconsent-review question since this `start_time`.

pegbertsch (approved):

PR #1868, opened 2020-07-27
[ROC-578] Allow for PM finalized value being null

robabram:
I spent two hours trying to find a participant ID that would cause the same exception as in the logs, but I couldn't find a single one.  I'm adding this check to fix the error.


kenny-skaggs:
Would it be good to set up a unit test to try out a participant that doesn’t have physical measurements?

robabram:
The actual scenario is the participant has probably 2 PM records where one appears finalized, but the finalized timestamp is null.  I searched high and low in prod for a record that could break the existing logic and couldn't find.  I agree that unittests would be good here, but I don't have time to spend on them yet.

wangy70 (approved):

PR #1869, opened 2020-07-28
[DA-1624] set_ehr_expired_when_reconsent

wangy70:


kenny-skaggs:
Should the consent expire status be cleared only when we get a new Yes response for the consent, or should we also clear it if they send us a No?

Additionally, we've seen PTSC send us old questionnaire responses. We could potentially get responses for participants that were made before the expiration when the participant might not have actually reconsented after the expiration. Should we check the authored times to be sure that the EHR consent we're receiving is new and not something that is old data?

wangy70:
@kenny-skaggs Good point! I think we need to compare the authored time. And for the No answer, I will check with Katie.

rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -337,10 +337,14 @@ def _update_participant_summary(
...
                                 participant_summary.consentForElectronicHealthRecordsFirstYesAuthored = authored
-                            if participant_summary.ehrConsentExpireStatus == ConsentExpireStatus.EXPIRED:
+                            if participant_summary.ehrConsentExpireStatus == ConsentExpireStatus.EXPIRED and \
+                                authored > participant_summary.ehrConsentExpireAuthored:
                                 participant_summary.ehrConsentExpireStatus = ConsentExpireStatus.UNSET
                                 participant_summary.ehrConsentExpireAuthored = None
                                 participant_summary.ehrConsentExpireTime = None
+                            elif participant_summary.ehrConsentExpireStatus == ConsentExpireStatus.EXPIRED and \
+                                authored < participant_summary.ehrConsentExpireAuthored:
+                                ehr_consent = False
```
kenny-skaggs, line 348: Oooh, good idea

kenny-skaggs (commented):
@wangy70, should this PR stay on hold until we know about what to do if they submit a later No response to the EHR module, or have you heard what should be done with that yet?

kenny-skaggs:
@wangy70, are we still waiting to hear back about whether we should clear the ehr expire status when the participant updates us with a No? Currently this will only clear the expire status if they answer with a Yes.

wangy70:
@kenny-skaggs  Sorry, just saw your comments. We had a meeting with PTSC this morning, and also double checked with HealthPro, the conclusion is that: no matter the EHR re-consent answer is YES or NO, we will set the expired status and authored to UNSET and NULL. I will make this change soon.

kenny-skaggs (approved):

PR #1873, opened 2020-07-30
[PDR-82] Use on DNA tests and disposed time when calculating enrollment_core_stored

robabram:


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -680,9 +680,12 @@ def _calculate_enrollment_timestamps(self, summary):
...
         for bbo in summary['biobank_orders']:
             for bboi in bbo['bbo_samples']:
-                if bboi['bbs_baseline_test'] == 1:
+                if bboi['bbs_dna_test'] == 1:
                     ordered_time = min(ordered_time, bboi['bbs_finalized'] or datetime.datetime.max)
-                    stored_time = min(stored_time, bboi['bbs_confirmed'] or datetime.datetime.max)
+                    # See: participant_summary_dao.py:calculate_max_core_sample_time() and
+                    #       _participant_summary_dao.py:126
+                    stored_time = min(stored_time, bboi['bbs_disposed'] or bboi['bbs_confirmed'] or
+                                      datetime.datetime.max)
```
kenny-skaggs: I don't see how the logic between this and RDR's participant_summary_dao lines up.

participant_summary_dao.py:126 only uses the disposed times if all of the samples for a test are disposed. Would this potentially use disposed times for a participant that still has non-disposed samples still for that test?

There's also different stages of logic in RDR's participant_summary_dao. The maximum disposal or confirmed times are stored in participant_summary for a given test. Could this potentially use an earlier time for a test because it's looking at the minimum for all of the dates for the tests rather than the maximums for each one?

robabram: I've asked Bhinnata to find the program documentation for calculating the enrollment core stored timestamp.  I'm also going to try and find the original ticket for it.

robabram: I've added a new commit.

rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -678,11 +678,28 @@ def _calculate_enrollment_timestamps(self, summary):
...
+                if bboi['bbs_dna_test'] == 1:
                     ordered_time = min(ordered_time, bboi['bbs_finalized'] or datetime.datetime.max)
-                    stored_time = min(stored_time, bboi['bbs_confirmed'] or datetime.datetime.max)
+                    # See: participant_summary_dao.py:calculate_max_core_sample_time() and
+                    #       _participant_summary_dao.py:126
+                    # Only catch the most recent test.
+                    if stored_sample_times[bboi['bbs_test']] == datetime.datetime.min:
+                        stored_sample_times[bboi['bbs_test']] = \
+                            max(bboi['bbs_disposed'] or bboi['bbs_confirmed'] or datetime.datetime.min,
+                                datetime.datetime.min)
```
kenny-skaggs: `stored_sample_times` is meant to be the maximum timestamp (disposed or confirmed time if a sample isn't disposed) from all the samples for a given test, right? It seems like it would only take the time from the first sample it looks at and not find the maximum for all of them.

robabram: I have pushed a new commit to more accurately replicate the logic in the participant summary dao.

kenny-skaggs (approved):

PR #1874, opened 2020-07-31
[DA-1679] Add Genomic Jobs to Cron; allow for manual runs

j-kanuch:
This PR adds the genomic jobs that are needed for the E2E testing to the cron scheduler. They are set up to be run manually.


rdr_service/offline/main.py
```diff
@@ -226,16 +225,47 @@ def genomic_new_participant_workflow():
...
+    genomic_pipeline.c2_participant_workflow()
+    return '{"success": "true"}'
+
+
+@app_util.auth_required_cron
+@_alert_on_exceptions
+def genomic_c1_participant_workflow():
+    """Temporarily running this manually for E2E Testing"""
+    now = datetime.utcnow()
+    if now.day == 0o1 and now.month == 0o1:
```
kenny-skaggs, line 240: Why use octal?

Not that I think it needs to change. I say go for it, especially for something temporary like this. But just curious 🐱

j-kanuch: I'm not sure why the other job functions originally used octal, but I kept it consistent with those.

pegbertsch (approved):

PR #1876, opened 2020-07-31
[DA-1676] MIGRATION: Add fields to GC Metrics; update upstream states for AW3 manifests

j-kanuch:
This PR updates the `genomic_gc_metrics_validation` table to include file paths needed in the AW3 manifest. Other stages of the pipeline that use this table were updated to test correct ingestion of the file paths. The AW3 queries were set up. These will be completed in a subsequent PR, as well as creating the actual manifests and the unit tests for the manifests.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -796,10 +798,14 @@ def reconcile_metrics_to_genotyping_data(self):
             for file_type in self.genotyping_file_types:
                 if not getattr(metric, file_type[0]):
                     filename = f"{metric.chipwellbarcode}{file_type[1]}"
-                    file_exists = self._check_genotyping_file_exists(file.bucketName, filename)
-                    setattr(metric, file_type[0], file_exists)
+                    file_exists = self._get_full_filename(file.bucketName, filename)
+
+                    if file_exists != 0:
```
pegbertsch, line 803: Minor/optional: Could these sequential if file_exists != 0 / if not file_exists checks can become an "if file_exists...else" block?  (Also on line 854/859). Perhaps return None instead of 0 for a "not found" result in _get_full_filename()?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1686,6 +1701,24 @@ def _setup_manifest_definitions(self):
             columns=self._get_manifest_columns(GenomicManifestTypes.CVL_W3),
         )
 
+        # DRC to Broad AW3 Array Manifest
+        self.MANIFEST_DEFINITIONS[GenomicManifestTypes.AW3_ARRAY] = self.ManifestDef(
+            job_run_field='arrAW3ManifestJobRunID',
```
pegbertsch, line 1706: Is it worth defining these job_run_field inline strings as constants somewhere?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1694,6 +1727,89 @@ def _get_source_data_query(self, manifest_type):
...
+                        sqlalchemy.join(ParticipantSummary,
+                                        GenomicSetMember,
+                                        GenomicSetMember.participantId == ParticipantSummary.participantId),
+                        GenomicGCValidationMetrics,
+                        GenomicGCValidationMetrics.genomicSetMemberId == GenomicSetMember.id
+                    )
+                ).where(
+                    (GenomicGCValidationMetrics.processingStatus == 'pass') &
+                    (GenomicSetMember.genomicWorkflowState != GenomicWorkflowState.IGNORE) &
+                    (GenomicSetMember.genomeType == "aou_array") &
```
pegbertsch, line 1761: These inline string values being compared for genomeType are also defined as constants in rdr_service/config.py -- is it worth importing/referencing those instead?

pegbertsch (approved):
A few minor/optional suggestions to consider.  Could be incorporated later if you agree.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Yes, I don't like how this is setup and will probably update it. Since other workflows use this method I needed to do some hacky changes. I'll fix in the next AW3 PR.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Yes, there are a large number of these so I'll incorporate in the next AW3 PR.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Yes, same as my previous comment.

PR #1895, opened 2020-08-12
[ROC-580] Restart Gunicorn workers before reaching GAE memory limit

kenny-skaggs:
### Background
Memory used by a python process can gradually increase over time without returning "freed" memory back to the OS. It almost looks like a memory leak, but there isn't anything that is leaking: everything is getting garbage collected when it should. The CPython interpreter keeps large chunks of memory until everything is cleared from them, and it sections off specific parts of those chunks to only be used by a certain data type. So if more memory is needed for a data type, CPython could request more from the operating system even if it has alot of unused memory (memory meant for another data type). There's more detail (and better descriptions) at https://realpython.com/python-memory-management/ and a description of how to handle this particular type of problem at https://chase-seibert.github.io/blog/2013/08/03/diagnosing-memory-leaks-python.html.

Eventually our app pulls in enough memory to have GAE decide we've had enough and it kills our instance. A typical fix for this would be to reduce the number of workers/threads we have going or reduce the number of requests they handle before closing themselves. But both of those solutions felt like they would have unnecessarily reduced the performance of the servers. I believe that the settings we have now for workers and threads are good for our average memory usage, but we're seeing issues when we have multiple requests for retrieving batches from ParticipantSummary or WorkspaceSnapshots.
### Changes in this PR
This PR adds to the Gunicorn configuration so that we can check the memory used by a GAE instance after each request has been processed. If the total memory used on the instance is more than 950MB (of the total 1024MB allowed) we have the worker restart after finishing the request. It's not ideal, but releasing the memory of a worker to the OS will reduce the total used by the instance. Ideally we would target the worker on the instance that is using the most memory, but we can watch the performance to see if this is enough or if further tweaks need to be made.


rdr_service/services/gunicorn_config.py
```diff
@@ -23,3 +24,29 @@
...
+    # In the future this can be tuned to use the process id of the worker to find how much memory that worker
+    # has allocated.
+    memory_used_bytes = self_mem_bytes + children_mem_bytes
+    if memory_used_bytes > memory_threshold_bytes:
+
+        # This is copied from Gunicorn's code for closing out a sync worker after reaching the max_requests limit
+        worker.alive = False
+
+        # Logs from the worker seem to appear beside the normal app logs, but without log levels attached to them.
+        worker.log.info(f"Auto-restarting worker after gunicorn instance {os.getpid()} was"
```
pegbertsch: Should the logging call be made before setting alive to False?

rdr_service/services/gunicorn_config.py
```diff
@@ -31,13 +30,23 @@
...
+    # has allocated.
+    memory_used_bytes = self_mem_bytes + children_mem_bytes
+    if memory_used_bytes > memory_threshold_bytes:
+
         # This is copied from Gunicorn's code for closing out a sync worker after reaching the max_requests limit
         worker.alive = False
+
+        # Logs from the worker seem to appear beside the normal app logs, but without log levels attached to them.
+        worker.log.info(f"Auto-restarting worker after gunicorn instance {os.getpid()} was"
+                        f"found to be using {memory_used_bytes} bytes.")
```
robabram: Would you might concider changing this to `f"found to be using {round(memory_used_bytes / 1024 / 1000, 2)} megabytes.")` to simplify the bytes number.

rdr_service/services/gunicorn_config.py
kenny-skaggs: The `post_request` method is called on the worker's thread, and the `alive` field is checked later (after this method to complete) to determine if the worker should continue running. The logger on the worker doesn't seem to check the alive status before processing statements. I think it's safe this way, but I can switch them around if you think it would be better for readability.

rdr_service/services/gunicorn_config.py
kenny-skaggs: Good idea. I'll switch everything over to using multiples of 1024 to be more accurate too.

rdr_service/services/gunicorn_config.py
pegbertsch: Just wanted to be sure the `worker.log` operation wasn't affected by whether `worker.alive` attribute was still True.

kenny-skaggs:
I did some more digging and found I'd made a few mistakes:
1. The resource module returns memory used as kilobytes, not bytes
2. The post_request hook runs in the worker's process, not gunicorn's. So memory used is for the worker and not the whole instance.
3. Gunicorn seems to require a certain signature and doesn't like the shortcut I used

I've updated the code to fix them. I've tried this latest code out in sandbox and it works as expected.

pegbertsch (approved):

PR #1900, opened 2020-08-13
[DA-1690] Tool for reversing approval of a deceased report

kenny-skaggs:
By design the API has doesn't have a way of reversing an approved deceased report. We'll be correcting any errors that come up manually. Since there are a few fields across a couple of data models that need to be set in a specific way I made this tool to make that job easier. I'll also update the playbook with an entry on what to do if we need to reverse an approved report.

Note: this will set whoever is running the tool as the reviewer of the report

### DIFF
--- a/rdr_service/tools/tool_libs/resurrect.py
+++ b/rdr_service/tools/tool_libs/resurrect.py
@@ -7,7 +7,8 @@ import argparse
 import logging
 import sys
 
-from rdr_service.dao.deceased_report_dao import DeceasedReportDao
+from rdr_service.dao.api_user_dao import ApiUserDao
+from rdr_service.model.api_user import ApiUser
 from rdr_service.model.deceased_report import DeceasedReport
 from rdr_service.model.participant_summary import ParticipantSummary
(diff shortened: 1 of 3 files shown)


rdr_service/tools/tool_libs/resurrect.py
```diff
@@ -0,0 +1,116 @@
...
+            report.denialReasonOther = self.args.reason_desc
+
+    @staticmethod
+    def update_participant_summary(session, participant_id):
+        participant_summary = session.query(ParticipantSummary).filter(
+            ParticipantSummary.participantId == participant_id
+        ).one_or_none()
+
+        if participant_summary is None:
+            print('WARNING: participant summary not found')
```
j-kanuch: Not a big deal, but if you use  _logger.warning('WARNING: participant summary not found') it will output this in color (yellow I believe), and optionally write to a file if "--log-file" is specified.

j-kanuch (approved):

rdr_service/tools/tool_libs/resurrect.py
kenny-skaggs: Good idea, thanks!

PR #1901, opened 2020-08-13
[DA-1577] Pb/da 1576 Implementation: Retention Eligibility Metrics 

wangy70:


rdr_service/model/participant_summary.py
```diff
@@ -56,6 +69,94 @@
...
+            LEAST(COALESCE(sample_status_1ed10_time, '9999-01-01'),
+                COALESCE(sample_status_2ed10_time, '9999-01-01'),
+                COALESCE(sample_status_1ed04_time, '9999-01-01'),
+                COALESCE(sample_status_1sal_time, '9999-01-01'),
+                COALESCE(sample_status_1sal2_time, '9999-01-01')
+            )
+        )
+        ELSE NULL
+     END
+"""
```
kenny-skaggs, line 113: The definition of `_COMPUTE_RETENTION_ELIGIBLE_SQL` is duplicated on lines 74 and 118. Same for `_COMPUTE_RETENTION_ELIGIBLE_TIME_SQL` on lines 93 and 137.

rdr_service/model/participant_summary.py
```diff
@@ -56,6 +69,94 @@
...
+
+# SQL Conditional for participant's retention eligibility computed column (1 = NOT_ELIGIBLE, 2 = ELIGIBLE)
+# TO DO:  This should be updated to verify no deceased status when that is implemented (see: DA-1601)
```
kenny-skaggs: Does deceased_status logic still need to be verified, or can the TODO be removed? Looks like it's including the deceased status field and that's the correct value for a participant having an UNSET deceased value (vs PENDING or APPROVED).

rdr_service/model/participant_summary.py
```diff
@@ -56,6 +69,94 @@
...
+_COMPUTE_RETENTION_ELIGIBLE_SQL = """
+    CASE WHEN
+      consent_for_study_enrollment = 1
+      AND (consent_for_electronic_health_records = 1 OR consent_for_dv_electronic_health_records_sharing = 1)
+      AND questionnaire_on_the_basics = 1
+      AND questionnaire_on_overall_health = 1
+      AND questionnaire_on_lifestyle = 1
+      AND samples_to_isolate_dna = 1
```
kenny-skaggs: `samples_to_isolate_dna` could hold values of 10 or 11 to say that the samples have been DISPOSED or CONSUMED. Should a participant not be retention eligible if that's the case?

wangy70: Based on the requirement, only `samples_to_isolate_dna`=1 will be count as retention eligible.

rdr_service/model/participant_summary.py
wangy70: Yea, I forgot to remove the comment, will remove it :)

rdr_service/model/participant_summary.py
wangy70: Ah, good catch, that may caused by the conflict merge, I will remove it.

rdr_service/model/participant_summary.py
wangy70: Done

rdr_service/model/participant_summary.py
wangy70: Done

kenny-skaggs (approved):

rdr_service/participant_enums.py
```diff
@@ -412,6 +421,14 @@ class DeceasedReportDenialReason(messages.Enum):
     OTHER = 4
 
 
+# DA-1576:  Retention Eligibility Metrics
```
pegbertsch, line 424: This looks like another duplication (class already defined on line 380)

PR #1916, opened 2020-08-20
[PDR-113] Fixes for enrollment status calculations.

robabram:


j-kanuch (approved):

j-kanuch (approved):

kenny-skaggs:
The Resource and BigQuery generators should match up, right? The Resource code doesn't have this change `data['consents'].sort(key=lambda consent_data: consent_data['consent_module_authored'], reverse=True)`. I'm not sure what else might be different.

PR #1920, opened 2020-08-25
[DA-1719] Add tz offset to returned timestamps.

j-kanuch:
The Genomic Outreach API response is returning timestamps that are timezone naive. The timezone offset needs to be included in timestamps in the JSON response.

### DIFF
--- a/rdr_service/dao/genomics_dao.py
+++ b/rdr_service/dao/genomics_dao.py
@@ -1,4 +1,6 @@
 import collections
+
+import pytz
 import sqlalchemy
 import logging
 
@@ -956,7 +958,7 @@ class GenomicOutreachDao(BaseDao):
 
         client_json = {
             "participant_report_statuses": report_statuses,
-            "timestamp": result['date']
+            "timestamp": result['date'].replace(tzinfo=pytz.UTC)
         }
         return client_json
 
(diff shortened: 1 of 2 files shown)


rdr_service/dao/genomics_dao.py
```diff
@@ -956,7 +958,7 @@ def to_client_json(self, result):
 
         client_json = {
             "participant_report_statuses": report_statuses,
-            "timestamp": result['date']
+            "timestamp": result['date'].replace(tzinfo=pytz.UTC)
```
kenny-skaggs: `result['date']` is timezone unaware so this works. But `replace` overwrites timezone data on a datetime object without changing anything else, even if it already has a timezone. So a date of 2020-05-01T03:00-06:00 would become 2020-05-01T03:00+00:00. I can't imagine this would ever see anything other than a timezone unaware `result['date']`. But _if_ there's ever a change in that, this could cause hard to detect bugs.

`pytz.utc.localize(result['date'])` functionally does the same as what you have, but it only accepts timezone unaware dates. So in the off chance anything changes about how the date is given to the dao, we'd see the unit tests failing and act accordingly.

This feels really nit-picky, but I thought I'd bring it up in case you thought it was a good idea.

j-kanuch: That's a good catch, I will make that fix. We are not expecting dates with timezone, but if that changes in the future, we would have a hard to find data issue on our hands. Better to localize than outright replace the timezone.

kenny-skaggs (approved):

PR #1934, opened 2020-08-28
[PDR-113] Add sort fallback when authored values are None.

robabram:


j-kanuch (approved):

wangy70 (approved):

kenny-skaggs:
What are the situations where this happens? Does it make sense to sort everything by created when there might still be some authored?

PR #1939, opened 2020-08-31
[PDR-113] Handle misc errors found during test rebuild.

robabram:


j-kanuch (approved):

rdr_service/resource/generators/participant.py
```diff
@@ -770,18 +775,21 @@ def _calculate_enrollment_status(self, summary):
...
+                        .get_intersection(ehr_date_range)
+
+                    if summary['consent_cohort'] == BQConsentCohort.COHORT_3.name:
+                        date_overlap = date_overlap.get_intersection(gror_date_range)
+
+                    # If there's any time that they had everything at once, then they should be a Core participant
+                    if date_overlap.any():
+                        status = EnrollmentStatusV2.CORE_PARTICIPANT
+                except TypeError:
+                    pass
```
kenny-skaggs: What error happened that this solves? There's an 'if' check at the start of working with the date ranges that I thought implies that anything here would be available.

robabram: Apparently None values can creep into the data sent to the 'get_intersection()' function which causes a TypeError exception.  My thinking with the exception handling here is; that if the participant doesn't have enough data and an exception is thrown, then they probably won't be made a Core Participant and its Ok to just move on.

robabram: Keep in mind, these errors are from the Stable environment which has some pretty terrible data there. It may be that in Prod this exception will never be thrown.

pegbertsch: I have wondered about these get_intersection() use cases.   I'm new to this but is module it a problem that we set add_start() but not add_stop()  to the ranges sometimes (often)?

kenny-skaggs: No, I built them out to be able to handle un-ending timeframes. Using consent for study as an example: to add a start but no end implies that consent was given and never revoked. I'm not sure, but I'm guessing these errors came up when None was added for start dates.

Robert and I talked in the slack group and he's adding a log message in case we see this in prod.

kenny-skaggs (approved):

PR #1941, opened 2020-09-01
Sync doc/requirements.txt with latest

pegbertsch:
There is a copy of our requirements.txt file under the doc/ directory to enable ReadTheDocs.   It needs to be kept in sync with our project requirements.txt.   This pulls in the most recent changes.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -73,7 +73,7 @@ oauthlib[signedtoken]==3.1.0  # via jira, requests-oauthlib
 packaging==20.4           # via sphinx
 parameterized==0.7.4      # via -r requirements.in
 pbr==5.4.5                # via jira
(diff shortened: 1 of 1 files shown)


kenny-skaggs:
So it needs to be a copy of our root-level requirements.txt file? Can we point the process that needs this to our project-level reqs file and only need to maintain one?

pegbertsch:
I think during the initial setup process to integrate with ReadTheDocs, it was determined that RTD expected to find the content in the same doc/ directory with the rest of the doc files.   This might be an area to go look at in more detail to see if there's a way we can solve this in a way that doesn't require duplication of the requirements file.

kenny-skaggs (approved):
Ok, I'll take a look at it if/when I get to the documentation this sprint.

PR #1944, opened 2020-09-01
1.78.4 hotfix - date_of_birth changed from string to date type.

robabram:


rdr_service/dao/bq_pdr_participant_summary_dao.py
```diff
@@ -214,8 +213,7 @@ def _calculate_ubr(self, ps_bqr, p_id):
                         break
 
             if consent_date:
-                age = int((dateutil.parser.parse(consent_date.isoformat()) -
-                                dateutil.parser.parse(ps_bqr.date_of_birth)).days / 365)
+                age = int((consent_date - ps_bqr.date_of_birth).days / 365)
```
kenny-skaggs, line 216: This doesn't account for leap years.

robabram: Leap years can be a future enhancement if needed.

robabram: The final calculation from this is just to see if the participant's age is between 18 and 65.

pegbertsch: Not sure if this was the only place you used dateutil.parser and if that is now an unused import.  Pylint may not like that on push.

kenny-skaggs: How accurate is it expected that to be? This will be incorrect for people up to 4 days before they turn 18 and 16 days before their 65th birthday.

What is `ubr_age_at_consent` used for? Is it worth the risk if `ubr_age_at_consent` isn't set to 1 for someone that gave consent before they were 18?

pegbertsch (approved):
Noted Kenny's comment.   If it's worthy of a backlog ticket to look at later we should file that so it doesn't get lost.

rdr_service/dao/bq_pdr_participant_summary_dao.py
robabram: Let me qualify this, its calculating the age of the participant at the time they consented to the program.  So leap year is not a factor since both values are fixed and don't change over time.  Its not calculating the participant's current age.

PR #1947, opened 2020-09-04
[DA-1726] add research_id to participant table

wangy70:
Add `research_id` column to `participant` table.


rdr_service/etl/export_etl_results.sh
```diff
@@ -32,5 +32,5 @@ fi
 echo "Exporting tables from the cdm database..."
 ./rdr_client/run_client.sh ${PROJECT_AND_ACCOUNT} export_tables.py \
    --directory ${DIRECTORY} --database cdm \
-   --tables care_site,condition_era,condition_occurrence,cost,death,device_exposure,dose_era,drug_era,drug_exposure,fact_relationship,location,measurement,observation,observation_period,payer_plan_period,person,procedure_occurrence,provider,visit_occurrence
+   --tables pid_rid_mapping,care_site,condition_era,condition_occurrence,cost,death,device_exposure,dose_era,drug_era,drug_exposure,fact_relationship,location,measurement,observation,observation_period,payer_plan_period,person,procedure_occurrence,provider,visit_occurrence
```
kenny-skaggs, line 35: I've only looked at the ETL portions of this, so don't let this hold anyone else back from reviewing the PR.

I have another PR out for the curation tool module. I got that to run well enough to avoid having to manually export and upload most of the tables (it works for anything but the observation table). Can you add the `pid_rid_mapping` table to that tool's list of tables as well? https://github.com/all-of-us/raw-data-repository/blob/9c43a1dc6e8bfe0b4b032c5bd4971cfe3988d7b2/rdr_service/tools/tool_libs/curation.py#L31

wangy70: Done

j-kanuch (approved):

PR #1949, opened 2020-09-04
[ROC-652] De-duplicating response answers in curation ETL

kenny-skaggs:
Curation would like to have the authored time rather than created time, as long as authored isn't Null.

The also only need the latest answer to a particular survey's question. For the most part we can filter out everything but the latest for a responses question and module codes. But the COPE survey's all use the same module code while sharing question codes. So for them I retrieve the Vibrent Form ID we get in (almost) every questionnaire payload. The COPE surveys have a unique Vibrent Form ID for each month, so grouping by that instead of the module value (COPE) will preserve the individual responses that each month has for a shared question code.

### DIFF
--- a/rdr_service/etl/etl.sql
+++ b/rdr_service/etl/etl.sql
@@ -537,17 +537,9 @@ DROP TABLE IF EXISTS cdm.questionnaire_vibrent_forms;
 CREATE TABLE cdm.questionnaire_vibrent_forms (
     questionnaire_id            bigint,
     version                     bigint,
-    vibrent_form_id             varchar(1024),
+    vibrent_form_id             varchar(200),
     INDEX vibrent_form_index    (questionnaire_id, version)
 );
-
-INSERT INTO cdm.questionnaire_vibrent_forms
-SELECT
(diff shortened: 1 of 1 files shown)


rdr_service/etl/etl.sql
```diff
@@ -529,6 +529,37 @@ CREATE TABLE cdm.dose_era
...
+);
+
+INSERT INTO cdm.questionnaire_vibrent_forms
+SELECT
+    qh.questionnaire_id,
+    qh.version,
+    json_unquote(json_extract(convert(qh.resource using utf8), '$.identifier[0].value')) vibrent_form_id
+FROM questionnaire_history qh
+;
+
```
robabram: You should create a compound index on cdm.questionnaire_vibrent_forms with the 'questionnaire_id' and 'version' columns here so joinning is fast.

kenny-skaggs: Good idea, done!

kenny-skaggs:
Putting this on hold for now while investigating mysql issues that occurred when running the ETL sql.

PR #1952, opened 2020-09-08
[DA-1726] Research id generate tool

wangy70:
research_id_generator.py
Parameters: --type [import|new] --input-file [file_path], when the type parameter is "import", it will import the research IDs from the specified mapping file, when the type parameter is "new", it will generate research IDs for those participant who don’t have a research ID yet.


kenny-skaggs:
Will we only be generating research_ids with this tool, or is this just for backfilling and a later PR will have IDs auto-generated when a participant gets created?

wangy70:
@kenny-skaggs This tool is only for backfilling. The PR for new created participants is already merged: https://github.com/all-of-us/raw-data-repository/pull/1947

robabram (approved):

PR #1957, opened 2020-09-09
[PDR-123] Merge multiple full/partial questionnaire responses.

robabram:
In this PR I created a Biobank Unittest Mixin for easier creation of Biobank Order and Stored Sample records.  Its not automatically added to the Base Unittest class.

Update existing Resource unittests and created a new unittest.

### DIFF
--- a/rdr_service/dao/bq_participant_summary_dao.py
+++ b/rdr_service/dao/bq_participant_summary_dao.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import re
 
+from collections import OrderedDict
 from dateutil import parser, tz
 from dateutil.parser import ParserError
 from sqlalchemy import func, desc, exc
@@ -932,7 +933,8 @@ class BQParticipantSummaryGenerator(BigQueryGenerator):
     @staticmethod
     def get_module_answers(ro_dao, module, p_id, qr_id=None):
         """
-        Retrieve the most recent questionnaire module answers for the given participant id.
+        Retrieve the questionnaire module answers for the given participant id.  This retrieves all responses to
+        the module and applies/layers the answers from each response to the final data dict returned.
         :param ro_dao: Readonly ro_dao object
         :param module: Module name
         :param p_id: participant id.
@@ -951,7 +953,7 @@ class BQParticipantSummaryGenerator(BigQueryGenerator):
                                 INNER JOIN questionnaire_concept qc on qr.questionnaire_id = qc.questionnaire_id
                                 INNER JOIN questionnaire q on q.questionnaire_id = qc.questionnaire_id
                         WHERE qr.participant_id = :p_id and qc.code_id in (select c1.code_id from code c1 where c1.value = :mod)
-                        ORDER BY qr.created DESC;
+                        ORDER BY qr.created;
                     """
 
         _answers_sql = """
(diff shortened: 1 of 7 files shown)


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -951,7 +953,7 @@ def get_module_answers(ro_dao, module, p_id, qr_id=None):
                                 INNER JOIN questionnaire_concept qc on qr.questionnaire_id = qc.questionnaire_id
                                 INNER JOIN questionnaire q on q.questionnaire_id = qc.questionnaire_id
                         WHERE qr.participant_id = :p_id and qc.code_id in (select c1.code_id from code c1 where c1.value = :mod)
-                        ORDER BY qr.created DESC;
+                        ORDER BY qr.created;
```
kenny-skaggs: Should these be ordered by `authored` date? That way I think we'd get their latest answer, regardless of when it made it to the RDR system.

robabram: In this case using the `created` date is the right thing.  In the RDR responses are applied to the participant summary in the order they arrive in.  To replicate that behavior, the generators need to order by `created`, not `authored`.

pegbertsch: This may be in part an RDR data quality issue.  I  ran the `_module_info_sql`  query on prod for `p_id = 837367063`  (participant whose partial response history initially highlighted the need to handle partial responses) and `mod =  'ConsentPII'.`  Their initial consent response `594431920` produced multiple identical result rows.  Does the `INNER JOIN questionnaire_concept` also need to match on `questionnaire_version`?

Don't think it has a functional impact since we'd just be rebuilding the same dict entry with each pass through the for loop on 988, but wondered if there is potentially a cumulative performance impact when doing a full rebuild.

rdr_service/resource/generators/participant.py
```diff
@@ -935,45 +936,48 @@ def _calculate_test_participant(self, summary):
...
+                                   q.version,
+                                   qr.authored,
+                                   qr.language,
+                                   qr.participant_id
+                            FROM questionnaire_response qr
+                                    INNER JOIN questionnaire_concept qc on qr.questionnaire_id = qc.questionnaire_id
+                                    INNER JOIN questionnaire q on q.questionnaire_id = qc.questionnaire_id
+                            WHERE qr.participant_id = :p_id and qc.code_id in (select c1.code_id from code c1 where c1.value = :mod)
+                            ORDER BY qr.created;
+                        """
```
pegbertsch: See comment above about the `_module_info_sql `query

rdr_service/dao/bq_participant_summary_dao.py
kenny-skaggs: Regarding the order-by: ordering by created to match how the participant summary is built makes sense if they need to match. I'll make a note to investigate further and see if that's how participant summary should function.

rdr_service/resource/generators/participant.py
robabram: I've modified the query to return the distinct set of questionnaire responses.

pegbertsch (approved):

PR #1964, opened 2020-09-11
[DOCS] Only build documentation once when deploying to prod

kenny-skaggs:
I've noticed that deploying to prod or stable will kick off a pair of builds on ReadTheDocs. I'm not sure yet what's going on with the stable build, but it looks like ReadTheDocs automatically starts a build when we change the default branch it uses. So I modified the deploy script to skip a second trigger of the docs-build when it's already updated the default branch.

### DIFF
--- a/rdr_service/tools/tool_libs/app_engine_manager.py
+++ b/rdr_service/tools/tool_libs/app_engine_manager.py
@@ -283,12 +283,14 @@ class DeployAppClass(object):
         docs = ReadTheDocsHandler(api_token)
 
         try:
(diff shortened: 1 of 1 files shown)


pegbertsch (approved):

kenny-skaggs:
Peggy and I talked a bit on slack. I'm realizing now that I'm not entirely sure what's causing what seem like extra builds of the documentation. My changes would fix things if there were multiple 'latest' builds, but we're only seeing one of those when expected. The only builds I'm not able to pin down to an event are the stable builds. Closing this PR and will continue to investigate.

PR #1972, opened 2020-09-16
[PDR-106] Add EHR status fields to PDR/BigQuery participant summary data

pegbertsch:
This adds the logic to populate fields that were already in the BQ pdr_participant schema, but were not being populated.  It's not ideal to have to pull the information out of the RDR participant_summary table, but it's the only source we have right now and PTSC needs this EHR status data in PDR.  This is an initial workaround to get us going until we have an alternate solution that is not dependent on RDR participant_summary.

The UpdateEhrStatus daily cron job will now kick off PDR/BigQuery rebuild requests for the affected participants from that day's job.

Some minor refactoring to move some code that handles batching the rebuild cloud task requests into its own utility method, as it was being replicated again for the UpdateEhrStatus cron job and was already used in a few other places.

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -6,7 +6,6 @@ from collections import OrderedDict
 from dateutil import parser, tz
 from dateutil.parser import ParserError
 from sqlalchemy import func, desc, exc
-from sqlalchemy.orm import load_only
 from werkzeug.exceptions import NotFound
 
 from rdr_service import config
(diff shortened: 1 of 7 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -167,6 +171,35 @@ def _prep_participant(self, p_id, ro_session):
...
+            ehr_status = EhrStatus(ps.ehrStatus)
+            data = {
+                'ehr_status': str(ehr_status),
+                'ehr_status_id': int(ehr_status),
+                'ehr_receipt': ps.ehrReceiptTime,
+                'ehr_update': ps.ehrUpdateTime
+            }
+
+        return data
+
```
kenny-skaggs, line 202: @robabram, I think we've talked about this before, but I don't remember what you've said about it. Would it make sense to have a common module for the Resource generator and BigQuery DAO to share code like this, rather than having duplication between them? If there's field naming difference them then maybe mapping code would help.

robabram: The BigQuery DAO stuff was supposed to be gone by now and replaced by the Resource APIs.

robabram (approved):

PR #1987, opened 2020-09-22
[PDR-106] Add ehr_status, ehr_receipt_time and ehr_update_time fields to

robabram:
This PR implements a mechanism for partial updates to BQ/Resource data.  It also enables partial update process this for the nightly EHR receipt cron job.

### DIFF
--- a/.idea/runConfigurations/Unittests__Partial.xml
+++ b/.idea/runConfigurations/Unittests__Partial.xml
@@ -14,7 +14,7 @@
     <option name="ADD_SOURCE_ROOTS" value="true" />
     <EXTENSION ID="PythonCoverageRunConfigurationExtension" runner="coverage.py" />
     <option name="SCRIPT_NAME" value="unittest" />
-    <option name="PARAMETERS" value="discover -s tests -p &quot;*questionnaire*&quot;" />
+    <option name="PARAMETERS" value="discover -s tests -p &quot;*update_ehr_status*&quot;" />
     <option name="SHOW_COMMAND_LINE" value="false" />
     <option name="EMULATE_TERMINAL" value="false" />
     <option name="MODULE_MODE" value="true" />
--- a/rdr_service/offline/update_ehr_status.py
+++ b/rdr_service/offline/update_ehr_status.py
@@ -7,6 +7,7 @@ from rdr_service.cloud_utils import bigquery
 from rdr_service.dao.ehr_dao import EhrReceiptDao
 from rdr_service.dao.organization_dao import OrganizationDao
 from rdr_service.dao.participant_summary_dao import ParticipantSummaryDao
+from rdr_service.model.participant_summary import ParticipantSummary
 from rdr_service.participant_enums import EhrStatus
 from rdr_service.offline.bigquery_sync import dispatch_participant_rebuild_tasks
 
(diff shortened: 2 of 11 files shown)


rdr_service/offline/update_ehr_status.py
```diff
@@ -62,8 +63,16 @@ def update_participant_summaries_from_job(job):
...
                          format(count, total_rows, batch_size))
-            pids = [param['pid'] for param in parameter_sets]
-            dispatch_participant_rebuild_tasks(pids, batch_size=batch_size)
+            # pids = [param['pid'] for param in parameter_sets]
+            patch_data = [{
+                'pid': p['pid'],
+                'patch': {
+                    'ehr_status': str(EhrStatus.PRESENT),
+                    'ehr_status_id': int(EhrStatus.PRESENT),
+                    'ehr_receipt': now,
```
pegbertsch: The EHR receipt time is not going to get updated by the bulk update if the field already had a value.  It's supposed to be a "first seen" persistent value, I think.   So this can't be updated to 'now' every night.   We have to get the ehr_receipt value from the participant summary record.

rdr_service/resource/tasks.py
```diff
@@ -32,16 +32,18 @@ def batch_rebuild_participants_task(payload):
     batch = payload['batch']
 
     logging.info(f'Start time: {datetime.utcnow()}, batch size: {len(batch)}')
+    # logging.info(json.dumps(batch, indent=2))
```
pegbertsch, line 35: Suggest maybe keeping this enabled as logging.debug data, instead of info?

rdr_service/tools/tool_libs/resource_tool.py
```diff
@@ -383,7 +499,7 @@ def run():
                                 action="store_true")  # noqa
 
 
-    genomic_parser = subparser.add_parser("genomic")
+    genomic_parser = subparser.add_parser("genomic", )
```
pegbertsch: I am not sure what the effect of this change is;  just checking that it isn't an inadvertent edit?

pegbertsch (commented):
May try to pull this branch so I can step through some of the new code and understand it better, but except for the cron job issue, I don't think I saw anything significant

.idea/runConfigurations/Unittests__Partial.xml
```diff
@@ -14,7 +14,7 @@
     <option name="ADD_SOURCE_ROOTS" value="true" />
     <EXTENSION ID="PythonCoverageRunConfigurationExtension" runner="coverage.py" />
     <option name="SCRIPT_NAME" value="unittest" />
-    <option name="PARAMETERS" value="discover -s tests -p &quot;*questionnaire*&quot;" />
+    <option name="PARAMETERS" value="discover -s tests -p &quot;*update_ehr_status*&quot;" />
```
pegbertsch, line 17: May be a "don't care" but is this file a candidate for .gitignore?

rdr_service/tools/tool_libs/resource_tool.py
robabram: That comma shouldn't be there, I'll remove it.

rdr_service/resource/tasks.py
robabram: We don't want to ever show the payloads in Production, so I would be careful and not enable this in Prod.

rdr_service/offline/update_ehr_status.py
```diff
@@ -59,11 +61,25 @@ def update_participant_summaries_from_job(job):
...
+            with summary_dao.session() as session:
+                cursor = session.query(ParticipantSummary.participantId, ParticipantSummary.ehrReceiptTime).all()
+                records = [r for r in cursor if r.participantId in pids]
+
+            patch_data = [{
+                'pid': rec.participantId,
+                'patch': {
+                    'ehr_status': str(EhrStatus.PRESENT),
+                    'ehr_status_id': int(EhrStatus.PRESENT),
+                    'ehr_receipt': rec.ehrReceiptTime if rec.ehrReceiptTime else now,
```
pegbertsch, line 79: Is always using rec.ehrReceiptTime appropriate?  Theoretically, if we're here, unless the bulk update failed this field in participant summary should always be populated for all the pids this cron job processed.

If there's an RDR bug/glitch where it's null for some reason, is it preferable to populate the value for PDR anyway?  (RDR/PDR would be out of sync but maybe that's intentional).

rdr_service/offline/update_ehr_status.py
robabram: Fixed this in new commit.

pegbertsch (approved):
Approving based on offline discussion that the unit tests were able to verify the new generator code.  

Also, please see comment about giving ehr_receipt a default value in the cron job if no value was read from RDR participant_summary table.  Just want to confirm that's intentional before this is merged.

.idea/runConfigurations/Unittests__Partial.xml
robabram: Sorry, forgot to revert that one before adding it to the commit.

rdr_service/offline/update_ehr_status.py
robabram: I think it would get corrected later on if there was an issue.

PR #2007, opened 2020-09-28
[DA-1754] MIGRATION - New AW4 manifest and genomic_set_member column: qc_status

j-kanuch:
This PR adds a column to the AW4 ingestion workflow for both Arrays and Whole Genome Sequencing. This PR also adds a column to the `genomic_set_member` table to store the new AW4 column.


pegbertsch (commented):
Does the new status column need to be added to the BQ/PDR  model and generator(s)?

j-kanuch:
@pegbertsch Yes, I will make a separate PR for adding to BQ/PDR and generators. I need to add the new A3 to that as well, which was added to the genomic_set_member table on Friday.

j-kanuch:
@pegbertsch I guess it wouldn't hurt to just add the

kenny-skaggs (approved):

PR #2009, opened 2020-09-29
[DA-1735] Updates for ETL script and export tool

kenny-skaggs:
This adds some functionality to the ETL script for removing duplicate answers from the results and using authored date rather than created date when available. For removing duplicate answer we create a helper table that has all authored dates for a participant's response to a question code. Duplicate answers are removed from the src_clean table by joining it to the response table and finding if there are any responses that are later: if there aren't then it's included in the results, if there are then it's left out. The COPE surveys all use the same module code, so to account for them I look for later answers by question code ids and external id's instead. The ETL script is also updated with a fix for setting the transaction isolation level. Without specifying SESSION, it was leaving the isolation level unmodified and locking the questionnaire response and answer records when performing the "insert into ... select ..." query.

I've also updated the `curation` tool so that we can use the `gcloud sql export csv` utility. This essentially instructs the Google Platform to store the results of a SQL query in a bucket. Using that will let us avoid having to export the tables to our computers and upload them manually.

### DIFF
--- a/rdr_service/etl/etl.sql
+++ b/rdr_service/etl/etl.sql
@@ -537,17 +537,9 @@ DROP TABLE IF EXISTS cdm.questionnaire_vibrent_forms;
 CREATE TABLE cdm.questionnaire_vibrent_forms (
     questionnaire_id            bigint,
     version                     bigint,
-    vibrent_form_id             varchar(1024),
+    vibrent_form_id             varchar(200),
     INDEX vibrent_form_index    (questionnaire_id, version)
 );
-
-INSERT INTO cdm.questionnaire_vibrent_forms
-SELECT
-    qh.questionnaire_id,
-    qh.version,
-    json_unquote(json_extract(convert(qh.resource using utf8), '$.identifier[0].value')) vibrent_form_id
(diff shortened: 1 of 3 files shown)


wangy70:
Does curation team only need the latest answers or they also need the history answers(even they may be the same answer)?

kenny-skaggs:
I think only the latest answers, regardless of whether they're the same or not. That was discussed in the comments of https://precisionmedicineinitiative.atlassian.net/browse/ROC-652. Here's the text of the ask: "Surveys de-duplicated, eg only one each of the first six survey QRs and one each of the three COPE month surveys (May, June, July)".

kenny-skaggs:
This was constructed to account for partial responses I believed we were seeing, but for surveys with branching logic it includes responses that shouldn't exist in the final set of answers. I'm going to update this to use the latest questionnaire response (trusting that we're getting a complete response each time, and then I'll take a close look at the responses later to make sure that's the case).

kenny-skaggs:
Ok, it's updated.

wangy70 (approved):

PR #2010, opened 2020-09-30
[DA-1678] Active passive retention

wangy70:
1. Calculate the retentionType on the fly when calling participant summary API.
2. Support to filter by retentionType=[ACTIVE|PASSIVE] for participant summary API.

### DIFF
--- a/rdr_service/dao/participant_summary_dao.py
+++ b/rdr_service/dao/participant_summary_dao.py
@@ -36,6 +36,7 @@ from rdr_service.model.participant_summary import (
     SUSPENDED_OR_DECEASED_PARTICIPANT_FIELDS,
     WITHDRAWN_PARTICIPANT_FIELDS,
     WITHDRAWN_PARTICIPANT_VISIBILITY_TIME,
+    RETENTION_WINDOW
 )
 from rdr_service.model.patient_status import PatientStatus
 from rdr_service.model.utils import get_property_type, to_client_participant_id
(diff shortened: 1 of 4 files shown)


rdr_service/dao/participant_summary_dao.py
```diff
@@ -894,6 +928,98 @@ def add_to_sqlalchemy_query(self, query, field):
...
+                or_(
+                    ParticipantSummary.consentCohort != ParticipantCohort.COHORT_1,
+                    ParticipantSummary.consentForGenomicsRORAuthored == None,
+                    ParticipantSummary.consentForGenomicsRORAuthored <= eighteen_month_ago
+                ),
+                or_(
+                    ParticipantSummary.consentCohort != ParticipantCohort.COHORT_2,
+                    ParticipantSummary.consentForGenomicsRORAuthored == None,
+                    ParticipantSummary.consentForGenomicsRORAuthored <= eighteen_month_ago
+                )
```
kenny-skaggs, line 1005: Not a critical thing, but are these checks of "not having authored date OR having it be more than eighteen months ago" to make sure that a participant isn't ACTIVE? Would it work to have the passive filter below be set up as "passive_criterion AND NOT active_criterion"?

wangy70: Yeah, that's what I did before, but it doesn't work, the result is not what we expected. I didn't find the reason yet.

kenny-skaggs: I think I see why that is. Any comparison of NULL to something results in NULL. So `SELECT NULL <= 8` and `SELECT NOT(NULL > 8)` both give NULL, which is falsy. Coalesces in the active_criterion would fix that.

kenny-skaggs (approved):

rdr_service/dao/participant_summary_dao.py
wangy70: Ah, yes, I think that the reason.

PR #2012, opened 2020-09-30
[DA-1769] Sync codes for all environments at once

kenny-skaggs:
This does a little refactoring to let the CodeSync tool class detect if we want to sync codes for every environment (with the _all project option). If _all is given as the project then the tool connects to all environments and imports the codes from the Redcap project and then exports the codes for Prod to google drive.

### DIFF
--- a/rdr_service/services/gcp_config.py
+++ b/rdr_service/services/gcp_config.py
@@ -21,7 +21,7 @@ GCP_PROJECTS = [
     "aou-pdr-data-prod"
 ]
 
-GCP_INSTANCES = {
+GCP_INSTANCES = {  # List of RDR's GCP projects mapped to their database instance names
     "all-of-us-rdr-prod": "all-of-us-rdr-prod:us-central1:rdrmaindb",
     "all-of-us-rdr-stable": "all-of-us-rdr-stable:us-central1:rdrmaindb",
     "all-of-us-rdr-staging": "all-of-us-rdr-staging:us-central1:rdrmaindb",
@@ -31,7 +31,6 @@ GCP_INSTANCES = {
     "all-of-us-rdr-ptsc-1-test": "all-of-us-rdr-ptsc-1-test:us-central1:rdrmaindb",
     "all-of-us-rdr-ptsc-2-test": "all-of-us-rdr-ptsc-2-test:us-central1:rdrmaindb",
     "all-of-us-rdr-ptsc-3-test": "all-of-us-rdr-ptsc-3-test:us-central1:rdrmaindb",
-
 }
 
 GCP_REPLICA_INSTANCES = {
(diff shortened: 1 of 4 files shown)


robabram:
I want to mention some thoughts about the tool base design you have added.  The original design goal of the new style tools was to be clean, simple and avoid unnecessary abstractions. If you work through the older bash scripts, there is a large number of file dependencies just for one tool to execute.  This new tool design goal does add a few redundant lines of code in each file, but it also avoids adding default behaviors and additional file dependencies to each tool.

With your _tool_base.py file you have added a dependency on a simple base class that by default will always activate the SQL proxy and also a 'get_session()' function (which might go better in the GCP Environment class with the other DB functions). You have also added default argparser arguments to the 'cli_run()' function, now I need to look in two files to see what the arguments for a tool are.  Also, what if I don't want the default arguments in my tool?

I guess I'm looking from you is a good reason for adding this abstraction layer to the tools. 

PS. If you haven't noticed yet, I don't like over abstraction in a project and the RDR project is riddled with over abstraction.  If I have to look at 4+ files just to understand a bit of code, then I look at it as a bad design.

kenny-skaggs:
Thanks for the feedback! A good dialog like this will lead to a more solid design that will make maintenance and extension of the code easier. I feel similarly about abstraction when it's not robust and it's difficult to trust it to do what it needs to do. But if done right, abstraction can be a really powerful tool at our disposal. I have no idea how the `activate_sql_proxy` function works or how it does what it does, but it's obvious from the name why it should be used and what it does at a high-level. It's those kinds of good abstractions I hope we can build more of.

My approach to this has been to reduce duplication of code by setting up shared patterns/logic in a base class/file. For example, each of the tools use the GCPprocessContext. If something changes in the future about how we want to set up GCP in our tools, then we could potentially only need to change it in one place rather than across multiple files. I've also been going at it with an iterative mindset. I haven't built anything in yet for a tool to not have a `project` argument. But if the need comes up then I don't see a reason we couldn't build out a way to allow for that.

Most of the tools I've been writing needed the sql_proxy activated, so it made sense to have that as the default. But it doesn't always have to be activated. `CodesExportClass` doesn't use it so it doesn't call to super's run. Maybe there's a way to have the sub-classes more clearly define whether to activate the proxy or not?

I've mostly been addressing the main point of why I think we should add this abstraction layer: and that's to be able to follow the DRY software principle and potentially get to a point where we can create tools by adding new modules that mostly contain just the code specific to the tasks they're designed to perform.

robabram:
I believe DRY can be overdone to the point where it makes code maintenance difficult and time consuming.  I don't feel like removing a dozen or so lines from the tool file to a base class is an improvement in this case.  This article kinds of sums up my thinking here; https://goiabada.blog/dont-obsess-over-code-dryness-e9ecc5224ff

robabram (approved):

kenny-skaggs:
@robabram, I made a few changes after running the sync. Let me know if they're good to merge as well. I tried to refactor where the sql proxy gets created, but having it happen in the ContextManager or a separate Base class both felt like they weren't ideal either.

robabram (approved):

PR #2014, opened 2020-10-01
[DA-1760] MIGRATION - Add dev_note, ignore_flag fields to Genomic tables; expand tooling

j-kanuch:
This PR contains the following updates:
- Added 1 field to `genomic_set_member`: `dev_note`
- Added 2 fields to `genomic_gc_validation_metrics`: `ignore_flag`, `dev_note`
- Added the aforementioned fields to the BQ/Resource generators.
- Updated gc metrics queries to filter on `ignore_flag != 1`
- Updated genomic member-state tool to require a dev note when running.
- Created new GC metrics update tool that can update arbitrary fields in `genomic_gc_validation_metrics`.
- Updated the import data tool to look in the correct directory when generating test participants (fixes the error in tools/setup_local_database.sh).


rdr_service/tools/import_data.sh
```diff
@@ -15,5 +15,5 @@ rdr_client/run_client.sh $@ import_codebook.py
 echo "Importing questionnaires..."
 tools/import_questionnaires.sh $@
 echo "Importing participants..."
-tools/import_participants.sh --file test/test-data/healthpro_test_participants.csv $@
+tools/import_participants.sh --file ../tests/test-data/healthpro_test_participants.csv $@
```
kenny-skaggs, line 18: 👍

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -608,6 +612,12 @@ def run(self):
         self.gcp_env.activate_sql_proxy()
         self.dao = GenomicSetMemberDao()
 
+        # Document these changes in the dev_note field
+        dev_note = input("Please enter a developer note for these updates:\n")
+        if dev_note in (None, ""):
+            _logger.error('A developer note is required for these changes.')
+            return 1
```
kenny-skaggs, line 619: Just sort of curious nit-picking: why use `input` here over a required CLI argument?

j-kanuch: My thought process was to ensure that when we update data, we are deliberately adding in a specific comment. It sort of serves as a reminder that we're manipulating data outside of the normal process. 
However, I do see use for it to be an argument too, as this would allow us to call it from other scripts in an automated fashion, though I don't like the idea of us having to update data like this very often. If it isn't too big an issue, I'd prefer to keep it as is for the time being. If it becomes cumbersome, we can update it in the future.

kenny-skaggs: Not a big issue at all, and after that I'm preferring that it stay the way it is. I agree that this should hopefully be used infrequently. At first I was thinking of it being used in other scripts, but this is a good way of making sure we're being careful. Otherwise it might be too easy to re-run as a stored command without paying attention to the note stored.

kenny-skaggs (approved):

PR #2017, opened 2020-10-05
Blocking Questionnaire and QuestionnaireResponses that use unrecognized codes

kenny-skaggs:
This changes the CodeDao to keep questionnaires and responses from creating new codes in the database, giving a 400 error with a message detailing what codes were unrecognized.

The QuestionnaireResponseDao and QuestionnaireDao both used lists of codes gathered from their payloads, but previously only the QuestionnaireResponseDao evaluated the answer codes. I've added block to the QuestionnaireDao so it maps answer option codes as well. That way if any answer codes in invalid on the questionnaire we can catch them when the Questionnaire is created, rather than later when we see them in a response.

### DIFF
--- a/tests/dao_tests/test_questionnaire_dao.py
+++ b/tests/dao_tests/test_questionnaire_dao.py
@@ -1,9 +1,10 @@
 import datetime
 
 from sqlalchemy.exc import IntegrityError
-from werkzeug.exceptions import NotFound, PreconditionFailed
+from werkzeug.exceptions import BadRequest, NotFound, PreconditionFailed
 
 from rdr_service.clock import FakeClock
+from rdr_service.code_constants import PPI_SYSTEM
 from rdr_service.dao.code_dao import CodeDao
 from rdr_service.dao.questionnaire_dao import (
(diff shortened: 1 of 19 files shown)


tests/dao_tests/test_questionnaire_dao.py
```diff
@@ -205,3 +208,44 @@ def test_parsing_external_id(self):
         })
 
         self.assertEqual('FORM_1A', model.externalId)
+
+    def test_questionnaire_payload_cannot_create_new_codes(self):
```
pegbertsch: Suggestion only:  I wondered whether it's worth creating an API test as well that does a POST of a questionnaire_response payload that you expect to be rejected because it contains unrecognized codes that were not in the Questionnaire test setup?

pegbertsch (approved):
My confidence in my knowledge of all the code that manages the code maps, etc.,  is not great.   I don't know if there are downstream effects I might be missing.

But the changes themselves make sense to me.

pegbertsch (approved):
2nd approval for updated tests

PR #2020, opened 2020-10-06
[DA-1770] Scheduled subset of genomic cron jobs.

j-kanuch:
This PR changes the schedule of the following genomic cron jobs:
- Genomic Pipeline AW0 (Cohort 3) Workflow
- Genomic Pipeline AW0 (Cohort 1) Workflow
- Genomic AW1 Workflow
- Genomic AW1 Failures Workflow
- Genomic GEM A1-A2 Workflow
- Genomic GEM A2 Workflow
- Genomic AW3 Workflow
- Genomic AW4 Workflow

These cron jobs were previously set to never run unless manually triggered. This PR will set them to run at the scheduled time. This is to support the GEM launch, the restarting of biospecimen collection, and the DRC Broad QC workflows.

### DIFF
--- a/rdr_service/cron_careevo.yaml
+++ b/rdr_service/cron_careevo.yaml
@@ -3,19 +3,20 @@ cron:
 - description: Participant count metrics (Do not manually start)
 - description: Flag ghost participants
 - description: Genomic Pipeline AW0 (Cohort 2) Workflow
-- description: Genomic Pipeline AW0 (Cohort 1) Workflow (Manual)
-- description: Genomic AW1 Workflow (Manual)
-- description: Genomic AW1 Failures Workflow (Manual)
+- description: Genomic Pipeline AW0 (Cohort 3) Workflow
+- description: Genomic Pipeline AW0 (Cohort 1) Workflow
+- description: Genomic AW1 Workflow
+- description: Genomic AW1 Failures Workflow
 - description: Genomic AW1C (CVL) Workflow (Manual)
 - description: Genomic AW1CF (CVL) Failures Workflow (Manual)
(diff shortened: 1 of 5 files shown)


pegbertsch (commented):
I don't have a specific concern here, but how were the job run times chosen, and was there any need to account for concurrency with other cron jobs?

j-kanuch:
> I don't have a specific concern here, but how were the job run times chosen, and was there any need to account for concurrency with other cron jobs?

For the most part, the times were determined by a combination of the required frequency (daily, weekly, every 6 hours), and concurrence with other jobs, as well as an estimate on how long they would take and where they sit in the whole genomic workflow. That said, some of the times were more arbitrary than others.

The 3 AW0 jobs are only to run once a week, but they are resource intensive and I did not want them to interfere with the Biobank import or the metrics jobs, which are also resource intensive, so I attempted to stagger them and have them run as late as possible before regular business hours. 
The GEM jobs (A1-A2 and A3) need to run every 6 hours because PTSC runs a job every 6 hour and our job needs to run before each one of those.
The AW1 and AW1F jobs should run daily, and these may vary in resources. I once again tried to get them later in the morning, after the Biobank Import and the Metrics jobs. I tried to stagger them with the AW0 jobs mentioned above for the one time a week they both run.
The AW3 and AW4 jobs were the most arbitrary. They are dependent on the AW1 job, but are also a parallel process to the GEM jobs. I don't have much reason for the specified times other than to avoid concurrency with the other jobs and to have them run after the AW1 job mentioned above as this is required by the genomic workflow.

pegbertsch (approved):

PR #2022, opened 2020-10-07
Explicitly releasing deceased report locks

kenny-skaggs:
The logs are showing that deceased report requests are erring out when trying to create multiple deceased reports for the same participant. That's how they should function, but the error they're returning doesn't look right.

When a named lock is taken by a transaction, the lock is supposed to be implicitly released when the transaction ends. So when two requests try to take the same lock one should get it and release it when done, then the next should get it and then see that the report is already there. So I wouldn't expect any scenarios when an attempt to obtain a lock times out.

In atleast one instance we see a request arrive to create a deceased report, then a second or two later the same request is made. The first request completes successfully, but the second request is unable to obtain the lock and fails on the timeout. It seems as if the locks aren't being released in the way I would expect. So this explicitly releases the lock when the updates to the database are complete.

I also found that attempting to return an error when within a session context seems to keep the request_log record from being created. I'm not sure yet how to handle that, but I made a note about it.

### DIFF
--- a/rdr_service/dao/deceased_report_dao.py
+++ b/rdr_service/dao/deceased_report_dao.py
@@ -554,12 +554,27 @@ class DeceasedReportDao(UpdatableDao):
         return observation.as_json()
 
     @staticmethod
+    def _deceased_report_lock_name(participant_id):
+        return f'rdr.deceased_report.p{participant_id}'
+
+    @staticmethod
+    def _release_report_lock(session, participant_id):
+        release_result = session.execute(
+            f"SELECT RELEASE_LOCK('{DeceasedReportDao._deceased_report_lock_name(participant_id)}')"
+        ).scalar()
+
+        if release_result is None:
+            logging.error(f'Deceased report lock did not exist for P{participant_id}!')
+        elif release_result == 0:
+            logging.error(f'Deceased report lock for P{participant_id} was not taken by this thread!')
+
+    @staticmethod
     def _can_insert_active_report(session, participant_id, lock_timeout_seconds=30):
         # Obtain lock for creating a participant's deceased report
         # If the named lock is free, 1 is returned immediately. If the lock is already taken, then it waits until it's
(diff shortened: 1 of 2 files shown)


robabram:
Can you explain the reasoning behind explicitly creating and releasing table locks here?  By default our MySQL instances set a table lock when trying to write/update any table.

kenny-skaggs:
I think I'm using named locks here instead of table locks (https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock).

Only one approved or pending deceased report (what I'll refer to as 'active') can exist for each participant (while multiple denied reports are allowed). New deceased reports will always be in an active state, so to insert a deceased report we first need to check whether there is already an active report for that participant. After the check, we then perform the insert. It was possible for two requests on separate instances to get past the check and think they're both inserting the only active deceased report, ending with a state of two active reports for one participant.

The lock acts as a mutex that can be shared across GAE instances and lets us ensure only one report is created (while any other concurrent requests made to wait on the lock would have to wait until the first completes before checking if they can create a deceased report). I could have used a table lock, but this feels better because it lets multiple deceased reports for different participants get processed concurrently.

Can you confirm that we use table locks by default when writing/updating records in a table? I think mysql out-of-the-box uses gap, index, or record locks depending on the query. Are we using table locks for updates to participant summary?

robabram:
Ok, I think your right about MySQL GAP locks.  It seems to me that the "table + history table" model we have used previously in RDR would take care of this with out introducing extra complexity.  In the "table + history table" model, there is only the most recent (only one) record per participant in the "table" and all activity is recorded in the "history table" using DB triggers.  So when updates are coming in from two or more AE instances, they simple overwrite the "table" record one at a time and the activity is all written to the history table.  So even if there are multiple requests coming in together at the millisecond level, MySQL's row locking will prevent data from being lost or an invalid data state for the record type from occurring.

robabram (approved):

PR #2023, opened 2020-10-07
[ROC-687] Fixes for data cleaning and header formatting

j-kanuch:
This PR fixes some errors in the CDM to CSV export process:
- Column headers were appearing in locations other than row 0
- Newline characters in the data were causing the CSV to have incomplete rows
- Double-quote characters in the data would be passed into the CSV, potentially breaking the formatting

### DIFF
--- a/rdr_service/tools/tool_libs/curation.py
+++ b/rdr_service/tools/tool_libs/curation.py
@@ -76,12 +76,37 @@ class CurationExportClass(object):
         # https://issuetracker.google.com/issues/64579566
         # NULL characters (\0) can also corrupt the output file, so they're removed.
         # And whitespace was trimmed before so that's moved into the SQL as well
-        field_list = [f"TRIM(REPLACE(COALESCE({name}, ''), '\\0', ''))" for name in column_names]
+        # Newlines and souble-quotes are also replace with spaces and single-quotes, respectively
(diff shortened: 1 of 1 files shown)


rdr_service/tools/tool_libs/curation.py
```diff
@@ -76,12 +76,37 @@ def export_table(self, table):
...
+        # Newlines and double-quotes are also replaced with spaces and single-quotes, respectively
+        field_list = [f"TRIM(REPLACE(REPLACE(REPLACE(COALESCE({name}, ''), '\\0', ''), '\n', ' '), '\\\"', '\\\''))"
+                      for name in column_names]
+
+        # Unions are unordered, so the headers do not always end up at the top of the file.
+        # The below format forces the headers to the top of the file
+        # This is needed because gcloud export sql doesn't support column headers and
+        # Curation would like them in the file for schema validation (ROC-687)
+        sql_string = f"""
+            SELECT {','.join(field_list)}
```
kenny-skaggs: Minor thing: but could this be the column name list? That way it doesn't have to run the trims and replaces against all the data again.

j-kanuch: Yes, I see what you are saying. I have updated and pushed the new version with `column_names` instead of `field_list`.

kenny-skaggs: Cool!

kenny-skaggs (approved):

PR #2024, opened 2020-10-07
[ROC-697] bug fix: update participant summary specimen site

wangy70:
Biobank order doesn't have attribute `barcode`, this criteria will always be True. It will cause the participant summary specimen site info be overwrite by salivary order and set to None. Fix it by seeing if there is any values for the site info.


kenny-skaggs:
Should there be a unit test for this?

wangy70:
Good idea, test case added.

kenny-skaggs (approved):

PR #2026, opened 2020-10-07
[PDR-76] Simplify resource schema definitions.

robabram:


pegbertsch (commented):
No specific concerns but I don't have much familiarity with Marshmallow usage yet.  Are the existing unit tests enough to give adequate coverage for these changes, and/or are there other tool-based tests that are providing coverage and validation?

robabram:
The unittests do cover the changes, because there was really no change in the information stored in a Schema.  These changes are primarily a simplification and optimization of how a Schema is defined and there was almost no change to any code that uses a Schema.

pegbertsch (approved):

PR #2027, opened 2020-10-08
[DA-1776] Code validation when importing from redcap

kenny-skaggs:
This adds the following checks when importing codes from Redcap:
- Any question codes that are single- or multi-select need to have at least one answer option
- Code values must only contain alphanumeric characters or the underscore
- Each survey must have a module code

If any of them are untrue, then the import is cancelled and error messages are printed out.

### DIFF
--- a/rdr_service/tools/tool_libs/codes_management.py
+++ b/rdr_service/tools/tool_libs/codes_management.py
@@ -327,10 +327,11 @@ class CodesSyncClass(ToolBase):
                     logger.error(f'The following question codes are missing answer options: {code_values_str}')
                     exit_code = 1
 
-            if exit_code == 1:
-                session.rollback()
-            elif not self.args.dry_run and not skip_file_export_write:
-                self.write_export_file(session)
+            if not self.args.dry_run:
+                if exit_code == 1:
+                    session.rollback()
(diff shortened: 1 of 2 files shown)


rdr_service/tools/tool_libs/codes_management.py
```diff
@@ -266,21 +297,42 @@ def run(self, skip_file_export_write=False):
...
+                    code_values_str = ", ".join([f'"{value}"' for value in invalid_codes_found])
+                    logger.error(f'Invalid code values found: {code_values_str}')
+                    exit_code = 1
+
+                if questions_missing_options:
+                    code_values_str = ", ".join([f'"{value}"' for value in questions_missing_options])
+                    logger.error(f'The following question codes are missing answer options: {code_values_str}')
+                    exit_code = 1
+
+            if exit_code == 1:
```
pegbertsch, line 330: I'm a little confused as to whether an update that needs to be rolled back could have occurred before calling write_export_file?

pegbertsch (approved):
A second reviewer wouldn't hurt since I'm not fully caught up on the structure of the REDCap data we're consuming.  But no specific concerns.

wangy70 (approved):

PR #2031, opened 2020-10-12
[DA-1785] support_filter_by_unset_for_retention_type

wangy70:


rdr_service/dao/participant_summary_dao.py
```diff
@@ -1015,6 +1015,19 @@ def add_to_sqlalchemy_query(self, query, field):
...
+                        field == RetentionStatus.NOT_ELIGIBLE,
+                        and_(
+                            passive_criterion,
+                            or_(
+                                ParticipantSummary.ehrReceiptTime == None,
+                                ParticipantSummary.ehrReceiptTime <= eighteen_month_ago
+                            )
+                        )
+                    )
+                )
```
kenny-skaggs, line 1030: I remember this slack conversation and this might line up with what I thought was right at the time, but looking back over everything I think Caroline's comment was a good clarification of everything: "A participant who is not eligible will always be not retained, but it is possible for someone who is eligible to also be not retained."

So she's saying that participants can be retention eligible, but not ACTIVE or PASSIVE. This is what I'm understanding now from the PRD:
- Participants are retention eligible if they meet some defined set of criteria (details not relevant here)
- Participants have an ACTIVE retention type if they are retention eligible and have done something something from a set of actions within the last 18 months
- Participants have a PASSIVE retention type if they are retention eligible and are not active, **but have EHR data in the last 18 months**

That last part of the definition of the PASSIVE type leaves another set of participants that _are_ retention eligible, don't have the ACTIVE retention type, and have not had EHR data in the last 18 months. That seems to fit with Caroline's comment and I believe it's that set of retention eligible participants that they're looking for with the UNSET retention type. (sorry for the really long comment)

wangy70: Do you mean when filter the retention type with `UNSET`, we need to exclude the participants who is not eligible?

kenny-skaggs: Yeah. I could be wrong, but I think filtering with retention type UNSET should give participants that are retention eligible but don't have the PASSIVE or ACTIVE retention type.

wangy70: This is a good point. I will check with Caroline about it.

wangy70: Just checked with Caroline, she wants to include no_eligible participants when filter retention type with `UNSET`.

kenny-skaggs: We talked it out on slack. I was mis-reading the sql alchemy and didn't realize that filters for eligible status could be used in tandem.

kenny-skaggs (approved):

PR #2034, opened 2020-10-13
[DA-1604] [Migration] Saving partial survey responses

kenny-skaggs:
This sets us up to be able to store partial questionnaire responses. We save the status of the response so that we can pass this on to curation. And if the status is anything other than COMPLETED then we don't use the response to update the participant summary.


j-kanuch (commented):
I have a question regarding partial questionnaire responses definition and the requirements. Does "partial" mean that it only has a subset of the questionnaire questions answered? 
Is the expectation then that the RDR stores multiple `in-progress` questionnaire responses, but only updates the participant summary if a questionnaire response is sent marked as `completed`?

kenny-skaggs:
Yeah, they would only have a portion of the questions answered. This is being set up for COPE surveys that users never finished, so we can get that through to curation. But it could apply to other questionnaires.

And yeah, participant summary data is from surveys that participants have 'submitted', so we wouldn't want to store anything there that hasn't been explicitly finalized, and we get those finalized questionnaire responses with the `completed` status.

j-kanuch:
OK, that makes sense. Thanks for the clarification.

j-kanuch (approved):

PR #2038, opened 2020-10-15
[DA-1790] MIGRATION - Store File Upload Date for Genomics files in genomic_file_processed

j-kanuch:
The PDR and SSAD teams require the dates that AW1 and AW2 manifest files were uploaded to the GCS buckets by the Biobank and GCs, respectively.  This data is not available in the RDR as a field for the analytics team to query. To support this, the this PR includes the following changes:
- The `upload_date` field was added to the `genomic_file_processed` table. 
- The file ingestion process now records the `upload_date`, which is based on the "updated" property of the GCS file object.
- Since existing data needs to be back-filled, this PR also includes a tool to back-fill the `upload_date` on previously processed files.
- Additionally, the Stable environment no longer needs the genomic ROC ticket alert, since we are no longer E2E testing this portion of the workflow on Stable, so the Stable environment was removed from the alert handler.

A subsequent PR will implement the `genomic_file_processed` model in the PDR BQ schemas and Resource generators.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -122,27 +123,29 @@ def generate_file_processing_queue(self):
...
                     self.job_run_id,
                     file_path,
                     self.bucket_name,
-                    file_name.split('/')[-1])
+                    file_data[0].split('/')[-1],
+                    upload_date=file_data[1])
 
                 self.file_queue.append(new_file_record)
 
     def _get_uningested_file_names_from_bucket(self):
```
kenny-skaggs: Not a huge issue, but should this be renamed for better understandability? I know the docstring has been updated, but the name still seems to imply that just names are being returned.

Otherwise it looks good and I'll approve it if you want to keep the name as is.

j-kanuch: I agree, I will change it to `_get_new_file_names_and_upload_dates_from_bucket()`.

kenny-skaggs (approved):

PR #2039, opened 2020-10-15
[DA-1782] add_enrollment_site_to_ps_api

wangy70:
`enrollment_site_id` is already in participant table, add it to participant_summary table now.


rdr_service/model/participant_summary.py
```diff
@@ -272,6 +272,11 @@ def organizationId(cls):
     @declared_attr
     def siteId(cls):
         return Column("site_id", Integer, ForeignKey("site.site_id"))
+
+    @declared_attr
+    def enrollmentSiteId(cls):
+        return Column("enrollment_site_id", Integer, ForeignKey("site.site_id"))
+
```
kenny-skaggs: The `site` field below was something I had set up for docs and ideally would stay near the `siteId` field.

Since the API removes the id fields and replaces them (so siteId becomes site, and organizationId becomes organization) I added fields as placeholders so we could put docstrings on them and have them picked up in the documentation builds.

What do you think of moving enrollmentSiteId a bit to keep the `site` and `siteId` fields together, and also adding an `enrollmentSite` placeholder with a docstring beneath it to further describe what the field contains (so it would follow the pattern of what `organizationId` and `organization` do above)?

wangy70: Thanks. It's done.

kenny-skaggs: 👍

kenny-skaggs (approved):

PR #2040, opened 2020-10-15
[PDR-122] Add biobank order/stored sample data to BQ PDR generators

pegbertsch:
The biobank data added is based on what was already being generated by the `_prep_biobank_info()` method in the generators, but was not included in the participant resource JSON data until now (see discussion in PDR-122).

The `_biobank_orders_sql` query defined/executed in the `_prep_biobank_info()` method was corrected to eliminate multiple result rows for the same `biobank_order_id`, in cases where the same order id value appeared multiple times in the `biobank_order_identifier` table.    **NOTE:  Please review the other unchanged queries in  `_prep_biobank_info()` to confirm my visual check that no other queries needed related changes.**   

For stored samples that have no associated biobank order, setting the biobank_order_id to 'UNSET' (instead of the previously used nonstandard string value).  None/NULL was not used because of possible side effects on other places in the generator/pid rebuild path that might use the `bbo_biobank_order_id` key from the resource data and expect a string value.

The `migrate-bq` tool fix resolves an issue where if BQ table/view names are substrings of what the user specified with the --names option, the tool might "match" the wrong table/view when testing for` in self.args.names.lower()`

### DIFF
--- a/rdr_service/dao/bq_participant_summary_dao.py
+++ b/rdr_service/dao/bq_participant_summary_dao.py
@@ -12,7 +12,7 @@ from werkzeug.exceptions import NotFound
 from rdr_service import config
 from rdr_service.code_constants import CONSENT_GROR_YES_CODE, CONSENT_PERMISSION_YES_CODE, CONSENT_PERMISSION_NO_CODE, \
     DVEHR_SHARING_QUESTION_CODE, EHR_CONSENT_QUESTION_CODE, DVEHRSHARING_CONSENT_CODE_YES, GROR_CONSENT_QUESTION_CODE, \
-    EHR_CONSENT_EXPIRED_YES, UNKNOWN_BIOBANK_ORDER_ID
+    EHR_CONSENT_EXPIRED_YES
 from rdr_service.dao.bigquery_sync_dao import BigQuerySyncDao, BigQueryGenerator
 from rdr_service.model.bq_base import BQRecord
 from rdr_service.model.bq_pdr_participant_summary import BQPDRParticipantSummary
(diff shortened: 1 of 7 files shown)


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -522,6 +522,7 @@ def _make_stored_sample_dict_from_row(bss_row, has_order=True):
                         where bss2.biobank_order_identifier = boi.`value`) as tests_stored
              from biobank_order bo left outer join biobank_order_identifier boi on bo.biobank_order_id = boi.biobank_order_id
              where boi.`system` = 'https://www.pmi-ops.org' and bo.participant_id = :pid
```
kenny-skaggs, line 524: I think this is tangential to your changes, but might be worth discussion. Will this need to include orders that are coming from Quest/CE?

rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -522,6 +522,7 @@ def _make_stored_sample_dict_from_row(bss_row, has_order=True):
...
+                   and boi.`value` in (select bss.biobank_order_identifier from biobank_stored_sample bss)
```
kenny-skaggs, line 525: Was this change only intended to reduce duplication? Is it also meant to filter out orders that don't have stored samples?

pegbertsch:
After discussing the issues raised by Kenny with him and Robert, and examining some edge cases/talking about upcoming process for Quest/CE biobank orders, I'm closing this PR to rework/refactor the `_prep_biobank_info() ` code.

PR #2043, opened 2020-10-16
[DA-1792] [MIGRATION] Renaming October cope survey to November

kenny-skaggs:
Changing the name in participant summary for the October COPE Survey to November to match the program-level naming convention for the survey. Also updating any references for the October survey I could find in the BQ code.


pegbertsch (approved):
We may need to do some manual cleanup of the BigQuery instances with the migrate-bq tool (--delete) if they already had a now-deprecated pdr_mod_cope_oct table created.   It may only affect sandbox.

kenny-skaggs:
Ok, thanks @pegbertsch. I'll make a note to remember that and coordinate with you as it's being deployed. Is there potentially a way to automate the migration?

PR #2046, opened 2020-10-20
[DA-1795] Update A1 Manifest site ID to UPPERCASE; add PDR generators to genomic file tool

j-kanuch:
This PR updates the site ID column of the A1 manifest to be uppercase. An unrelated change was made to the genomic file processed back fill tool as well to add the PDR generators.

### DIFF
--- a/tests/cron_job_tests/test_genomic_pipeline.py
+++ b/tests/cron_job_tests/test_genomic_pipeline.py
@@ -1789,7 +1789,7 @@ class GenomicPipelineTest(BaseTestCase):
             self.assertEqual("yes", rows[0]['consent_for_ror'])
             self.assertEqual(test_member_1.consentForGenomicsRORAuthored, parse(rows[0]['date_of_consent_for_ror']))
             self.assertEqual(test_member_1.chipwellbarcode, rows[0]['chipwellbarcode'])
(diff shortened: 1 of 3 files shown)


tests/cron_job_tests/test_genomic_pipeline.py
```diff
@@ -1789,7 +1789,7 @@ def test_gem_a1_manifest_end_to_end(self):
             self.assertEqual("yes", rows[0]['consent_for_ror'])
             self.assertEqual(test_member_1.consentForGenomicsRORAuthored, parse(rows[0]['date_of_consent_for_ror']))
             self.assertEqual(test_member_1.chipwellbarcode, rows[0]['chipwellbarcode'])
-            self.assertEqual(test_member_1.gcSiteId, rows[0]['genome_center'])
+            self.assertEqual('JH', rows[0]['genome_center'])
```
pegbertsch, line 1792: Was this an intended change to a hardcoded string, rather than something like test_member_1.gcSiteId.upper()?

pegbertsch (approved):
Approved assuming the hardcoded string in the test case was an intended change.

tests/cron_job_tests/test_genomic_pipeline.py
j-kanuch: Yes, this was intended. Thanks for checking!

PR #2048, opened 2020-10-21
[ROC-711] Split EHR Status cron job; catching exceptions on BQ rebuild.

j-kanuch:
This PR splits the UpdateEhrStatus cron job into two separate jobs:
1. UpdateEhrStatusOrganization
  - This updates the Organizations receipt time in the `ehr_receipt` table.
2. UpdateEhrStatusParticipant
  - This updates the participants' `ehr_receipt_time` in the `participant_summary` table

UpdateEhrStatusParticipant was also updated to catch exceptions during the BQ rebuild call for debugging the 502 failures we are seeing during the nightly run.

### DIFF
--- a/rdr_service/offline/update_ehr_status.py
+++ b/rdr_service/offline/update_ehr_status.py
@@ -92,13 +92,13 @@ def update_participant_summaries_from_job(job):
                 dispatch_participant_rebuild_tasks(patch_data, batch_size=batch_size)
 
             except BadGateway as e:
-                LOG.error(f'Bad Gateway: {e}')
+                LOG.error(f'Bad Gateway: {e}', exc_info=True)
 
             except InternalServerError as e:
-                LOG.error(f'Internal Server Error: {e}')
+                LOG.error(f'Internal Server Error: {e}', exc_info=True)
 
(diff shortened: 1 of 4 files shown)


rdr_service/offline/update_ehr_status.py
```diff
@@ -79,7 +88,17 @@ def update_participant_summaries_from_job(job):
...
+                dispatch_participant_rebuild_tasks(patch_data, batch_size=batch_size)
+
+            except BadGateway as e:
+                LOG.error(f'Bad Gateway: {e}')
+
+            except InternalServerError as e:
+                LOG.error(f'Internal Server Error: {e}')
+
+            except HTTPException as e:
+                LOG.error(f'HTTP Exception: {e}')
```
kenny-skaggs: What do you think of passing `exc_info=True` to the logs so they’ll print out full stack traces too? (https://docs.python.org/3/library/logging.html#logging.Logger.debug)

j-kanuch: Yes, I will add that to the logs as well.

kenny-skaggs (approved):

PR #2050, opened 2020-10-22
[PDR-122] Add biobank order and status information to PDR (refactored)

pegbertsch:
This is a refactor of the original PDR-122 changes reviewed in [PR #2040](https://github.com/all-of-us/raw-data-repository/pull/2040) (closed/discontinued)

The `_prep_biobank_info()` method in the generators was refactored to simplify the complex SQL queries that could sometimes produce incorrect results (see PR 2040).  The logic for assembling the biobank order/sample data now exists primarily in the Python code.  

It looks for all the biobank orders associated with the participant, all the biobank stored samples (with or without an associated biobank order) for the participant's biobank ID, and then assembles all the biobank order / ordered sample / stored sample history into the participant's PDR resource data.

Note that the nested biobank samples (bbo_samples) data in the participant's PDR resource data can contain samples that have been ordered but do not yet have a stored sample record (as well as "orderless" stored samples). The related participant ` biospec ` PDR data can still be generated exactly as it was before.


rdr_service/model/bq_pdr_participant_summary.py
```diff
@@ -309,3 +312,58 @@ class BQPDRPatientStatuesView(BQView):
...
+               nt.bbs_confirmed,
+               nt.bbs_status,
+               nt.bbs_status_id,
+               nt.bbs_disposed,
+               nt.bbs_disposed_reason,
+               nt.bbs_disposed_reason_id
+           FROM (
+              SELECT *,
+                  ROW_NUMBER() OVER (PARTITION BY participant_id ORDER BY modified desc, test_participant desc) AS rn
+                FROM `all-of-us-rdr-sandbox`.rdr_ops_data_view.pdr_participant
```
kenny-skaggs: Should this be referencing sandbox?

rdr_service/resource/generators/participant.py
```diff
@@ -594,59 +619,29 @@ def _make_stored_sample_dict_from_row(bss_row, has_order=True):
...
+        # pseudo order with an order id of 'UNSET'
+        if len(bss_missing_orders):
+            orderless_stored_samples = list()
+            for bss_row in bss_missing_orders:
+                orderless_stored_samples.append(_make_sample_dict_from_row(bss=bss_row, bos=None))
+
+            orders.append({
+                'biobank_order_id': 'UNSET',
+                'tests_stored': len(orderless_stored_samples),
+                'samples': orderless_stored_samples
```
kenny-skaggs, line 640: I'm confused about some of the functionality here. Here (with the samples that don't have orders) the `samples` field is set to the list of stored samples. But up on line 624 the stored samples aren't put in the structure for orders that have samples, `samples` is set to an empty list. But in bq_participant_summary_dao.py, both structures (orders that have samples, and orderless samples) have the stored samples in the `samples` list. Is the difference intentional?

kenny-skaggs (commented):
I left some comments. I mostly focused on bq_participant_summary.py and generators/participant.py. @robabram should probably be the one to approve the changes for schemas and migrations.

rdr_service/model/bq_pdr_participant_summary.py
pegbertsch: Good catch!  I think that was an unintentional cut/paste bug.  Fixing now

rdr_service/resource/generators/participant.py
pegbertsch: That empty list was a bug.  Fixed in later commit

robabram (approved):

PR #2051, opened 2020-10-22
Logging HTTPExceptions in offline app

kenny-skaggs:
It looks like Flask was sending out HTTPException responses before we had a chance to log them. I'll describe the Flask code below, but (kind of counter-intuitively) we need to set `TRAP_HTTP_EXCEPTIONS` to True in the offline app to be able to have the errors printed out in the logs. I think the default app doesn't have this issue because it's wrapped in flask_restful.

A drawback to this is that any cron job failures because of HTTPExceptions (BadRequest, BadGateway, Forbidden, ...) will all receive the response of 500. But I think that's ok since the only client will be GAE calling cron jobs. And we'll get to see what the status code was (and more importantly the stack trace and error message) in the logs.

In Flask's app.py file, line 1938 defines `full_dispatch_request` which handles requests (https://github.com/pallets/flask/blob/93dd1709d05a1cf0e886df6223377bdab3b077fb/src/flask/app.py#L1938). Requests are handled in a `try` block and line 1952 calls `handle_user_exception` if something happens. That method is defined up on line 1781 (https://github.com/pallets/flask/blob/93dd1709d05a1cf0e886df6223377bdab3b077fb/src/flask/app.py#L1781). At line 1815 it will check to see if it's a type of HTTPException and if it should not trap it. When it doesn't trap it, then it returns from 1816 and sends the response (without giving us a chance to log it).

But if set the config up so that it will trap it, then this continues on to 1821 (as long as it hasn't found an error handler for it). This then re-raises the exception which throws the exception up to where `full_dispatch` is called on line 2447 (https://github.com/pallets/flask/blob/93dd1709d05a1cf0e886df6223377bdab3b077fb/src/flask/app.py#L2447). The exception is caught again there and line 2450 calls `handle_exception`. That method is defined on 1824 (https://github.com/pallets/flask/blob/93dd1709d05a1cf0e886df6223377bdab3b077fb/src/flask/app.py#L1824) and on line 1859 the signal that an exception occurred is sent out. Coincidentally it also logs the exception itself on line 1871.

### DIFF
--- a/tests/cron_job_tests/test_offline_app_setup.py
+++ b/tests/cron_job_tests/test_offline_app_setup.py
@@ -14,13 +14,14 @@ class OfflineAppTest(BaseTestCase):
 
     def test_offline_http_exceptions_get_logged(self):
         """Make sure HTTPException are logged when thrown"""
+        expected_error_message_text = 'This exception message should appear in the logs'
 
         # Need to raise an HTTPException on a cron call, picking an arbitrary one that is easy to mock
         with mock.patch('rdr_service.offline.main.mark_ghost_participants') as mock_cron_call, \
                 mock.patch('rdr_service.app_util.check_cron', return_value=True),\
-                mock.patch('rdr_service.services.gcp_logging.logging') as mock_logging:
+                mock.patch('rdr_service.main.logging') as mock_logging:
             def throw_exception():
-                raise HTTPException('This exception message should appear in the logs')
+                raise HTTPException(expected_error_message_text)
             mock_cron_call.side_effect = throw_exception
 
             # Call to trigger the exception
@@ -35,4 +36,4 @@ class OfflineAppTest(BaseTestCase):
             self.assertIsNotNone(error_log_call, 'An error log should have been made')
 
             error_message = error_log_call.args[0]
-            self.assertIn('This exception message should appear in the logs', error_message)
+            self.assertIn(expected_error_message_text, error_message)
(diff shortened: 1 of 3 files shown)


tests/cron_job_tests/test_offline_app_setup.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_offline_http_exceptions_get_logged(self):
+        """Make sure HTTPException are logged when thrown"""
+
+        # Need to raise an HTTPException on a cron call, picking an arbitrary one that is easy to mock
+        with mock.patch('rdr_service.offline.main.mark_ghost_participants') as mock_cron_call, \
+                mock.patch('rdr_service.app_util.check_cron', return_value=True),\
+                mock.patch('rdr_service.services.gcp_logging.logging') as mock_logging:
+            def throw_exception():
+                raise HTTPException('This exception message should appear in the logs')
```
pegbertsch: Minor and optional, but maybe the message text can be defined in a class attribute, so it's easy to visually see where the same text is used/checked against?

pegbertsch (approved):
No issues with the changes or the explanation for them, but wouldn't mind a second opinion on the side effects mentioned.  (They seem harmless to me).

tests/cron_job_tests/test_offline_app_setup.py
kenny-skaggs: Good suggestion, because this also turned up an interesting side effect. I originally wrote it assuming that the message of the exception I was also appearing in the logs. But after I tried to implement your suggestions I'm seeing now that only the stack-trace got printed out in `flask_restful_log_exception_error`.

So to change the line to something like `raise HTTPException(expected_error_message)` would also change what the logs show since the stack-trace is showing the code and not the error message itself.

I'm updating offline/main.py now and the test to make sure we get the exception message too.

pegbertsch (approved):

PR #2056, opened 2020-10-27
[DA-1801] Genomic Tool Updates for Collection Tube and Project ID

j-kanuch:
This PR updates the genomic tools to fix the following issues:
1. The `process-runner` tool does not write the correct `project_id` when updating `big_query_sync`.
2. The RDR needs a way to update `genomic_set_member.collection_tube_id` from a list of (biobank_id, new collection_tube_id) pairs.

For #2, this PR creates a new `genomic collection-tube` tool. This tool takes a CSV file of (biobank_id, new collection_tube_id) pairs and updates `genomic_set_member`.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -308,11 +312,11 @@ def _ingest_gc_manifest(self, data, _site):
...
 
                     else:
-                        logging.error(f'Invalid collection tube ID: {collection_tube_id}'
-                                        f' or genome_type: {genome_type}')
-
-                        # Aborting the job if invalid collection tube ID found.
-                        return GenomicSubProcessResult.ERROR
+                        #return GenomicSubProcessResult.ERROR
+                        logging.error(f"Missing collection tube ID: {collection_tube_id}, "
+                                      f"biobank id: {row_copy['biobankidsampleid'].split('_')[0]}, "
```
pegbertsch, line 317: Is anything lost for debugging when dropping the `sampleid` part of the `biobankidsampleid` value from the logged error?

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -673,6 +674,172 @@ def update_genomic_set_member_state(self, member, state):
...
+                for _genome_type in (config.GENOME_TYPE_WGS, config.GENOME_TYPE_ARRAY):
+                    member = self.dao.get_member_from_biobank_id(_bid, _genome_type)
+
+                    if member is None:
+                        # Member with BID doesn't exist, skip it and log it.
+                        nonexistent_bids.append((_bid, _genome_type))
+                        continue
+
+                    else:
+                        member.devNote = dev_note
```
kenny-skaggs: I could be wrong, but I think at this point the session that loaded `member` is closed so setting devNote isn't guaranteed to be saved. Later in `update_genomic_set_member_collection_tube` it gets merged to a new session, so I think that path would save the devNote. But if the else-block is followed then I think the devNote would not be saved.

This probably isn't a big deal, and it might be doing what you want it to anyway.

rdr_service/genomic/genomic_job_components.py
j-kanuch: For this particular log message, only the `biobank_id` and `collection_tube_id` are needed. The sample ID is not important here. We are getting the sample ID from the Biobank, but the lookup is failing because of `collection_tube_id` and `biobank_id`.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -673,6 +674,172 @@ def update_genomic_set_member_state(self, member, state):
...
+        :param new_tube_id:
+        :return: boolean
+        """
+        ss_dao = BiobankStoredSampleDao()
+        sample = ss_dao.get(new_tube_id)
+
+        if self.args.sample_override:
+            return True
+
+        return bid == sample.biobankId
```
pegbertsch: Is it guaranteed there will be a stored sample result returned?   Need to check sample exists first and then compare?

rdr_service/tools/tool_libs/genomic_utils.py
j-kanuch: At that point, the member object should be in memory with the `devNote` attribute set. I'm passing a reference to the member object to the `update_genomic_set_member_collection_tube` method. I haven't run into any issues testing this field on sandbox. 
However, I think I need to move it into the next condition: `if new_tube_member is None and valid_tube:`
Otherwise, when it compares `new_tube_member` to `member` on line 749, the devNote attribute may cause this condition to be false.

kenny-skaggs: 👍

pegbertsch (approved):

rdr_service/tools/tool_libs/genomic_utils.py
j-kanuch: Good catch, I've updated the code.

PR #2059, opened 2020-10-28
[PDR-158] Fix module submittal status values in generators

pegbertsch:
The main change is to add logic to  `_prep_modules()  `that will look for answers to consent questions in order to determine the appropriate module submittal status.    It used to hardcode any module response that was detected to SUBMITTED.

Some of the tests were updated to check the appropriate submittal statuses were assigned based on the test's consent data setup.

Note that this fix changes the outcome of the generators' enrollment status calculation in cases where we now detect that (for example) an EHR consent 'yes' was not provided.  That may mean fewer scenarios where the generator determines enrollment_status of CORE_PARTICIPANT or FULLY_CONSENTED.  

 I went through the logic for the `test_previous_ehr_unsure_with_dv_yes` test case and determined that:
1. There is no answer code in the code table for ConsentPermission_NotSure (and no current cases where a participant's `consent_for_study_enrollment ` status is SUBMITTED_NOT_SURE).
2.  Given the fix to the _prep_modules() logic, the participant would not reach CORE_PARTICIPANT status based on the existing test setup and trying to use a "not sure" response.

I modified the test (and the test name) to reflect that.  Instead it will use an initial EHR 'yes' consent that is later reverted.

### DIFF
--- a/rdr_service/dao/bq_participant_summary_dao.py
+++ b/rdr_service/dao/bq_participant_summary_dao.py
@@ -50,7 +50,9 @@ _consent_module_question_map = {
     'GROR': 'ResultsConsent_CheckDNA',
     'PrimaryConsentUpdate': 'Reconsent_ReviewConsentAgree',
     'ProgramUpdate': None,
-    'COPE': 'section_participation'
+    'COPE': 'section_participation',
+    'cope_nov': 'section_participation',
+    'GeneticAncestry': 'GeneticAncestry_ConsentAncestryTraits'
 }
 
 # _consent_expired_question_map must contain every module ID from _consent_module_question_map.
@@ -61,7 +63,9 @@ _consent_expired_question_map = {
     'GROR': None,
     'PrimaryConsentUpdate': None,
     'ProgramUpdate': None,
-    'COPE': None
+    'COPE': None,
+    'cope_nov': None,
+    'GeneticAncestry': None
 }
 
 # Possible answer codes for the consent module questions and what submittal status the answers correspond to
@@ -81,7 +85,10 @@ _consent_answer_status_map = {
     # COPE_A_13
     CONSENT_COPE_NO_CODE: BQModuleStatusEnum.SUBMITTED_NO_CONSENT,
     # COPE_A_231
-    CONSENT_COPE_DEFERRED_CODE: BQModuleStatusEnum.SUBMITTED_NOT_SURE
+    CONSENT_COPE_DEFERRED_CODE: BQModuleStatusEnum.SUBMITTED_NOT_SURE,
+    'ConsentAncestryTraits_Yes': BQModuleStatusEnum.SUBMITTED,
+    'ConsentAncestryTraits_No': BQModuleStatusEnum.SUBMITTED_NO_CONSENT,
+    'ConsentAncestryTraits_NotSure': BQModuleStatusEnum.SUBMITTED_NOT_SURE
 }
 
 class BQParticipantSummaryGenerator(BigQueryGenerator):
@@ -358,7 +365,6 @@ class BQParticipantSummaryGenerator(BigQueryGenerator):
 
                 # check if this is a module with consents.
                 if module_name in _consent_module_question_map:
-
                     # Calculate Consent Cohort from ConsentPII authored
                     if consent_dt is None and module_name == 'ConsentPII' and row.authored:
                         consent_dt = row.authored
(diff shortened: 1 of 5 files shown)


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -315,60 +346,66 @@ def _prep_modules(self, p_id, ro_session):
...
-                    'mod_status_id': BQModuleStatusEnum.SUBMITTED.value,
-                })
+                }
+                # Default status, may be updated based on consent answer
+                module_status = BQModuleStatusEnum.SUBMITTED
 
                 # check if this is a module with consents.
-                if module_name not in _consent_module_question_map:
-                    continue
+                if module_name in _consent_module_question_map:
```
pegbertsch, line 367: Note:  The diff ended up looking more significant than the content changes were.   Logically, the code is very similar to the previous version (except for handling the answer lookup to find the status value to use on line 400), and the fact that the if statement was reoriented from a  'not in' to an 'in' check

rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -49,9 +60,29 @@
     'EHRConsentPII': 'EHRConsentPII_ConsentExpired',
     'GROR': None,
     'PrimaryConsentUpdate': None,
-    'ProgramUpdate': None
+    'ProgramUpdate': None,
+    'COPE': None
```
kenny-skaggs: `COPE` will capture the May, June, and July Cope surveys. Should the November Cope survey (`cope_nov`) be added to these lists as well?

rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -315,60 +346,66 @@ def _prep_modules(self, p_id, ro_session):
...
+                            consent['consent_value'] = 'ConsentPermission_Yes'
+                            consent['consent_value_id'] = self._lookup_code_id('ConsentPermission_Yes', ro_session)
+                        else:
+                            consent_value = qnan.get(_consent_module_question_map[module_name], None)
+                            consent['consent_value'] = consent_value
+                            consent['consent_value_id'] = self._lookup_code_id(consent_value, ro_session)
+                            consent['consent_expired'] = \
+                                qnan.get(_consent_expired_question_map[module_name] or 'None', None)
+                            # Check for a specific submittal status based on the answer value (default to SUBMITTED)
+                            module_status = _consent_answer_status_map.get(consent_value, BQModuleStatusEnum.SUBMITTED)
```
kenny-skaggs: Is there any value in having a warning logged for any answer code that isn't found in `_consent_answer_status_map`? Maybe it would help flag any unexpected answer codes that are interpreted as SUBMITTED.

tests/resource_tests/generator_tests/test_participant_enrollment.py
```diff
@@ -237,7 +256,7 @@ def test_previous_ehr_unsure_with_dv_yes(self):
 
         # Get Core status through EHR consents
         self._submit_ehrconsent(self.participant_id,
-                                response_code=CONSENT_PERMISSION_NOT_SURE,
+                                response_code=CONSENT_PERMISSION_YES_CODE,
```
kenny-skaggs, line 259: Somewhere along the way I got the idea that an EHR submission of UNSURE meant that a DV_EHR of YES could allow for CORE status. Looking through the code now I can't remember what gave me that idea. Thanks for fixing up this test. I'll investigate further and make sure no other changes are needed and if anyone is affected by my mistake.

rdr_service/dao/bq_participant_summary_dao.py
kenny-skaggs: Thanks, that helped alot

rdr_service/dao/bq_participant_summary_dao.py
pegbertsch: I don't see any COPE module name strings defined in `code_constants.py,` and I didn't see any 'Module Name' values other than COPE in the RDR `code` table.   Is` cope_nov `about to be added but not in our codebase yet?

follow-up:  Still need to push changes to fix this.

rdr_service/dao/bq_participant_summary_dao.py
pegbertsch: Looking at this now.   I may also need to add some new/additional content for GeneticAncestry consents to this PR.

robabram (approved):

rdr_service/dao/bq_participant_summary_dao.py
pegbertsch: Fixed in latest commit

PR #2087, opened 2020-11-10
[PDR-165] Remove replayed questionnaire responses from PDR data

pegbertsch:
Had some significant "duplication" in PDR module/consent data due to some known replay events, such as PTSC replaying ~65K GROR consents because of missing electronic signatures in the associated PDF files.

Additionally, found a prior replay event from April (ROC-447/ROC-475) where PTSC deprecated the original GROR consent questionnaire;  the updated version had a different consent question/answer structure.  So the replays were mapped to the new questionnaire ID structure, but still had the same authored date as the original response for the deprecated questionnaire ID.  Added some special case code for deprecated GROR responses that will allow them to still be recognized as "replays"/duplicates of the updated response, despite the structural differences.

The PDR generator code that assembles the lists of module/consent summaries will look for responses that match on everything except creation date, and only include the most recently created.

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -1175,6 +1175,22 @@ class ParticipantSummaryGenerator(generators.BaseGenerator):
                 return None
 
             # Query the answers for all responses found.
+            # Note on special logic for GROR module:  the original GROR consent questionnaire was quickly replaced by
+            # a revised questionnaire with a different consent question/answer structure.  GROR consents (~200)
+            # that came in for the old/deprecated questionnaire_id were resent by PTSC using the new questionnaire_id
+            # (See ROC-447/ROC-475)
+            #
+            # When processing a deprecated GROR response, add a key/value pair to the data
+            # simulating what the consent answer would look like in the revised consent.  E.g., if the
+            # deprecated GROR consent response had these question codes/boolean answer values (only one will be True/1):
+            #   'CheckDNA_Yes': '0',
+            #   'CheckDNA_No': '1',
+            #   'CheckDNA_NotSure': '0'
+            # ... then this key/value pair will be added to simulate the revised GROR consent question code/answer code:
+            #    'ResultsConsent_CheckDNA': 'CheckDNA_No'
+            #
+            # This way the answers returned can have the same logic applied to them by _prep_modules(), for all GROR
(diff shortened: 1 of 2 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -357,14 +367,24 @@ def _prep_modules(self, p_id, ro_session):
         consent_dt = None
 
         if results:
+            # Track the last module/consent data dictionaries generated, so we can detect and omit replayed responses
+            last_mod_processed = {}
+            last_consent_processed = {}
             for row in results:
+                consent_added = False
                 module_name = self._lookup_code_value(row.codeId, ro_session)
+                module_name = self._lookup_code_value(row.codeId, ro_session)
```
kenny-skaggs: Is this duplicate line (extra `module_name = self._lookup_code_value(row.codeId, ro_session)`) intentional?

rdr_service/resource/generators/participant.py
```diff
@@ -357,14 +367,24 @@ def _prep_modules(self, p_id, ro_session):
...
                     'module': module_name,
                     'baseline_module': 1 if module_name in self._baseline_modules else 0,  # Boolean field
                     'module_authored': row.authored,
                     'module_created': row.created,
                     'language': row.language,
+                    'status': module_status.name,
+                    'status_id': module_status.value
                 }
                 # Default status, may be updated based on consent answer
                 module_status = BQModuleStatusEnum.SUBMITTED
```
kenny-skaggs: This was deleted in the BQ code. Should it be deleted here?

rdr_service/resource/generators/participant.py
pegbertsch: I think it was a bad cut paste from when I copied the BQ code.  Fixing.

rdr_service/resource/generators/participant.py
pegbertsch: It should be added back in in the BQ code.  Lines 384-385?
It was due to the refactoring of some related logic, which needed the fields to be present in the dictionary by default.  So I added these lines back when the dictionary is initially populated.  But there is a redundant line here on 390 which is related to the same cut/paste.  Fixing

kenny-skaggs (approved):

PR #2110, opened 2020-11-30
Revert "Revert "[DA-1784] allow hpo lite access UNSET participants""

wangy70:
Reverts all-of-us/raw-data-repository#2093

NIH confirmed the HPO Lite users are able to see the UNSET participants.

### DIFF
--- a/rdr_service/api/participant_summary_api.py
+++ b/rdr_service/api/participant_summary_api.py
@@ -8,6 +8,8 @@ from rdr_service.dao.base_dao import _MIN_ID, _MAX_ID
 from rdr_service.dao.participant_summary_dao import ParticipantSummaryDao
 from rdr_service.model.hpo import HPO
 from rdr_service.model.participant_summary import ParticipantSummary
+from rdr_service.config import getSettingList, HPO_LITE_AWARDEE
+from rdr_service.code_constants import UNSET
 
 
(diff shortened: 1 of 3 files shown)


kenny-skaggs:
Should there be a unit test for this?

wangy70:
I tried to have a test case for this one, but I found it's hard to do. Because we skipped the OAuth process in our testing framework, but this one need to go through the OAuth and get the `user_email` and then get the awardee name from the configuration(we have to mock the configuration too). That's why I give up :(  Let me know if you have any idea about how to test it.

kenny-skaggs:
In the deceased report tests I needed to alter the roles for the user that the tests use (example@example.com). I used `override_setting` to change what oauth returned as the information for the user. You could add the awardee role for the test and then at the end of the test override the settings again to set them back.

In a few places in the tests I called https://github.com/all-of-us/raw-data-repository/blob/devel/tests/api_tests/test_deceased_report_api.py#L29 to set the roles and the test class's `tearDown` method (a few lines above the code from the last link) would restore the roles.

You could use the config override to set the hpo-lite awardee list too.

wangy70:
Cool, I will try it, thanks!

wangy70:
Test case is added.

kenny-skaggs (approved):

PR #2115, opened 2020-12-01
[DA-1824] AW0 manifest modifications

kenny-skaggs:
This updates the AW0 cron job for cohort 3 to run every Wednesday. And updates the workflow for each cohort to prioritize samples in the following order: 1ED04, 1ED10, and 1SAL2. So if a participant has an 1ED10 sample and a 1SAL2, the 1ED10 would be used in the workflow and the 1SAL2 would be left out.

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -1986,14 +1986,6 @@ class GenomicBiobankSamplesCoupler:
                 and ssed.test in ("1ED04", "1ED10")
                 and ssed.status < 13
             ORDER BY oeds.collected DESC
-                and ed04.collected = (
-                    SELECT MAX(os.collected)
-                    FROM biobank_ordered_sample os
-                        JOIN biobank_order o ON o.biobank_order_id = os.order_id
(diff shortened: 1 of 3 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -1749,7 +1737,54 @@ def _get_new_biobank_samples(self, from_date):
...
+            previously_found_sample = preferred_samples.get(sample.participant_id, None)
+            if previously_found_sample is not None:
+                preferred_sample = self._determine_best_sample(previously_found_sample, sample)
+
+            preferred_samples[sample.participant_id] = preferred_sample
+
+        return list(preferred_samples.values())
+
+    @staticmethod
+    def _determine_best_sample(sample_one, sample_two):
```
j-kanuch, line 1759: I like what you did here

j-kanuch (approved):

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: Thanks :)

j-kanuch (approved):

PR #2123, opened 2020-12-03
[DA-1820] MIGRATION - AW2F Manifest Workflow

j-kanuch:
This PR implements the AW2F workflow for the generation of the AW2F manifest. This manifest is dependent on receipt of the AW1 and AW2 manifests. This includes a data migration to support this and future 'feedback' manifests.

The purpose of this workflow is to track when the DRC has received a record of AW2 feedback from the Genome Centers for an AW1 manifest. The Biobank is requesting for the RDR system to add a new manifest to provide feedback to the Biobank of samples that were successfully processed by the Genome Centers (GCs) or failed due to contamination.

Significant updates:
- two new RDR tables to track manifests and feedback records:
--`GenomicManifestFile`
--`GenomicManifestFeedback`
- Inserting a `GenomicManifestFile` and `GenomicManifestFileFeedback` record during the AW1 ingestion process
- Calculating `contamination_category` when the AW2 is ingested
- new Cron job that scans for AW1 manifests that have all records returned as processed by the GCs
- new workflow to generate the AW2F manifest
- Some refactoring to allow for pub/sub pattern using Cloud Tasks/Cloud Functions for ingesting AW1 (a future PR will implement this for the AW2)
- Adding the aou_cloud repository to the requirements but specifying the branch `rdr-compatible` to fix missing and incompatible requirements.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -715,6 +745,36 @@ def _get_qc_status_from_value(self, aw4_value):
...
+        elif (0.01 <= raw_contamination <= 0.03) and ps.consentForGenomicsROR == QuestionnaireStatus.SUBMITTED:
+            return GenomicContaminationCategory.EXTRACT_WGS
+
+        # No Extract if contamination between 1 and 3 % inclusive and GROR is not Yes
+        elif (0.01 <= raw_contamination <= 0.03) and ps.consentForGenomicsROR != QuestionnaireStatus.SUBMITTED:
+            return GenomicContaminationCategory.NO_EXTRACT
+
+        # Extract Both if contamination > 3%
+        elif raw_contamination > 0.03:
+            return GenomicContaminationCategory.NO_EXTRACT
```
kenny-skaggs: The comment here seems to conflict with what is returned. Should it be returning `GenomicContaminationCategory.EXTRACT_BOTH`?

j-kanuch: You are correct, I will update it.

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -326,33 +397,73 @@ def run_cvl_reconciliation_report(self):
...
+            else:
+                result = self.manifest_compiler.generate_and_transfer_manifest(manifest_type, _genome_type)
+
+            if result['code'] == GenomicSubProcessResult.SUCCESS:
                 logging.info(f'Manifest created: {self.manifest_compiler.output_file_name}')
-                # Insert the file record
+
+                new_file_path = f'{self.bucket_name}/{self.manifest_compiler.output_file_name}'
+
+                now_time = clock.CLOCK.now()
```
pegbertsch: As used here, this is equivalent to datetime.utcnow() (used on line 96), correct?

rdr_service/participant_enums.py
```diff
@@ -714,6 +718,16 @@ class GenomicManifestTypes(messages.Enum):
...
     AW3_WGS = 9
+    AW2F = 10
+
+
+class GenomicContaminationCategory(messages.Enum):
+    UNSET = 0
+    NO_EXTRACT = 1
+    EXTRACT_WGS = 2
+    EXTRACT_BOTH = 3
+    TERMINAL_NO_EXTRACT = 4
```
pegbertsch, line 729: Is TERMINAL_NO_EXTRACT for future use?   (Not used except as part of a SQLAlchemy case statement to map these values to strings?)

rdr_service/genomic/genomic_job_controller.py
j-kanuch: Yes, I'm not sure why I'm doing it two different ways. I can update them both to use datetime.utcnow()

rdr_service/participant_enums.py
j-kanuch: Yes, this category is significantly more complicated and is going to be implemented in the upcoming sprint.

pegbertsch (approved):
Minor questions;  do not impact functionality

pegbertsch (approved):

PR #2128, opened 2020-12-08
[DA-1814] MIGRATION - January COPE fields

kenny-skaggs:
This adds the January COPE fields to the participant summary and the related structures to the BigQuery code.

There were also two separate dictionaries of participant summary data that several tests used for checking ParticipantSummaryAPI results. Each of those places needed to be updated with the new fields every time something was added to participant summary, otherwise the tests wouldn't pass. This PR has changes for consolidating the set of defaults into one structure that all the dependent tests read from.

This also updates the December BQ table to use the December schema rather than the November one.


j-kanuch (approved):

pegbertsch (commented):
I think there's another place we need to add BQPDRCOPEJan:
rdr_service/resource/tasks.py modules list on lines 52-60

PR #2133, opened 2020-12-09
[DA-1838] hpo_lite_pairing_import

wangy70:
Import HPO Lite pairing data from REDCap project and pair the participants.


rdr_service/tools/tool_libs/hpo_lite_pairing_sync.py
```diff
@@ -0,0 +1,27 @@
...
+        importer = HpoLitePairingImporter()
+        importer.import_pairing_data(parsed_since_date)
+
+
+def add_additional_arguments(arg_parser):
+    arg_parser.add_argument('--since-date', help='Request all records sync the given date/time', default=None)
+
+
+def run():
+    return cli_run(tool_cmd, tool_desc, HpoLitePairingSyncTool, add_additional_arguments)
```
kenny-skaggs, line 27: 😄

rdr_service/offline/import_hpo_lite_pairing.py
```diff
@@ -0,0 +1,124 @@
...
+                    # save the import record
+                    import_record = HpoLitePairingImportRecord(
+                        participantId=participant_id,
+                        pairedDate=pair_date,
+                        orgId=org.organizationId,
+                        created=CLOCK.now()
+                    )
+                    import_record.uploadingUser = upload_user
+
+                    self.base_dao.insert_with_session(session, import_record)
```
kenny-skaggs, line 118: What do you think of committing the session after using it for the record (right after the insert, before the `for` loop starts on the next record)?

That way, if the following records have any errors then previously successful pairings don't get rolled back. And it will also release the locks on the participant taken by the `_update_pairing_with_session` method's call to get it for update.

kenny-skaggs (commented):
Looks good. Just had a comment on managing the session's transaction while importing.

rdr_service/offline/import_hpo_lite_pairing.py
wangy70: Good point, done!

kenny-skaggs (approved):

PR #2134, opened 2020-12-10
[PDR-166 fix] Refactor BQPDRQuestionnaireResponseGenerator

pegbertsch:
Refactored code to replace the deprecated `sp_get_questionnaire_answers` and `sp_get_code_module_items` procedure calls, as they may no longer return the desired results for DRC-managed codebooks where codes are shared between modules .  Addition to the original design we discussed (to correct issues found during testing):

Handling for multi-select question codes where a user can provide more than one answer to the same question code in a questionnaire_response_id payload (example: user selects multiple ethnicity options).  The stored procedure used to handle this with a GROUP_CONCAT and merge all selections into a single answer value that was a comma-separated list

Also cleaned up the participant generator logic to better recognize "invalid" consents (when no consent answer supplied in the response payload).

Note on testing:  Rebuilt all participants on stable who had a questionnaire_on_cope_dec_authored date.   Did queries against the base pdr_participant and pdr_mod_* tables on aou-pdr-data-stable BQ instance to compare results of the most recently built record and previously built records which have not yet been purged from the BQ base tables.

The new builds can sometimes contain answers/details for nested questions that were not previously picked up by the `sp_get_questionnaire_answers` stored procedure (e.g., specific ethnicity options under a broader ethnicity response, such as BlackSpecific_Barbadian)


rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -47,45 +104,91 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
             logging.info('Generator: ignoring questionnaire module id [{0}].'.format(module_id))
             return None, list()
 
-        qnans = self.ro_dao.call_proc('sp_get_questionnaire_answers', args=[module_id, p_id])
-        if not qnans or len(qnans) == 0:
-            return None, list()
-
-        bqrs = list()
```
pegbertsch, line 55: Note: this block has moved to line 157 in the new version, but logically remains the same

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -25,6 +29,59 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
+        # Query to get a list of questionnaire_response_id values for this participant and module
+        _participant_module_responses_sql = """
+            select qr.questionnaire_id, qr.questionnaire_response_id, qr.created, qr.authored, qr.language,
+                   qr.participant_id
+            from questionnaire_response qr
+            where qr.participant_id = :p_id and questionnaire_id IN (
+                select q.questionnaire_id from questionnaire q
+                inner join questionnaire_history qh on q.version = qh.version
+                       and qh.questionnaire_id = q.questionnaire_id
+                inner join questionnaire_concept qc on qc.questionnaire_id = q.questionnaire_id
```
kenny-skaggs, line 58: It might not affect anything, but to be safe `questionnaire_concept` should probably be joined on the version too.

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -25,6 +29,59 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
+                        qra.value_integer, qra.value_decimal,
+                        qra.value_boolean, qra.value_string, qra.value_system,
+                        qra.value_uri, qra.value_date, qra.value_datetime) as answer
+            FROM questionnaire_response qr
+                     INNER JOIN questionnaire_response_answer qra
+                                ON qra.questionnaire_response_id = qr.questionnaire_response_id
+                     INNER JOIN questionnaire_question qq
+                                ON qra.question_id = qq.questionnaire_question_id
+                     INNER JOIN questionnaire q
+                                ON qq.questionnaire_id = q.questionnaire_id
```
kenny-skaggs: Is the questionnaire being joined in for anything? I was trying to trace back to what joining it this way might affect, but it doesn't seem like any data is read from the `questionnaire` table.

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -25,6 +29,59 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
+            WHERE qr.questionnaire_response_id = :qr_id
+            order by qr.created DESC;
```
kenny-skaggs: I may be misunderstanding how the later code works, but should this order by `qr.authored ASC` so that when the dictionary is built the result is the latest answer we have for the participant (regardless of the order that things have be replayed to us). Or atleast `qr.created ASC` to match how the processing would work in RDR's questionnaire_response_dao?

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -47,45 +104,91 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
-                    setattr(bqr, fld_name, str(fld_value))
-
-                # Truncate zip codes to 3 digits
-                if fld_name in ('StreetAddress_PIIZIP', 'EmploymentWorkAddress_ZipCode') and len(fld_value) > 2:
-                    setattr(bqr, fld_name, fld_value[:3])
-
-            bqrs.append(bqr)
-            if latest:
-                break
+        # This section refactors how the module answer data is built, replacing  the sp_get_questionnaire_answers
```
pegbertsch, line 107: Note on testing:
During development, I would use the resource rebuild-pids tool to exercise the new code, and build the questionnaire data both using the new SQL logic and the old stored procedure call in the same make_bqrecord() invocation.   The resulting bqr records (new vs. old) were compared field by field to make sure they had matching answer values. 

This would verify the new logic is correctly replicating exactly what the stored procedure would for other modules (Basics, LifeStyle, OverallHealth, etc.), and how I caught mismatches such as the new logic not initially handling multi-select answers.

That gave me confidence the refactored code should be producing results consistent with our existing pdr_mod_* data.

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: These were the queries we worked up in the slack meeting, cut and pasted.    I'll review them again

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: I think reverse chronological order is important here for the PDR generator logic and how it handles partial response payloads (wants to get the most recently sent answers first).  I was thinking we should work backwards from most recently sent payload, regardless of authored date.  Will get a second opinion from Robert

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: Looking.

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: The main SELECT statement in the `sp_get_questionnaire_answers` stored proc we're replacing uses `ORDER BY created DESC` but per Robert that may be a holdover that preceded introduction of authored dates.  I think the appropriate update is to `ORDER BY authored DESC, created DESC `.   I'm also going to re-check the participant data generators as I think this should also be incorporated there

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: Fixed in most recent changes

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: Considered this in the most recent changes, but still using created (processed earliest to latest now) to match the previous logic that used the stored procedure call results

rdr_service/dao/bq_questionnaire_dao.py
pegbertsch: Fixed in most recent changes

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -48,45 +101,90 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
+                    else:
+                        setattr(bqr, fld_name, str(fld_value))
+
+                    # Truncate zip codes to 3 digits
+                    if fld_name in ('StreetAddress_PIIZIP', 'EmploymentWorkAddress_ZipCode') and len(fld_value) > 2:
+                        setattr(bqr, fld_name, fld_value[:3])
+
+                bqrs.append(bqr)
+                if latest:
+                    break
```
kenny-skaggs, line 184: Since these are processed in order of oldest responses to newest, does passing in `latest` as True give the oldest response only?

rdr_service/model/bq_questionnaires.py
```diff
@@ -76,45 +77,75 @@ def get_fields(self):
...
+            inner join (
+                select distinct qq.code_id
+                from questionnaire_question qq where qq.questionnaire_id in (
+                    select qc.questionnaire_id from questionnaire_concept qc
+                            where qc.code_id = (
+                                select code_id from code c2 where c2.value = :module_id and system = :system
+                            )
+                )
+            ) qq2 on qq2.code_id = c.code_id
+            where c.code_type = 3
```
kenny-skaggs: This will miss any codes that were previously Topic type codes, but have been repurposed as question codes. I think for the logic here it's safe to assume that the code type doesn't matter if the code_id is used by a questionnaire_question.

pegbertsch: Fixed in latest version

kenny-skaggs (approved):

PR #2142, opened 2020-12-17
[DA-1849, 1847] Updates to Genomic AW1 and AW2 workflows

j-kanuch:
This PR implements several changes in support of the AW1 and AW2 workflows:
 - The AW1 and AW2 cron jobs were disabled and replaced with cloud functions and cloud tasks.
 - A cloud function was required for each GCS bucket resource: An AW1 and AW2 function for each of the 3 GCs, totalling 6 cloud functions.
 - The validation rules of AW2 files were modified to ignore column order and to set contamination to 0 if it is returned as a negative.
 - The AW2 workflow was modified to inherit from UpsertableDao.
 - The UpsertableDao was modified to return the upserted object.
 - The GEM A1 query was updated to remove duplicates.
 - A new Cloud Task API endpoint was created for the AW2 workflow.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -557,12 +557,18 @@ def _process_gc_metrics_data_for_insert(self, data_to_ingest):
...
 
-            # TODO: limit call rate to 10 characters for now until clarification from GCs
             try:
                 row_copy['callrate'] = row_copy['callrate'][:10]
             except KeyError:
                 pass
 
+            # Clean up contamination if data is less than 0
+            try:
+                if '-' in row_copy['contamination']:
```
kenny-skaggs: Is contamination previously validated to be a number? If not, then there's a chance that this could silently hide that contamination values are invalid. Like if someone decided to try to use a range of "0.02 - 0.05". Definitely an edge case, but this might ignore it and zero it out.

rdr_service/gcloud_functions/genomic_aw1_jh_function/main.py
```diff
@@ -0,0 +1,94 @@
...
+    cloud_resource = 'aou-rdr-sandbox-mock-data'
+
+    if _project_suffix == 'sandbox':
+        cloud_resource = 'aou-rdr-sandbox-mock-data'
+
+    if _project_suffix == 'stable':
+        cloud_resource = 'genomics-baylor-dryrun'
+
+    if _project_suffix == 'prod':
+        cloud_resource = 'prod-genomics-baylor'
```
kenny-skaggs: The file names for the aw1 and aw2 functions for each GC line up except for Baylor. For Baylor there's an aw1_jh and aw2_bcm function. Is that correct, and if not does it matter?

j-kanuch: I see that this is somewhat confusing. The buckets are named "baylor", but Baylor actually manages 2 genomic center IDs `gc_site_id`: JH and BCM. The AW1 files typically come through as JH (Johns Hopkins), and the AW2 files are returned as BCM (Baylor College of Medicine).  I tried to be consistent with my function naming by naming them based on the `gc_site_id`. The function name doesn't really matter, as long as they're unique, but it may be better to name them with the name of the bucket they are monitoring: i.e. broad, baylor, and northwest.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Contamination is not previously validated to be a number, but it probably should be. I will make the update.

j-kanuch: I updated this to validate that `contamination` is a number. This will ensure we can perform the contamination category calculations. I added a TODO in the RDR model to update the datatype to a float. I did not perform this yet as it will impact both the RDR and PDR.

kenny-skaggs (approved):

PR #2143, opened 2020-12-17
[DA-1845] MIGRATION - Adding TERMINAL_NO_EXTRACT contamination category

kenny-skaggs:
This adds the TERMINAL_NO_EXTRACT contamination category result to the category calculation method. The TERMINAL_NO_EXTRACT value is used when the contamination category for a sample would result in one of the EXTRACT values, but there is no other samples for the participant or any other sample was already found to be contaminated.

In order to support this, a new table was created that records samples that have been found to have been contaminated (or have had the collection tube id swapped for another).


rdr_service/genomic/genomic_job_components.py
```diff
@@ -745,35 +751,65 @@ def _get_qc_status_from_value(self, aw4_value):
...
-    def calculate_contamination_category(self, raw_contamination, member):
+    @staticmethod
+    def _participant_has_potentially_clean_samples(session, biobank_id):
+        """Check for any stored sample for the participant that is not contaminated
+        and is a 1ED04, 1ED10, or 1SAL2 test"""
+        query = session.query(BiobankStoredSample).filter(
+            BiobankStoredSample.biobankId == biobank_id
+        ).outerjoin(GenomicSampleContamination).filter(
+            GenomicSampleContamination.id.is_(None),
+            BiobankStoredSample.test.in_(['1ED04', '1ED10', '1SAL2'])
```
j-kanuch, line 763: We may want to check that `BiobankStoredSample.status` is not one of these:
SAMPLE_NOT_RECEIVED = 13
    SAMPLE_NOT_PROCESSED = 14
    ACCESSINGING_ERROR = 15
    LAB_ACCIDENT = 16
    QNS_FOR_PROCESSING = 17
    QUALITY_ISSUE = 18

kenny-skaggs: Done! Thanks for catching that.

j-kanuch (approved):

PR #2148, opened 2020-12-23
[DA-1859] Tool and Cloud Task for Calculating Contamination Category

j-kanuch:
This PR implements a tool to update `gc_validation_metrics` with `contamination_category` based on the latest AW2 contamination number and current GROR consent. The tool can be ran locally or can be pushed to a cloud task queue. To support the usage of cloud tasks, a new cloud task class and endpoint was created: `CalculateContaminationCategoryApi` and `calculate_contamination_category_task`, respectively.

### DIFF
--- a/rdr_service/tools/tool_libs/genomic_utils.py
+++ b/rdr_service/tools/tool_libs/genomic_utils.py
@@ -1312,10 +1312,15 @@ class CalculateContaminationCategoryClass(GenomicManifestBase):
             if count > 0:
                 batch_count += 1
                 data = {"member_ids": batch}
-                _task.execute('calculate_contamination_category_task',
-                              payload=data,
-                              queue=task_queue,
-                              project_id=self.gcp_env.project)
+
+                if self.args.dryrun:
+                    _logger.info("In Dryrun mode, skip submitting cloud task.")
+
+                else:
+                    _task.execute('calculate_contamination_category_task',
+                                  payload=data,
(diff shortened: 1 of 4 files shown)


rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -1221,6 +1223,135 @@ def update_member_aw1_file_processed_id(self, mid, fid):
...
+                    GenomicGCValidationMetrics.genomicSetMemberId == mid
+                ).one_or_none()
+
+                if record is not None:
+                    # calculate new contamination category
+                    contamination_category = genomic_ingester.calculate_contamination_category(
+                        record.GenomicSetMember.collectionTubeId,
+                        float(record.GenomicGCValidationMetrics.contamination),
+                        record.GenomicSetMember
+                        )
```
kenny-skaggs, line 1354: I don't know if this might make a difference for your workflow, but the `calculate_contamination_category` method modifies the database by adding samples to the contaminated table if they should be there, regardless of whether the tool is being run as a dry-run or not. Maybe it's fine as it is, but I wanted to make sure you were aware of it.

j-kanuch: Good observation, I've updated the logic to only call `calculate_contamination_category` when `dryrun` is false.

kenny-skaggs (approved):

PR #2152, opened 2020-12-28
[DA-1855] Update WGS reconciliation file naming.

j-kanuch:
This PR updates the file naming conventions for the genomic data reconciliation WGS workflow. A TODO was added to update the local revision number in the future.

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -1331,8 +1331,9 @@ class GenomicReconciler:
         """
         metrics = self.metrics_dao.get_with_missing_seq_files(self.controller.last_run_time)
 
-        # TODO: Update filnames when clarified
-        external_ids = "LocalID_InternalRevisionNumber"
+        # TODO: this number may change in the future, currently all files have 1 as the revision number
+        # Will need to follow up with GCs on what this number represents
+        local_revision_id = "1"
 
         total_missing_data = []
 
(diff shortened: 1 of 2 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -1331,8 +1331,9 @@ def reconcile_metrics_to_sequencing_data(self):
         """
         metrics = self.metrics_dao.get_with_missing_seq_files(self.controller.last_run_time)
 
-        # TODO: Update filnames when clarified
-        external_ids = "LocalID_InternalRevisionNumber"
+        # TODO: this number may change in the future, currently all files have 1 as the revision number
+        # Will need to follow up with GCs on what this number represents
+        local_revision_id = "1"
```
kenny-skaggs, line 1336: If these change in the future they would need to be modified in a hotfix or release, right? Would it be better to put them in the config?

j-kanuch: I believe that this is going to require a fix to allow more than 1 revision number. I am assuming that data files that fail QC are going to be re-sequenced and then uploaded with something like "_2". However, I haven't gotten clarification on this yet. All files currently have "_1". I don't think putting it in the config would be helpful because we'll still need to fix the logic that searches the bucket. This is sort of a placeholder for a subsequent PR once I get that clarification.

kenny-skaggs: Ah, ok. That makes sense. Thanks for clarifying.

kenny-skaggs (approved):

PR #2160, opened 2021-01-04
[DA-1852] Restarting genomics a1 workflow for Jan 5

kenny-skaggs:
This reverts PR #2139.

I will also reconfigure the cron jobs on Prod after opening this PR.

### DIFF
--- a/rdr_service/cron_default.yaml
+++ b/rdr_service/cron_default.yaml
@@ -109,6 +109,11 @@ cron:
   timezone: America/New_York
   schedule: every day 06:00
   target: offline
+- description: Genomic GEM A1 Workflow
+  url: /offline/GenomicGemA1Workflow
+  timezone: America/New_York
+  schedule: every tuesday 12:00
+  target: offline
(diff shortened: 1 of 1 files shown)


rdr_service/cron_default.yaml
```diff
@@ -109,6 +109,11 @@ cron:
   timezone: America/New_York
   schedule: every day 06:00
   target: offline
+- description: Genomic GEM A1 Workflow
+  url: /offline/GenomicGemA1Workflow
+  timezone: America/New_York
+  schedule: every tuesday 12:00
```
j-kanuch: As of 13:45 today, 1/4/2021, Color will be updating their processing time to 6 AM. This cron entry sets our A1 job to run at 12, noon. Based on this new information, we will need to update this to be earlier. I suggest 00:00. I just found out about this via Slack from Cynthia, a PM for Color.

j-kanuch (approved):

j-kanuch (approved):

PR #2172, opened 2021-01-08
Remove BQ generators for manifest file and feedback

j-kanuch:
This PR temporarily disables the BQ generators for the `genomic_manifest_file` and `genomic_manifest_feedback` tables. This will allow for local runs of the AW2 ingestion process without disrupting the Big Query sync job due to a mismatch between the production and local schemas.

### DIFF
--- a/rdr_service/dao/genomics_dao.py
+++ b/rdr_service/dao/genomics_dao.py
@@ -11,8 +11,7 @@ from werkzeug.exceptions import BadRequest, NotFound
 
 from rdr_service import clock, config
 from rdr_service.dao.base_dao import UpdatableDao, BaseDao, UpsertableDao
-from rdr_service.dao.bq_genomics_dao import bq_genomic_set_member_update, \
-    bq_genomic_manifest_feedback_update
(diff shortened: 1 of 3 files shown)


wangy70 (approved):

pegbertsch (commented):
I'm wondering if this PR is still needed and/or what the plan is for it.    

Does it require a hotfix, or is the idea just to have these changes merged to devel for convenience in running tools locally that skip the genomic manifest PDR data builds, without having to disable these calls on a local branch?

I'm also assuming the plan is to revert the changes as part of the 1.87.1 RDR release?

j-kanuch:
> I'm wondering if this PR is still needed and/or what the plan is for it.
> 
> Does it require a hotfix, or is the idea just to have these changes merged to devel for convenience in running tools locally that skip the genomic manifest PDR data builds, without having to disable these calls on a local branch?
> 
> I'm also assuming the plan is to revert the changes as part of the 1.87.1 RDR release?

This doesn't require a hotfix, the idea was to merge these to devel for convenience in running the tools locally. This would be reverted as part of the 1.87.1 RDR release.

PR #2177, opened 2021-01-13
[DA-1869] Refactor Array and WGS Data File Reconciliation

j-kanuch:
This PR updates the genomic data reconciliation workflow for both array and whole genome sequencing. The following changes were made:
- The reconciliation was updated to be performed on each bucket separately.
- The process now only lists the data files once per bucket and stores the file names in a list.
- A mapping of GC name to `gc_site_id` was added to the base config.
- The `genomic_gc_validation_metrics` queries for missing files were updated to filter on `gc_site_id`.
- The file searching for the WGS reconciliation was updated to use regular expressions.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -1497,13 +1540,21 @@ def _check_genotyping_file_exists(self, bucket_name, filename):
...
+        filenames = [name for name in self.file_list if name.lower().endswith(filename.lower())]
+        return filenames[0] if len(filenames) > 0 else 0
+
+    def _get_full_filename_with_expression(self, expression):
+        """ Searches file_list for names that match the expression
+        :param expression: pattern to match
+        :return: file name with highest revision number
+        """
+        filenames = [name for name in self.file_list if re.search(expression, name)]
+        filenames.sort(reverse=True)
```
kenny-skaggs: Why do you reverse the list here?

j-kanuch: There are almost always only going to be 1 file, but if there are multiple, this is to return the file name with the highest revision number. 
For example, if there are 2 files matching the regex pattern, they may look something like "UW_A12345_1001_1.cram" and "UW_A12345_1001_2.cram". Sorting the list ensures they are in order of revision number since that is the only number that would change. I could sort without reversing and take the last element of the list instead. It would do the same thing.

kenny-skaggs: Ok, thanks. I thought it might have something to do with that, but wanted to be sure.

It looks like listing files in google storage returns in lexicographical order. If there are enough revisions of a file then I think this could begin returning the wrong one. Like if there were 11 revisions they would appear in the list as name_1, name_10, name_11, name_2, name_3...

j-kanuch: That is true. It would be unlikely that there are 11 files, but to be safe we'll probably want to naturally sort the filenames list. I can update.

j-kanuch: I've updated the code to sort in natural order by descending revision number.

kenny-skaggs (approved):

PR #2182, opened 2021-01-18
[DA-1873] MIGRATION - Renaming January COPE to February

kenny-skaggs:
If there's another COPE survey, we'll come up with a different naming scheme. This updates all our references to the January COPE so they match that it's February now instead. I _think_ I got all the references in the BQ/Resource code, please triple check that.


tests/dao_tests/test_questionnaire_response_dao.py
```diff
@@ -827,10 +827,10 @@ def test_january_cope_survey(self):
 
         self._create_cope_questionnaire(identifier='JanCope')
```
pegbertsch: Does this line need updating?

kenny-skaggs: Thanks for noticing it. Maybe it could be fun to leave it as a sort of easter-egg in the test for the name change? Currently the December and February test cases are set up to follow the behavior of not recognizing the externalId/identifier and using the authored date instead. So that string could be anything not defined on the cope_form_id_map created by the test on line 111.

But I could see it potentially leading to confusion and will change it if you think that's possible.

pegbertsch: Probably not functionally necessary;  just trying to make it less confusing for maintenance purposes.
I think there's actually an external_id value in production for CE questionnaires that says 'October COPE Survey' for the cope_nov module, which is unfortunate/a little confusing.  

Approving this as is, and you can decide if it's worth updating or not.

pegbertsch (approved):

PR #2183, opened 2021-01-19
[DA-1883] Targeted ingestion tool for AW1 and AW2 data

j-kanuch:
This tool allows for the ingestion of a single row from an AW1 or AW2 manifest. Using a provided `genomic_set_member.id`, the manifest file path, and the manifest data type (AW1 or AW2), the tool will ingest only the row associated with the `genomic_set_member` record.

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -275,14 +275,9 @@ class GenomicFileIngester:
             return GenomicSubProcessResult.NO_FILES
         return GenomicSubProcessResult.ERROR
 
-    def _ingest_gc_manifest(self, data, _site):
-        """
-        Updates the GenomicSetMember with GC Manifest data
-        :param data:
-        :param _site: gc_site ID
-        :return: result code
-        """
-        gc_manifest_column_mappings = {
+    @staticmethod
+    def get_aw1_manifest_column_mappings():
+        return {
             'packageId': 'packageid',
             'sampleId': 'sampleid',
             'gcManifestBoxStorageUnitId': 'boxstorageunitid',
@@ -306,6 +301,15 @@ class GenomicFileIngester:
             'gcManifestFailureMode': 'failuremode',
             'gcManifestFailureDescription': 'failuremodedesc',
         }
+
+    def _ingest_gc_manifest(self, data, _site):
+        """
+        Updates the GenomicSetMember with GC Manifest data
+        :param data:
+        :param _site: gc_site ID
+        :return: result code
+        """
+        gc_manifest_column_mappings = self.get_aw1_manifest_column_mappings()
         try:
             count = 0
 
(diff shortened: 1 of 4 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -424,6 +428,126 @@ def _ingest_gc_manifest(self, data, _site):
...
+
+        # Set other fields not in AW1 file
+        member.reconcileGCManifestJobRunId = self.job_run_id
+        member.aw1FileProcessedId = self.file_obj.id
+        member.gcSite = self._get_site_from_aw1()
+
+        # Only update the member's state if it was AW0
+        if member.genomicWorkflowState == GenomicWorkflowState.AW0:
+            member.genomicWorkflowState = GenomicWorkflowState.AW1
+            member.genomicWorkflowStateModifiedTime = clock.CLOCK.now()
```
kenny-skaggs, line 461: I'm a little confused by the comment on this block. From what I understand the code updates the modified time if the workflow state is AW0 or AW1. Is that date the 'state' that the comment is referring to? If the workflow state is something other than AW0 or AW1, is it ok that other fields except the date would be modified? Does the workflow state need to be changed at some point in the ingestion?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -424,6 +428,126 @@ def _ingest_gc_manifest(self, data, _site):
...
+            row['callrate'] = row['callrate'][:10]
+        except KeyError:
+            pass
+
+        # Set contamination field to fload
+        row['contamination'] = float(row['contamination'])
+
+        # Percentages shouldn't be less than 0
+        if row['contamination'] < 0:
+            row['contamination'] = 0
```
kenny-skaggs: From a maintenance/reusability perspective: there's code elsewhere that runs through a file and does this kind of processing for each row in the file, right? Is the only difference is that it is coupled to code that uses it for multiple rows in a file? What do you think of reusing that code (and refactoring it a bit if needed)? (that way, if anything changes about how to process the file, there's only one place to make updates)

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -1466,6 +1466,113 @@ def process_contamination_category_locally(self):
...
+class IngestionClass(GenomicManifestBase):
+    """
+    Perform a targeted ingestion of AW1 or AW2 data
+    for an arbitrary set of participants based on genomic_set_member.id
+    """
+    def run(self):
+
+        # Validate arguments
+        if not self.args.csv and not self.args.member_ids:
+            _logger.error('Either --csv or --samples must be provided.')
```
kenny-skaggs: Should this say `--member-ids` instead of `--samples`?

rdr_service/genomic/genomic_job_components.py
j-kanuch: Good question, I should clarify that the "state" the comment is referring to is the `genomicWorkflowState`. I will update the comments. The `genomicWorkflowStateModifiedTime` should only be updated if the `genomicWorkflowState` is updated. There is another `modified` field that is updated whenever the `genomic_set_member` is updated.  These two dates should be separate as data can be updated without the `genomicWorkflowState` changing. Downstream processes key off of `genomicWorkflowStateModifiedTime`, for example the GEM A1 workflow looks for this when processing records that have moved into the GEM_READY state.
The reason why we only want to update the state if the `genomicWorkflowState` is AW0 here is that these records may have moved forward in the process and have a state that is downstream of AW1. We don't want to regress their  `genomicWorkflowState` in this case.  For example, if we've ingested AW2 data for a record, we don't want to move it from AW2 back to AW1.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Good point, I meant to address this once I got the code working. The major differences between the new and existing code are that the existing code checks for control samples, processes multiple records, and needs to do a lookup on the `genomic_set_member object`. 
I can cut down on the duplicated code by refactoring the repeated code sections into their own functions.

rdr_service/tools/tool_libs/genomic_utils.py
j-kanuch: Yes, that was a late name change that got missed. Thanks!

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: Ah, sorry. I mis-read this as a two line if-check. That all makes sense, thanks!

kenny-skaggs (approved):

PR #2188, opened 2021-01-20
[DA-1886] Genomics AW3 File and Cron Job Updates

j-kanuch:
This PR splits the AW3 cron job into separate jobs for the Array and WGS workflows. This PR also updates the WGS data file extensions for the VCF files after reviewing the data during the Broad DRC test.


pegbertsch (approved):
Not seeing anything that would impact the handling of creation of the `bigquery_sync` records,  so approving based on that assumption.

PR #2191, opened 2021-01-22
[DA-1884] MIGRATION - Importing survey structure information from Redcap

kenny-skaggs:
This updates the codebook import process to create data structures that represent how a particular survey is written in Redcap. This is a solution for updating how we represent codes parenting each other. Questions and answer options within a survey are represented with their own records, and are associated with their codes. So question and answer codes are tied to module codes through a many-to-many structure using the Survey, SurveyQuestion, and SurveyQuestionOption objects.

These new objects also save question type information from Redcap, giving the question type that they are and any additional validation information.


pegbertsch (approved):
I still am exploring these code changes / trying it out in my local environment to see if I have any unanswered questions but looks good so far.  Would like to see this get included in the 1.88.1 release as a "soft rollout"

I think it's safe to merge since AFAIK there aren't any new dependencies yet on the new tables/Survey structures yet that would impact current RDR functionality.

PR #2193, opened 2021-01-22
Refactor: Using tool_base class in sync_consent tool

kenny-skaggs:
This updates the sync_consent tool to use the ToolBase class. I'm planning on updating the tool to share more code with the same code that runs for the cron job, and to be able to make a call to the server's ManuallyTriggerConsentSync endpoint when we need to zip consent files for the VA.

### DIFF
--- a/rdr_service/tools/tool_libs/sync_consent.py
+++ b/rdr_service/tools/tool_libs/sync_consent.py
@@ -1,26 +1,16 @@
-#! /bin/env python
-#
-# Copy Consent EHR files to HPO buckets.
-#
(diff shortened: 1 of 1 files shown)


dtharpeuno (approved):
Should have named that Base class `ToolTime`

👍

kenny-skaggs:
Months or years from now when I'm trying to shoehorn a feature into the tools that tracks how long they've run, know that it was all because I really wanted to create something called ToolTime (and don't be surprised if it breaks alot and has unsafe coding practices).

PR #2195, opened 2021-01-25
Fixes for BQ schema issues

pegbertsch:
Fixes for BQ schema generation issues found during PDR rebuild:
* Update the SQL for the BQ `v_pdr_participant_module` view to include the new `mod_external_id` field
* Restrict the queries to find questionnaire question codes based on the system value (to resolve issues with noncomformant data that exists in stable environment)
* Make the checks for duplicate field names when generating a module schema case-insentive

The stable and production BQ schemas have already been migrated with these changes incorporated

### DIFF
--- a/rdr_service/model/bq_questionnaires.py
+++ b/rdr_service/model/bq_questionnaires.py
@@ -107,10 +107,11 @@ class _BQModuleSchema(BQSchema):
                 from questionnaire_question qq where qq.questionnaire_id in (
                     select qc.questionnaire_id from questionnaire_concept qc
                             where qc.code_id = (
-                                select code_id from code c2 where c2.value = :module_id and system = :system
+                                select code_id from code c2 where c2.value = :module_id and c2.system = :system
                             )
                 )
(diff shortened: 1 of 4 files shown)


rdr_service/model/bq_questionnaires.py
```diff
@@ -181,7 +182,7 @@ def get_fields(self):
                 # flag duplicate fields.
                 found = False
                 for fld in fields:
-                    if fld['name'] == name:
+                    if fld['name'].lower() == name.lower():
```
kenny-skaggs, line 185: It looks like there might be a duplicate of this in the resource generators (https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/resource/schemas/questionnaires.py#L160). Should it be updated there too?

pegbertsch: Yes, I just noticed that too -- thanks!  Fixed in latest commit

kenny-skaggs (approved):

PR #2198, opened 2021-01-26
[DA-1882] Ingest retention eligible data

wangy70:
Ingest the retention metrics data from buckets using google cloud function and task. And remove our current retention status/type logic, use the imported data instead.


rdr_service/dao/retention_eligible_metrics_dao.py
```diff
@@ -0,0 +1,36 @@
...
+                and record.retentionType == retention_eligible_metrics_obj.retentionType \
+                and record.retentionEligibleTime == retention_eligible_metrics_obj.retentionEligibleTime:
+                return record.id, False
+            else:
+                return record.id, True
+        else:
+            return None, None
+
+    def upsert_all_with_session(self, session, retention_eligible_metrics_records):
+        records = list(retention_eligible_metrics_records)
```
kenny-skaggs: Why list them?

kenny-skaggs (dismissed):

wangy70:
Hold this one until we get the file dropped in production env.

rdr_service/alembic/versions/813b74f84338_retention_eligible_import_table.py
```diff
@@ -0,0 +1,92 @@
...
+def downgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('participant_summary', 'retention_type')
+    op.drop_column('participant_summary', 'retention_eligible_time')
+    op.drop_column('participant_summary', 'retention_eligible_status')
+    op.add_column('participant_summary',
+                  sa.Column('retention_eligible_status', rdr_service.model.utils.Enum(RetentionStatus),
+                            sa.Computed(
+                                '\n    CASE WHEN\n      consent_for_study_enrollment = 1\n      AND (consent_for_electronic_health_records = 1 OR consent_for_dv_electronic_health_records_sharing = 1)\n      AND questionnaire_on_the_basics = 1\n      AND questionnaire_on_overall_health = 1\n      AND questionnaire_on_lifestyle = 1\n      AND withdrawal_status = 1\n      AND suspension_status = 1\n      AND samples_to_isolate_dna = 1\n    THEN 2 ELSE 1\n    END\n',
+                                persisted=True), nullable=True))
```
robabram: Can these computed SQL statements be put in a multi-line Python string variable above in this file for readability, then use the variable in the 'sa.Computed()' function?

kenny-skaggs: Robert, these strings are part of the auto-generated migration and would be for adding back the computed columns if we were to downgrade. I imagine in that scenario we'd refer to the string at https://github.com/all-of-us/raw-data-repository/pull/2198/files#diff-ec950a38bbe9afabace937c2d149d8625b967edae25d41d47df7a676608e704dL74 for readability if needed and would rather leave the migration as is. How does that sound?

kenny-skaggs (approved):

rdr_service/gcloud_functions/retention_eligible_import_function/main.py
```diff
@@ -0,0 +1,89 @@
...
+
+        cloud_file_path = f'{self.event.bucket}/{self.event.name}'
+
+        data = {
+            "file_path": cloud_file_path,
+            "bucket_name": self.event.bucket,
+            "upload_date": self.event.timeCreated,
+        }
+
+        _logger.info("Pushing cloud task...")
```
pegbertsch, line 45: Optional:  Would it be useful to add details from the data dict to the log message to indicate which file the task is expected to process?

pegbertsch (approved):

PR #2199, opened 2021-01-26
Setting up a new lint process

kenny-skaggs:
This updates the lint process to be a little bit more robust and more convenient. Our current process lints all code files as they are on our workstation each time we push to a git branch. This new process will instead lint only the staged files when committing. All files will be linted in CircleCI as part of the unit test check (just in case something slips through).

Everyone will need to follow these steps to update their git hooks:
- cd to `raw-data-repository/.git/hooks`
- Modify the `pre-push` file and remove the line that has `git-hooks/pre-commit`. Alternatively you can delete the `pre-push` file if there's nothing else in there that would run.
- Create a new file named `pre-commit` with the following contents:
```
#!/bin/bash -e
git-hooks/pre-commit
```


.circleci/config.yml
```diff
@@ -18,6 +18,12 @@ commands:
   unittest-app-steps:
     steps:
       - checkout
+      - run:
+          name: Install pylint
+          command: sudo pip install pylint
```
dtharpeuno: Do you need this since the `Setup Python Environment` runs `setup_python_env.sh` which activates the venv and installs pylint? Maybe move the new step `Code static analysis checks` after the `Setup Python Environment` step?

dtharpeuno (approved):
💯

.circleci/config.yml
kenny-skaggs: Good catch, thanks! I thought I'd checked to see if it was part of the requirements file, but I guess I didn't check very closely.

PR #2203, opened 2021-01-28
[DA-1889] MIGRATION - Refactor AW1 workflow and support Control Samples

j-kanuch:
This PR updates the following:
- Refactor the AW1 workflow for more maintainability
- Create a new `genomic_set_member` record if the `parent_sample_id` in the AW1 manifest is found to be in the list of NIST control samples
- Change datatype of `genomic_set_member.biobank_id` to support the control sample `biobank_id` provided in the AW1
- Update the BQ schema to use a new string field for `biobank_id_str`.
- Update the genomic state machine to return the current state if the transition function doesn't recognize the value of the `signal` parameter.

A summary of the control sample workflow follows:
- Control samples are set up from a predetermined list from the Biobank. 
- They are created as `genomic_set_member` records with only the `sample_id` and `genomic_workflow_state` populated.
- The `genomic_workflow_state` is `CONTROL_SAMPLE` (34) for these records
- During AW1 processing, if a `genomic_set_member` isn't found in the RDR on lookup of `collection_tube_id` from the AW1, the workflow looks at the `parent_sample_id` from the AW1 to check if this sub-sample is parented to a control sample.
- If the above check succeeds, a new `genomic_set_member` record is created with the attributes from the AW1 file (`biobank_id`, `collection_tube_id`, etc.)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -302,131 +302,120 @@ def get_aw1_manifest_column_mappings():
...
+                    genomic_set_member_update(member.id)
 
-                # Skip already processed members if failure mode isn't defined in manifest
-                if member.reconcileGCManifestJobRunId is not None and row_copy['failuremode'] in (None, ''):
                     continue
 
-                member.gcSiteId = _site
+                else:
+                    # Not a control sample, continue with next check
+                    pass
```
kenny-skaggs: Do you think it could help readability to remove the `continue` above and have the rest of the iteration within the `else` block?

rdr_service/resource/generators/genomics.py
```diff
@@ -92,6 +92,12 @@ def make_resource(self, _pk, backup=False):
             row = ro_session.execute(text('select * from genomic_set_member where id = :id'), {'id': _pk}).first()
             data = self.ro_dao.to_dict(row)
 
+            # Delete biobank_id key since we send to biobank_id_str
+            try:
+                del data['biobank_id']
+            except KeyError:
+                pass
+
```
kenny-skaggs, line 101: Does this not need to set the `biobank_id_str` field?

rdr_service/genomic/genomic_job_components.py
j-kanuch: I originally had it the way you suggest, but I found that having the `continue` helped make it more readable since we want to terminate the iteration here explicitly. Nesting the remainder of the iteration into the `else` here added more complexity because of the condition on line 343. However, I can nest the block from 360-384 into this condition and that should increase readability.

rdr_service/resource/generators/genomics.py
j-kanuch: Yes, I missed that here. Updating now.

rdr_service/resource/generators/genomics.py
```diff
@@ -92,6 +92,13 @@ def make_resource(self, _pk, backup=False):
             row = ro_session.execute(text('select * from genomic_set_member where id = :id'), {'id': _pk}).first()
             data = self.ro_dao.to_dict(row)
 
+            # Set biobank_id_str and delete biobank_id
+            try:
+                data['biobank_id_str'] = data['biobank_id']
+                del data['biobank_id']
+            except KeyError:
+                pass
+
```
kenny-skaggs, line 101: Sorry for not thinking of it before, but I think this needs to be done in the BQ copy of this code too (https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/bq_genomics_dao.py#L105).

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: I see what you mean. I'd missed that there was a branch that could have skipped over all of it.

kenny-skaggs (dismissed):

rdr_service/resource/generators/genomics.py
j-kanuch: Yes, I see what you mean. I forgot that there are 2 places this needed to be done. Updating now.

kenny-skaggs (approved):

PR #2206, opened 2021-01-29
Remove C1, C2 AW0 job, change C3 schedule

j-kanuch:
This PR removes the Cohort 1 and Cohort 2 AW0 cron jobs and temporarily suspends the Cohort 3 AW0 cron job. The GEM A1 cron job is currently temporarily suspended on the devel branch. Both of these cron jobs will be enabled again when the CE participants are removed from the genomics system.

### DIFF
--- a/rdr_service/cron_default.yaml
+++ b/rdr_service/cron_default.yaml
@@ -59,20 +59,10 @@ cron:
   schedule: every day 02:45
   timezone: America/New_York
   target: offline
-- description: Genomic Pipeline AW0 (Cohort 1) Workflow
-  url: /offline/GenomicC1AW0Workflow
-  timezone: America/New_York
-  schedule: every monday 07:15
-  target: offline
-- description: Genomic Pipeline AW0 (Cohort 2) Workflow
-  url: /offline/GenomicC2AW0Workflow
-  schedule: every monday 07:00
-  timezone: America/New_York
-  target: offline
 - description: Genomic Pipeline AW0 (Cohort 3) Workflow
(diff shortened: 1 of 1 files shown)


rdr_service/cron_default.yaml
```diff
@@ -59,20 +59,10 @@ cron:
...
-  target: offline
-- description: Genomic Pipeline AW0 (Cohort 2) Workflow
-  url: /offline/GenomicC2AW0Workflow
-  schedule: every monday 07:00
-  timezone: America/New_York
-  target: offline
 - description: Genomic Pipeline AW0 (Cohort 3) Workflow
   url: /offline/GenomicC3AW0Workflow
   timezone: America/New_York
-  schedule: every wednesday 06:30
```
dtharpeuno, line 75: Do we have a piece of custom code that deciphers these strings? Or what is being used for that process?

dtharpeuno (approved):

rdr_service/cron_default.yaml
j-kanuch: This is the syntax that Google Cloud expects to decipher the cron schedule.  I personally think the cron style of scheduling is more straight forward (i.e. * * * 5, etc.). Below is a link to some documentation on the matter.
https://cloud.google.com/appengine/docs/flexible/python/scheduling-jobs-with-cron-yaml

dtharpeuno: Yea just have never seen that style before, was wondering if we had some NLP processing in our repo for that value.

PR #2207, opened 2021-01-29
Refactor: Creating ToolTestMixin to share code for testing tools

kenny-skaggs:
This PR creates a ToolTestMixin class to share code for testing ToolBase classes. I've also made the ToolBase class publicly available (rather than marked as private to the tool_libs module) so the mixin can reference it without warnings.


dtharpeuno (approved):
👍

PR #2215, opened 2021-02-04
Making auth check more robust

kenny-skaggs:
We occasionally see random ssl and connection (broken pipe) errors when trying to make a request to google's API for validating the authentication token. This adds logic to catch those errors and use the retry-loop to attempt the validation again.


dtharpeuno (approved):
👍

PR #2218, opened 2021-02-05
[ROC-823] Temporarily mapping old zip code answer fields

kenny-skaggs:
The curation team found some zip code answers that were given in a questionnaireResponseAnswer field on the valueInteger field rather than the valueString field. Curation is expecting these on the valueString field and would need to transfer them over if we didn't clean it up. This PR adds to the ETL query logic to map them correctly (based on the question code value). I plan on submitting a ticket to have these responses re-sent to us to correctly structure the data. When that is done we'll be able to remove this.


rdr_service/tools/tool_libs/curation.py
```diff
@@ -287,7 +291,11 @@ def _get_base_src_clean_answers_select(session):
                 func.left(QuestionnaireResponseAnswer.valueString, 1024),
                 QuestionnaireResponseAnswer.valueDate,
                 QuestionnaireResponseAnswer.valueDateTime,
-                answer_code.display
+                answer_code.display,
+                case([  # Use valueInteger if the question code should be re-mapped
+                    (question_code.value.in_(zipcode_question_codes_to_remap),
```
j-kanuch, line 296: I may be misreading this, but shouldn't this be `notin_` since you don't want to map zip codes to `valueInteger`?

kenny-skaggs: I think this is where we'd want to check that it is one of the codes to map (rather than using `notin_`).

This block maps these QuestionnaireResponseAnswer values from `valueString`, `valueDate`, `valueDateTime`, and `display` to the SrcClean `value_string` field. If the coalesce for setting the SrcClean.value_string column finds all of those fields to be NULL, then it checks to see if the question code matches one of the values we want to map. 

The two codes in `zipcode_question_codes_to_remap` sometimes have the answer in `valueInteger` instead. So if everything else is NULL, and the `in_` check results in True. Then we want to pass QuestionnaireResponseAnswer's `valueInteger` to SrcClean's `value_string` because curation is expecting zip codes in the `value_string` field.

Let me know if that clears it up. I could be thinking about this the wrong way around too.

j-kanuch: Thanks, that makes sense.

j-kanuch (approved):

PR #2228, opened 2021-02-11
Fixes for building local database

kenny-skaggs:
The `setup_local_database.sh` script was seeing some issues from recent changes. There's a possible OperationalError that can come up when checking for the `tmp_questionnaire_response` table. And the consent responses get rejected when trying to create local participants.


dtharpeuno (approved):
🔥 thanks... ran into this yesterday

pegbertsch (approved):

PR #2230, opened 2021-02-11
[DA-1880/1907] MIGRATION - Load Raw AW1 Data

j-kanuch:
This PR builds a new workflow for ingesting the raw AW1 data into a new table:`genomic_aw1_raw`. This is in support of the error logging and data quality pipeline.
The workflow is executed from the Google Cloud Functions that ingest the AW1 data and is triggered from a cloud tasks API endpoint. A tool was added to `genomic_utils.py` to execute the AW1 loading as well.


rdr_service/model/genomics.py
```diff
@@ -361,6 +361,59 @@ class GenomicManifestFeedback(Base):
...
+    visit_description = Column("visit_description", String(255), nullable=True)
+    sample_source = Column("sample_source", String(255), nullable=True)
+    study = Column("study", String(255), nullable=True)
+    tracking_number = Column("tracking_number", String(255), nullable=True)
+    contact = Column("contact", String(255), nullable=True)
+    email = Column("email", String(255), nullable=True)
+    study_pi = Column("study_pi", String(255), nullable=True)
+    test_name = Column("test_name", String(255), nullable=True)
+    failure_mode = Column("failure_mode", String(255), nullable=True)
+    failure_mode_desc = Column("failure_mode_desc", String(255), nullable=True)
```
kenny-skaggs, line 410: I'm a fan of switching the naming convention style.

If you want to fully take advantage of it you can remove the first argument and not need to re-specify the name, so it could be `failure_mode_desc = Column(String(255), nullable=True)` instead

kenny-skaggs (approved):

rdr_service/model/genomics.py
j-kanuch: Nice, in a follow-up PR I'll do just that.

PR #2236, opened 2021-02-15
Moving unit test sql server port

kenny-skaggs:
9306 seems to be a default port for Mysql and we can occasionally see another server running there. This updates the unit test code to use another port (10010 by default, but can also be overridden with a environment variable).

The code uses the db_config file when looking for how to connect to the database. That can be overridden, but only if the file originally had something for the key anyway. So these changes needed to leave something in the db_config setup so that it could later be found and overridden.

### DIFF
--- a/tests/.test_configs/db_config.json
+++ b/tests/.test_configs/db_config.json
@@ -1,7 +1,7 @@
 {
     "db_connection_string": "mysql+mysqldb://rdr:rdr!pwd@127.0.0.1/?charset=utf8",
     "backup_db_connection_string": "mysql+mysqldb://rdr:rdr!pwd@127.0.0.1/?charset=utf8",
-    "unittest_db_connection_string": "mysql+mysqldb://root@127.0.0.1:9306/?charset=utf8",
+    "unittest_db_connection_string": "<overridden in tests>",
     "db_password": "rdr!pwd",
     "db_connection_name": "",
     "db_user": "rdr",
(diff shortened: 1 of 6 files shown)


tests/helpers/mysql_helper.py
```diff
@@ -45,7 +45,11 @@
 from tests.helpers.mysql_helper_data import PITT_HPO_ID, PITT_ORG_ID, AZ_HPO_ID, AZ_ORG_ID
 
 BASE_PATH = '{0}/rdr-mysqld'.format(tempfile.gettempdir())
-MYSQL_PORT = 9306
+MYSQL_HOST = "127.0.0.1"  # Do not use 'localhost', we want to force using an IP socket.
+MYSQL_PORT = os.getenv('RDR_UNITTEST_SQL_SERVER_PORT', 10010)
+
+db_conn_str = f'mysql+mysqldb://root@{MYSQL_HOST}:{MYSQL_PORT}/?charset=utf8'
+config.override_setting('unittest_db_connection_string', db_conn_str)
```
dtharpeuno: Nice!

tests/.test_configs/db_config.json
```diff
@@ -1,7 +1,7 @@
 {
     "db_connection_string": "mysql+mysqldb://rdr:rdr!pwd@127.0.0.1/?charset=utf8",
     "backup_db_connection_string": "mysql+mysqldb://rdr:rdr!pwd@127.0.0.1/?charset=utf8",
-    "unittest_db_connection_string": "mysql+mysqldb://root@127.0.0.1:9306/?charset=utf8",
+    "unittest_db_connection_string": "<overridden in tests>",
```
dtharpeuno, line 4: Does this get evaluated somehow or is it just there for a comment?

dtharpeuno (dismissed):

tests/.test_configs/db_config.json
kenny-skaggs: It's mostly there as a placeholder. To get the code that overrides the connection string to work there needs to be something in the config file for it to override.

dtharpeuno (approved):

PR #2240, opened 2021-02-16
[DA-1853] Ingest aw5 manifest

wangy70:


rdr_service/model/bq_genomics.py
```diff
@@ -362,21 +362,36 @@ class BQGenomicGCValidationMetricsSchema(BQSchema):
     site_id = BQField('site_id', BQFieldTypeEnum.STRING, BQFieldModeEnum.NULLABLE)
     chipwellbarcode = BQField('chipwellbarcode', BQFieldTypeEnum.STRING, BQFieldModeEnum.NULLABLE)
     idat_green_received = BQField('idat_green_received', BQFieldTypeEnum.INTEGER, BQFieldModeEnum.NULLABLE)
+    idat_green_deleted = BQField('idat_green_deleted', BQFieldTypeEnum.INTEGER, BQFieldModeEnum.NULLABLE)
```
j-kanuch: Can the new "deleted" fields be added to the end of the schema definition instead of in their current place? When making schema changes to these BQ tables, if the new fields are added to the end, we can just rebuild the records in the PDR as the data is updated. Otherwise, we would need to truncate the tables and then rebuild all records.

rdr_service/resource/schemas/genomics.py
```diff
@@ -221,21 +221,36 @@ class GenomicGCValidationMetricsSchema(Schema):
     site_id = fields.String(validate=validate.Length(max=80))
     chipwellbarcode = fields.String(validate=validate.Length(max=80))
     idat_green_received = fields.Int16()
+    idat_green_deleted = fields.Int16()
```
j-kanuch: Same as above for these fields. Can we move the new `*_deleted` fields to the end of the definition?

rdr_service/model/bq_genomics.py
wangy70: Done

rdr_service/resource/schemas/genomics.py
wangy70: Done

j-kanuch (approved):

PR #2242, opened 2021-02-17
Raising operational error for client request

kenny-skaggs:
In a recent PR I had operational errors (lock wait timeouts) use the retry loop if something happened. But if the loop hit the same operational error each time it would no longer send back the details, it would instead send the message that it retried multiple times and failed.

This updates the code to check if it's on the last try. If it is then it raises the error (as if it hadn't been in the retry loop).

### DIFF
--- a/rdr_service/dao/base_dao.py
+++ b/rdr_service/dao/base_dao.py
@@ -9,8 +9,6 @@ from contextlib import closing
 
 import sqlparse
 
-from sqlalchemy.exc import OperationalError
-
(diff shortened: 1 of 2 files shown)


rdr_service/dao/base_dao.py
```diff
@@ -9,8 +9,6 @@
 
 import sqlparse
 
-from sqlalchemy.exc import OperationalError
-
```
kenny-skaggs, line 13: Duplicates import down on line 19.

dtharpeuno (approved):

PR #2245, opened 2021-02-17
Batch process samples in genomic manifest

dtharpeuno:
We need to speed up the insertion process into the db when we batch process the samples for manifest generation. This PR utilizes the sqlalchemy bulk_save_objects method and batch inserts every 1k of object in enumeration.

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -2234,8 +2234,11 @@ class GenomicBiobankSamplesCoupler:
                     validationFlags=valid_flags,
                     ai_an='N' if samples_meta.valid_ai_ans[i] else 'Y',
                     genomeType=self._ARRAY_GENOME_TYPE,
-                    genomicWorkflowState=GenomicWorkflowState.AW0_READY
+                    genomicWorkflowState=GenomicWorkflowState.AW0_READY,
+                    created=clock.CLOCK.now(),
+                    modified=clock.CLOCK.now(),
                 )
+
                 # Also create a WGS member
                 new_wgs_member_obj = deepcopy(new_array_member_obj)
                 new_wgs_member_obj.genomeType = self._WGS_GENOME_TYPE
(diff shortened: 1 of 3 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -2185,66 +2185,76 @@ def process_samples_into_manifest(self, samples_meta, cohort, saliva=False, loca
...
+
+                if count % 1000 == 0:
+                    session.bulk_save_objects(processed_array_wgs)
+                    bulk_ids = self.member_dao.get_members_from_set_id(new_genomic_set.id)
+                    bq_genomic_set_member_batch_update(bulk_ids, project_id=self.controller.bq_project_id)
+                    genomic_set_member_batch_update(bulk_ids)
+                    processed_array_wgs.clear()
+
+            if count and processed_array_wgs:
+                session.bulk_save_objects(processed_array_wgs)
```
kenny-skaggs, line 2262: Does `bulk_save_objects` trigger the SqlAlchemy ORM events set up on the model class (set up here https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/model/genomics.py#L54). If not, would we want to set the created and modified times in the code?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2185,66 +2185,76 @@ def process_samples_into_manifest(self, samples_meta, cohort, saliva=False, loca
...
+                # Also create a WGS member
+                new_wgs_member_obj = deepcopy(new_array_member_obj)
+                new_wgs_member_obj.genomeType = self._WGS_GENOME_TYPE
+
+                processed_array_wgs.extend([new_array_member_obj, new_wgs_member_obj])
+                count += 1
+
+                if count % 1000 == 0:
+                    session.bulk_save_objects(processed_array_wgs)
+                    bulk_ids = self.member_dao.get_members_from_set_id(new_genomic_set.id)
```
kenny-skaggs: I understand the need to load the ids for the batch of genomic members that were just inserted, but would this also load the ids from all previously saved batches for this genomic set?

I'm thinking the first batch saves and the ids would load as expected. But then the second batch would load all the ids from the first and second batches (and send re-send the ids from the first batch to the bq update methods).

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Good catch! Will update.

rdr_service/genomic/genomic_job_components.py
dtharpeuno: You are correct. Logic has been updated to reflect.

kenny-skaggs (approved):

PR #2248, opened 2021-02-19
[DA-1909] MIGRATION - Genomic Incident table

j-kanuch:
This PR creates a new table: `genomic_incident`. The following workflows were updated:
- The AW1 ingestion workflow was updated to insert a `genomic_incident` record if all sample look-ups fail to find an associated `genomic_set_member` record. 
- The AW2 ingestion workflow was updated to insert a `genomic_incident` record if all sample look-ups fail to find an associated `genomic_set_member` record.
- The AW1 and AW2 workflows no longer raise exceptions when the aforementioned look-ups fail.
- The `_exit` method on the `GenomicJobController` context manager was updated to insert a `genomic_incident` record if the `job_result` was unsuccessful.


rdr_service/dao/genomics_dao.py
```diff
@@ -1446,3 +1446,15 @@ def get_from_filepath(self, filepath):
...
+class GenomicIncidentDao(UpdatableDao):
+    def __init__(self):
+        super(GenomicIncidentDao, self).__init__(
+            GenomicIncident, order_by_ending=['id'])
+
+    def get_id(self, obj):
+        pass
+
+    def from_client_json(self):
+        pass
```
dtharpeuno, line 1460: Is this to mitigate the `NotImplementedError` on the parent class?

j-kanuch: Yes, exactly.

dtharpeuno (approved):
👍

PR #2249, opened 2021-02-19
[DA-1914] Add optional character to regex pattern to support UW naming convention.

j-kanuch:
This PR updates the WGS data reconciliation workflow file naming pattern. UW has a different naming convention than the other 2 GCs: they include a "v" before the version, i.e. "..._v1.cram".  Because "v" is not an integer, the sorting function needed to be updated. Because of the inconsistencies between the GCs, the workflow was modified to recognize both file naming conventions.


dtharpeuno (approved):
Nice

PR #2253, opened 2021-02-23
[ROC-834] Adding Dec and Feb COPE surveys to retention type calculation

kenny-skaggs:
This adds the December and February COPE surveys to the retention type calculation.

### DIFF
--- a/rdr_service/dao/participant_summary_dao.py
+++ b/rdr_service/dao/participant_summary_dao.py
@@ -820,6 +820,10 @@ class ParticipantSummaryDao(UpdatableDao):
                  model.questionnaireOnCopeJuneAuthored > eighteen_month_ago) or \
                 (model.questionnaireOnCopeMayAuthored and
                  model.questionnaireOnCopeMayAuthored > eighteen_month_ago) or \
+                (model.questionnaireOnCopeDecAuthored and
+                 model.questionnaireOnCopeDecAuthored > eighteen_month_ago) or \
+                (model.questionnaireOnCopeFebAuthored and
+                 model.questionnaireOnCopeFebAuthored > eighteen_month_ago) or \
                 (model.consentCohort == ParticipantCohort.COHORT_1 and
                  model.consentForStudyEnrollmentAuthored != model.consentForStudyEnrollmentFirstYesAuthored and
                  model.consentForStudyEnrollmentAuthored > eighteen_month_ago) or \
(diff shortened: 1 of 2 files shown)


pegbertsch (commented):
Do we need to confirm that the questionnaireOnCope* (status) field is an exact match to SUBMITTED and the QuestionnaireResponse is COMPLETED?   Scott implied he's using the combination of SUBMITTED status and a COMPLETED response in his PTSC reports but I don't know if it was for retention.

wangy70:
Can we hold this one? This part will be replaced by the retention metrics file soon. I'm working on that one, just want to avoid more conflicts...

kenny-skaggs:
> Do we need to confirm that the questionnaireOnCope* (status) field is an exact match to SUBMITTED and the QuestionnaireResponse is COMPLETED? Scott implied he's using the combination of SUBMITTED status and a COMPLETED response in his PTSC reports but I don't know if it was for retention.

@pegbertsch I'm not sure. I don't believe the reported issue was related to whether a participant had consented to the COPE surveys or not, and I'm not sure that was part of the original definitions for the fields. From what I understand of the retention status, I don't believe it regards data but is more a reflection of participant activity. So even if a participant submits an answer of 'No consent', they'll still at least be taking an action on something and their retention status should reflect that. But that's all guess-work. In any case I think Yu is working on changes to have us read the values from Vibrent, so they'll be in charge of the definitions soon.

> Can we hold this one? This part will be replaced by the retention metrics file soon. I'm working on that one, just want to avoid more conflicts...

@wangy70 There are sites that are waiting on this correction. Do you know when the changes for getting the retention fields from Vibrent will be ready? My hope is to get this corrected in the next hotfix.

wangy70:
@kenny-skaggs  OK, if that's the case, please go ahead submit your PR, I will resolve the conflicts later. I'm not sure when the new change will be ready, there are still some communication around it.

pegbertsch:
@kenny-skaggs I have asked Scott about the COPE survey SUBMITTED_NO_CONSENT case (if they declined or deferred taking the survey)and if that still qualifies as "active" retention but have not gotten a response yet

pegbertsch (approved):
Approving based on slack discussion that we can fix the active/passive status for SUBMITTED_NO_CONSENT COPE survey statuses later, if necessary.   PTSC may also be counting those as an active retention qualifier vs. passive (but that is still being confirmed)

PR #2257, opened 2021-02-24
[PDR-261] Fix logic for PDR generator ubr_age_at_consent flag setting

pegbertsch:
Adjust `ubr_age_at_consent` calculation so people over 65 (but not yet 66) are counted in UBR group

Also:  add options to `resource participant` tool restrict rebuilding of participant data to just the `pdr_participant`  data record (`--no-modules` option), or just the questionnaires (`pdr_mod_*` records ) for a participant but not their `pdr_participant` data (`--modules-only `option)


pegbertsch:
Kenny brought up a question about leap year calculations.   Would like to still submit this change for now as I think it mitigates the reported issue, but confirm with PTSC how they are handling age calculations w.r.t. leap year cases.

kenny-skaggs (approved):

PR #2261, opened 2021-02-25
[DA-1923] ps_api_support_core_minus_pm

wangy70:
Add a new enrollment status `CORE_MINUS_PM` to participant summary API


kenny-skaggs:
From what I understand CORE_MINUS_PM should behave the same as CORE status. And when a participant reaches CORE status they can't do anything do have that status removed (such as revoking EHR consent). Should a participant be able to lose their CORE_MINUS_PM status?

wangy70:
It's not mentioned in the requirement, let me check with them.

wangy70:
@kenny-skaggs  Updated, CORE_MINUS_PM can not be downgrade now.

kenny-skaggs (approved):

PR #2265, opened 2021-03-01
[DA-1911] Data-dictionary updater

kenny-skaggs:
This is work in progress for automatically updating the data-dictionary with the database schema. It uses the schema gathered from the database, and pairs that with information from tables that are mapped to ORM models and others that are hardcoded. The Sphinx code parser is used to get to the docstrings on class-level attributes for the ORM models. There it looks for special strings (such as `@rdr_dictionary_show_unique_values`) to know how and where to display data for the column in the data-dictionary.

This is most of the work for updating the data-dictionary. But it still needs to be set up as part of the deploy process and it needs to be able to write the change-log. Further PRs will wrap up the rest of the work for it.

### DIFF
--- a/rdr_service/services/data_dictionary_updater.py
+++ b/rdr_service/services/data_dictionary_updater.py
@@ -283,4 +283,5 @@ class DataDictionaryUpdater:
         ) as sheet:
             self._populate_schema_tabs(sheet)
             self._populate_hpo_key_tab(sheet)
+            self._populate_site_key_tab(sheet)
             self._populate_questionnaire_key_tab(sheet)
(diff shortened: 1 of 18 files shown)


rdr_service/services/data_dictionary_updater.py
```diff
@@ -0,0 +1,287 @@
...
+        self.internal_table_list = [
+            'alembic_version',
+            'metrics_age_cache_bak',
+            'metrics_enrollment_status_cache_bak',
+            'metrics_gender_cache_bak',
+            'metrics_lifecycle_cache_bak',
+            'metrics_race_cache_bak',
+            'metrics_region_cache_bak',
+            'temp_debug_bigquery_sync',
+            'tmp_participants'
```
dtharpeuno, line 38: Might want to add `ptsc_cohort` to this list.

kenny-skaggs: Good idea, thanks for noticing it.

dtharpeuno (approved):
🔥 🔥 🔥

PR #2266, opened 2021-03-02
File name insertion Genomic Manifest File

dtharpeuno:
In previous PR forgot to add this along with migration, backfill for data is planned, but need this to be inserted with new records

### DIFF
--- a/rdr_service/genomic/genomic_job_controller.py
+++ b/rdr_service/genomic/genomic_job_controller.py
@@ -121,7 +121,7 @@ class GenomicJobController:
                 bucketName=path_list[0],
                 recordCount=0,  # Initializing with 0, counting records when processing file
                 rdrProcessingComplete=0,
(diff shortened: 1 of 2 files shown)


rdr_service/genomic/genomic_job_controller.py
```diff
@@ -494,6 +495,7 @@ def generate_manifest(self, manifest_type, _genome_type, **kwargs):
                     recordCount=result['record_count'],
                     rdrProcessingComplete=1,
                     rdrProcessingCompleteDate=now_time,
+                    fileName=new_file_path.split('/')[2]
```
j-kanuch: We may want to change this to `new_file_path.split('/')[-1]` in case there is another sub-folder in the path. That way it always pulls the file name.

j-kanuch (approved):

PR #2268, opened 2021-03-02
[DA-1894] Filter records from AW2F if they are missing contamination

j-kanuch:
This is a small PR to update the AW2F query. This adds a filter to remove samples if they are missing contamination data.


rdr_service/genomic/genomic_job_components.py
```diff
@@ -2993,6 +2994,8 @@ def _get_source_data_query(self, manifest_type):
                 ).where(
                     (GenomicSetMember.genomicWorkflowState != GenomicWorkflowState.IGNORE) &
                     (GenomicGCValidationMetrics.ignoreFlag == 0) &
+                    or_(GenomicGCValidationMetrics.contamination.isnot(None),
```
dtharpeuno: Was about to ask if the second line was needed but ran a query and seems like there is some empty strings in the contamination records. Should we get rid of the second line and if its an empty string the field run a backfill to change it back to NULL?

dtharpeuno (approved):

rdr_service/genomic/genomic_job_components.py
j-kanuch: Ideally there should be contamination data for every sample that is ingested in an AW2. We need to follow-up on the samples that are currently missing contamination data as they may need to be updated by the GCs. These filters are fail-safes so that we don't send them to the Biobank.

kenny-skaggs: The intent of the change is to make sure returned data doesn't have a related GenomicGCValidationMetrics that is missing contamination data (`contamination` is NULL or an empty string), right?

Does the `or_` needs to be removed? `GenomicGCValidationMetrics.contamination is not Null OR GenomicGCValidationMetrics.contamination != ''` still evaluates to True if `GenomicGCValidationMetrics.contamination` is an empty string.

j-kanuch: Yea, this needs to be an AND condition.

PR #2270, opened 2021-03-03
[DA-1927] - Genomic long read pilot workflow

dtharpeuno:
Adding workflow for genomics long read pilot:
-- added to genomic utils (tools)
-- slight refactoring in other logic, mainly for inserts on genomic set member table
-- other PEP8 warnings

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -2289,7 +2289,7 @@ class GenomicBiobankSamplesCoupler:
                 processed_members.append(dup_member_obj)
                 count = i
 
-                if count % 100 == 0:
+                if count % 100 == 0 and count > 0:
                     self.genomic_members_insert(
                         members=processed_members,
                         session=session,
@@ -2384,7 +2384,7 @@ class GenomicBiobankSamplesCoupler:
                 processed_array_wgs.extend([new_array_member_obj, new_wgs_member_obj])
                 count = i
 
-                if count % 1000 == 0:
+                if count % 1000 == 0 and count > 0:
                     self.genomic_members_insert(
                         members=processed_array_wgs,
                         session=session,
@@ -2468,7 +2468,12 @@ class GenomicBiobankSamplesCoupler:
 
     def genomic_members_insert(self, *, members, session, set_id, bids):
         """
-
+        Bulk save of member for genomic_set_member as well as PDR
+        batch updating of members
+        :param: members
+        :param: session
+        :param: set_id
+        :param: bids
         """
         try:
             session.bulk_save_objects(members)
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -2387,9 +2387,9 @@ class GenomicBiobankSamplesCoupler:
 
                 bids.append(bid)
                 processed_array_wgs.extend([new_array_member_obj, new_wgs_member_obj])
-                count = i
+                count = i + 1
 
-                if count % 1000 == 0 and count > 0:
+                if count % 1000 == 0:
                     self.genomic_members_insert(
                         members=processed_array_wgs,
                         session=session,
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -2247,22 +2247,20 @@ class GenomicBiobankSamplesCoupler:
         based on downstream filters.
         :return:
         """
-        participants = self._get_long_read_participants()
+        participants = self._get_long_read_participants(max_num=384)
 
         if len(participants) > 0:
             return self.process_genomic_members_into_manifest(
-                participants=participants,
-                max_num=384
+                participants=participants
             )
 
         logging.info(f'Long Read Participant Workflow: No participants to process.')
         return GenomicSubProcessResult.NO_FILES
 
-    def process_genomic_members_into_manifest(self, *, participants, max_num=None):
+    def process_genomic_members_into_manifest(self, *, participants):
         """
         Compiles AW0 Manifest from already submitted genomic members.
         :param participants:
-        :param max_num:
         :return:
         """
 
@@ -2299,9 +2297,6 @@ class GenomicBiobankSamplesCoupler:
                         bids=[pm.biobankId for pm in processed_members]
                     )
                     processed_members.clear()
-
-                if max_num and count == max_num:
-                    break
 
             if count and processed_members:
                 self.genomic_members_insert(
@@ -2397,6 +2392,7 @@ class GenomicBiobankSamplesCoupler:
                         bids=bids
                     )
                     processed_array_wgs.clear()
+                    bids.clear()
 
             if count and processed_array_wgs:
                 self.genomic_members_insert(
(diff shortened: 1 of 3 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -2312,25 +2387,24 @@ def process_samples_into_manifest(self, samples_meta, cohort, saliva=False, loca
...
-                    bq_genomic_set_member_batch_update(member_ids, project_id=self.controller.bq_project_id)
-                    genomic_set_member_batch_update(member_ids)
+                    self.genomic_members_insert(
+                        members=processed_array_wgs,
+                        session=session,
+                        set_id=new_genomic_set.id,
+                        bids=bids
+                    )
                     processed_array_wgs.clear()
-                    bids.clear()
```
kenny-skaggs, line 2325: Does this not need to clear the `bids` list anymore?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2312,25 +2387,24 @@ def process_samples_into_manifest(self, samples_meta, cohort, saliva=False, loca
...
-                members = self.member_dao.get_members_from_set_id(new_genomic_set.id, bids=bids)
-                member_ids = [m.id for m in members]
-                bq_genomic_set_member_batch_update(member_ids, project_id=self.controller.bq_project_id)
-                genomic_set_member_batch_update(member_ids)
+                self.genomic_members_insert(
+                    members=processed_array_wgs,
+                    session=session,
+                    set_id=new_genomic_set.id,
+                    bids=bids
+                )
```
kenny-skaggs, line 2404: I like the idea of breaking this out into it's own function 😄

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2543,6 +2638,32 @@ def _get_new_c1_participants(self, from_date):
...
+            ).join(
+                ParticipantRaceAnswers,
+                ParticipantRaceAnswers.participantId == ParticipantSummary.participantId,
+            ).join(
+                Code,
+                ParticipantRaceAnswers.codeId == Code.codeId,
+            ).filter(
+                Code.value == 'WhatRaceEthnicity_Black',
+                GenomicSetMember.genomeType == 'aou_wgs',
+                GenomicSetMember.genomeType != 'long_read',
```
kenny-skaggs: Is the intention with this to filter out any participants that were already submitted for a long-read manifest?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2543,6 +2638,32 @@ def _get_new_c1_participants(self, from_date):
...
+                GenomicSetMember.genomicWorkflowState != GenomicWorkflowState.IGNORE,
+                ParticipantSummary.participantOrigin == 'vibrent',
+                ParticipantSummary.ehrUpdateTime.isnot(None)
```
kenny-skaggs: I think Sunny's idea was to use participants that had recent EHR update times. From what I understand the hope is to get participants that have submitted alot of EHR files, but we just don't know the number of distinct files. We could use the latest, but that could get really new participants that have just submitted one.

What if we ordered the list of participants by the amount of time between their first EHR upload (`ehrReceiptTime`) and their latest upload (`ehrUpdateTime`) with the biggest time spans at the top?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2543,6 +2638,32 @@ def _get_new_c1_participants(self, from_date):
...
+            ).all()
```
kenny-skaggs: Will the number of participants being used for a long_read manifest always be limited to a cap (set to 384 where these participants are processed)? Would it work to have a limit clause on this query (rather than returning all GenomicSetMembers that match the criteria)? That could help with performance if we aren't expecting to use them all.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2729,7 +2849,7 @@ def _get_source_data_query(self, manifest_type):
                     (GenomicGCValidationMetrics.idatGreenMd5Received == 1) &
                     (GenomicGCValidationMetrics.vcfReceived == 1) &
                     (GenomicGCValidationMetrics.vcfMd5Received == 1) &
-                    (GenomicSetMember.aw3ManifestJobRunID == None)
+                    (GenomicSetMember.aw3ManifestJobRunID.is_(None))
```
kenny-skaggs, line 2858: 🎉

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Yea sounds good. Will add it. Still getting familiar with the ORM.

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Correct. Since we have to run this multiple times, like once a month, don't want to pick up the ones that have been denoted with that genome type already.

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Yea discussed it with Josh, just going with checking for `ehrUpdateTime` at this time.

rdr_service/genomic/genomic_job_components.py
dtharpeuno: 👍

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Your right. Adding it back. Thanks for seeing that.

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: I think there would need to be something more for filtering out `aou_wgs` genome type set members that already have a corresponding `long_read` genome type set member. This will pick up a set member with a genome type of `aou_wgs` the first time around. And after that first pass (if I'm understanding this right), each participant used would have two set members: one for `aou_wgs` and the other for `long_read`. Then the second time this query could still use the `aou_wgs` types that it used before because their type isn't `long_read`.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2239,11 +2241,79 @@ def create_c1_genomic_participants(self, from_date, local=False):
...
+                    biobankId=participant.biobankId,
+                    genomicSetId=new_genomic_set.id,
+                    participantId=participant.participantId,
+                    nyFlag=participant.nyFlag,
+                    sexAtBirth=participant.sexAtBirth,
+                    collectionTubeId=participant.collectionTubeId,
+                    validationStatus=participant.validationStatus,
+                    validationFlags=participant.validationFlags,
+                    ai_an=participant.ai_an,
+                    genomeType='long_read',
```
j-kanuch: Can you put this in a variable? Similar to these:
` _ARRAY_GENOME_TYPE = "aou_array"`
` _WGS_GENOME_TYPE = "aou_wgs"`

j-kanuch (approved):

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Sure.
https://github.com/all-of-us/raw-data-repository/pull/2270/commits/32a1ed82658f62b9b1b3d02acd909f652dd2484d

PR #2273, opened 2021-03-04
[DA-1938] Modify Ingestion Cloud Functions

j-kanuch:
This PR adds a new task queue for genomics with 0 retries. All genomics cloud functions were updated to use this task queue. The Northwest AW1 cloud function was updated to ignore files in the `downloaded` sub-folder.


dtharpeuno (approved):
Nice. Can we merge this asap, maybe go out with hotfix, so we can mitigate this issue with Northwest doing this.

PR #2276, opened 2021-03-04
Update call to create_task() 

j-kanuch:
This small PR updates the call to `create_task()`. The resource rebuild tool fails when positional arguments are given for `parent` and `task` due to other expected parameters. This was updated to call `create_task()` with keyword arguments to fix the error.


pegbertsch (approved):
Approved as is.

Since the theory is that the error seen (incorrect number of positional args?)  might only occur without this change depending on what version of the cloud SDK is installed in the local environment:  If we can confirm that later, we may want to review other tools to see if any other calls might be impacted.

PR #2285, opened 2021-03-09
[ROC-838] Mapping unset demographics answers to skip

kenny-skaggs:
In September of 2018 the PTSC introduced the Skip code, prior to that we didn't receive answers for the skipped demographics questions. These changes are to provide a uniform dataset by displaying Skip for the demographics questions that we didn't receive an answer for. We should continue providing UNSET if a participant doesn't have a response for TheBasics survey, or if we're masking the responses for a participant that is withdrawn.


rdr_service/dao/participant_summary_dao.py
```diff
@@ -822,14 +824,17 @@ def get_client_id(self):
             client_id = 'example'  # account for temp configs that dont create the key
         return client_id
 
-    def to_client_json(self, model):
+    def to_client_json(self, model: ParticipantSummary):
```
dtharpeuno, line 827: oh that `type hinting` 👍

dtharpeuno (approved):

PR #2288, opened 2021-03-09
[ROC-880] MIGRATION - Storing codeable concept text from physical measurement observations

kenny-skaggs:
HPRO can send a physical measurement observation (captured by our Measurement class) with a FHIR ValueCodeableConcept object. If the code value indicates a protocol modification, we should also be receiving a text description that provides a reason for the deviation from protocol. This adds a column so that we can make that easily accessible to sql queries.

I'll make a follow up PR to create a temporary tool to backfill the existing physical measurements using the resource column in the database.

As far as I can tell, the physical measurement observations aren't sent to PDR.


rdr_service/dao/physical_measurements_dao.py
```diff
@@ -529,6 +530,14 @@ def from_observation(observation, full_url, qualifier_map, first_pass):
...
             value_code_system = value_coding.system
             value_code_value = value_coding.code
+
+            value_code_description = observation.valueCodeableConcept.text
+            desc_char_count = len(value_code_description)
+            char_limit = Measurement.valueCodeDescription.type.length
+            if desc_char_count > char_limit:
+                logging.warning(f'Truncating codeable concept description of length {desc_char_count}')
+                value_code_description = value_code_description[:char_limit]
+
```
dtharpeuno, line 540: I see that the field on the model is set to `512`,  should it be bigger since you are having to do this check? Is it okay that the value truncated?

dtharpeuno (approved):
👍

rdr_service/dao/physical_measurements_dao.py
kenny-skaggs: I set up the check as a safeguard to make sure it doesn't crash if it's too long (I don't think sqlalchemy protects us from an exception if it's too long but I haven't verified that). So I'm not expecting anything to get truncated. But if it does we'll become aware of it and be able to resolve it while the physical measurements is still able to get into the system and be used.

PR #2289, opened 2021-03-09
[ROC-880] Backfill tool for measurement codeable concept text

kenny-skaggs:
This makes it more convenient to backfill the description field for the physical measurement observations. The code for parsing the physical measurement request is derived from https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/physical_measurements_dao.py#L530 and https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/physical_measurements_dao.py#L621

### DIFF
--- a/rdr_service/tools/tool_libs/backfill_physical_measurement_observation_code_text.py
+++ b/rdr_service/tools/tool_libs/backfill_physical_measurement_observation_code_text.py
@@ -1,0 +1,57 @@
+from datetime import datetime
+
+from rdr_service.dao.physical_measurements_dao import _OBSERVATION_RESOURCE_TYPE, PhysicalMeasurementsDao
+from rdr_service.lib_fhir.fhirclient_1_0_6.models import observation as fhir_observation
+from rdr_service.model.measurements import Measurement, PhysicalMeasurements
+from rdr_service.tools.tool_libs.tool_base import cli_run, ToolBase, logger
+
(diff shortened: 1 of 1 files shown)


rdr_service/tools/tool_libs/backfill_physical_measurement_observation_code_text.py
```diff
@@ -0,0 +1,58 @@
...
+        with self.get_session() as session:
+            found_measurements = True
+            while found_measurements:
+                found_measurements = False
+                physical_measurements_query = session.query(
+                    PhysicalMeasurements.physicalMeasurementsId,
+                    PhysicalMeasurements.resource
+                ).filter(
+                    PhysicalMeasurements.physicalMeasurementsId > latest_id
+                ).order_by(PhysicalMeasurements.physicalMeasurementsId).limit(500)
```
dtharpeuno, line 26: What is the range for the `physicalMeasurementsId`?

dtharpeuno (approved):
Nice use of the `while` loop!

rdr_service/tools/tool_libs/backfill_physical_measurement_observation_code_text.py
kenny-skaggs: Looks like it might be randomly generated and ranges between 100,000,000 and 1,000,000,000 right now.

PR #2292, opened 2021-03-10
Fix down revision in alembic

dtharpeuno:
-- fixing revision for alembic builds


rdr_service/alembic/versions/061dc1adde35_.py
```diff
@@ -21,7 +21,7 @@
 
 # revision identifiers, used by Alembic.
 revision = '061dc1adde35'
-down_revision = 'bb29335721aa'
+down_revision = 'd5e7affa8201'
```
j-kanuch, line 24: Can you also update the down revision in the doc-string on line 4?

dtharpeuno: https://github.com/all-of-us/raw-data-repository/pull/2292/commits/5307af24d3591347fef0220ac8347011f8058f99

j-kanuch (approved):

PR #2298, opened 2021-03-11
Aborting workers rather than quitting them

kenny-skaggs:
Part of the `handle_quit` method for Gunicorn's thread workers is to shutdown the thread pool. It seems like if a request can come in at just the right time, a Gunicorn worker that is shutting down can still accept the request but then crash when trying to interact with the thread pool.

The `handle_abort` method does everything that `handle_quit` does, except shutting down the thread pool. This may mean that if a worker is still able to accept a request, then it will process it as it's shutting down. But that should be ok.

I've tested this locally and was able to reproduce the 502 error received when the thread pool error happens. Switching to the abort method resolves the problem.

### DIFF
--- a/rdr_service/services/gunicorn_config.py
+++ b/rdr_service/services/gunicorn_config.py
@@ -50,4 +50,4 @@ def post_request(worker, request, environment, response):  # pylint: disable=unu
         # WARNING: As of now the parameters sig and frame aren't used by the thread worker (our current default). If
         # we change to another worker type, or if Gunicorn updates the handle_quit code to do something with them,
         # then we may need to pass something in.
-        worker.handle_quit(None, None)
+        worker.handle_abort(None, None)


dtharpeuno (approved):
Do share the way you were able to test this locally please.

kenny-skaggs:
With lots of setup 😝

I modified two of the cron job endpoints to make them simple functions that didn't run any of the rest of the code. One would return a string immediately (and didn't do anything else), the other would import time and then `time.sleep` for 5 minutes. I also modified the gunicorn.config file to use 4 workers with 4 threads (to get gunicorn to use the thread worker type and to emulate how the server has multiple workers). And lastly modified how it reboots the workers so that it would reboot with every 4th call (rather than checking for memory).

Then I used Gunicorn to run the offline (cronjob) flask app locally by running `gunicorn -c rdr_service/services/gunicorn_config.py --timeout 18000 rdr_service.offline.main:app`. I had used the offline app before for other Gunicorn issues, so that seemed easiest to continue with. Then in postman I would use one tab to make a call to the endpoint that took 5 minutes to return. And in another tab I would make one call after another.

I could see when a worker was rebooting from the command-line that was running gunicorn. Before the change, if I was able to get a request sent from postman right as a reboot cycle was starting, then I would get a ConnectionReset error. After the change a request at the same time would be delayed until a new worker was in place.

PR #2300, opened 2021-03-11
Bypass update for long read 

dtharpeuno:
-- need to bypass the updating of the state via the signals tool for long read for now


j-kanuch (approved):
Good call on skipping the state update post-export for the long read.

PR #2302, opened 2021-03-12
[PDR-243] Correct PDR participant biospec data to reflect FINALIZED

pegbertsch:
RDR participant summary uses a combination of finalized details and BiobankOrderStatus details (UNSET/AMENDED/CANCELLED) to translate a participant's biobank history into the `participant_summary.biospecimen_status` value (looks for orders that are not CANCELLED and also finalized and maps that to OrderStatus value FINALIZED, or UNSET otherwise).

PDR had been pulling the BiobankOrderStatus enum class values directly from the biobank order details and so wasn't reflecting the OrderStatus mapping.   This updates the logic to provide an analogous OrderStatus value in the data.

### DIFF
--- a/rdr_service/dao/bq_participant_summary_dao.py
+++ b/rdr_service/dao/bq_participant_summary_dao.py
@@ -878,8 +878,8 @@ class BQParticipantSummaryGenerator(BigQueryGenerator):
                 'bbo_finalized_site': row.finalized_site,
                 'bbo_finalized_site_id': row.finalized_site_id,
                 'bbo_finalized_time': row.finalized_time,
-                'bbo_finalized_status': str(OrderStatus(finalized_status)),
-                'bbo_finalized_status_id': int(OrderStatus(finalized_status)),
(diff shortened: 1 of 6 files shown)


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -854,20 +855,31 @@ def _make_sample_dict_from_row(bss=None, bos=None):
                 if stored_sample:
                     stored_count += 1
 
+            # PDR-243:  calculate an UNSET or FINALIZED OrderStatus to include with the biobank order data.  Aligns
```
pegbertsch, line 858: Note:  The "calculated" OrderStatus values are being added to both the BQBiobankOrderSchema  data (here, as `finalized_status`) and in the BQPDRBiospecimenSchema (where a previous field is being corrected to use this new `finalized_status` for its value).

kenny-skaggs (approved):

PR #2304, opened 2021-03-12
Fixing column offset for data-dictionary update

kenny-skaggs:
I missed that the main schema tab on the data-dictionary has column A hidden. So this shifts everything to start the rows at B instead.


dtharpeuno (approved):
will you have to add this as a hotfix since the updater is in stable now?

kenny-skaggs:
Nope, no hotfix needed. This all runs as part of the deploy tool on our machines. None of it is triggered in the cloud.

PR #2311, opened 2021-03-17
[DA-1953] Refactor AW1 raw tool; implement AW2 raw tool

j-kanuch:
This PR refactors the AW1 raw ingestion tool and implements the AW2 branch. Methods were renamed where applicable to share between AW1 and AW2 ingestions.  PDR support was added after loading the `GenomicSetMember` and `GenomicGCValidationMetrics` objects.

### DIFF
--- a/rdr_service/genomic/genomic_job_controller.py
+++ b/rdr_service/genomic/genomic_job_controller.py
@@ -398,7 +398,7 @@ class GenomicJobController:
             raise ValueError(f'contamination must be a number for member_id: {member.id}')
 
         # Calculate contamination_category using an ingester
-        ingester = GenomicFileIngester(_controller=self)
+        ingester = GenomicFileIngester(_controller=self, job_id=self.job_id)
         category = ingester.calculate_contamination_category(member.collectionTubeId,
                                                              raw.contamination, member)
         raw.contamination_category = category
--- a/rdr_service/genomic/genomic_job_controller.py
+++ b/rdr_service/genomic/genomic_job_controller.py
@@ -18,7 +18,8 @@ from rdr_service.config import (
     GENOME_TYPE_ARRAY,
     MissingConfigException)
 from rdr_service.dao.bq_genomics_dao import bq_genomic_job_run_update, bq_genomic_file_processed_update, \
-    bq_genomic_manifest_file_update, bq_genomic_manifest_feedback_update
+    bq_genomic_manifest_file_update, bq_genomic_manifest_feedback_update, \
+    bq_genomic_gc_validation_metrics_batch_update, bq_genomic_set_member_batch_update
 from rdr_service.genomic.genomic_data_quality_components import ReportingComponent
 from rdr_service.genomic.genomic_set_file_handler import DataError
 from rdr_service.genomic.genomic_state_handler import GenomicStateHandler
@@ -40,7 +41,8 @@ from rdr_service.dao.genomics_dao import (
     GenomicManifestFileDao, GenomicManifestFeedbackDao, GenomicIncidentDao, GenomicSetMemberDao, GenomicAW1RawDao,
     GenomicAW2RawDao)
 from rdr_service.resource.generators.genomics import genomic_job_run_update, genomic_file_processed_update, \
-    genomic_manifest_file_update, genomic_manifest_feedback_update
+    genomic_manifest_file_update, genomic_manifest_feedback_update, genomic_gc_validation_metrics_batch_update, \
+    genomic_set_member_batch_update
 from rdr_service.genomic.genomic_mappings import raw_aw1_to_genomic_set_member_fields, \
     raw_aw2_to_genomic_set_member_fields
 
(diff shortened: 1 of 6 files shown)


dtharpeuno (approved):

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -246,22 +251,24 @@ def ingest_specific_manifest(self, filename):
         except RuntimeError:
             self.job_result = GenomicSubProcessResult.ERROR
 
-    def ingest_member_ids_from_aw1_raw_table(self, member_ids):
+    def ingest_member_ids_from_awn_raw_table(self, n, member_ids):
         """
         Pulls AW1 data from genomic_aw1_raw and loads to genomic_set_member
+        :param n: 1 or 2 for AW1, AW2
```
pegbertsch, line 262: Is there ever going to be a need for more than just these two raw tables?

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -297,41 +304,116 @@ def ingest_member_ids_from_aw1_raw_table(self, member_ids):
...
-        result_msg += f'    Missing IDs: {missing}'
-        result_msg += f'    Multipls found for IDs: {multiples}'
+                        metrics_obj = self.set_validation_metrics_from_raw(r)
 
-        return result_msg
+                        metrics_obj = s.merge(metrics_obj)
+                        s.commit()
+                        metrics.append(metrics_obj.id)
+
+                    s.merge(r[0])
```
pegbertsch: General comment for future consideration:  moving forward, I think it helps make code easier to digest/understand on first read  if variable names (even the with/as aliases) are more descriptive (vs. single-char names s, r, n, etc.)

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -297,41 +304,116 @@ def ingest_member_ids_from_aw1_raw_table(self, member_ids):
...
         # Iterate through mapped fields
-
         _map = raw_aw1_to_genomic_set_member_fields
 
         for key in _map.keys():
             member.__setattr__(_map[key], getattr(raw, key))
 
         return member
 
+    def set_validation_metrics_from_raw(self, rec: tuple):
```
pegbertsch, line 353: Is this method specific to AW2?   Make method comments more descriptive as to what raw data its using?

pegbertsch, line 358: So this method is only called in the `ingest_member_ids_from_awn_raw_tables() `AW2 case but it gets AW1 raw data as an input? (Slightly confused by the method name vs. how/when it's called by the code, not sure what the parameter data looks like structurally).

Are "validation metrics" only applicable to AW2?

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -297,41 +304,116 @@ def ingest_member_ids_from_aw1_raw_table(self, member_ids):
...
+                    completed_members.append(r[0].id)
+
+            # BQ Updates
+            if n == 2:
+                # Metrics
+                bq_genomic_gc_validation_metrics_batch_update(metrics, project_id=self.bq_project_id)
+                genomic_gc_validation_metrics_batch_update(metrics)
+
+            # Members
+            bq_genomic_set_member_batch_update(metrics, project_id=self.bq_project_id)
```
pegbertsch, line 338: I'm wondering how big these batches end up being.   We may want to talk (with Robert to make sure I'm not suggesting something ill-advised?) about whether we can try to make use of the resource-rebuild task queues to farm out these requests to rebuild the data for bigquery/resource generators.   

This would specifically be for performance improvements, but it could make the local tool runs much less costly.

pegbertsch (approved):
I still have some confusion around the workflows associated with these tools but no specific issues.

I think we should discuss an additional refactor (can be done independently from this PR) to optimize how the genomics data is rebuilt for PDR (BigQuery and resource generator data), utilizing cloud tasks to offload the processing during local tool invocations?

rdr_service/genomic/genomic_job_controller.py
j-kanuch: There may be more raw tables in the future, but the AW1 and AW2 would still be the primary tables that we would source data from using these tools.

rdr_service/genomic/genomic_job_controller.py
j-kanuch: That was a mistake in the docstring. This Is only called in the case of AW2. "Validation Metrics" refers to the data in the AW2. I'll update the docstring.

rdr_service/genomic/genomic_job_controller.py
j-kanuch: Yes, that is a good practice. I'll update "s" and "r" to be "session" and "record", respectively. The variable "n" represents the "n" in AWN and is defined in the docstring. I'll update the comment to make that more clear.

rdr_service/genomic/genomic_job_controller.py
j-kanuch: Closing the loop on this, we'll want to implement the resource-rebuild task queues for this. We can implement this in a future PR.

rdr_service/genomic/genomic_job_controller.py
pegbertsch: (Nit) It made me wonder whether another parameter name may be more descriptive/extensible should the use cases for this grow.

PR #2313, opened 2021-03-18
[DA-1954] MIGRATION Indexes for email/login_phone_number on Participant Summary

dtharpeuno:
Need to add indexes on email and login_phone_number fields for faster querying/response from new CheckLogin api route


kenny-skaggs (approved):

pegbertsch (approved):
Can you flag this as a MIGRATION in the PR description?  (e.g., [DA-1954] MIGRATION Indexes for...)

Also, we may want to do some profiling in the lower environments on how long it's likely to take to add indexes to the `participant_summary` table when we're ready to deploy to prod.  This may be a potentially disruptive migration that impacts RDR API service during the deploy.

PR #2315, opened 2021-03-19
Removing unnecessary validation check

kenny-skaggs:
TL;DR: The error will only raise if something went wrong when finding the questionnaire again. If the answer has a questionId then it was already found on the questionnaire. I also fixed up some naming issues I noticed when looking through the tests.

This removes the code that raises a BadRequest when an answer was found with a questionId from a different QuestionnaireHistory than expected. In reality this would only raise an error if there was something wrong with how the questionnaire was found. Here's my reasoning:
- An error is raised if the `answer.questionId` for any of the answers doesn't exist in `q_question_ids`
- `q_question_ids` is a list of question ids retrieved from the `questionnaire_history` object that we load on line 335
- `questionnaire_history` on line 335 is loaded from the db using the questionnaire id and semantic version that we have stored on the `questionnaire_response` object being inserted
- The `questionnaire_response` object is created way down in the `from_client_json` method on line 808 (818 before the deletion).
- `questionnaire_response` gets the semantic version and id of the `questionnaire` that is loaded on line 790 (800 before the deletion).
- That `questionnaire` is also used when creating the QuestionnaireResponseAnswer objects. The `_extract_codes_and_answers` method gets all the questions from the questionnaire and stores them in a dictionary keyed by their link id
- The `_populate_codes_and_answers` method then goes through all the answers in the response, creating QuestionnaireResponseAnswer objects and setting their `questionId` by looking up the question from questionnaire by using the link-id. If the answer uses a link-id that doesn't match up with a question on the survey, then no QuestionnaireResponseAnswer is created.

So, because the questionIds for the response answers are derived from the questionnaire that the response is for. The only way this check could find any answers that have question ids that aren't in the questionnaire history is if a different questionnaire was loaded on line 335 than the questionnaire used when parsing the response.

### DIFF
--- a/tests/api_tests/test_questionnaire_response_api.py
+++ b/tests/api_tests/test_questionnaire_response_api.py
@@ -612,12 +612,12 @@ class QuestionnaireResponseApiTest(BaseTestCase):
         participant_id = self.create_participant()
         questionnaire_id = self.create_questionnaire("questionnaire1.json")
         q = QuestionnaireDao()
-        quesstionnaire = q.get(questionnaire_id)
-        make_transient(quesstionnaire)
-        quesstionnaire.status = QuestionnaireDefinitionStatus.INVALID
+        questionnaire = q.get(questionnaire_id)
+        make_transient(questionnaire)
+        questionnaire.status = QuestionnaireDefinitionStatus.INVALID
         with q.session() as session:
-            existing_obj = q.get_for_update(session, q.get_id(quesstionnaire))
-            q._do_update(session, quesstionnaire, existing_obj)
+            existing_obj = q.get_for_update(session, q.get_id(questionnaire))
+            q._do_update(session, questionnaire, existing_obj)
         q.get(questionnaire_id)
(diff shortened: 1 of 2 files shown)


tests/api_tests/test_questionnaire_response_api.py
```diff
@@ -612,12 +612,12 @@ def test_invalid_questionnaire(self):
         participant_id = self.create_participant()
         questionnaire_id = self.create_questionnaire("questionnaire1.json")
         q = QuestionnaireDao()
-        quesstionnaire = q.get(questionnaire_id)
-        make_transient(quesstionnaire)
-        quesstionnaire.status = QuestionnaireDefinitionStatus.INVALID
+        questionnaire = q.get(questionnaire_id)
```
dtharpeuno, line 615: ➕ for correct spelling!

dtharpeuno (approved):

wangy70 (approved):

PR #2317, opened 2021-03-22
Printing times for when migrations start and end

kenny-skaggs:
This changes the log line format for alembic so that the time (to the second) is printed with each log line, and also adds a log statement giving the time after all the migrations are complete. Since alembic prints out a log statement when starting a migration, this will let us look at the logs and know how long each migration took when deploying.


dtharpeuno (approved):
⚡

pegbertsch (approved):

PR #2320, opened 2021-03-23
Fix for google api networking issues

kenny-skaggs:
This addresses two issues I've noticed with using the google API:
- Sometimes we'll get a timeout when trying to download the current data-dictionary. Increasing python's socket timeout fixes that.
- If comparing the data-dictionary to the database schema takes too long (or if I take a while to respond to the command line prompt for what the changes were) then we can get a broken pipe error when trying to upload the values. That seems to be related to the google API dropping the idle connection. Using the API library to build the connection right before it's needed resolves that problem.

### DIFF
--- a/rdr_service/services/google_sheets_client.py
+++ b/rdr_service/services/google_sheets_client.py
@@ -1,5 +1,6 @@
 from googleapiclient import discovery
 from oauth2client.service_account import ServiceAccountCredentials
+import socket
 
 from rdr_service.services.gcp_utils import gcp_get_iam_service_key_info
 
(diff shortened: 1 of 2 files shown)


j-kanuch (approved):

rdr_service/services/google_sheets_client.py
```diff
@@ -44,6 +41,24 @@ def __init__(self, spreadsheet_id, service_key_id, tab_offsets=None):
             'offset_str': offset
         } for tab_name, offset in tab_offsets.items()} if tab_offsets else {}
 
+    def _build_service(self):
+        # The Google API client uses sockets, and the requests can take longer than the default timeout.
+        # The proposed solution is to increase the default timeout manually
+        # https://github.com/googleapis/google-api-python-client/issues/632
+        # The socket seems to be created when calling discover.build, so this temporarily increases the timout for
```
dtharpeuno: :nit timeout

dtharpeuno (approved):

rdr_service/services/google_sheets_client.py
kenny-skaggs: Thanks for catching that. Apparently it's not the first time I've done that 😂

PR #2323, opened 2021-03-24
[ROC-914] keep PM time in PS when cancel a PM

wangy70:
We want to keep the `physicalMeasurementsTime` , `physicalMeasurementsFinalizedTime` and `physicalMeasurementsFinalizedSiteId` in participant summary even when the PM has been cancelled.

### DIFF
--- a/rdr_service/dao/physical_measurements_dao.py
+++ b/rdr_service/dao/physical_measurements_dao.py
@@ -4,6 +4,7 @@ import logging
 from rdr_service.lib_fhir.fhirclient_1_0_6.models import observation as fhir_observation
 from rdr_service.lib_fhir.fhirclient_1_0_6.models.fhirabstractbase import FHIRValidationError
 from sqlalchemy.orm import subqueryload
+from sqlalchemy.sql import or_
 from sqlalchemy.orm.attributes import flag_modified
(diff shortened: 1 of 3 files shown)


kenny-skaggs:
Can you post parts of the discussion around this here or in the ROC ticket? I'm trying to understand what led to this change myself, but it would also be really helpful for anyone in the future that needs to look back and understand why it is that we keep these fields when the measurements have been cancelled.

wangy70:
There is no discussion yet, from the ticket description, I think we want to keep the `physicalMeasurementsTime` even after it's cancelled. I tagged Sunny in the ticket to confirm the expected behavior, waiting for the response :)

pegbertsch (commented):
I think this needs to be verified/accounted for in the PDR generator logic as well.
(See:  `_prep_physical_measurements()` in `dao/bq_participant_summary_dao.py` and `resource/generators/participant.py)`.  We probably need to enhance the unit tests to confirm consistency between the RDR data and the generated PDR data.

wangy70:
There may be some change for the requirement, let's hold this and wait for Sunny's feedback.

wangy70:
We have some misunderstanding about the requirement, After discuss with Sunny, we don't need to do any change for this one. Close this PR.

PR #2324, opened 2021-03-24
[DA-1935] Slack notifications / more incident reporting Genomics pipeline

dtharpeuno:
-- Need the ability to receive slack notifications when validation processes fail in Genomics workflows
   
   - Added a class for handling slack notification in webhooks via `SlackMessageHandler` class
       * Basic implementation for interacting with webhooks and posting messages via `requests`
   - Tied slack notifications into genomic incident reporting, can send notification with incident by passing 
    `slack=True` in kwargs for method
   - Also better tracking of missing columns in manifest ingestion by `key` comparison in validation and sending missing ones 
     in messages
   - Adding dict for rdr webhooks in config
   - Couple of data generators added, although not used in this PR, were needed
   - Various tests

Configuration change requests for production:
https://precisionmedicineinitiative.atlassian.net/browse/DA-1935?focusedCommentId=83669


rdr_service/alembic/versions/32d43b9b5f77_.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('genomic_incident', sa.Column('slack_notification', sa.SmallInteger(), nullable=False))
```
j-kanuch: We'll want to add the default value of `0` here.

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -854,23 +856,38 @@ def load_raw_awn_data_from_filepath(self, file_path):
...
+        :return:
         """
-        with self.incident_dao.session() as session:
-            return session.add(GenomicIncident(**kwargs))
+        insert_kwargs = {key: value for key, value in kwargs.items()
+                         if key in GenomicIncident.__table__.columns.keys()}
+        incident = self.incident_dao.insert(GenomicIncident(**insert_kwargs))
+
+        if kwargs.get('slack') is False:
+            return
```
j-kanuch: I don't think the default behavior should be to send an alert. We insert records to `genomic_incident` for every row that fails to ingest, so for every ingestion we would need to remember to add the kwarg for slack alerts.

dtharpeuno: 2350cec964f2226912ac6f75845502e0f1cb5ea6

j-kanuch (approved):

wangy70 (approved):

PR #2328, opened 2021-03-25
[DA-1955] More Long Read Pilot Changes

dtharpeuno:
-- Need to add more changes for criteria in concern to long-read pilot workflow
	< 1% contamination


rdr_service/genomic/genomic_job_components.py
```diff
@@ -2691,6 +2697,7 @@ def _get_long_read_participants(self, limit=None):
                 Code.value == 'WhatRaceEthnicity_Black',
                 GenomicSetMember.genomeType.in_(['aou_wgs']),
                 GenomicSetMember.genomicWorkflowState != GenomicWorkflowState.IGNORE,
+                or_(GenomicGCValidationMetrics.contamination <= 0.01, GenomicGCValidationMetrics.contamination == ""),
```
j-kanuch: We don't want to include `contamination == ""`. This means that the GC did not provide contamination data which means that we don't know if the sample is contaminated or not. We also want to include a filter for `GenomicGCValidationMetrics.ignore_flag == 0`. I should have mentioned both of these conditions earlier since we are now joining to `GenomicGCValidationMetrics`.

dtharpeuno: Sure np, Ill remove/add it.

dtharpeuno: aeb52fd98db9efa1c7da2b937eb28dc6ee4e33a3

j-kanuch (approved):

wangy70 (approved):

PR #2330, opened 2021-03-26
MIGRATION  Remove unneeded column in Genomic File Model

dtharpeuno:
-- We are using an incident report to now to mitigate tracking so no need for this missing files bool on the model


j-kanuch (approved):
Note: For future PRs, prefix the PR title with "MIGRATION" if it contains an Alembic migration, so that other devs know in case they have migrations that could potentially cause conflicts.

wangy70 (approved):

PR #2332, opened 2021-03-26
Add new Broad GenomicGCValidationMetrics fields to PDR

j-kanuch:
This PR add 4 new fields to the PDR `GenomicGCValidationMetrics` schemas.


pegbertsch (commented):
I think this is safe to merge.   Appears the generators should pick up these fields automatically if they match the MySQL column names but we'll want to perform genomics BQ migration and data rebuild test in a lower environment to confirm the values are being picked up.  Then we can perform the migration in production BigQuery and do a targeted rebuild once the RDR release makes production.

pegbertsch (approved):

PR #2334, opened 2021-03-26
Unit test logging updates

kenny-skaggs:
This doesn't need to be merged for the release.

This makes two updates for unit test logging. The first is to add the timestamp to logs that are printed during unit tests. It could be helpful in general, but came to mind when I thought of the second change.

The other change is to add the LoggingDatabaseActivity context manager. Any code run within a `with LoggingDatabaseActivity()` block will print the SQL that is run against the database. There could be other applications for it, but I plan on using it when working on tests to verify the raw SQL generated by the ORM and to see how many queries are being generated from lazy loading.

### DIFF
--- a/tests/helpers/diagnostics.py
+++ b/tests/helpers/diagnostics.py
@@ -1,0 +1,11 @@
+import logging
+
+
+class LoggingDatabaseActivity:
+    def __enter__(self):
+        self._sqlalchemy_logger = logging.getLogger('sqlalchemy.engine')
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):
Nice. Fun fact you can also access the `raw` generated sql when debugging/working by accessing the `__str__` method on the query obj.

ex: `results.__str__()`

pegbertsch (approved):

PR #2342, opened 2021-03-31
[ROC-920] update workspace active project number algorithm

wangy70:
1. change active workspace number algorithm: 
Add the following criteria: `active` and `has verified institution`.

2. fix bug for multiple owners case:
Workspace allows to have multiple owns. When calculate workspace verified institution, if a workspace has creator info, then use creator info; if no creator info, we need to go through each owners, any of them has verified institution, the workspace will be marked as `has verified institution`.

### DIFF
--- a/rdr_service/dao/workbench_dao.py
+++ b/rdr_service/dao/workbench_dao.py
@@ -3,7 +3,7 @@ import json
 from werkzeug.exceptions import BadRequest
 from dateutil.parser import parse
 import pytz
-from sqlalchemy import desc, or_, func, distinct
+from sqlalchemy import desc, or_, and_, func, distinct
 from sqlalchemy.orm import subqueryload, joinedload
 from rdr_service.dao.base_dao import UpdatableDao
 from rdr_service import clock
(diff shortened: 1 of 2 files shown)


rdr_service/dao/workbench_dao.py
```diff
@@ -452,8 +452,24 @@ def get_workspaces_with_user_detail(self, **kwargs):
...
+                session.query(distinct(WorkbenchWorkspaceUser.workspaceId))
+                    .filter(WorkbenchWorkspaceUser.isCreator == 1)
+                    .subquery()
+            )
             count_query = (session.query(distinct(WorkbenchWorkspaceApproved.workspaceSourceId))
-                           .filter(WorkbenchWorkspaceApproved.excludeFromPublicDirectory == 0,
+                           .join(WorkbenchWorkspaceUser, WorkbenchResearcher, WorkbenchInstitutionalAffiliations)
+                           .filter(WorkbenchWorkspaceUser.workspaceId == WorkbenchWorkspaceApproved.id,
+                                   WorkbenchWorkspaceUser.researcherId == WorkbenchResearcher.id,
+                                   WorkbenchInstitutionalAffiliations.researcherId == WorkbenchResearcher.id,
```
kenny-skaggs: Because of the set of foreign keys between `WorkbenchWorkspaceUser`, `WorkbenchWorkspaceApproved`, `WorkbenchResearcher`, and `WorkbenchInstitutionalAffiliations`, SqlAlchemy is already aware of how to join them all and is generating them in inner joins because of line 461. So the filters for `WorkbenchWorkspaceUser.workspaceId == WorkbenchWorkspaceApproved.id`, `WorkbenchWorkspaceUser.researcherId == WorkbenchResearcher.id` and `WorkbenchInstitutionalAffiliations.researcherId == WorkbenchResearcher.id` can be removed to help clarify the intent of the query.

rdr_service/dao/workbench_dao.py
```diff
@@ -452,8 +452,24 @@ def get_workspaces_with_user_detail(self, **kwargs):
...
+                                   WorkbenchWorkspaceApproved.excludeFromPublicDirectory == 0,
+                                   WorkbenchWorkspaceApproved.status == 1,
+                                   WorkbenchInstitutionalAffiliations.isVerified == 1,
+                                   or_(WorkbenchWorkspaceUser.isCreator == 1,
+                                       and_(
+                                           WorkbenchWorkspaceUser.role == 3,
```
kenny-skaggs: Can you use `int(WorkbenchWorkspaceUserRole.OWNER)` here instead of `3` to make it clearer what's being filtered?

rdr_service/dao/workbench_dao.py
wangy70: Wow, cool, I never know that. Done!

rdr_service/dao/workbench_dao.py
wangy70: Good point, done!

kenny-skaggs (approved):

dtharpeuno (approved):

PR #2347, opened 2021-03-31
Upgrading google-auth so it's compatible with the new version of rsa

kenny-skaggs:
When the rsa library was updated the pip install started throwing a warning that google auth was incompatible (google-auth 1.15.0 depends on rsa's version being >=3.1.4 and <4.1). CircleCI and Google builds were continuing to work, but ReadTheDocs builds were failing (https://readthedocs.org/projects/all-of-us-raw-data-repository/builds/13396770/).

This PR updates google-auth to the latest version (1.28.0, https://pypi.org/project/google-auth). And it adds a step in the Circle-CI process to upgrade pip. Using the newer version of pip will error out rather than warn about the incompatibility and install anyway (https://app.circleci.com/pipelines/github/all-of-us/raw-data-repository/2840/workflows/b062d515-06ef-4987-8c1b-dc2dded96d3d/jobs/10573).

It also installs the requirements file last so that Circle-CI fails on the install step if installing the requirements fails.

### DIFF
--- a/ci/setup_python_env.sh
+++ b/ci/setup_python_env.sh
@@ -7,5 +7,5 @@ source venv/bin/activate
 export PYTHONPATH=`pwd`
 echo "PYTHONPATH=${PYTHONPATH}"
 pip install --upgrade pip
-pip install -r requirements.txt
 pip install safety
+pip install -r requirements.txt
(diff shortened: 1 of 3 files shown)


ci/setup_python_env.sh
```diff
@@ -6,5 +6,6 @@ python3 -m venv venv
 source venv/bin/activate
 export PYTHONPATH=`pwd`
 echo "PYTHONPATH=${PYTHONPATH}"
-pip install -q -r requirements.txt
-pip install safety
\ No newline at end of file
+pip install --upgrade pip
```
dtharpeuno, line 9: 👍

j-kanuch (approved):

pegbertsch (approved):

PR #2349, opened 2021-04-01
Move genomic enums

dtharpeuno:
Moving all genomics defined enums to thier own file


rdr_service/api/cloud_tasks_api.py
```diff
@@ -21,7 +21,7 @@
 from rdr_service.model.genomics import GenomicSetMember, GenomicGCValidationMetrics
 from rdr_service.offline import genomic_pipeline
 from rdr_service.offline.sync_consent_files import cloudstorage_copy_objects_task
-from rdr_service.participant_enums import GenomicJob, GenomicManifestTypes
+from rdr_service.genomic_enums import GenomicJob, GenomicManifestTypes
```
kenny-skaggs: We don't really have a defined organization for import statements, but this is something I've been following:
> any third-party or built-in imports (like datetime, or sqlalchemy) in alphabetical order
> single empty line
> any imports from RDR modules in alphabetical order

What do you think of that? Having a defined organization of importing could help when looking through a big list of them. The enum import would move up in the list to help move toward the method I've kind of been using.

kenny-skaggs (commented):
I think this is a really good idea. There are alot of genomic enums and it makes alot of sense for them to have their own module.

j-kanuch (approved):
Thanks for doing this!

PR #2350, opened 2021-04-01
[DA-1928] MIGRATION - Removing tmp_questionnaire_response table in CDM database

kenny-skaggs:
These changes are the first step in the work for flagging duplicate questionnaires. This is to get the duplicate flag in the RDR.

The CDM database is deleted with each run of the curation ETL. So we've needed to keep rebuilding the list of responses that were duplicates from PTSC. This PR creates an `is_duplicate` flag on the QuestionnaireResponse table and transitions all references of the tmp_questionnaire_response table over to using that flag.

A follow up PR will update the PDR code. Several changes were needed there, so I didn't want to make this PR too big with them.

This is changing how the code does what it does (where it reads that a response is a duplicate), rather than changing behavior. So (other than updating some setup for the unittests) this relies on the tests already covering the code.


kenny-skaggs:
The migration for this also includes a hash column, that'll be populated in a later PR and used to detect duplicates.

rdr_service/model/questionnaire_response.py
```diff
@@ -42,6 +42,14 @@ class QuestionnaireResponse(Base):
     """The actual time the participant completed the questionnaire"""
     language = Column("language", String(2), nullable=True)
     """Language that the response was completed in"""
+
+    # Can be used to indicate equality between sets of answers
+    answerHash = Column('answer_hash', String(32), nullable=True)
+    """@rdr_dictionary_internal_column"""
+
+    isDuplicate = Column("is_duplicate", Boolean, server_default=expression.false())
```
dtharpeuno, line 50: 👍

rdr_service/model/questionnaire_response.py
```diff
@@ -42,6 +42,14 @@ class QuestionnaireResponse(Base):
...
+    """Indicates that the response is a duplicate of another that was previously received and should be ignored"""
```
dtharpeuno, line 51: Should we add these for every addition so the data dict will pick up the declaration/explanation?

dtharpeuno (approved):

rdr_service/model/questionnaire_response.py
kenny-skaggs: Yeah! Ideally we'll incrementally get all the fields documented, but any new fields that are added should have documentation. That'll help us and other teams understand what it is and how to use it.

j-kanuch (approved):

PR #2351, opened 2021-04-01
MIGRATION - Fix genomic_set_member_history table field lengths

j-kanuch:
The field lengths for `genomic_set_member` were modified in a previous PR. The `genomic_set_member_history` field lengths were not updated. AW1 ingestions are failing with "data too long for column..." errors due to the history table inserts failing. This PR makes the same modifications as the earlier migration, but to the `genomic_set_member_history` table.


kenny-skaggs (approved):

dtharpeuno (approved):

dtharpeuno:
Do we need to send this out as a hotfix?

j-kanuch:
> Do we need to send this out as a hotfix?

@dtharpeuno Ideally we would, but it most-likely won't be available on prod until after the weekend. Because these are all ALTER statements, we may be able to avoid a hotfix. I will bring it up in our morning meeting to discuss with the team.

PR #2353, opened 2021-04-02
[DA-1962] File path updates for genomic file processed

dtharpeuno:
**Resolves** *(link to ticket)*
https://precisionmedicineinitiative.atlassian.net/browse/DA-1962

**Description of changes/additions**
Updates to the file path processing for genomic file processed  

**Tests** *(below selection and/or description)*

Tested locally and edited unit tests. Also deployed to sandbox to test Update statement in migration


rdr_service/genomic/genomic_job_components.py
```diff
@@ -225,6 +224,7 @@ def generate_file_queue_and_do_ingestion(self):
                     genomic_file_processed_update(file_ingested.id)
 
                 except IndexError:
+                    self.file_queue.popleft()
                     logging.info('No files left in file queue.')
```
kenny-skaggs: Why add the `popleft` call in the except-block?

I'm a little confused about this code in general. I'm guessing the `except IndexError` is meant to catch if the `file_queue[0]` fails. But how would it fail when the while condition is already checking if the file_queue has items?

dtharpeuno: There was an error in one of the method calls in the while loop which was just reiterating the loop and never removing the problem from the queue. So added that to mitigate when trying to debug the issue. But fixed the problem downstream.
a7aaae48aa5adc9c8511078573ea6fff86a2658e

kenny-skaggs (approved):

kenny-skaggs:
The migration file will need to be updated to have `8ea65655941d` as the down revision (from the PR I just merged https://github.com/all-of-us/raw-data-repository/pull/2355)

dtharpeuno:
> The migration file will need to be updated to have `8ea65655941d` as the down revision (from the PR I just merged #2355)

Thanks, Ill take care of it.

wangy70 (approved):

PR #2354, opened 2021-04-05
bug_fix_research_directory_api

wangy70:
**Resolves** *(link to ticket)*
Bug fix for Research Directory API

**Description of changes/additions**
This is a bug caused by last fix, for those workspaces with no creator, we need to make sure the owner has verified institution.

**Tests** *(below selection and/or description)*
Update test case `test_workspace_audit_and_research_directory_api.py` to cover this case.

### DIFF
--- a/rdr_service/dao/workbench_dao.py
+++ b/rdr_service/dao/workbench_dao.py
@@ -539,7 +539,8 @@ class WorkbenchWorkspaceDao(UpdatableDao):
                 # if has creator, calculate by creator; if no creator, calculate by owners
                 if creator_user_id == researcher.userSourceId and researcher_has_verified_institution:
                     workspace_has_verified_institution = True
-                elif creator_user_id is None and researcher.userSourceId in owner_user_ids:
+                elif creator_user_id is None and researcher.userSourceId in owner_user_ids and \
+                    researcher_has_verified_institution:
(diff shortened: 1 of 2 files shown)


rdr_service/dao/workbench_dao.py
```diff
@@ -539,7 +539,8 @@ def get_workspaces_with_user_detail(self, **kwargs):
                 # if has creator, calculate by creator; if no creator, calculate by owners
                 if creator_user_id == researcher.userSourceId and researcher_has_verified_institution:
                     workspace_has_verified_institution = True
-                elif creator_user_id is None and researcher.userSourceId in owner_user_ids:
+                elif creator_user_id is None and researcher.userSourceId in owner_user_ids and \
+                    researcher_has_verified_institution:
```
dtharpeuno, line 543: Can we just remove the var init for `researcher_has_verified_institution` on line 516, since we aren't using it anywhere downstream, and just use `affiliation.isVerified` on this test in the condition.

dtharpeuno: Nevermind I see the reason for the var.

dtharpeuno (approved):

kenny-skaggs (approved):

PR #2358, opened 2021-04-05
[DA-1928] Cron job flagging responses as duplicates

kenny-skaggs:
## Partially Resolves *[DA-1928](https://precisionmedicineinitiative.atlassian.net/browse/DA-1928)*
One more follow up PR will be made to create a quick tool for backfilling the answer hash values for the QuestionnaireResponses we already have in the database.

This adds a new class to be used for flagging questionnaire responses that are replays of each other. It follows the previous algorithm used: no responses are marked unless there are 10 or more duplicates, the latest response is left unmarked (`isDuplicate = 0`) so that we'll use the last one received. This is also configured to run as a cron job, so won't update any questionnaire responses that are already known as duplicates.

This also saves the MD5 digest of the answers json for comparison.

## Tests
- [x] unit tests for the newly added class

### DIFF
--- a/rdr_service/services/response_duplication_detector.py
+++ b/rdr_service/services/response_duplication_detector.py
@@ -13,7 +13,8 @@ class ResponseDuplicationDetector:
         Used to check the database for any new questionnaire response duplicates.
 
         :param duplication_threshold: The number of matching responses needed in a group before any of them will be
-            considered duplicates. Defaults to 10.
+            considered duplicates. Any responses that have already been marked as duplicates count toward this total.
+            Defaults to 10.
         """
         self.duplication_threshold = duplication_threshold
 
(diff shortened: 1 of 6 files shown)


rdr_service/services/response_duplication_detector.py
```diff
@@ -0,0 +1,76 @@
...
+                                  older_response: Type[QuestionnaireResponse]):
+        return and_(
+            newer_response.created > older_response.created,
+            newer_response.externalId == older_response.externalId,
+            newer_response.answerHash == older_response.answerHash,
+            newer_response.participantId == older_response.participantId
+        )
+
+    def flag_duplicate_responses(self):
+        with get_database().session() as session:
```
dtharpeuno, line 32: 🔥

wangy70 (approved):

j-kanuch (approved):

PR #2359, opened 2021-04-05
[PDR-285] Update Workbench Researcher views to use user_source_id as primary key.

robabram:
Change primary key in workbench researcher views.


wangy70 (approved):

dtharpeuno (approved):
👍

PR #2364, opened 2021-04-06
Recording database migration results in Jira ticket

kenny-skaggs:
## Resolves *no ticket*
Records the migration results, including times for each migration, to the Jira ticket.

This stores the output of the Alembic command on the class, so then the deploy class can add it to the Jira ticket.

## Tests
- [ ] no unit tests

I've tested the deploy code against sandbox while debugging to see the output string.


wangy70 (approved):

dtharpeuno (approved):
Nice

PR #2367, opened 2021-04-07
[PDR-254] Make test_participant status visible/filterable in BQ module tables

pegbertsch:
To facilitate filtering of test participant questionnaire response data in BigQuery, a `test_participant` column will be added to the set of common fields that exist in each BigQuery table that is derived from the `BQModuleSchema` class

- Update generator logic that determines the test participant status in `pdr_participant` data

- Update BQPDRQuestionnaireResponseGenerator query that looks for participant module responses for the `pdr_mod_* `data, to also return a calculated `test_participant `field

- Corrects the `TEST_HPO_ID` constant declaration.  This integer constant did not appear to be referenced anywhere except test code (`TEST_HPO_NAME` string constant was used in code, including being passed to the `HPODao.get_by_name()` method to look up the HPO integer ID).   Unit tests all passed

## Tests
Added new generator data validation to `test_participant_api.test_switch_to_test_account` unit test


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -284,6 +284,9 @@ def _prep_participant(self, p_id, ro_session):
         cohort_2_pilot_flag = \
             ParticipantCohortPilotFlag.COHORT_2_PILOT if cohort_2_pilot else ParticipantCohortPilotFlag.UNSET
 
+        # If RDR paired the pid to hpo TEST or flagged as either ghost or test participant, treat as test participant
+        test_participant = p.isGhostId == 1 or p.isTestParticipant == 1 or p.hpoId == TEST_HPO_ID
+
```
kenny-skaggs, line 293: Should this check the login_phone_number and email (if we have them)?

rdr_service/participant_enums.py
```diff
@@ -35,7 +35,7 @@
 TEST_EMAIL_PATTERN = "%@example.com"
 # The name of the 'test' HPO that test participants are normally affiliated with.
 TEST_HPO_NAME = "TEST"
-TEST_HPO_ID = 19
+TEST_HPO_ID = 21
```
kenny-skaggs, line 38: It might be better to load the HPO by the name, just in case this gets out of date again.

rdr_service/dao/bq_participant_summary_dao.py
pegbertsch: If this doesn't identify the pid as a test participant, we will still end up calling `_calculate_test_participant()` in the generator code later (see: line 171), which can still flag it based on email or phone.   That hasn't been disabled, but there is a TODO to try and turn it off eventually.

rdr_service/participant_enums.py
pegbertsch: I wish I didn't have to keep adding more joins and/or queries (through the HPODao) to the generator logic vs. filtering on the `participant` table data that's already part of the query.  Is there a legitimate risk this will change again?

rdr_service/dao/bq_participant_summary_dao.py
kenny-skaggs: 👍

rdr_service/participant_enums.py
kenny-skaggs: I'm not sure how it changed the first time. But if it does change for some reason we're not likely going to think to update this code. And a random HPO's participants would be flagged as test participants.

pegbertsch: I changed things to use results from HPO table and filter on TEST_HPO_NAME (as is done in other RDR API code)

kenny-skaggs (approved):

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -48,12 +48,21 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
+        # PDR-254:  Updating the module response data to include a test_participant flag as a common field
+        # Intended to make filtering module response data by test_participant status more efficient in BigQuery
         _participant_module_responses_sql = """
             select qr.questionnaire_id, qr.questionnaire_response_id, qr.created, qr.authored, qr.language,
-                   qr.participant_id, qh2.external_id, qr.status
+                   qr.participant_id, qh2.external_id, qr.status,
+                   CASE
+                       WHEN p.is_test_participant = 1  or p.is_ghost_id = 1 or h.name = :test_hpo THEN 1
+                       ELSE 0
+                   END as test_participant
```
pegbertsch, line 60: Note:  I believe RDR `participant` table is now a highly reliable source of truth in using the ghost / test participant / HPO pairing details to determine PDR test participant status.  But, there may still be a risk of not identifying module responses as belonging to newly created test participants if the participant was created with a valid phone number but an `@example.com` email address.   Is it worth further complicating this query to check for those cases?

rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -1267,28 +1273,31 @@ def datetime_to_date(val):
...
         :param summary: summary data
         :return: dict
         """
-        test_participant = summary['is_ghost_id']
-
-        # Check for @example.com in email address
-        if not test_participant:
-            if summary.get('test_participant') == 1:
-                test_participant = 1
-            # Check to see if the participant is in the Test HPO.
```
robabram, line 1282: Why wouldn't we keep this block in?  These are checks to see if the participant has already been marked Test Participant in RDR.

pegbertsch: The checks for ghost, test participant, or HPO pairing were moved into `_prep_participant()` and I felt it was redundant/more confusing to have them in both places.   So I made this specific to the additional email/phone checks.

This method no longer gets called automatically.   It only gets called if the summary data already created by `_prep_participant()` has not already flagged the participant as test.   Personally, I'm hoping we can get rid of this, and we may be able to if RDR is updated to always do the pairing to TEST based on test email or phone

robabram (approved):

PR #2372, opened 2021-04-09
[PDR fix] Correct handling of EHR consent expiration in PDR data

pegbertsch:
PDR generators handle partial module responses by carrying forward previously submitted answers to question codes that were not sent in a subsequent payload.   This caused a problem with EHR consents because we carried forward the hidden `EHRConsentPII_ConsentExpired` question code and `EHRConsentPII_ConsentExpired_Yes` answer to a subsequent renewed EHR consent response.   This may have impacts on other PDR data elements, e.g., when `enrollment_status` is calculated.

In case we run into more cases like this, the fix will maintain a list of codes that should be "unlayered" for a given module. We will not carry prior answers to those forward to the next module response payload

Also did some minor refactoring of unit test helper routines used in validating output from the PDR data generators, and addressed what could be bug waiting to happen with building participant data if a participant is unpaired 
## Tests
- [x] unit tests
Added `test_ehr_consent_expired_and_renewed` test case to  verify the new logic around unlayered question codes

### DIFF
--- a/tests/dao_tests/test_bigquery_sync_dao.py
+++ b/tests/dao_tests/test_bigquery_sync_dao.py
@@ -218,16 +218,22 @@ class BigQuerySyncDaoTest(BaseTestCase, QuestionnaireTestMixin):
         return biobank_order
 
     @staticmethod
-    def get_modules_by_name(module_name=None, module_list=None):
+    def get_generated_items(item_list, item_key=None, item_value=None, sort_key=None):
         """
-            Extracts module entries from the participant resource generator data modules list
-            Returns a filtered list of entries that match the module name, sorted by authored date
+            Extracts requested items from a provided list of dicts (e.g., ps_json['consents'])
+            Returns a filtered list of dict entries that match the item name, sorted if requested
+            Example:
+                get_items(ps_json['modules'], item_key='mod_module', item_value='OverallHealth',
+                                  sort_key='mod_authored')
         """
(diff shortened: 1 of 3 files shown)


tests/dao_tests/test_bigquery_sync_dao.py
```diff
@@ -429,3 +436,40 @@ def test_no_on_ehr_overrides_yes_on_dv(self):
...
+        self.assertEqual(ehr_consents[1].get('consent_value', None), CONSENT_PERMISSION_NO_CODE)
+        self.assertEqual(ehr_consents[1].get('consent_expired', None), EHR_CONSENT_EXPIRED_YES)
+
+        # Verify the last EHR consent renewal;  'consent_expired' value should not be carried forward from last consent
+        self.assertEqual(ehr_consents[2].get('consent_module_authored', None), self.TIME_3)
+        self.assertEqual(ehr_consents[2].get('consent_value', None), CONSENT_PERMISSION_YES_CODE)
+        # This field should be None for consent payloads that don't contain the expiration hidden question code
+        self.assertIsNone(ehr_consents[2].get('consent_expired', ''))
+
+        self.assertIsNotNone(ps_json)
```
pegbertsch: This was leftover from adding a line where I could set a breakpoint.  Will remove

kenny-skaggs (approved):

robabram (approved):

PR #2375, opened 2021-04-12
General refactoring in genomic pipeline

dtharpeuno:
## Resolves 
N/A

Just some general/refactoring cleanup that I wanted to do for more code readability and consolidation
TODO: after this will try and get to refactoring the raw SQL queries into the ORM syntax.

## Tests
No tests added @ this time, just making sure existing tests are still passing for current workflows


rdr_service/genomic/genomic_queries.py
```diff
@@ -1,481 +0,0 @@
...
...810288873,812524724,814899961,819265696,819950283,820073586,823758344,830302099,831226243,836837874,837721794,838889749,841403776,844613792,848159433,851224539,851621512,852294303,853072726,853712780,855034665,857694544,858161509,859351011,859568629,865187888,865784343,866140322,868852263,869844960,877677445,879078463,880167360,880568948,884161802,885461547,893773868,894252703,895672943,905449614,907123717,907405741,910111063,913827794,914200698,914732151,918627535,918911212,920156311,921352328,922432405,924531063,924542857,926555538,926818829,926876459,927094113,933731079,945880676,950456601,951122259,953465025,955165843,955764343,962466291,966404836,973375903,974122861,979468523,979907745,984560898,985298078,985584918,985624484,987888386,989747214,990592038,990715447,993699312,994927821,995295154,995686976,999527800,999988289,116826403,148469844,186903103,228059843,298721343,311337192,329020340,336078084,338831211,367705012,382605411,387230641,413496554,447932598,468961337,481461052,489283953,533267737,568390849,645025662,672914529,837721794,838889749,865187888,866140322,924531063,926818829,955764343,270776467,295092346,303973848,322963871,593191281,696155559,705391345,711822745)
```
kenny-skaggs, line 65: I'm not sure what these participants are, but they don't get transferred to the new SQL query class. Is that intentional?

dtharpeuno: There was a massive AW0 generation that happened a few months ago and these were participants that were flagged from research into the cohort discrepancies as not being, at the time, correctly defined as to what cohort group they were actually supposed to be in. So since this manifest generation/query was run, most of these participants have been correctly assigned to their cohort group, so the list is pretty much out of date at this time.

kenny-skaggs: Should the list be updated rather than removed? Are there any that should continue to be filtered while we still track down what cohort they should be a part of?

dtharpeuno: Don't think so. The large manifest generation was "suppose" to a one-time run that the Biobank asked for and I think it's better to remove it from the code for readability and let it be resurrected if need be, which is a very slim chance, from git.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -3344,82 +3007,51 @@ def get_def(self, manifest_type):
...
+            }
+        }
+        return self.ManifestDef(
+            job_run_field=def_config[manifest_type]['job_run_field'],
+            source_data=self._get_source_data_query(manifest_type),
+            destination_bucket=f'{self.bucket_name}',
+            output_filename=def_config[manifest_type]['output_filename'],
+            columns=self.manifest_columns_config[manifest_type],
+            signal=def_config[manifest_type]['signal'],
+        )
```
j-kanuch, line 3054: Nice

j-kanuch (approved):

kenny-skaggs (approved):

PR #2377, opened 2021-04-12
Fix AW2 contamination non-float bug

j-kanuch:
This PR fixes a bug in the AW2 ingestion workflow when `contamination` is not supplied. The RDR must allow `contamination` to be a non-float in cases where `processing_status` is not `pass`. Previously, the code was returning a `GenomicSubprocessResult` object, but this is not correct. Now the code is returning the `row` object when `contamination` is a non-float and the `processing_status` is not `pass`. When the `processing_status` is `pass`, and the `contamination` value is a non-float, the code writes a `genomic_incident` record.

## Tests
- [] test_ingest_specific_aw2_file
- [] test_aw2_genomic_incident_inserted
- [] test_aw2f_manifest_generation_e2e

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -677,8 +677,19 @@ class GenomicFileIngester:
                 row['contamination'] = 0
 
         except ValueError:
-            logging.error(f'contamination must be a number for sample_id: {row["sampleid"]}')
-            return GenomicSubProcessResult.ERROR
+            if row['processingstatus'].lower() != 'pass':
+                return row
+
+            _message = f'contamination must be a number for sample_id: {row["sampleid"]}'
+
+            self.controller.create_incident(source_job_run_id=self.job_run_id,
+                                            source_file_processed_id=self.file_obj.id,
+                                            code=GenomicIncidentCode.DATA_VALIDATION_FAILED.name,
+                                            message=_message,
+                                            biobank_id=member.biobankId,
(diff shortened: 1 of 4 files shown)


dtharpeuno (approved):

rdr_service/genomic/genomic_job_components.py
```diff
@@ -677,8 +677,19 @@ def prep_aw2_row_attributes(self, row: dict, member: GenomicSetMember):
...
+            _message = f'contamination must be a number for sample_id: {row["sampleid"]}'
+
+            self.controller.create_incident(source_job_run_id=self.job_run_id,
+                                            source_file_processed_id=self.file_obj.id,
+                                            code=GenomicIncidentCode.DATA_VALIDATION_FAILED.name,
+                                            message=_message,
+                                            biobank_id=member.biobankId,
+                                            sample_id=row['sampleid'],
+                                            )
+
```
kenny-skaggs, line 692: If the ValueError occurs and the row's `processingstatus` is pass, then the code would continue to try to calculate the contamination category (below) and would raise another ValueError. I may be missing something, but I think this would cause a crash before it's done processing the file.

j-kanuch: You are right. We want to return here after `create_incident()` and then skip the row in the main ingestion loop. Since the incident is written to the incident table, there is a record of the erroneous row.

kenny-skaggs (approved):

PR #2378, opened 2021-04-13
[DA-1965] Sanitize rows in manifest

dtharpeuno:
## Resolves
https://precisionmedicineinitiative.atlassian.net/browse/DA-1965

There was some 500 errors firing for the last few weeks on a manifest that was uploaded from GC that was missing header. The validation methods we have in place were passing since all the reqs that the logic is looking for were satisfied. This PR cleans the bad data before it reaches the validation 

## Tests
- [x] unit tests


wangy70 (approved):

rdr_service/genomic/genomic_job_components.py
```diff
@@ -695,14 +711,9 @@ def update_member_for_aw2(self, member: GenomicSetMember):
...
 
+        state = GenomicWorkflowState.AW2
         member.aw2FileProcessedId = self.file_obj.id
-
-        # Only update the state if it was AW1
-        if member.genomicWorkflowState == GenomicWorkflowState.AW1:
-            member.genomicWorkflowState = GenomicWorkflowState.AW2
-            member.genomicWorkflowStateModifiedTime = clock.CLOCK.now()
-
-        self.member_dao.update(member)
```
j-kanuch: The logic here prevents state from being updated on re-ingestion if we need to process the file again for whatever reason. If member.genomicWorkflowState is downstream of AW2, for instance GEM_READY, and we need to re-ingest the file, without this logic the state would be reverted back to AW2.

j-kanuch (approved):

PR #2381, opened 2021-04-14
Refactoring gcp_db_daemon

kenny-skaggs:
## Resolves *no ticket*
In order to make connections to new databases for testing out Mysql 8 and Postgres we'll want to add new ports for them in the proxy connection setup. To make that easier I refactored the gcp_db_daemon code so that it will be simpler to set up new ports.

The methods `print_instances` and `get_instances` (renamed to `get_instances_arg_str`) both had information about the port numbers being used for each environment (for the primary and replica databases). This moves the port number information into a mapping dictionary (`environment_proxy_port_map`). The SQL database instance name is also set up in the dictionary. `build_gcp_instance_connection_name` is also added to help with different database names.

The `print_instances` method previously printed out all the connection information, even if those environments were activated. This updates it to only print out the ports that were activated.

## Tests
- [ ] unit tests

No unit tests were created and there don't seem to be any. But I've tested locally by comparing instance connection name lists sent to cloud_sql_proxy, making sure they're identical before and after these changes.

### DIFF
--- a/rdr_service/services/gcp_db_daemon.py
+++ b/rdr_service/services/gcp_db_daemon.py
@@ -45,12 +45,27 @@ def run():
                 RdrEnvironment.PTSC_3_TEST:     ProxyPortData(primary=9980, replica=9985)
             }
 
+            self.environments_to_activate = [RdrEnvironment.PROD, RdrEnvironment.STABLE, RdrEnvironment.STAGING]
+
+            if self._args.enable_sandbox:
+                self.environments_to_activate.append(RdrEnvironment.SANDBOX)
+            if self._args.enable_test:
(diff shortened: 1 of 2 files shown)


rdr_service/services/gcp_db_daemon.py
```diff
@@ -1,29 +1,34 @@
...
+    name: str
+    port: int
+
+
+@dataclass
+class ProxyData:
+    primary: DatabaseProxy
+    replica: DatabaseProxy = None  # Not every environment has a replica db
+
+
```
dtharpeuno, line 31: 🔥

dtharpeuno (approved):
Awesome!!!!!

wangy70 (approved):

PR #2382, opened 2021-04-15
[DA-1968] ParticpantSummary API updates for parameters

dtharpeuno:
## Resolves *[ticket DA-1968]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-1968

Changes for participant summary API:
1. Need to filter out bad params/filters that do not exist on the summary obj and do not return and raise an exception if only bad params are passed
2. Add a pair constraint in the request if either the lastName or dateOfBirth is in the request, for checking that both exist in params

## Tests
- [x] unit tests

Screen Shot 2021-04-14 at 4 34 41 PM


rdr_service/api/participant_summary_api.py
```diff
@@ -66,6 +70,33 @@ def _make_bundle(self, results, id_field, participant_id):
...
 
+    @staticmethod
+    def _check_constraints():
+        message = None
+        invalid = False
+
+        pairs = {
+            'lastName': {
+                'fields': ['lastName', 'dateOfBirth'],
+                'message': ''
```
kenny-skaggs: Does the message field here do anything?

rdr_service/api/participant_summary_api.py
```diff
@@ -53,8 +53,12 @@ def get(self, p_id=None):
                     raise Forbidden
             return super(ParticipantSummaryApi, self)._query("participantId")
 
-    def _make_query(self):
-        query = super(ParticipantSummaryApi, self)._make_query()
+    def _make_query(self, check_invalid=True):
+        pair_constraint, message = self._check_constraints()
+        if pair_constraint:
```
kenny-skaggs: nit: `pair_constraint` sounds like an object or something that can give information about a constraint (like what's required or what the constraint is for) rather than an indicator that the constraint has failed. What about `constraint_failed`?

tests/api_tests/test_participant_summary_api.py
```diff
@@ -465,6 +465,77 @@ def test_check_login(self):
...
+        )
+        self.assertEqual(response_bad_filter.status_code, 400)
+        self.assertEqual(response_bad_filter.json['message'], 'No valid fields were provided')
+
+        response_no_filter = self.send_get("ParticipantSummary")
+        self.assertEqual(len(response_no_filter['entry']), num_summary)
+
+    def test_constraints_dob_and_lastname(self):
+        num_summary = 3
+        date_str = datetime.date(1978, 10, 9)
```
kenny-skaggs: nit: I might be using the wrong convention, but I would assume a variable holds a string if it ends with `_str` like this. But `date_str` holds a Date object. Should it be changed?

tests/api_tests/test_participant_summary_api.py
```diff
@@ -465,6 +465,77 @@ def test_check_login(self):
...
+                    lastName=f"Tester_{num}",
+                    dateOfBirth=date_str,
+                )
+
+        response_only_dob = self.send_get(
+                f"ParticipantSummary?dateOfBirth={date_str}",
+                expected_status=http.client.BAD_REQUEST
+            )
+        self.assertEqual(response_only_dob.status_code, 400)
+        self.assertEqual(response_only_dob.json['message'], 'Missing lastName in request')
```
kenny-skaggs: Can we update the error message to better specify that querying by DOB requires lastName as well? If I'm a user that isn't aware of this requirement and just try to filter by DOB but get the error message then it isn't very clear to me that it's a requirement because of the DOB (I could think it's always required). Same for lastName requiring DOB.

rdr_service/api/participant_summary_api.py
```diff
@@ -53,8 +53,12 @@ def get(self, p_id=None):
...
+            raise BadRequest(f"{message}")
```
kenny-skaggs, line 61: The ticket specifies that when HPRO searches we want to enforce that DOB and lastName be used together. I believe we still want to allow PTSC and the awardees to search by lastName or DOB without requiring both.

dtharpeuno: Sure, was going on the ref to messages that were outlined in the doc and still edited it. Will change.

dtharpeuno: Sure will add it now.

tests/api_tests/test_participant_summary_api.py
dtharpeuno: Was using the verbiage outlined in the doc as ref and still edited it. But yea Ill edit the message.

rdr_service/api/participant_summary_api.py
dtharpeuno: Good catch.

tests/api_tests/test_participant_summary_api.py
```diff
@@ -2399,8 +2493,6 @@ def testQuery_manyParticipants(self):
             self.assertResponses("ParticipantSummary?_count=2&_sort=firstName", [[ps_1, ps_3], [ps_2]])
             self.assertResponses("ParticipantSummary?_count=2&_sort:asc=firstName", [[ps_1, ps_3], [ps_2]])
             self.assertResponses("ParticipantSummary?_count=2&_sort:desc=firstName", [[ps_2, ps_3], [ps_1]])
-            self.assertResponses("ParticipantSummary?_count=2&_sort=dateOfBirth", [[ps_2, ps_1], [ps_3]])
-            self.assertResponses("ParticipantSummary?_count=2&_sort:desc=dateOfBirth", [[ps_3, ps_1], [ps_2]])
```
kenny-skaggs: For the other assertions being removed: what do you think of changing the user's role so that the tests can continue to verify what they've been verifying? Are there other places in the tests that still verify that things like `gt` filtering work for DOB?

But for these two one in particular: why were the checks for ordering by DOB removed?

dtharpeuno: Sure,  ill add them back with the user's role being updated.

kenny-skaggs (approved):

wangy70 (approved):

PR #2390, opened 2021-04-20
[PDR QC] Ignore duplicate answers in same questionnaire response (PDR generators)

pegbertsch:
While investigating deltas between PDR and RDR consent counts: found several GROR `questionnaire_response` payloads in RDR that had duplication of question and answer codes in the JSON data (See: `questionnaire_response_id` 680418686).  

In the PDR generators the duplicate answers became a comma-separated list (e.g., 'CheckDNA_Yes,CheckDNA_Yes') which no longer matched any recognized consent answer codes used to determine consent status, so status ended up SUBMITTED_INVALID.

Added logic to ignore duplicate question/answer values from the same response in the generators, in case there are any additional payloads with the same issue.

## Tests
- [x] unit tests - Added `test_gror_consent_with_duplicate_answers`

Added `PDRGeneratorTestMixin` to integrate the BigQuery and resource generators into the test class and simplify verifying generator data during unit tests.   Refactored existing tests that were already doing some PDR generator output validation and moved other helper methods into the Mixin class.

### DIFF
--- a/rdr_service/dao/bq_participant_summary_dao.py
+++ b/rdr_service/dao/bq_participant_summary_dao.py
@@ -1342,6 +1342,7 @@ class BQParticipantSummaryGenerator(BigQueryGenerator):
 
         _answers_sql = """
             SELECT qr.questionnaire_id,
+                   qra.question_id,
                    qq.code_id,
                    (select c.value from code c where c.code_id = qq.code_id) as code_name,
                    COALESCE((SELECT c.value from code c where c.code_id = qra.value_code_id),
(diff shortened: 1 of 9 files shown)


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -1355,7 +1356,9 @@ def get_module_answers(ro_dao, module, p_id, qr_id=None, return_responses=False)
                                 ON qra.question_id = qq.questionnaire_question_id
                      INNER JOIN questionnaire q
                                 ON qq.questionnaire_id = q.questionnaire_id
-            WHERE qr.questionnaire_response_id = :qr_id;
+            WHERE qr.questionnaire_response_id = :qr_id
+            -- Order by question and calculated answer so any duplicates from the same response are adjacant
+            ORDER BY qra.question_id, 4
```
kenny-skaggs: Would it be better to use the column name rather than the number in the `order by` to make it clearer what is getting ordered?

pegbertsch: Yes, I should've caught that I could use that alias/calculated column name.    Thanks!

kenny-skaggs (approved):

robabram (approved):

PR #2396, opened 2021-04-21
Using Backoff to handle 502 errors when updating data-dictionary

kenny-skaggs:
## Resolves *no ticket*
The Google Sheets API seems to frequently return a 502 error when trying to download the data-dictionary values. But if retried a few times the request will succeed.

This uses Backoff to retry the API after 30 seconds. It also adds a retry-loop to the upload, just in case.

## Tests
- [x] unit tests

Unit tests to make sure the backoff is set up correctly.

### DIFF
--- a/rdr_service/services/google_sheets_client.py
+++ b/rdr_service/services/google_sheets_client.py
@@ -1,4 +1,6 @@
+import backoff
 from googleapiclient import discovery
+from googleapiclient.errors import HttpError
 from oauth2client.service_account import ServiceAccountCredentials
 import socket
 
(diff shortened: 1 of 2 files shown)


rdr_service/services/google_sheets_client.py
```diff
@@ -85,6 +87,7 @@ def _get_offset_string(self, tab_id):
         })
         return tab_offset_data['offset_str']
 
+    @backoff.on_exception(backoff.constant, HttpError, max_tries=4, jitter=None, interval=30)
```
dtharpeuno, line 90: This is cool.

dtharpeuno (approved):

j-kanuch (approved):

PR #2413, opened 2021-04-29
[ROC-974] Only enforce constraints if awardee is not present.

j-kanuch:
A recent update to the participant summary API required date of birth and last name parameters to be provided when Health Pro requested a participant look-up. This PR updates this requirement to only be enforced when the `awardee` parameter is not provided. Unit tests were updated as well to test the new behavior.

## Tests
- [] unit tests
- tests.api_tests.test_participant_summary_api.ParticipantSummaryApiTest.test_constraints_dob_and_lastname

### DIFF
--- a/rdr_service/api/participant_summary_api.py
+++ b/rdr_service/api/participant_summary_api.py
@@ -56,9 +56,10 @@ class ParticipantSummaryApi(BaseApi):
             return super(ParticipantSummaryApi, self)._query("participantId")
 
     def _make_query(self, check_invalid=True):
-        constraint_failed, message = self._check_constraints()
-        if constraint_failed:
-            raise BadRequest(f"{message}")
+        if not request.args.get("awardee"):
+            constraint_failed, message = self._check_constraints()
+            if constraint_failed:
+                raise BadRequest(f"{message}")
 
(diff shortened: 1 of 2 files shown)


rdr_service/api/participant_summary_api.py
```diff
@@ -56,9 +56,10 @@ def get(self, p_id=None):
             return super(ParticipantSummaryApi, self)._query("participantId")
 
     def _make_query(self, check_invalid=True):
-        constraint_failed, message = self._check_constraints()
-        if constraint_failed:
-            raise BadRequest(f"{message}")
+        if not request.args.get("awardee"):
+            constraint_failed, message = self._check_constraints()
+            if constraint_failed:
+                raise BadRequest(f"{message}")
```
kenny-skaggs, line 62: Approving since this works for what we need for the hotfix, but I'm worried about how this would fit in with future changes for the ParticipantSummaryApi class. If we need to add more constraints to the query, we'd likely want to add them to the `_check_constraints` method. But if the 'awardee' wasn't part of constraints we'd want to add, then this would need to be modified so we could get the new constraints checked.

It might be better to modify the `_check_constraints` logic to skip the current checks we have if the `awardee` is used.

kenny-skaggs (approved):

PR #2417, opened 2021-04-30
[DA-1986] Updating number of Gunicorn workers and threads

kenny-skaggs:
## Resolves *[DA-1986](https://precisionmedicineinitiative.atlassian.net/browse/DA-1986)*
The servers don't seem limited by CPU or any other resources. It just looks like there weren't enough Gunicorn workers/threads to accept and process the incoming requests.

This updates the Gunicorn configuration code to reduce the number of workers (to potentially avoid overhead from process swapping and benefit from having more memory available for each worker) as well as increasing the number of threads per worker.

Each instance has 2 CPU cores. So previously the code would set up each instance with 4 workers and 4 threads per worker (so 16 threads capable of accepting requests per instance). This updates the code to run with 2 workers (significantly helping memory, but not required since threads share memory) and 32 threads per worker. So now each instance will have 64 threads capable of processing requests.

## Tests
- [ ] unit tests

This was load tested on Sandbox. The increased number of threads allows the server to handle a number of requests that would cause the original settings to drop requests.

### DIFF
--- a/rdr_service/services/gunicorn_config.py
+++ b/rdr_service/services/gunicorn_config.py
@@ -12,7 +12,7 @@ max_requests_jitter = 50
 if os.getenv('GAE_ENV', '').startswith('standard'):
     _port = os.environ.get('PORT', 8081)
     workers = multiprocessing.cpu_count()
-    threads = multiprocessing.cpu_count() * 16
+    threads = workers * 16
 
 
 bind = "0.0.0.0:{0}".format(_port)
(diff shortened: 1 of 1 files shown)


dtharpeuno (approved):
Nice, maybe share how you were able to load test sandbox? A lot of consecutive CURLs to API route?

kenny-skaggs:
> Nice, maybe share how you were able to load test sandbox? A lot of consecutive CURLs to API route?

Yeah, I had uploaded a bunch of PUT requests to the biobank specimen endpoint. I put together a script that would run from my computer and used threads to send concurrent requests to the server. I was able to see the same errors that we're seeing with the server getting overwhelmed when I had 25 separate threads each making their own set of 5 consecutive requests.

wangy70 (approved):

PR #2422, opened 2021-05-04
[PDR Fix] Biobank sample pipeline may not trigger PDR data updates

pegbertsch:

Two issues that were not properly rebuilding `bigquery_sync` participant data after state changes in participant's RDR data due to biobank sample pipeline activity:

1. `update_bigquery_sync_participants()` was being called with a  timestamp reflecting local system time, which was used to query against the UTC timestamps in the RDR to find recently updated records.   (May be okay in production if GAE instances are set to UTC  but unsure if there are conditions where a different local TZ could be in effect)

2. The pipeline logic updated the `participant_summary` table but was looking for recently modified `participant` table records for the PDR data rebuild.   This would miss participants whose biobank details were updated by the pipeline job

## Tests
- [x] unit tests
Updated `cron_job_tests.test_biobank_samples_pipeline.test_end_to_end` to test for the expected `bigquery_sync` table changes

### DIFF
--- a/rdr_service/offline/biobank_samples_pipeline.py
+++ b/rdr_service/offline/biobank_samples_pipeline.py
@@ -29,6 +29,7 @@ from rdr_service.model.biobank_stored_sample import BiobankStoredSample
 from rdr_service.model.code import Code
 from rdr_service.model.config_utils import from_client_biobank_id, get_biobank_id_prefix
 from rdr_service.model.participant import Participant
+from rdr_service.model.participant_summary import ParticipantSummary
 from rdr_service.model.questionnaire import QuestionnaireQuestion
 from rdr_service.model.questionnaire_response import QuestionnaireResponse, QuestionnaireResponseAnswer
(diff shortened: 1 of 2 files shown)


rdr_service/offline/biobank_samples_pipeline.py
```diff
@@ -86,7 +87,7 @@ def upsert_from_latest_csv():
         csv_reader = csv.DictReader(csv_file, delimiter="\t")
         written = _upsert_samples_from_csv(csv_reader)
 
-    ts = datetime.datetime.now()
+    ts = clock.CLOCK.now()
```
pegbertsch: Note: this is the method used to generate the timestamp applied to the `participant_summary` records in `upsert_from_latest_csv()`, so it was changed from a `datetime` method here for consistency

wangy70 (approved):

robabram (approved):

PR #2423, opened 2021-05-04
Adding code coverage reports to Codacy

kenny-skaggs:
## Resolves *no ticket*
Codacy has the ability to display code coverage data, this sets CircleCI up to send coverage reports for tracking. I think with these settings, Codacy would only display the code coverage for the devel branch. I'm still trying to make sense of how it works and that might be easier when I can see them working for the devel branch.

`.coveragerc` is used to configure options for the coverage (such as branch detection) and is also picked up by Pycharm. `ci/run_unittests.sh` is modified to generate coverage reports for the unit tests while running them and then send the reports to Codacy (using an environment variable to associate it with the RDR project).

## Tests
- [ ] unit tests

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+command_line = -m unittest discover -v -s tests
+branch = True
+source =
+    rdr_service
+omit=
+    # ignore fhir client libs
+    rdr_service/lib_fhir/*
+    rdr_service/alembic/*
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):
Screen Shot 2021-05-04 at 10 19 47 AM

Cool!

robabram (approved):

PR #2432, opened 2021-05-05
Add condition for existing manifest record

j-kanuch:
This PR adds a condition to only create a new `genomic_manifest_file` record if one currently doesn't exist for the specified file path.

## Tests
- [] unit tests


wangy70 (approved):

dtharpeuno (approved):
Nice thanks for fixing that.

PR #2434, opened 2021-05-06
Update compare-ingestion tool

j-kanuch:
This PR updates the genomic compare-ingestion tool. The AW2 query was modified to use `genomic_file_processed` and the output file now includes a column for the manifest file path. Some logic was modified to skip empty rows.

## Tests
- [] unit tests


dtharpeuno (approved):
👍

PR #2437, opened 2021-05-07
[DA-1985] Using Flask-Limiter to set up rate limiting

kenny-skaggs:
## Resolves *[DA-1985](https://precisionmedicineinitiative.atlassian.net/browse/DA-1985)*
This uses Flask-Limiter (https://flask-limiter.readthedocs.io/en/stable/) to protect the API from being overwhelmed by too many requests from a particular client (such as the recent issue seen with workbench).

Another recent PR for tuning the Gunicorn settings (#2417) enables the server to handle more requests per second. So the changes for this PR are mostly an extra layer to help protect the API's responsiveness.

From April 26th to May 6th, workbench had a few instances of going over 15 requests per second. PTSC was close once with 9 requests in a second, and HPRO reached 8. With these changes, when a service account makes more than 15 requests per second to the same endpoint the API will return a 429 status code and give the details that there can only be 15 requests per second.

The default setting of 15 requests per second can be overridden by a config change. And the config will also store the connection information for the Memorystore cache that will allow each instance to share a count of the number of requests received for a service account and endpoint.

## Tests
- [x] unit tests


rdr_service/app_util.py
```diff
@@ -389,3 +391,15 @@ def datetime_as_naive_utc(value):
...
 def is_care_evo_and_not_prod():
     return GAE_PROJECT != "all-of-us-rdr-prod" and get_account_origin_id() == "careevolution"
+
+
+def install_rate_limiting(app):
+    cache_location = config.getSettingJson('cache_storage_location', default='memory://')
+    default_rate_limit = config.getSettingJson('default_rate_limit', default='15/second')
+    Limiter(
+        app,
+        key_func=lambda: get_oauth_id() or get_remote_address(),
```
dtharpeuno, line 401: Big fan of `lambda` functions, but not sure it's needed here, the or statement should suffice if a result is returned from either of those methods.
key_func=`get_oauth_id() or get_remote_address()`

dtharpeuno (approved):
This is cool, does `TooManyRequests` return a 429?

kenny-skaggs:
> This is cool, does `TooManyRequests` return a 429?

Yeah, `TooManyRequests` is representative of the 429 code.

rdr_service/app_util.py
kenny-skaggs: The `key_func` parameter shouldn't be the result of the methods, but instead needs to be a callable that gives the result of the or-statement. `get_oauth_id` and `get_remote_address` wouldn't actually be called here when setting up the Limiter, but instead are part of the definition of the function given as the `key_func` parameter.

wangy70 (approved):

PR #2441, opened 2021-05-11
[DA-2007] Merge BigQuery and Resource participant generator code

robabram:
* Merged BQ participant generator code with Resource generator
* Removed PDR version of Resource Participant
* Added PII filters to participant schema
* Added tool to verify BQ participant is correct when using Resource generator.
* Added first pass on recording participant activity into a list.  Hopefully will make calculating participant enrollment status easier.

## Tests
- Changes were tested on Sandbox by running a full rebuild for all participants.


rdr_service/dao/bq_participant_summary_dao.py
```diff
@@ -142,46 +34,44 @@ class BQParticipantSummaryGenerator(BigQueryGenerator):
...
-            # # Depreciated for now: calculate enrollment status times
-            # summary = self._merge_schema_dicts(summary, self._calculate_enrollment_timestamps(summary))
-            # calculate distinct visits
-            summary = self._merge_schema_dicts(summary, self._calculate_distinct_visits(summary))
-            # calculate test participant status (if it was not already set by _prep_participant() )
-            # TODO:  Can this be removed now in favor of determination from _prep_participant()?
-            if summary['test_participant'] == 0:
-                summary = self._merge_schema_dicts(summary, self._check_for_test_credentials(summary))
+        # Add sub-table field prefixes back in and map a few other fields.
+        for k, v in SUB_PREFIXES.items():  # pylint: disable=unused-variable
```
kenny-skaggs, line 66: Why use `for k, v in SUB_PREFIXES.items()` and have pylint ignore the unused `v` rather than just get the keys with `for k in SUB_PREFIXES`?

rdr_service/resource/helpers.py
```diff
@@ -109,3 +110,23 @@ def get_intersection(self, other_collection):
...
+    Load the file app_data/rural_zipcodes.txt and return a list of rural zipcodes.
+    """
+    paths = ('app_data', 'rdr_service/app_data', 'rest-api/app_data')
+    codes = list()
+
+    for path in paths:
+        if os.path.exists(os.path.join(path, 'rural_zipcodes.txt')):
+            with open(os.path.join(path, 'rural_zipcodes.txt')) as handle:
+                # pylint: disable=unused-variable
+                for count, line in enumerate(handle):
```
kenny-skaggs, line 126: Why use `enumerate` if the `count` variable isn't used?

rdr_service/dao/bq_participant_summary_dao.py
robabram: It's just the style I'm used to, I don't remember that working in Python 2.x.

rdr_service/resource/helpers.py
robabram: I don't remember why I used 'enumerate' there, but it works fine as is.

kenny-skaggs: It works, but do you agree that it's less readable and affects maintainability? I know it's small, but I'm hoping we can have more discussions as a team and learn from each other.

rdr_service/resource/generators/participant.py
```diff
@@ -606,6 +704,11 @@ def _prep_modules(self, p_id, ro_session):
                 # consent_added == True means we already know it wasn't a replayed consent
                 if consent_added or not self.is_replay(last_mod_processed, module_data, created_key='module_created'):
                     modules.append(module_data)
+                    try:
+                        activity.append(_act(row.authored, ActivityGroup.QuestionnaireModule,
+                                                 ParticipantActivity[module_name], **mod_ca))
```
kenny-skaggs: We've been moving toward letting codes be case-insensitive. Can this be updated to account for that?

rdr_service/resource/generators/participant.py
```diff
@@ -707,9 +824,12 @@ def _prep_physical_measurements(self, p_id, ro_session):
                 'finalized_site': self._lookup_site_name(row.finalizedSiteId, ro_session),
                 'finalized_site_id': row.finalizedSiteId,
             })
+            _act(row.finalized, ActivityGroup.Profile, ParticipantActivity.PhysicalMeasurements)
```
kenny-skaggs: Should the result of the `_act` call be appended to the `activity` list?

rdr_service/resource/generators/participant.py
robabram: Done

rdr_service/resource/generators/participant.py
robabram: Yes, good catch.

rdr_service/resource/helpers.py
robabram: @kenny-skaggs, look I don't want to be insensitive to your concerns here and I certainly don't believe this code is the most Pythonic way to enumerate through the file line by line.  But I do believe that the code functions correctly and is readable and anyone with Python knowledge will understand exactly what it is doing.

This block of code was written in 2019 and has functioned correctly and without errors since then.  On a scale of 1 to 10, my priority on updating this code to be the way you want it to look is -100. Peggy and I are trying to accomplish multiple big tasks in PDR under a time crunch, this is just not a priority for me right now.  I hope that you will understand that I would rather meet deadlines with working code than deliver pretty code late.

But if it is bothering you so badly please change it and create a PR, I will be happy to approve it. Otherwise it can wait until the RDR 2.0 rewrite.

kenny-skaggs (approved):
No other obvious functionality errors found. @pegbertsch may want to review the algorithms and data processing

rdr_service/resource/generators/participant.py
```diff
@@ -595,6 +692,7 @@ def _prep_modules(self, p_id, ro_session):
 
                         module_data['status'] = module_status.name
                         module_data['status_id'] = module_status.value
+                        mod_ca['ConsentAnswer'] = consent['consent_value']
```
pegbertsch, line 695: I'm not sure why we need this field here, but I've also been wondering if we're reaching a point where having separate consent nested records is becoming unnecessary.   We could theoretically filter the module data based on a new module data flag field that says whether that module is a consent or not.   (May require a little more than that to make a backwards compatible v_pdr_participant_consent view that filters v_pdr_participant_module results looking for consents only).  But that may also have a downside w/query performance.

rdr_service/resource/generators/participant.py
```diff
@@ -777,6 +903,25 @@ def _make_sample_dict_from_row(bss=None, bos=None):
...
+            """
+            act_ = list()
+            confirmed_ts = processed_ts = datetime.datetime.min
+            for sample in samples:
+                if sample['confirmed'] and sample['confirmed'] > confirmed_ts:
+                    confirmed_ts = sample['confirmed']
+                if sample['processed'] and sample['processed'] > confirmed_ts:
+                    processed_ts = sample['processed']
+            if confirmed_ts != datetime.datetime.min:
+                act_.append(_act(confirmed_ts, ActivityGroup.Biobank, ParticipantActivity.BiobankReceived))
```
pegbertsch, line 920: Since specific samples (DNA tests) factor into enrollment status calculations, do we need to distinguish if the activity involved a DNA sample?

tests/api_tests/test_questionnaire_response_api.py
```diff
@@ -526,8 +526,7 @@ def test_gror_consent(self):
                            resource,
                            expected_status=http.client.BAD_REQUEST)
 
-    @mock.patch('rdr_service.dao.bq_participant_summary_dao.logging')
-    def test_gror_consent_with_duplicate_answers(self, mock_logging):
+    def test_gror_consent_with_duplicate_answers(self):
```
pegbertsch, line 529: I think we can replace this with a mock of the resource generator logging.  The intent was to confirm that the generator logged the message about skipping duplicate responses to the same question.

tests/api_tests/test_questionnaire_response_api.py
```diff
@@ -574,8 +573,6 @@ def test_gror_consent_with_duplicate_answers(self, mock_logging):
 
         summary = self.send_get("Participant/%s/Summary" % participant_id)
         self.assertEqual(summary['consentForGenomicsROR'], 'SUBMITTED')
-        # Generic verification PDR generator logged a warning.  TODO:  Possible to test logged message w/regex?
```
pegbertsch, line 577: replace with check of the resource generator logging?

rdr_service/tools/tool_libs/bq_res_validator.py
```diff
@@ -0,0 +1,178 @@
...
+
+            res = res_gen.make_resource(pid)
+            res_data = res.get_data()
+
+            bq = bq_gen.make_bqrecord(pid)
+            bq_data = bq.to_dict()
+
+            self.compare_dicts(bq_data, res_data)
+
+        return 0
```
pegbertsch, line 151: Possible suggestion for this tool:   compare the results from the "new" bq generator to what's in the bigquery_sync  pdr_participant resource data for the  pid and confirm backwards compatibility.

rdr_service/resource/generators/participant.py
robabram: I think the next iteration of the activity logging could roll-up the test types we need so there is only one activity record per biobank order.

rdr_service/resource/generators/participant.py
robabram: This question answer value goes into to the activity log so you can know what they answered for each consent module.

tests/api_tests/test_questionnaire_response_api.py
robabram: Cool, I wasn't quite how to update that mock patch.

tests/api_tests/test_questionnaire_response_api.py
robabram: How about in the next resource PR?

rdr_service/tools/tool_libs/bq_res_validator.py
robabram: Before I deleted the BQ generator code, I had this tool comparing the BQ generator output to the Resource generator output.  Once it lined up perfectly, I deleted the BQ generator and refactored this tool to compare the outputs again to make sure they lined up perfectly.  I'm feeling very good about this merge of the two generators, but it wouldn't hurt to expand this tool in the future to do a bigger comparison across all the participants.

rdr_service/resource/generators/participant.py
pegbertsch: @robabram I don't have a problem with this, but since we have the status  in the module data as well (SUBMITTED -- aka yes if this is a consent module, SUBMITTED_NO_CONSENT, SUBMITTED_NOT_SURE) just wondering if we also need the answer code or can work off the module status and response status (IN_PROGRESS or COMPLETED).  Will approve as is for now.  Given that RDR tends to use those SUBMITTED status enums in `participant_summary` thought we might want to use those in the activity history, too.

tests/api_tests/test_questionnaire_response_api.py
pegbertsch: That's fine.

pegbertsch (approved):

PR #2442, opened 2021-05-12
[DA-2015] Check for arrays AW3 before submitting WGS data in AW3

j-kanuch:

Broad DRC has requested that RDR not sent WGS samples in the AW3 until the corresponding record with the `aou_array` genome type has had data provided by the GCs. 
The `gc_manifest_parent_sample_id`, which is supplied by the Biobank in the AW1 (`parent_sample_id`), is used to match the `aou_wgs` record in `genomic_set_member` to the `aou_array` sample. This matching is accomplished via a self-join to `genomic_set_member`.
Rather than computing all Array conditions, the WGS AW3 query was updated to use the existence of the  Array sample's AW3 job run ID (`aw3ManifestJobRunID`) as a proxy for these criteria. The Array AW3 job runs daily and requires that all data files exist. This criteria is identical to the Broad DRC's request and therefore justifies the use of the AW3 job run ID as a proxy for this criteria.

## Tests
- [] unit tests
`test_aw3_wgs_manifest_generation` was expanded to test the array sample AW3.

### DIFF
--- a/rdr_service/genomic/genomic_data.py
+++ b/rdr_service/genomic/genomic_data.py
@@ -1,4 +1,5 @@
 import sqlalchemy
+from sqlalchemy.orm import aliased
 
 from rdr_service.config import GENOME_TYPE_ARRAY, GENOME_TYPE_WGS
 from rdr_service.genomic_enums import GenomicSubProcessResult, GenomicWorkflowState, GenomicManifestTypes, \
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):
👍

kenny-skaggs (approved):

PR #2443, opened 2021-05-12
[DA-2020] Remove A1 manifest validation from Slack alerts

j-kanuch:
The A2 ingestion cron job was picking up the A1 manifest. The A2 job was failing the A1 manifest due to incorrect naming convention and then posting alerts to the Slack `#rdr-genomic-alerts` channel. This PR explicitly sets the `controller.sub_folder_name` in the A2 ingestion workflow so that it does not pick up the A1 manifest.

## Tests
- [] unit tests

### DIFF
--- a/rdr_service/offline/genomic_pipeline.py
+++ b/rdr_service/offline/genomic_pipeline.py
@@ -233,8 +233,8 @@ def gem_a2_manifest_workflow():
     Entrypoint for GEM A2 Workflow
     """
     with GenomicJobController(GenomicJob.GEM_A2_MANIFEST,
-                              bucket_name=config.GENOMIC_GEM_BUCKET_NAME) as controller:
-        controller.sub_folder_name = config.GENOMIC_GEM_A2_MANIFEST_SUBFOLDER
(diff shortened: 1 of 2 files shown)


rdr_service/offline/genomic_pipeline.py
```diff
@@ -234,6 +234,7 @@ def gem_a2_manifest_workflow():
     """
     with GenomicJobController(GenomicJob.GEM_A2_MANIFEST,
                               bucket_name=config.GENOMIC_GEM_BUCKET_NAME) as controller:
+        controller.sub_folder_name = config.GENOMIC_GEM_A2_MANIFEST_SUBFOLDER
```
dtharpeuno: Thanks for finding this. Can we just set it as a parameter in the controller when it's being invoked?

dtharpeuno (approved):

rdr_service/offline/genomic_pipeline.py
j-kanuch: Yes, good suggestion. I pushed a new commit that implements this.

wangy70 (approved):

PR #2447, opened 2021-05-17
[DA-2024] modify_public_metrics_cronjob

wangy70:
## Resolves *[DA-2024](https://precisionmedicineinitiative.atlassian.net/browse/DA-2024)*
1. Calculate public metrics cache for stage one(recent 30 days) with the latest production data.
2. Reuse the public metics historical cache for stage two(from 2017.5 to 30 days ago).
3. No change for metric cache of health pro dashboard.

For public metrics, we don't need to recalculate the withdrawal, test, ghost count for the historical data, the logic is straightforward and easy now. But this change still follow the current design, because the public metrics shares the logic with healthPro dashboard metrics. Once the healthPro dashboard retired(maybe in two weeks), we need to refactor the design and use a simple way to calculate the public metrics.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/metrics_cache_dao.py
+++ b/rdr_service/dao/metrics_cache_dao.py
@@ -67,6 +67,18 @@ class MetricsCacheJobStatusDao(UpdatableDao):
             record = query.first()
             return record
 
+    def get_last_complete_stage_two_data_inserted_time(self, table_name, cache_type=None):
+        with self.session() as session:
+            query = session.query(MetricsCacheJobStatus.dateInserted)
+            query = query.filter(MetricsCacheJobStatus.cacheTableName == table_name,
+                                 MetricsCacheJobStatus.stage_one_complete.is_(True),
+                                 MetricsCacheJobStatus.stage_two_complete.is_(True))
+            if cache_type:
+                query = query.filter(MetricsCacheJobStatus.type == str(cache_type))
+            query = query.order_by(desc(MetricsCacheJobStatus.id))
+            record = query.first()
+            return record
+
 
 class MetricsEnrollmentStatusCacheDao(BaseDao):
     def __init__(self, cache_type=MetricsCacheType.METRICS_V2_API, version=None):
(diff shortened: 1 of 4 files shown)


j-kanuch (approved):

kenny-skaggs:
I'm trying to understand the intention behind the `copy_historical_cache_data` methods? Is it just that the tables need to have the data with the newer `date_inserted` dates? Do the old copies of the data need to be persisted, or would it potentially be more time and memory effective to just update the `date_inserted` values of the older records?

wangy70:
> I'm trying to understand the intention behind the `copy_historical_cache_data` methods? Is it just that the tables need to have the data with the newer `date_inserted` dates? Do the old copies of the data need to be persisted, or would it potentially be more time and memory effective to just update the `date_inserted` values of the older records?

Haha, I also hesitated between `insert` and `update` for a while. Eventually I chose `insert`, I think it's more safe to have some copies of the original historical data. If something wrong, we can recover from the original data. But I'm also ok to change to `update`, let me know if you still think `update` is better. I will make the change:)

kenny-skaggs:
> Haha, I also hesitated between `insert` and `update` for a while. Eventually I chose `insert`, I think it's more safe to have some copies of the original historical data. If something wrong, we can recover from the original data. But I'm also ok to change to `update`, let me know if you still think `update` is better. I will make the change:)

I haven't fully walked through it, but I'm wondering how much database space this might create and how much time it might take to copy them. If I'm understanding it correctly each day this runs would have a record for its own day (inserted in stage one) and would insert a new copy of the records inserted for the previous day (which would be a row for every day since the start of the program). Is that correct?

wangy70:
> > Haha, I also hesitated between `insert` and `update` for a while. Eventually I chose `insert`, I think it's more safe to have some copies of the original historical data. If something wrong, we can recover from the original data. But I'm also ok to change to `update`, let me know if you still think `update` is better. I will make the change:)
> 
> I haven't fully walked through it, but I'm wondering how much database space this might create and how much time it might take to copy them. If I'm understanding it correctly each day this runs would have a record for its own day (inserted in stage one) and would insert a new copy of the records inserted for the previous day (which would be a row for every day since the start of the program). Is that correct?

Yes, you are right. There will be millions of records to copy.  Do you think `update` will be faster than `insert`?

kenny-skaggs:
> Yes, you are right. There will be millions of records to copy. Do you think `update` will be faster than `insert`?

I think update would be a bit faster. But I guess I'm more worried about the amount of duplication. Some of the tables are a gigabyte or more and only have about 7 days worth of data inserted. If most of the storage space for those tables is used for this historical data then I could see them growing to more than 5 gigs.

If we use the update instead and the only value that would be changed is the `date_inserted` field I think we could restore the state if needed. Or am I missing something there?

wangy70:
> > Yes, you are right. There will be millions of records to copy. Do you think `update` will be faster than `insert`?
> 
> I think update would be a bit faster. But I guess I'm more worried about the amount of duplication. Some of the tables are a gigabyte or more and only have about 7 days worth of data inserted. If most of the storage space for those tables is used for this historical data then I could see them growing to more than 5 gigs.
> 
> If we use the update instead and the only value that would be changed is the `date_inserted` field I think we could restore the state if needed. Or am I missing something there?

Make sense, I will change it to `update`.

wangy70:
Done

kenny-skaggs (approved):

PR #2453, opened 2021-05-19
[DA-2027] Message Broker API for color with mocked PTSC response

wangy70:
## Resolves *[DA-2027](https://precisionmedicineinitiative.atlassian.net/browse/DA-2027)*
Since PTSC’s test env is not ready, this PR implement Message Broker API for color with mocked PTSC response. So that color can integrate with Message Broker API as early as possible.

1. Create message broker tables
2. Setup message broker API and store message broker record
3. Hard code PTSC's response, will update this part once PTSC's test env is ready.

## Tests
- [x] unit tests


rdr_service/main.py
```diff
@@ -357,6 +358,8 @@ def _log_request_exception(sender, exception, **extra):  # pylint: disable=unuse
 # Data generator API used to load fake data into the database.
 api.add_resource(DataGenApi, API_PREFIX + "DataGen", endpoint="datagen", methods=["POST", "PUT"])
 
+# Message broker API
+api.add_resource(MessageBrokerApi, API_PREFIX + "Message/Broker", endpoint="message.broker", methods=["POST"])
```
dtharpeuno: Is this the path naming convention that was decided on? Are there going to be more endpoints to the message API route?

rdr_service/dao/message_broker_dao.py
```diff
@@ -0,0 +1,94 @@
...
+        dest_url = self._get_message_dest_url(message.eventType, message.messageDest)  # pylint: disable=unused-variable
+        request_body = message.requestBody  # pylint: disable=unused-variable
+
+        # PTSC test env is not ready for this implementation
+        # TODO - get access to dest endpoint and sent message to dest
+        # mock response from PTSC
+
+        response_code = '200'
+        response_body = {'result': 'mocked result'}
+        response_error = ''
```
dtharpeuno, line 82: Should the response logic be in the API class?

rdr_service/model/message_broker.py
```diff
@@ -0,0 +1,64 @@
...
+    modified = Column("modified", UTCDateTime6, nullable=True)
+    """The last modified time for this record."""
+
+    eventType = Column("event_type", String(128))
+    """message event type"""
+    destination = Column("destination", String(80))
+    """message destination, decided by participant origin"""
+    url = Column("url", String(512))
+    """message destination endpoint"""
+
```
dtharpeuno, line 59: Does this need to tie back to the `MessageBrokerRecord` via a foreign key so the records are linked?

rdr_service/dao/message_broker_dao.py
wangy70: This is not the final response to color. This is the middle response from PTSC, and we need to wrap it and return to color. The final response for color is in `to_client_json` which overwrite the base class method and follow our current framework :)

rdr_service/main.py
wangy70: Based on current design, this is the only endpoint for message broker, we will use different `eventType` in the payload for different events.

dtharpeuno: Should we make it `MessageBroker` then?

rdr_service/model/message_broker.py
wangy70: They are not the foreign key relationship. They all can have some value that the other table don't have. For `MessageBrokerRecord` , it's possible that we got some message with wrong event type, but we still need to store the record.

rdr_service/dao/message_broker_dao.py
```diff
@@ -0,0 +1,94 @@
...
+from werkzeug.exceptions import BadRequest
+from rdr_service.dao.base_dao import BaseDao
+from rdr_service.dao.participant_dao import ParticipantDao
+from rdr_service.model.message_broker import MessageBrokerRecord, MessageBrokerMetadata
+
+
+class MessageBrokerDao(BaseDao):
+    def __init__(self):
+        self.message_metadata_dao = MessageBrokerMetadataDao()
+        super(MessageBrokerDao, self).__init__(MessageBrokerRecord)
```
dtharpeuno: :nit Can we call super here first since we are not doing any validations/calcs needed before we are calling `super`.

rdr_service/model/message_broker.py
```diff
@@ -0,0 +1,64 @@
...
+    __table_args__ = (UniqueConstraint('event_type', 'destination', 'url', name='unique_message_target'),)
+
+
+event.listen(MessageBrokerRecord, "before_insert", model_insert_listener)
+event.listen(MessageBrokerRecord, "before_update", model_update_listener)
```
dtharpeuno, line 64: Do we not need these for `MessageBrokerMetadata`?

wangy70: Good catch, I missed that.

rdr_service/dao/message_broker_dao.py
wangy70: Done

rdr_service/main.py
wangy70: Done

dtharpeuno (approved):
Nice job!

robabram (approved):

PR #2459, opened 2021-05-21
Fixing slow_sql alerts

kenny-skaggs:
## Resolves *no ticket*
The code for finding duplicate questionnaire responses was occasionally triggering the slow_sql alert.

The script doesn't really need to check on QuestionnaireResponses that have already been analyzed, only the new ones since the last run. This updates the script to look at the new responses and compare them against all the responses we have in order to see if the new responses have duplicates.

I pulled the query for loading the responses out into it's own method to make it easier to test. I also have it looking at the responses created in the last 2 days (creating a little overlap just to be sure we don't miss anything since the cron job runs daily). To take advantage of an index that makes finding responses by created date, the questionnaire id needed to be referenced and a hint needed to be provided to Mysql.

## Tests
- [x] unit tests


j-kanuch (approved):

dtharpeuno (approved):
NIce query!

PR #2461, opened 2021-05-25
[DA-2044] Adding vaccine survey fields to participant summary

kenny-skaggs:
## Resolves *[DA-2044](https://precisionmedicineinitiative.atlassian.net/browse/DA-2044)*
Adds fields to the participant summary for the first COPE Vaccine Minute survey.

Added columns to the participant summary and set them in the questionnaire response dao when getting a response for the survey.

## Tests
- [x] unit tests


rdr_service/alembic/versions/93fce807f225_fields_in_participant_summary_for_.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('participant_summary', sa.Column('questionnaire_on_vaccine_minute_1', rdr_service.model.utils.Enum(QuestionnaireStatus), nullable=True))
```
pegbertsch: should this be cope_vaccine_minute_1?

rdr_service/code_constants.py
```diff
@@ -97,6 +97,9 @@
 CONSENT_COPE_NO_CODE = "COPE_A_13"
 CONSENT_COPE_DEFERRED_CODE = "COPE_A_231"
 
+# COPE Minute Survey Codes
+VACCINE_MINUTE_1_MODULE_CODE = "cope_vaccine1"
```
pegbertsch: Wondering if we should include COPE_ in the named constants, etc.   (On the off chance there is some future module pertaining to a different vaccine)

tests/dao_tests/test_questionnaire_response_dao.py
```diff
@@ -847,6 +868,28 @@ def test_february_cope_survey(self):
...
+            authored_datetime=authored_date
+        )
+
+        # Check that the summary fields have the survey data
+        summary: ParticipantSummary = self.session.query(ParticipantSummary).filter(
+            ParticipantSummary.participantId == participant.participantId
+        ).one()
+        self.assertEqual(QuestionnaireStatus.SUBMITTED, summary.questionnaireOnVaccineMinute1)
+        self.assertEqual(authored_date, summary.questionnaireOnVaccineMinute1Authored)
+
```
pegbertsch, line 902: Optional: There's an opportunity to use the PDRGeneratorTestMixin to validate we've also captured the questionnaire response details in the pdr_participant (nested module) data and/or the pdr_mod_cope_vaccine1 data.  (I can also look at adding it in a separate PDR)

rdr_service/alembic/versions/93fce807f225_fields_in_participant_summary_for_.py
kenny-skaggs: Do mean having the column names be `questionnaire_on_cope_vaccine_minute_1` and `questionnaire_on_cope_vaccine_minute_1_authored`?

rdr_service/code_constants.py
kenny-skaggs: I think I might have started with that, but took it out since it was a bit long. At the moment there's only the one minute survey, and a constant in the code would be easy enough to update if another minute survey was started for another vaccine. But I can update it if you still think it'd be better to have it contain COPE now rather than later.

tests/dao_tests/test_questionnaire_response_dao.py
kenny-skaggs: The PDR code would need to be updated first, right? I think that's the scope for DA-2043.

rdr_service/alembic/versions/93fce807f225_fields_in_participant_summary_for_.py
pegbertsch: Yes.  Also, based on the codebook, I didn't think "minute" was part of the module name.  Not sure if prior precedent on these fields is to mirror the module code value string?

tests/dao_tests/test_questionnaire_response_dao.py
pegbertsch: For the participant data, if the Questionnaire/QuestionnaireResponse data exists in the test database then I think the  participant data generators should pick it up in the participant's nested module data.   There may still be another field to verify later if we update the generators to treat it like a baseline/ppi module?

I also already added skeleton support for the new module  as part of [PR #2452](https://github.com/all-of-us/raw-data-repository/pull/2452/filesl), for the BQPDRQuestionnaireResponseGenerator (we don't have a resource generator yet for the module response data).  This was when we thought the launch was going to be 6/3.  I may need to update the DA ticket.  

I can make the unit test changes later.

rdr_service/alembic/versions/93fce807f225_fields_in_participant_summary_for_.py
kenny-skaggs: I agree on COPE. I was trying to reduce the size of the name, but I don't think it's worth losing the COPE in the name.

I have minute in there because I've seen discussions of it referring to it as the "Minute Survey" without saying much about Covid or vaccine, and I thought it might be helpful for anyone trying to find the data if "minute" was there too. I'll try to track down what the messaging is between different partners for the name of the survey.

I could be wrong, but I didn't think we were necessarily basing these fields on the code names. Would consumers of the summary api (like HPRO and the sites) be aware of the code names?

tests/dao_tests/test_questionnaire_response_dao.py
kenny-skaggs: Ah, sorry. I forgot that was added.

Thanks for pointing out (and setting up) the PDRGeneratorTestMixin. I'll update the unit test.

kenny-skaggs: Ok, should be added. Let me know if there's anything that should be changed for how I tested.

rdr_service/code_constants.py
kenny-skaggs: Added COPE

rdr_service/alembic/versions/93fce807f225_fields_in_participant_summary_for_.py
kenny-skaggs: Added COPE. I'm still leaning toward leaving 'minute' there since Vibrent seems to be referring to it as the Minute Survey. But let me know what you think.

pegbertsch: I called it pdr_mod_cope_vaccine1 for the PDR schema in the RDR 1.96.1 release PDR changes;  I don't necessarily think they have to align but I need to decide whether to change it in the hotfix.  Thoughts?

tests/dao_tests/test_questionnaire_response_dao.py
```diff
@@ -847,6 +868,38 @@ def test_february_cope_survey(self):
...
+        # Check that the summary fields have the survey data
+        summary: ParticipantSummary = self.session.query(ParticipantSummary).filter(
+            ParticipantSummary.participantId == participant.participantId
+        ).one()
+        self.assertEqual(QuestionnaireStatus.SUBMITTED, summary.questionnaireOnCopeVaccineMinute1)
+        self.assertEqual(authored_date, summary.questionnaireOnCopeVaccineMinute1Authored)
+
+        participant_res_data = self.make_participant_resource(participant.participantId)
+        vaccine_module_data = None
+        for module_data in participant_res_data['modules']:
```
pegbertsch: For future reference:  There is a convenience method in the mixin where you can do something like:

 ```
vaccine_module_data = self. get_generated_items(participant_res_data['modules'],  item_key='module',
                                       item_value=COPE_VACCINE_MINUTE_1_MODULE_CODE)
```
...and get back a list of any matching module entries (would have to dereference expected index/element)

pegbertsch (approved):

rdr_service/alembic/versions/93fce807f225_fields_in_participant_summary_for_.py
kenny-skaggs: I think they would align enough as they are, especially if they're for different audiences.

tests/dao_tests/test_questionnaire_response_dao.py
kenny-skaggs: Thanks! I was looking for something like that, but must have not been looking closely enough.

pegbertsch (approved):

wangy70 (approved):

PR #2465, opened 2021-05-27
[DA-2042] fix_hos_update_cache_issue

wangy70:
## Resolves *[DA-2042](https://precisionmedicineinitiative.atlassian.net/browse/DA-2042)*
Refresh HPO list cache when doing update or insert

When multiple HOP requests comes, they may hit different instances, in some chance the cache maybe inconsistent between multiple instances, this will cause issue. This fix force refresh the cache to resolve this problem. 

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/cache_all_dao.py
+++ b/rdr_service/dao/cache_all_dao.py
@@ -86,5 +86,7 @@ class CacheAllDao(UpdatableDao):
             return []
         return [self.get(id_) for id_ in ids]
 
-    def get_all(self):
+    def get_all(self, refresh_cache=False):
+        if refresh_cache:
+            self._invalidate_cache()
         return list(self._get_cache().id_to_entity.values())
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):

kenny-skaggs:
Thanks, this is a good way to make sure the cache doesn't have any problems with quick changes. What do you think of also updating the `Not updating...` log statements (like https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/organization_hierarchy_sync_dao.py#L110)  to warnings? That way it's a little less silent if something else strange might be going on.

dtharpeuno:
> Thanks, this is a good way to make sure the cache doesn't have any problems with quick changes. What do you think of also updating the `Not updating...` log statements (like https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/organization_hierarchy_sync_dao.py#L110) to warnings? That way it's a little less silent if something else strange might be going on.

Second this.

wangy70:
> Thanks, this is a good way to make sure the cache doesn't have any problems with quick changes. What do you think of also updating the `Not updating...` log statements (like https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/organization_hierarchy_sync_dao.py#L110) to warnings? That way it's a little less silent if something else strange might be going on.

Done!

kenny-skaggs (approved):

robabram (approved):

PR #2480, opened 2021-06-04
[MIGRATION] DA-2060 Adding incident track for data file ingestions

dtharpeuno:
## Resolves *[ticket DA-2060]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2060

With the recent rollout of the cloud task for ingesting the gvcf data files, PubSub notification workflow is running correctly but if does not find a corresponding genomic metric record, cannot save file paths to database. Need to be able to save these records ingestion attempts and be notified when this happens.

## Tests
- [x] unit tests


kenny-skaggs (approved):

j-kanuch (approved):
Good idea to track it this way.

PR #2481, opened 2021-06-04
[DA-2055] Adding code for validating Vibrent primary consent files

kenny-skaggs:
## Partially resolves *[DA-2055](https://precisionmedicineinitiative.atlassian.net/browse/DA-2055)*
There's going to be alot of code for setting up a system for validating consent PDF files. I'll try to keep the PRs relatively small. This first one sets up a framework for testing how the files are parsed, and adds a few classes for working with PDF files.

There isn't anything yet in the code that depends on these classes. That will be added in one of the last PRs for this work (when setting up a nightly cron job for validating new consents).

This introduces a few different classes to provide functionality for working with the PDF files. The PDF data itself will be parsed by the pdfminer.six library and the `Pdf` class will extend it a bit by adding some convenience methods. The `ConsentFile` abstract base class provides generic methods and functionality needed for all consent files (parsing signatures and dates), but (since each file type is different) the base class relies on subclasses to find the PDF elements that give the elements to retrieve data from. Eventually this validation code will be extended to check CE files as well, so the `PrimaryConsentFile` class holds functionality common to both (checking for VA consent wording). Then the `VibrentPrimaryConsentFile` class encapsulates the functionality specific for where to find elements on the Primary consent files from Vibrent.

## Tests
- [x] unit tests

There are several possible formats for the consent file and we may need to extend the code to work with more in the future. So the tests are set up to check for all layout found so far and allow for adding more to be checked as we find them.

### DIFF
--- a/rdr_service/services/consent_files.py
+++ b/rdr_service/services/consent_files.py
@@ -1,0 +1,185 @@
+from abc import ABC, abstractmethod
+from dateutil import parser
+from geometry import Rect
+from google.cloud.storage.blob import Blob
+from io import BytesIO
+from pdfminer.high_level import extract_pages
+from pdfminer.layout import LTChar, LTFigure, LTImage, LTTextBox
+from typing import List
+
+
+class ConsentFile(ABC):
+    def __init__(self, pdf: 'Pdf'):
+        self.pdf = pdf
+
+    def get_signature_on_file(self):
+        signature_elements = self._get_signature_elements()
+        signature = None
+        for element in signature_elements:
+            first_child = Pdf.get_first_child_of_element(element)
+
+            if isinstance(first_child, LTChar):
+                possible_signature = ''
+                for char in element:
+                    possible_signature += char.get_text()
+                possible_signature = possible_signature.strip()
+                if possible_signature != '':
+                    signature = possible_signature
+                    break
+
+            elif isinstance(first_child, LTImage):
+                signature = True
+                break
+
+        return signature
+
+    def get_date_signed(self):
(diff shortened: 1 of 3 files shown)


requirements_base.txt
```diff
@@ -72,6 +72,7 @@ oauthlib[signedtoken]==3.1.0  # via jira, requests-oauthlib
 packaging==20.4           # via sphinx
 parameterized==0.7.4      # via -r requirements.in
 pbr==5.4.5                # via jira
+pdfminer.six==20201018
```
dtharpeuno, line 75: What is the approval process for this?

rdr_service/services/consent_files.py
```diff
@@ -0,0 +1,170 @@
...
+        for element in page:
+            element_rect = Rect.from_edges(
+                left=element.x0,
+                right=element.x1,
+                bottom=element.y0,
+                top=element.y1
+            )
+            if element_rect.intersection(search_box) is not None:
+                elements.append(element)
+
```
dtharpeuno, line 112: This is cool

dtharpeuno (approved):
🔥

requirements_base.txt
kenny-skaggs: From what I understand we need approval for the licenses themselves and not necessarily the packages. The `ci/check_licenses.py` script (run by CircleCI as the Check Licenses step) checks the license of all installed packages against the `ci/license_whitelist.txt` and `ci/license_exceptions.txt` files. As long as the project's license is there then it's already been approved.

j-kanuch (approved):

PR #2483, opened 2021-06-07
[DA-2028] Message broker ptsc api

wangy70:
1. update the payload format based on PTSC's document.
2. handle the API access authentication/authorization.
    a) PTSC will configure an OpenID connect client in their identity provider.
    b) Calls from DRC will use a token obtained using the client credentials grant, used for machine to machine 
    authentication/authorization.
    c) Token should be included in the HTTP Authorization header using the Bearer scheme.

This PR is implemented based on the document and not test in any real environment, because the PTSC's test env is not ready yet. 

## Tests
only mocked test


rdr_service/message_broker/message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+
+            if response['status'] in ('200', '201'):
+                auth_info.accessToken = content['access_token']
+                now = clock.CLOCK.now()
+                expired_at = now + timedelta(seconds=content['expires_in'])
+                auth_info.expiredAt = expired_at
+                self.dest_auth_Dao.update(auth_info)
+                return content['access_token']
+            else:
+                raise BadRequest(f'can not get access token for dest: {self.message.messageDest}')
```
kenny-skaggs: A 400 BadRequest could be interpreted as an error on the requester's part and imply that something about the request needs to be changed. Would a 502 BadGateway be better to describe that the request couldn't be fulfilled because of upstream issues?

rdr_service/message_broker/message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+    def get_access_token(self):
+        """Returns the access token for the API endpoint."""
+        auth_info = self.dest_auth_Dao.get_auth_info(self.message.messageDest)
+        if not auth_info:
+            raise BadRequest(f'can not find auth info for dest: {self.message.messageDest}')
+
+        now = clock.CLOCK.now()
+        five_mins_later = now + timedelta(minutes=5)
+        if auth_info.accessToken and auth_info.expiredAt > five_mins_later:
+            return auth_info.accessToken
```
kenny-skaggs, line 28: Should a test be added that checks that an expired access token is refreshed?

rdr_service/message_broker/message_broker.py
```diff
@@ -0,0 +1,105 @@
+from werkzeug.exceptions import BadRequest
+import httplib2
```
dtharpeuno: Since we have `requests` in our venv, and its a wider used lib, can we utilize that for the `http` interactions?
https://python.libhunt.com/compare-requests-vs-httplib2

rdr_service/message_broker/message_broker.py
dtharpeuno: ^^ same

dtharpeuno (approved):
Nice job, especially doing this off a doc with no env to interact with. Just left a comment about `http` lib.

rdr_service/message_broker/message_broker.py
wangy70: Done!

rdr_service/message_broker/message_broker.py
wangy70: Done!

rdr_service/message_broker/message_broker.py
wangy70: Done!

wangy70:
> Nice job, especially doing this off a doc with no env to interact with. Just left a comment about `http` lib.

Done, thanks!

kenny-skaggs (approved):

PR #2487, opened 2021-06-09
[DA-2056] Adding model and distinct module for consent validation

kenny-skaggs:
## Partially resolves *[DA-2056](https://precisionmedicineinitiative.atlassian.net/browse/DA-2056)*
To keep the consent code more organized, I created a module of its own within the services module. This updates the `consent_files` module, renaming it to `files` and placing it in the new `consent` module. The next PR will add a `validation` module.

I've also added a `consent_file` database model that will record consent file validation results as well as other data relevant to the consent file (signature and date, and in future PRs whether it has been synced and what the file path is).

When making the migration file, I also noticed that one of the genomic classes didn't line up with the schema. I checked with Darryl and the module field for GenomicMemberReportState should be nullable.

## Tests
- [x] unit tests


rdr_service/model/consent_file.py
```diff
@@ -0,0 +1,42 @@
...
+
+class ConsentFile(Base):
+    __tablename__ = 'consent_file'
+    id = Column("id", Integer, primary_key=True, autoincrement=True, nullable=False)
+    created = Column(UTCDateTime)
+    modified = Column(UTCDateTime)
+    participant_id = Column(Integer, ForeignKey(Participant.participantId))
+    type = Column(Enum(ConsentType))
+
+    file_exists = Column(Boolean)
```
j-kanuch, line 29: Should `file_exists` have a default value of `False`?

kenny-skaggs: At the moment I wasn't planning on it. I was planning on having the code explicitly set whether the file was there or not. Do you think it should have a default value?

j-kanuch: I think it's OK without the default value. I noticed that the other boolean fields had a default value of `False` and I just wanted to make sure this was on purpose.

j-kanuch (approved):

rdr_service/model/consent_file.py
kenny-skaggs: Ah, ok. I see what you mean now, thanks. Yeah, the others default to False to make it easier when creating a validation record for a file that isn't there. That way it can be set to `file_exsts=False` and not need to worry about what should be said for the signature and signing date.

wangy70 (approved):

PR #2489, opened 2021-06-10
Fix for parsing migration files for unittest schema changes

kenny-skaggs:
## Resolves *no ticket*
I've been running into CI issues with unittests and this resolves it. The unittest database schemas are generated from the model files rather than from the migrations. And then the migrations are checked for a function named `unittest_schemas`, these functions exist in a small number of migration files and provide additional sql that needs to be applied to the test database in order to have it more closely match the production database.

The code for analyzing the migrations was written by hand and was failing to find the down_revision for migrations that have more than one down_revision. I was attempting to add a new migration (in the PR I closed yesterday, I'll re-open it when this is merged) and the migration was being treated as the initial migration. The code would look through all migrations and put together the dependencies of them. And then start with the initial migration to walk through and find any `unittest_schemas` methods. But my new migration was being seen as the first, so no schema changes were found. Because of that errors were coming up in code that depended on them being there (like for a missing `ppi_participant_view`).

I'm unclear on how this is the first time we're running into this since this isn't the first migration with two down revisions that we've merged.

This updates the code that looks for the `unittest_schemas` method in migration files by replacing the custom code with alembic's SDK.

## Tests
- [ ] unit tests

### DIFF
--- a/tests/helpers/mysql_helper.py
+++ b/tests/helpers/mysql_helper.py
@@ -219,6 +219,9 @@ def _run_unit_test_migrations(engine):
     with temporary_sys_path('rdr_service'):  # the revision files need to find modules (like model) in rdr_service
         migrations_directory = os.path.join(os.getcwd(), "rdr_service", "alembic")
         migrations_api = ScriptDirectory(migrations_directory)
+
+        # walk_revisions returns revisions in order of newest to oldest,
+        # reversing to start with the first and work up to the latest
         for revision in reversed(list(migrations_api.walk_revisions())):
             with warnings.catch_warnings():  # Ignore warnings from 'DROP IF EXISTS' sql statements
                 warnings.simplefilter("ignore")
--- a/tests/helpers/mysql_helper.py
+++ b/tests/helpers/mysql_helper.py
@@ -6,7 +6,6 @@
 #
 #
 #
-from alembic.script import ScriptDirectory
 import atexit
 import csv
 import inspect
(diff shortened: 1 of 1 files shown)


tests/helpers/mysql_helper.py
```diff
@@ -6,22 +6,20 @@
 #
 #
 #
+from alembic.script import ScriptDirectory
```
dtharpeuno: nit: Can we move this below the standard lib imports

tests/helpers/mysql_helper.py
```diff
@@ -210,81 +208,29 @@ def reset_mysql_instance(with_data=True, with_consent_codes=False):
...
     4 times as slow as using sqlalchemy. Loading DB Functions and Views this
     way into the schema works much faster.
     :param engine: Database engine object
     """
-    alembic_path = os.path.join(os.getcwd(), "rdr_service", "alembic", "versions")
-    if not os.path.exists(alembic_path):
-        raise OSError("alembic migrations path not found.")
+    with temporary_sys_path('rdr_service'):  # the revision files need to find modules (like model) in rdr_service
+        migrations_directory = os.path.join(os.getcwd(), "rdr_service", "alembic")
+        migrations_api = ScriptDirectory(migrations_directory)
```
dtharpeuno, line 220: Way to lean on alembic classes instead. Leave it to the pros!

dtharpeuno (approved):

wangy70 (approved):

PR #2493, opened 2021-06-15
[DA-2056] Cron jobs for managing consent validation

kenny-skaggs:
## Resolves *[DA-2056](https://precisionmedicineinitiative.atlassian.net/browse/DA-2056)*
This wraps up DA-2056 by scheduling a pair of cron jobs. One runs daily and is responsible for checking the consents from the previous 26 hours. For each new consent that has been authored in the past 26 hours at least one record is made: either a valid file is found and will be marked as ready for sync, multiple records for each invalid file will be created to specify what was found as wrong for the files, or a special type of record will be created denoting that no consent files were found.

The other job runs roughly every 7 days and checks for any resolutions for invalid files. Any records are loaded that note invalid or missing files. Then for each participant and consent type that the records indicate, the consent files are checked again. If a new file has been uploaded and is ready to sync, then a record is created for it and each previous validation record is marked as obsolete. If no new files are ready for syncing, then any newly validated files are recorded and are all stored for the next check.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/services/consent/validation.py
+++ b/rdr_service/services/consent/validation.py
@@ -20,7 +20,7 @@ class ConsentValidationController:
         self.participant_summary_dao = participant_summary_dao
         self.storage_provider = storage_provider
 
-        self.va_hpo_id = hpo_dao.get_by_name('VA').hpo_id
+        self.va_hpo_id = hpo_dao.get_by_name('VA').hpoId
 
     def check_for_corrections(self):
         """Load all of the current consent issues and see if they have been resolved yet"""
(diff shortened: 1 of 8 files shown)


rdr_service/services/consent/validation.py
```diff
@@ -1,10 +1,171 @@
...
+            return [ready_file]
+        else:
+            return results
+
+    @classmethod
+    def _has_new_consent(cls, consent_status, authored, min_authored):
+        return consent_status == QuestionnaireStatus.SUBMITTED and authored > min_authored
+
+    def _build_validator(self, participant_summary: ParticipantSummary) -> 'ConsentValidator':
+        consent_factory = files.ConsentFileAbstractFactory.get_file_factory(
```
wangy70, line 128: Should this type hint be an object instead of a string 'ConsentValidator' ?

kenny-skaggs: Normally they'd just be direct references. But in this case the interpreter wouldn't know about the `ConsentValidator` class yet because it's defined later in the same file. So a string can be used as a placeholder which is evaluated later. https://www.python.org/dev/peps/pep-0484/#forward-references

wangy70: Thanks for the explanation, good to know!

wangy70 (approved):

j-kanuch (approved):

PR #2498, opened 2021-06-21
[DA-2066] Pre-calculate enrollment status dates for metrics data in PDR

robabram:
* Rewrite enrollment status calculation code in Participant generator to be simpler and more complete.
* Added new unittests for new enrollment status code.
* Added 'non_participant_answer' field to module data (MITRE request for CATI data).
* Merged Consents fields into Modules data to begin depreciation of Consents data.

## Tests
- [] unit tests


rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,303 @@
...
+    _biobank_samples = None
+    _physical_measurements = None
+    _modules = None
+
+    # Timestamps for when each status was achieved, these are set in the self.save_calc() method.
+    registered_ts = None
+    participant_ts = None
+    participant_plus_ehr_ts = None
+    core_participant_minus_pm_ts = None
+    core_participant_ts = None
```
kenny-skaggs: These are all fields meant for specific instances, correct? Here they're class variables and the behavior might get a little confusing. Can they be moved to an `__init__` method?

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,303 @@
...
+                status = PDREnrollmentStatusEnum.CoreParticipantMinusPM
+            if status == PDREnrollmentStatusEnum.CoreParticipantMinusPM and \
+                    physical_measurements and \
+                    (cohort != ConsentCohortEnum.COHORT_3 or gror_consented):
+                status = PDREnrollmentStatusEnum.CoreParticipant
+
+            # Set the permanent enrollment status value if needed. Enrollment status can go down
+            # unless the enrollment status has reached 'CoreParticipant'.
+            # TODO: If participant reaches CoreParticipantMinusPM status, can the status level go back down?
+            if self.status != PDREnrollmentStatusEnum.CoreParticipant:
```
kenny-skaggs: Participants should also not lose CoreMinusPM status if they achieve that. https://pmi-engteam.slack.com/archives/CG2HWQ3E2/p1614297761080900?thread_ts=1614291316.071300&cid=CG2HWQ3E2

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,303 @@
...
+            if info.calculated is False and ev.event == ParticipantEventEnum.PhysicalMeasurements and \
+                    ev.finalized is not None:
+                info.calculated = True
+                info.first_ts = ev.last_ts = ev.timestamp
+                info.values.append(ev)
+                break
+
+        return self.save_calc('_physical_measurements', info)
+
+    def calc_modules(self, events):
```
kenny-skaggs: Since this specifically works with the baseline modules, and doesn't include things like COPE and FamilyHistory, can the name be more explicit?

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,303 @@
...
+            # Calculate enrollment status
+            status = PDREnrollmentStatusEnum.Unset
+            if signed_up:
+                status = PDREnrollmentStatusEnum.Registered
+            if status == PDREnrollmentStatusEnum.Registered and consented:
+                status = PDREnrollmentStatusEnum.Participant
+            if status == PDREnrollmentStatusEnum.Participant and ehr_consented:
+                status = PDREnrollmentStatusEnum.ParticipantPlusEHR
+            if status == PDREnrollmentStatusEnum.ParticipantPlusEHR and biobank_samples and \
+                    (modules and len(modules.values) >= 3) and \
```
kenny-skaggs: Can the list of baseline modules be loaded from the config and used to find the number expected? That way if it ever updates there's less of a risk of it not getting changed here.

rdr_service/resource/generators/participant.py
```diff
@@ -1005,13 +966,9 @@ def get_biobank_sample_activity(samples):
             for test in bbo_samples:
                 if test['dna_test'] == 1:
                     dna_tests += 1
-                    if test['confirmed']:
-                        dna_tests_confirmed += 1
                 # PDR-134:  Add baseline tests counts
-                if test['baseline_test'] == 1:
+                elif test['baseline_test'] == 1:
```
kenny-skaggs: Just wanted to confirm that a dna_test (like 2ED10) should not be counted as a dna test and a baseline test.

rdr_service/resource/generators/participant.py
```diff
@@ -1062,16 +1015,46 @@ def get_biobank_sample_activity(samples):
...
+                        sum([r['baseline_test'] for r in order['samples'] if r['confirmed'] is not None])
+                    order['isolate_dna_confirmed'] = \
+                        sum([r['dna_test'] for r in order['samples'] if r['confirmed'] is not None])
+                    # Find minimum confirmed date of DNA tests if we have any.
+                    if order['isolate_dna']:
+                        try:
+                            tmp_ts = min([r['confirmed'] for r in order['samples'] if r['confirmed'] is not None])
+                            act_key = ParticipantEventEnum.BiobankConfirmed
+                            act_ts = tmp_ts
+                        except ValueError:  # No confirmed timestamps in list.
```
kenny-skaggs, line 1051: Based on the comment I'd think this except would fire if there aren't any confirmed timestamps, but is it possible for some ordered and stored samples to be in the list, and then later in the list have orderless stored samples?

tests/resource_tests/calculator_tests/test_enrollment_status_calculator.py
```diff
@@ -0,0 +1,192 @@
...
+    {'timestamp': datetime(2018, 5, 17, 2, 11, 37), 'group': 'Biobank', 'group_id': 20,
+     'event': p_event.BiobankOrder, 'dna_tests': 0, 'basline_tests': 0},
+    {'timestamp': datetime(2018, 5, 21, 18, 9, 12), 'group': 'Profile', 'group_id': 1,
+     'event': p_event.PhysicalMeasurements, 'finalized': datetime(2018, 3, 8, 18, 9, 12)},
+    {'timestamp': datetime(2019, 6, 13, 0, 0), 'group': 'Profile', 'group_id': 1,
+     'event': p_event.EHRLastReceived}
+]
+
+
+class EnrollmentStatusCalculatorTest(BaseTestCase):
```
kenny-skaggs, line 47: Looks great that the tests don't depend on the database. If you include this then they can avoid any overhead of checking on any data that needs to be cleared, and it could make it encourage keeping future tests from depending on the db.

```
    def __init__(self, *args, **kwargs):
        super(EnrollmentStatusCalculatorTest, self).__init__(*args, **kwargs)
        self.uses_database = False
```

tests/resource_tests/calculator_tests/test_enrollment_status_calculator.py
```diff
@@ -0,0 +1,192 @@
...
+        self.assertEqual(self.esc.status, PDREnrollmentStatusEnum.CoreParticipant)
+        self.assertEqual(self.esc.cohort, ConsentCohortEnum.COHORT_3)
+
+    def test_cohort_3_gror_yes_no_answer(self):
+        """ Shift activity dates so we look like a cohort 3 participant with GROR consent with multiple answers. """
+        # Make a deep copy so we don't mess with the BASIC_ACTIVITY timestamp values.
+        activity = copy.deepcopy(BASIC_ACTIVITY)
+        activity.append(
+            {'timestamp': datetime(2018, 3, 6, 21, 24, 10), 'group': 'QuestionnaireModule', 'group_id': 40,
+             'event': p_event.GROR, 'ConsentAnswer': None, 'answer': 'CheckDNA_Yes',
```
kenny-skaggs: Why does the data structure have a `ConsentAnswer` field and an `answer` field? I tried to search and I found where `ConsentAnswer` is initialized to `None` but I couldn't see where it was used or populated.

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        :param activity: A list of activity dictionary objects created by the ParticipantSummaryGenerator.
+        """
+        self.activity = sorted([JSONObject(r) for r in activity if r['timestamp']], key=lambda i: i.timestamp)
+        # Work through activity by slicing to determine current enrollment status.
+        # This method allows us to iterate once through the data and still catch participants
+        # that might have been considered a Core Participant at one point, but would not by
+        # looking at their current state.
+        for x in range(1, len(self.activity)+1):
+            events = self.activity[0:x]
```
kenny-skaggs, line 81: I like the idea of stepping through the event timeline and checking the participant's status 👍

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: I'm not sure exactly what you mean by 'specific instances', but they are supposed to be class variables.  After the run() method is called each one of those may be populated with a timestamp.  Each of these variables is used, even if set to None, to populate the participant resource JSON.

kenny-skaggs: Can they be moved to instance variables rather than class variables? They work now since they can only be assigned to on the instances (shadowing the class variables here), but if any of them were defaulted to a mutable object (like a dictionary or a list) then that could lead to hard to track down errors.

rdr_service/resource/generators/participant.py
robabram: Good catch, I'll revert it.

rdr_service/resource/generators/participant.py
robabram: For the BiobankConfirmed event to be valid, there must be at least one sample with a confirmed date.  In the lower environments (IE: sandbox), there are lots of low quality stored sample records with out a 'confirmed' timestamp.  I wouldn't expect this try/except to hit at all in prod, even with orderless stored samples.

tests/resource_tests/calculator_tests/test_enrollment_status_calculator.py
robabram: I'll add that code, but this makes me think we should re-think adding any default database using mixins to the BastTestCase class. Probably an RDR 2.0 refactor.

tests/resource_tests/calculator_tests/test_enrollment_status_calculator.py
robabram: This was an artifact from all my refactoring the event code.  'ConsentAnswer' was added in 'generators/participant.py:607'.  I removed all references to it.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: Thanks, I think this simplifies the code greatly from what we had before in the generator.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: Done, see new commit.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: Done, see new commit.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: Done, see new commit.

kenny-skaggs: Sorry, I should have been clearer. What I meant with the comment here was that the config has the list of baseline modules expected, and right now there are 3 of them but that could change in the future. Rather than having 3 hardcoded here, can the number of baseline modules be loaded from the config?

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: For now, each of these timestamp class properties should be either a 'datetime' value or None.  If they were to be set to a collection type object in the future, then I think the class code would need refactoring and that would probably fix your concern.  Would you like me to create an '__init__()' method and set them explicitly to None so a new instance of the class defaults them to None (already the class default)?

kenny-skaggs: Is there a reason they're set up as class variables? What I'm imagining is that the `__init__` method could be created and they would be defined as instance variables there and their class variable definitions would be removed.

robabram: That is weird to me, class properties are also placeholders so developers know what properties are important on all instances of the class and also for auto code completion in an IDE.

robabram: I feel like you are talking about the issue in this code where the class property can affect new instances of the class.  
```
class A:
    prop = 0
    def run(self):
        self.prop += 1
    def class_run(self):
    	A.prop += 1
        
z = A()
z.run()
print(z.prop)
y = A()
y.run()
print(y.prop)
z.class_run()
n = A()
n.run()
print(n.prop)
-----
1
1
2
```

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: The new commit does that.  It loads the baseline modules from the config and creates a list of PDREnrollmentStatusEnum enums from it.  So if a new baseline module is added to the config it automatically gets added to the enrollment status calculation without requiring any code changes.

rdr_service/resource/calculators/participant_enrollment_status.py
kenny-skaggs: Yes, I am. With that example and the code under review it's not immediately apparent when the variable would be a class or instance level variable. Defining the instance variables in the `__init__` method is the generally accepted way of declaring fields for an object instance in python. And the Pycharm IDE makes autocompletion available for it too.

rdr_service/resource/calculators/participant_enrollment_status.py
kenny-skaggs: So the `>= 3` here wouldn't need to change if a 4th baseline module was added and needed to be completed for a participant to be considered CORE?

robabram: I missed the check with the hardcode '3', that is fixed now.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: I'm going to have to disagree with you, it may be a suggestion but it's definitely not a "best practice".  If it was, defining class properties would have be factored out of Python by now.  This way works fine and the results are exactly the same if I were to move them to the init method.  I am not aware of any code in the RDR that  modifies a class property by "class" name, every class property modification is done using an instance object of the class.  My personal preference is to put important properties on the class as class properties, rather than in the init method.  I think it is fine the way you do it and I would not ask you to change your ways to mirror mine. Lastly, you made me curious, so I looked through the SQLAlchemy code in the project virtual environment and apparently they are fine with defining class properties in their code.

kenny-skaggs: To clarify: I'm saying that defining instance variables in the `__init__` and not having them shadow class variables are best practices. I'm not saying there's anything wrong with class level variables, there are times that they're useful. But I'm not really understanding their purpose here. The code is using the names on a per-instance basis and only uses the class variables as a default, so there doesn't seem to be a functional purpose for having them defined on the class. And when looking for any important instance variables the init method is a known place to look (it's a best practice to place the __init__ at the top for that reason).

I agree that they work for what's being done with them now, since Python would search the class for attribute definitions if they weren't found on the instance. But mostly wanted to point this out as a design/maintainability issue since the defaults could only ever be immutable, and it's not "pythonic" so could cause unexpected issues with builtin functions for any future code that would want to use instances of this class.

robabram: The class properties I have defined are (maybe not all of them) used to get the results of the enrollment status calculation and also used in the unittests.   Since I know they are going to be accessed by code outside the class, I have set them as class properties so they always exist on every instance of the object and I have set them with the default values I want. So I don't have to explicitly set a default value or any value at all in the class init method.  The nature of the calculation code is to go as far as it can and set the class property values that were calculated and be done.  Not all class properties will be calculated in each run.   With your way, defining them as instance properties in the init method, would require the python interpreter to set defaults for all of the properties for every new instance of the class, instead of setting them once as the default for all instances of the class.  Sounds like a waste of CPU cycles to me.  If you are really hung up over this design pattern then we need to have a meeting between you, Sunny and me to discuss, because frankly I think you are now wasting my time discussing coding semantics that are irrelevant and I'm not getting work done.

rdr_service/resource/calculators/participant_enrollment_status.py
kenny-skaggs: I think there might be a way to speed this up. From what I'm understanding this iterates through subsets of participant activity events. So, for a rough example, if a participant has the following events: `Primary Consent, EHR, GROR, Biosamples`. The algorithm would start by looking just at the Primary Consent and figure out the status from that. Then it would look at the status after the Primary Consent and the EHR. Then it would consider the status after Primary Consent, EHR, and GROR.

Each time it analyzes a new subset of events, it calls all of the methods for calculating the components of status (`calc_signup`, `calc_consent`, `calc_ehr_consent`, ...). And each of those methods loops through all of the events in the subset to search for the ones that are of interest to that method.

I think this looping and searching could be avoided by stepping through each event at the root level once, maintaining the activity seen in the timeline so far on the instance of the calculator, and calculating the status after each event has been analyzed. And each method would only be called with the event that it's looking for, rather than having the methods look through the entire list. So the code in the run function would look something like this:
```
for event in self.activity:
    if event.event == ParticipantEventEnum.SignupTime:
        self.signed_up = self.calc_signup(event)
    ...
```

robabram: That is how the old enrollment status calculation code basically worked.  The issue there and possibly with your suggestion is we might have to add back in the additional code logic to see if a participant had reached core status at some point in the past, but wouldn't be if you look at their current data.  By working through the events in order regardless of event type, that removes the need for handling edge cases like participants that reached core status but then revoked their EHR consent later.  Another big plus is this version is much simpler to debug the logic than the older code.

kenny-skaggs: What I'm describing is a bit different than how the old code worked, since the old code only calculated the status after all activity events had been processed. What I'm suggesting essentially boils down to iterating through the list of events once, calling the just the calc method that is designed for a given event's type (that would take the for-loop out of the `calc_...` methods), and check the status after each event has been analyzed (like it's doing now when it checks the status after a list of events was analyzed). What I'm suggesting would have one for-loop that goes through the events one at a time and still has the aggregate information for the events that had been seen before. Where the current implementation starts from the beginning each time another event is added (recalculating the same state that was found in the last iteration).

This is what the top-level method could look like:
```
for event in self.activity:
    if event.event == ParticipantEventEnum.SignupTime:
        self.signed_up = self.calc_signup(event)
    elif ev.event == ParticipantEventEnum.ConsentPII:
        self.consented, cohort = self.calc_consent(ev)
    elif ev.event == ParticipantEventEnum.EHRConsentPII or ev.event == ParticipantEventEnum.DVEHRSharing:
        self.ehr_consented = self.calc_ehr_consent(ev)
    elif ev.event == ParticipantEventEnum.GROR:
        self.gror_consented = self.calc_gror_consent(ev)
    elif ev.event == ParticipantEventEnum.BiobankConfirmed and ev.dna_tests > 0:
        self.biobank_samples = self.calc_biobank_samples(ev)
    elif ev.event == ParticipantEventEnum.PhysicalMeasurements and ev.finalized is not None:
        self.physical_measurements = self.calc_physical_measurements(ev)
    elif ev.group_id == 40:
        self.modules = self.calc_baseline_modules(ev)
```

And each calc method would be simplified to only handle an event that is already known to be the correct type. This is what the `calc_signup` method could look like:
```
         info = EnrollmentStatusInfo()
         info.calculated = True
         info.first_ts = ev.last_ts = ev.timestamp
         info.add_value(ev)
         return self.save_calc('_signup', info)
```

This would still catch any participants that have had CORE but don't because the calculation would still happen after each event and `self.status` wouldn't be downgraded if it was one of the CORE statuses.

robabram: I'm not 100% sure this doesn't introduce opennings for edge cases to calculate wrong (I'll think more deeply about this on Monday).  The most obvious edge case is when there are many EHRConsent submissions with a No in the middle.  If it can be proven that it doesn't calculate incorrectly with edge case data, then I would be Ok with this change.

robabram: What I would like to avoid in this code is having to add new code specifically to cover edge cases.  That made the older code pretty ugly to debug.  Although it hurts me a little bit to do the loop with slicing (performance), it so far been shown to correctly calculate the enrollment status on the edge cases we have thrown at it.

pegbertsch: My take:  I think we could look for performance improvements based on iterating over the chronologically sorted activity list once.  I think of this like a state machine, where the combination of the current enrollment status and the next event in the activity list determine the next state, and only certain events have an impact on certain states.  E.g., if you're CORE_MINUS_PM, you only care if you see a physical measurements event and don't have to process anything else.  If you reach CORE, you've reached the "terminal" state.

It would probably help to draw a visual diagram of the states and possible transitions based on tracked activities.

I think this could be considered for a future PR and does not have to be included now.   Changing the logic later shouldn't have to trigger any data rebuilds since it's just a different way to get to the same result.

robabram: I agree with Peggy, I think the current code is in a good place where we can move forward and in the future we can iteratively improve it.

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        info = EnrollmentStatusInfo()
+        for ev in events:
+            if ev.event == ParticipantEventEnum.ConsentPII:
+                info.calculated = True
+                info.first_ts = ev.last_ts = ev.timestamp
+                info.add_value(ev)
+                break
+
+        # Calculate consent cohort group
```
pegbertsch, line 190: Just a reminder, we're trying to coalesce on PTSC's cohort assignments and deprecate calculations based on the cutoff dates.   ("TO DO")

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,321 @@
...
+        for ev in events:
+            if ev.event in [ParticipantEventEnum.EHRConsentPII, ParticipantEventEnum.DVEHRSharing]:
+                # See if we need to reset the info object.
+                if ev.answer in [CONSENT_PERMISSION_NO_CODE, DVEHRSHARING_CONSENT_CODE_NO, EHR_CONSENT_EXPIRED_YES,
+                                 DVEHRSHARING_CONSENT_CODE_NOT_SURE]:
+                    self._ehr_consented = None  # Reset any saved info.
+                    info = EnrollmentStatusInfo()
+                    continue
+                # See if we should set the consent info.
+                if info.calculated is False and \
```
pegbertsch, line 221: Could be an elif here?  (optional)

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,321 @@
...
+        info = EnrollmentStatusInfo()
+        for ev in events:
+            if ev.event == ParticipantEventEnum.GROR:
+                # See if we need to reset the info object.
+                if ev.answer in [CONSENT_GROR_NO_CODE, CONSENT_GROR_NOT_SURE]:
+                    self._gror_consented = None  # Reset any saved info.
+                    info = EnrollmentStatusInfo()
+                    continue
+                # See if we should set the consent info.
+                if info.calculated is False and ev.answer == CONSENT_GROR_YES_CODE:
```
pegbertsch, line 246: could be an elif? (optional)

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,321 @@
...
+        Determine if biobank has confirmed DNA test for participant.
+        Criteria:
+          - Physical Measurements have been received and finalized.
+        :param events: List of events
+        :return: EnrollmentStatusInfo object
+        """
+        info = EnrollmentStatusInfo()
+        for ev in events:
+            if info.calculated is False and ev.event == ParticipantEventEnum.PhysicalMeasurements and \
+                    ev.finalized is not None:
```
pegbertsch: Needs resolution:

Should we be using the status field to filter out CANCELLED measurements? I found 1 case in RDR that has a finalized timestamp but also has a cancelled_time and a status of CANCELLED

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,321 @@
...
+            for mod_ev in self._module_enums:
+                if ev.event == mod_ev:
+                    if ev.timestamp < info.first_ts:
+                        info.first_ts = ev.timestamp
+                    if ev.timestamp > info.last_ts:
+                        info.last_ts = ev.timestamp
+                    info.add_value(ev)
+
+        # If we have seen all the baseline modules, set calculated to True
+        if info.values is not None and len(info.values) == len(self._module_enums):
```
pegbertsch, line 326: Needs resolution:

Can there be cases where a participant submitted one of the modules multiple times, but has not submitted at least one of all the baseline modules?  I.e., length check could pass despite missing modules?

(Note:  I don't think we currently have any cases of this but unsure if it's possible)

rdr_service/resource/calculators/participant_enrollment_status.py
```diff
@@ -0,0 +1,321 @@
...
+    """
+    Calculate participant enrollment status.
+    """
+    status: PDREnrollmentStatusEnum = PDREnrollmentStatusEnum.Unset
+    # events = None  # List of EnrollmentStatusEvent objects.
+    activity = None  # List of activity created from Participant generator.
+
+    cohort = None
+
+    # First info object for each part used in calculating enrollment status.
```
pegbertsch, line 47: Needs resolution:

Do we need to track withdrawal status here?  I.e., is it possible to find a withdrawal (or suspension?) event before the biobank samples are confirmed such that it could prevent them from reaching core status if we stopped advancing their status after seeing the withdrawal?

Not sure if a deceased report or suspension could also impact advancing enrollment status.  Admittedly, a corner case here.

rdr_service/resource/generators/participant.py
```diff
@@ -1062,16 +1015,45 @@ def get_biobank_sample_activity(samples):
...
+
+                if order['samples']:
+                    order['baseline_tests_confirmed'] = \
+                        sum([r['baseline_test'] for r in order['samples'] if r['confirmed'] is not None])
+                    order['isolate_dna_confirmed'] = \
+                        sum([r['dna_test'] for r in order['samples'] if r['confirmed'] is not None])
+                    # Find minimum confirmed date of DNA tests if we have any.
+                    if order['isolate_dna']:
+                        try:
+                            tmp_ts = min([r['confirmed'] for r in order['samples'] if r['confirmed'] is not None])
```
pegbertsch, line 1048: Needs resolution:

I think it's pretty typical for all samples in an order to share timestamps like confirmed...but if a single biobank order can share both DNA and non-DNA sample types,  should this be:

min([r['confirmed'] for r in order['samples'] if ( r['confirmed'] is not None and r['dna_test']))

rdr_service/resource/schemas/participant.py
```diff
@@ -58,37 +58,43 @@ class ModuleStatusSchema(Schema):
...
     consent_value = fields.String(validate=validate.Length(max=80))
     consent_value_id = fields.Int32()
-    consent_module = fields.String(validate=validate.Length(max=80))
-    consent_module_authored = fields.DateTime()
-    consent_module_created = fields.DateTime()
     consent_expired = fields.String(validate=validate.Length(max=80))
-    consent_module_external_id = fields.String(validate=validate.Length(max=120))
-    consent_response_status = fields.EnumString(enum=QuestionnaireResponseStatus)
-    consent_response_status_id = fields.EnumInteger(enum=QuestionnaireResponseStatus)
+    non_participant_answer = fields.String(validate=validate.Length(max=60))
```
pegbertsch, line 66: We need to communicate any re-ogranization of fields for the BQ v_pdr_participant_module and v_pdr_participant_consent views especially to Scott;  I think he has queries that depend on those views that will break if we start removing fields from one and adding them to another (e.g.,have removed  consent_module, consent_module_authored, etc. here).

Should we leave those fields in here for now, until we get rid of the v_pdr_participant_consent view?

rdr_service/resource/schemas/participant.py
```diff
@@ -320,7 +332,8 @@ class ParticipantSchema(Schema):
     races = fields.Nested(RaceSchema, many=True)
     genders = fields.Nested(GenderSchema, many=True)
     modules = fields.Nested(ModuleStatusSchema, many=True)
-    consents = fields.Nested(ConsentSchema, many=True)
+    # TODO: Deprecated, but leave around until BigQuery table support is removed.
+    # consents = fields.Nested(ConsentSchema, many=True)
```
pegbertsch, line 336: Per comments above, are we ready to remove this yet or do we need to give users time to update the queries?

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: Understood.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: I'm not sure if a withdrawal would suspend calculating the enrollment status or not. We could easily add it in the future if needed though.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: My preference is to not use the `elif` here, but I think it could go either way.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: My preference is to not use the `elif` here, but I think it could go either way.

rdr_service/resource/calculators/participant_enrollment_status.py
robabram: Good catch, this does code does need to make sure we have one of each baseline module.  I'll make a new commit.

rdr_service/resource/schemas/participant.py
robabram: I want to leave it in for a short while, just in case we need to revert for some reason.

pegbertsch (approved):

kenny-skaggs (approved):

PR #2502, opened 2021-06-22
[DA-2058] Script for modifying consent validation records

kenny-skaggs:
## Resolves *[DA-2058](https://precisionmedicineinitiative.atlassian.net/browse/DA-2058)*
This updates the consent tool to allow for modifying consent validation records. It will take the id of the record and optional values to set (I've only added the ones I imagine we'll need for now). Then it'll list out what the update will be (to make sure it's the correct record and the changes are correct) before applying them.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/consent_dao.py
+++ b/rdr_service/dao/consent_dao.py
@@ -12,9 +12,9 @@ class ConsentDao(BaseDao):
     def __init__(self):
         super(ConsentDao, self).__init__(ConsentFile)
 
-    def get(self, obj_id):
+    def get(self, obj_id) -> ConsentFile:
         with self.session() as session:
-            return session.query(ConsentFile).filter(ConsentFile.id == obj_id).one()
+            return session.query(ConsentFile).filter(ConsentFile.id == obj_id).one_or_none()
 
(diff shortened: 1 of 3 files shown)


rdr_service/dao/consent_dao.py
```diff
@@ -13,6 +13,10 @@ class ConsentDao(BaseDao):
     def __init__(self):
         super(ConsentDao, self).__init__(ConsentFile)
 
+    def get(self, obj_id) -> ConsentFile:
+        with self.session() as session:
+            return session.query(ConsentFile).filter(ConsentFile.id == obj_id).one_or_none()
+
```
dtharpeuno: Doesn't the `get` method on the parent class provide this already?

dtharpeuno (approved):

rdr_service/dao/consent_dao.py
kenny-skaggs: Good catch! I'm not really sure why I did it this way. But you're right, BaseDao's `get` gives the exact same thing. I'll update it.

j-kanuch (approved):

PR #2516, opened 2021-06-30
[DA-2066] Catch null key values in lower enviornments.

robabram:
Update v_pdr_participant_module view.

## Resolves *[DA-2066](https://precisionmedicineinitiative.atlassian.net/browse/DA-2066)*

Minor fixes for issues discovered from testing  in Stable.
* Check for BQ schema data dict has null key values (lower environment problem)
* Add new fields to BQ 'v_pdr_participant_module' view.

### DIFF
--- a/rdr_service/model/bq_base.py
+++ b/rdr_service/model/bq_base.py
@@ -8,6 +8,7 @@ import importlib
 import inspect
 import itertools
 import json
+import logging
 import operator
 import re
 from enum import Enum, EnumMeta
@@ -567,6 +568,9 @@ class BQRecord(object):
             :return: dict
             """
             for key, val in src.items():
+                if not isinstance(key, str):
+                    logging.warning('Dict key is not a string.')
+                    continue
                 # validate key against schema if needed
                 if schema and not getattr(schema, key, None):
                     # raise KeyError('{0} key not in schema'.format(key))
--- a/rdr_service/model/bq_pdr_participant_summary.py
+++ b/rdr_service/model/bq_pdr_participant_summary.py
@@ -325,7 +325,8 @@ class BQPDRModuleView(BQView):
            nt.mod_module, nt.mod_baseline_module,
            CAST(nt.mod_authored AS DATETIME) as mod_authored, CAST(nt.mod_created AS DATETIME) as mod_created,
            nt.mod_language, nt.mod_status, nt.mod_status_id, nt.mod_response_status, nt.mod_response_status_id,
-           nt.mod_external_id
+           nt.mod_external_id, nt.mod_questionnaire_response_id, nt.mod_consent, nt.mod_consent_value,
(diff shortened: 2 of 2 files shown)


dtharpeuno (approved):

rdr_service/model/bq_base.py
```diff
@@ -567,6 +568,9 @@ def update(dest, src, schema):
             :return: dict
             """
             for key, val in src.items():
+                if not isinstance(key, str):
+                    logging.warning('Dict key is not a string.')
```
pegbertsch, line 572: How noisy will this be based on what you saw on stable?  Or, you don't expect it to really happen on prod so no risk of high volume of warnings?

Edit:  per offline conversation, this warning is likely to be limited to lower environments due to some lower quality test data.  Not expecting a flood of warnings in prod.  Resolving.

rdr_service/model/bq_pdr_participant_summary.py
```diff
@@ -325,7 +325,8 @@ class BQPDRModuleView(BQView):
            nt.mod_module, nt.mod_baseline_module,
            CAST(nt.mod_authored AS DATETIME) as mod_authored, CAST(nt.mod_created AS DATETIME) as mod_created,
            nt.mod_language, nt.mod_status, nt.mod_status_id, nt.mod_response_status, nt.mod_response_status_id,
-           nt.mod_external_id
+           nt.mod_external_id, nt.mod_questionnaire_response_id, nt.mod_consent, nt.mod_consent_value,
```
pegbertsch, line 328: Thanks for catching this.  Are we sure this is the only managed view where the SELECT statement is explicitly populated vs. "auto-generated" field list in the SELECT?  Since we're nuking all tables / views including genomics, etc.?

Edit:  I think this is the only view schema that needs updating in the model, and/or I'm fairly confident we can resolve issues by re-running migrations from a local branch without having to cut a release to fix.   Closing this conversation.

rdr_service/model/bq_base.py
kenny-skaggs, line 572: Just in case this is triggered on Prod, should there be something in the log statement about what is being set and what it's being set on (to help track down what happened if it does get printed in Prod)?

rdr_service/model/bq_pdr_participant_summary.py
robabram: I don't expect this to hit in prod, I think it's just some low quality questionnaire response data in the lower environments causing this to hit.  I don't think there will be a lot of hits in the lower environments.

rdr_service/model/bq_base.py
robabram: I don't expect this to hit in prod, I think it's just some low quality questionnaire response data in the lower environments causing this to hit. I don't think there will be a lot of hits in the lower environments.

pegbertsch (approved):

PR #2524, opened 2021-07-15
[DA-2081] MIGRATION - Refactor WGS and Array AW2-Data Reconciliation process

j-kanuch:
This PR refactors the genomic data file reconciliation process for both the WGS and Array workflows. The previous implementation would request the files in the genome centers' data buckets and then update the `genomic_gc_validation_metrics` record when a file was found for a particular record. The new implementation changes this process in the following ways:
- When files are uploaded to the data buckets, a pub/sub message is sent to the `genomic_data_files_upload` topic
- This triggers a Google Cloud Function: `genomic_ingest_data_files_function`
- The cloud function then executes a Google Cloud Task via the `IngestDataFilesTaskApi`, sending the `file_path` and `bucket_name` in the payload.
- The task uses the `GenomicJobController` to execute a new method: `accession_data_files()`
- `accession_data_files()` will write the file metadata to a new table `genomic_gc_data_file`.

The reconciliation process is still executed via the two existing cron jobs for Array and WGS, respectively: 
- `/offline/GenomicArrayReconciliationWorkflow`
- `/offline/GenomicWGSReconciliationWorkflow`

The cron jobs have changed in that they execute a new method on the `GenomicReconciler` component: `reconcile_metrics_to_data_files`. This method uses the `genomic_gc_data_file` table to look up whether a specified data file exists.  The file attributes and mappings were moved to the `genomic_mappings` module for better readability.

Unit tests, models, and a new DAO were updated or created to support the above changes.

## Tests
- [x] unit tests


rdr_service/api/genomic_cloud_tasks_api.py
```diff
@@ -256,11 +256,14 @@ class IngestDataFilesTaskApi(BaseGenomicTaskApi):
     """
     def post(self):
         super(IngestDataFilesTaskApi, self).post()
+        logging.info(f'self data: {self.data}')
```
dtharpeuno: Do we need two logs here?

rdr_service/gcloud_functions/genomic_ingest_data_files_function/main.py
```diff
@@ -44,16 +44,43 @@ def run(self):
...
+            "vcf.gz.md5sum",  # raw/regular vcf md5
+            "hard-filtered.vcf.gz",   # hard-filtered vcf
+            "hard-filtered.vcf.gz.tbi",   # hard-filtered vcf index
+            "hard-filtered.vcf.gz.md5sum",   # hard-filtered vcf md5
+            "cram",   # cram
+            "cram.md5sum",   # cram md5
+            "cram.crai",    # cram index
+            'hard-filtered.gvcf.gz',  # gvcf
+            'hard-filtered.gvcf.gz.md5sum'  # gvcf md5
+        ]
```
dtharpeuno, line 79: Nice!

rdr_service/gcloud_functions/genomic_ingest_data_files_function/main.py
```diff
@@ -44,16 +44,43 @@ def run(self):
...
+
+        file_path = self.event.attributes.bucketId + "/" + self.event.attributes.objectId.replace(" ", "")
```
dtharpeuno, line 81: Do we need to edit the file path here, since we are still splitting the same way?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1823,243 +1823,146 @@ def __init__(self, run_id, job_id, archive_folder=None, file_mover=None,
...
+                            setattr(_obj, file_type_config['file_received_attribute'], 1)  # received
+                            setattr(_obj, file_type_config['file_path_attribute'], f'gs://{file.file_path}')
 
-        self.file_list = [f.name for f in files]
+                            metric_touched = True
 
-        total_missing_data = []
-        metric_touched = False
+            # if missing files
+            # missing_data_files.append(default_filename)
```
dtharpeuno: Think we can delete this.

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -577,10 +649,8 @@ def run_reconciliation_to_data(self, *, genome_type):
                     gc_site_id = site_id_mapping['northwest']
 
                 # Run the reconciliation by GC
-                if genome_type == 'array':
-                    self.job_result = self.reconciler.reconcile_metrics_to_array_data(_gc_site_id=gc_site_id)
-                elif genome_type == 'wgs':
-                    self.job_result = self.reconciler.reconcile_metrics_to_wgs_data(_gc_site_id=gc_site_id)
+                self.job_result = self.reconciler.reconcile_metrics_to_data_files(genome_type,
+                                                                                  _gc_site_id=gc_site_id)
```
dtharpeuno, line 666: Better!

rdr_service/genomic/genomic_mappings.py
```diff
@@ -48,3 +48,175 @@
     "processing_status": "processingStatus",
     "notes": "notes",
 }
+
+genomic_data_file_mappings = {
```
dtharpeuno, line 52: 🔥

dtharpeuno (approved):
This is great. NIce job!
Left a few questions.

rdr_service/api/genomic_cloud_tasks_api.py
j-kanuch: No, this was for debugging. I will remove.

rdr_service/gcloud_functions/genomic_ingest_data_files_function/main.py
j-kanuch: I needed to add the bucket here so that the full path is included in the DB. I found out late yesterday that `self.event.attributes.objectId` doesn't have the bucket name. Because of this, the full file paths were not being written correctly to the `genomic_gc_validation_metrics` table.

dtharpeuno: Do we need to backfill the table? Was that causing the issue with the files ingestion on not being able to find the metrics record, if so my bad!

dtharpeuno: Sorry didn't scroll down I see the full file path was being sent in the payload.

j-kanuch: The new table will need to be back-filled. This is something I'll work on while waiting for this to get on production. The missing metrics is a separate issue that is either because the GC uploads the data file before giving us the AW2 manifest, or we haven't ingested the manifest yet. Having the files stored in a table should help mitigate the missing metrics issue, though I don't think it will go away entirely. Either way, the reconciliation will continue to attempt to reconcile them until they are found.

rdr_service/genomic/genomic_job_components.py
j-kanuch: Good eye, I will remove.

pegbertsch (commented):
@j-kanuch So this is an RDR-only table that  enables how the other existing tables that also get synced to PDR are populated?

j-kanuch:
> @j-kanuch So this is an RDR-only table that enables how the other existing tables that also get synced to PDR are populated?



> @j-kanuch So this is an RDR-only table that enables how the other existing tables that also get synced to PDR are populated?



> @j-kanuch So this is an RDR-only table that enables how the other existing tables that also get synced to PDR are populated?

@pegbertsch Correct. This table records the files in the Google Cloud Buckets so that our process doesn't need to call the cloud storage API during the reconciliation. The end product is that the `genomic_gc_validation_metrics` fields for the files received and paths are populated from the new table.

kenny-skaggs (approved):

PR #2526, opened 2021-07-16
Consent changes and optimization

kenny-skaggs:
## Resolves *no ticket*
This addresses some local issues with the consent scripts and also handles a few memory issues that have been seen in the server.

- When creating a session locally, the default is to connect to the local database and there doesn't seem to be a way to globally change that. So this passes a session through the consent code to make sure it's working against the server.
- This adds ValidationOutputStrategies to make code for finding validation results and then what to do with that data more 'plug-and-play'.
- Some more refactoring was done in the consent file module to help make it easier for when we implement CE's validation.
- There was a memory issue when syncing files, so this loads only the ones that we know will sync (there are many more that won't be synced).

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/consent_dao.py
+++ b/rdr_service/dao/consent_dao.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import List
+from typing import Collection
 
 from sqlalchemy import or_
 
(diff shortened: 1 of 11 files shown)


rdr_service/services/consent/files.py
```diff
@@ -28,76 +28,183 @@ def __init__(self, participant_id: int, storage_provider: GoogleCloudStorageProv
...
+            for blob_wrapper in self.consent_blobs
+            if self._is_ehr_consent(blob_wrapper)
+        ]
+
+    def get_gror_consents(self) -> List['GrorConsentFile']:
+        return [
+            self._build_gror_consent(blob_wrapper)
+            for blob_wrapper in self.consent_blobs
+            if self._is_gror_consent(blob_wrapper)
+        ]
```
dtharpeuno, line 76: NIce!

dtharpeuno (approved):

rdr_service/dao/consent_dao.py
```diff
@@ -13,65 +14,73 @@ class ConsentDao(BaseDao):
...
+        ).join(
+            Participant,
+            Participant.participantId == ParticipantSummary.participantId
+        ).filter(
+            ParticipantSummary.participantOrigin == 'vibrent',
+            Participant.isGhostId.isnot(True),
+            Participant.isTestParticipant.isnot(True),
+            or_(
+                ParticipantSummary.email.is_(None),
+                ParticipantSummary.email.notlike('%@example.com')
```
pegbertsch, line 31: If you're going to explicitly check for test email, should you also explicitly check for existing pairing to TEST hpo?
Not sure what the overlap is but a raw count of participant table records where `hpo_id = 21 `and `is_test_participant` is not true is >1K

rdr_service/dao/consent_dao.py
```diff
@@ -13,65 +14,73 @@ class ConsentDao(BaseDao):
...
+            )
+        )
+        if end_date is None:
+            query = query.filter(
                 or_(
-                    ParticipantSummary.email.is_(None),
-                    ParticipantSummary.email.notlike('%@example.com')
+                    ParticipantSummary.consentForStudyEnrollmentFirstYesAuthored >= start_date,
```
pegbertsch, line 37: If there's a bounded date range, I'm not sure I understand why we're limiting primary consent checks to FirstYes by that end date.    Does this still pick up validations for primary consent updates for Cohort 1?

rdr_service/services/consent/files.py
```diff
@@ -28,76 +28,183 @@ def __init__(self, participant_id: int, storage_provider: GoogleCloudStorageProv
...
     def _get_source_bucket(self) -> str:
         ...
 
+    @abstractmethod
+    def _get_source_prefix(self) -> str:
+        ...
+
 
 class VibrentConsentFactory(ConsentFileAbstractFactory):
     CABOR_TEXT = (
```
pegbertsch, line 120: For future consideration: I think we may want to "share" these text identifers between Vibrent/CE ConsentFactory;  I think they both have to be using the same IRB-approved text.

rdr_service/services/consent/validation.py
```diff
@@ -13,6 +15,171 @@
...
+        if verbose:
+            output_line.write(f'{str(file.id).ljust(8)} - ')
+        output_line.write(f'P{file.participant_id} - {str(file.type).ljust(10)} ')
+
+        if not file.file_exists:
+            output_line.write('missing file')
+        else:
+            errors_with_file = []
+            if not file.is_signature_valid:
+                errors_with_file.append('invalid signature')
```
pegbertsch, line 153: For future consideration:
I think we should be defining a dict object with all the specific errors we know how to detect so they are centralized in one place

pegbertsch (approved):
Approving but still pretty new to this code.  I think it could really help to add docstrings to the class and method definitions to document their purpose.

rdr_service/dao/consent_dao.py
kenny-skaggs: I think I had specifically left the Test HPO out just in case they were real participants that had been inadvertently paired to Test. But the same could happen for ghost or test too. I think it would be good to revisit this part when discovering participants that need consents copied for reasons other than authored date (such as changing orgs) if that's ok with you.

rdr_service/dao/consent_dao.py
kenny-skaggs: No, not at the moment. Validation for consent updates for cohort 1 still needs to be implemented.

rdr_service/services/consent/files.py
kenny-skaggs: I think that's a great idea. There may be some other functionality that could me moved into the base class too when we work on the CE factory.

rdr_service/services/consent/validation.py
kenny-skaggs: Yeah. I also liked some of the ideas we were discussing about not having these as hardcoded strings. We'll definitely want to revisit this.

PR #2533, opened 2021-07-22
[DA-2073] exclude_under_18_from_etl

wangy70:
## Resolves *[DA-2073](https://precisionmedicineinitiative.atlassian.net/browse/DA-2073)*
Filter the curation ETL export to exclude participants that were under 18 at the time of consent. 

Added a filter to the ETL SQL.

## Tests
[x] test cases

### DIFF
--- a/rdr_service/tools/tool_libs/curation.py
+++ b/rdr_service/tools/tool_libs/curation.py
@@ -2,15 +2,15 @@
 #
 # Template for RDR tool python program.
 #
-
 import logging
-from sqlalchemy import and_, case, insert, or_
+from sqlalchemy import and_, case, insert, or_, text
 from sqlalchemy.orm import aliased
 from sqlalchemy.sql import func
 from sqlalchemy.sql.expression import literal_column
 from sqlalchemy.sql.functions import coalesce, concat
 from typing import Type
 
+from rdr_service.clock import CLOCK
 from rdr_service import config
 from rdr_service.code_constants import PPI_SYSTEM, CONSENT_FOR_STUDY_ENROLLMENT_MODULE,\
     EMPLOYMENT_ZIPCODE_QUESTION_CODE, STREET_ADDRESS_QUESTION_CODE, STREET_ADDRESS2_QUESTION_CODE, ZIPCODE_QUESTION_CODE
@@ -377,13 +377,9 @@ class CurationExportClass(ToolBase):
             ),
             QuestionnaireResponse.status != QuestionnaireResponseStatus.IN_PROGRESS,
             QuestionnaireResponse.isDuplicate.is_(False),
-            or_(
-                ParticipantSummary.dateOfBirth.is_(None),
-                and_(
-                    ParticipantSummary.dateOfBirth.isnot(None),
-                    func.year(func.from_days(func.datediff(ParticipantSummary.consentForStudyEnrollmentAuthored,
-                                                           ParticipantSummary.dateOfBirth))) >= 18
-                )
+            and_(
+                ParticipantSummary.dateOfBirth.isnot(None),
+                func.timestampdiff(text('YEAR'), ParticipantSummary.dateOfBirth, CLOCK.now()) >= 18
             )
         )
 
--- a/rdr_service/tools/tool_libs/curation.py
+++ b/rdr_service/tools/tool_libs/curation.py
@@ -4,6 +4,7 @@
 #
 
 import logging
+import datetime
 from sqlalchemy import and_, case, insert, or_
 from sqlalchemy.orm import aliased
 from sqlalchemy.sql import func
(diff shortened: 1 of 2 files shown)


rdr_service/tools/tool_libs/curation.py
```diff
@@ -372,13 +376,21 @@ def _get_base_src_clean_answers_select(session):
                 QuestionnaireResponseAnswer.valueString.isnot(None)
             ),
             QuestionnaireResponse.status != QuestionnaireResponseStatus.IN_PROGRESS,
-            QuestionnaireResponse.isDuplicate.is_(False)
+            QuestionnaireResponse.isDuplicate.is_(False),
+            or_(
+                ParticipantSummary.dateOfBirth.is_(None),
+                and_(
+                    ParticipantSummary.dateOfBirth.isnot(None),
+                    func.year(func.from_days(func.datediff(ParticipantSummary.consentForStudyEnrollmentAuthored,
```
pegbertsch: Two functional questions:

1.  Is this the vetted algorithm for accurate age calculation for SQLAlchemy?  It's been a challenge figuring out what algorithm works best in different dialects of SQL (e.g., over on PDR) such that it accounts for leap years, etc.  If there's a justification/discussion online somewhere for doing age calculations in SQLAlchemy this way, it might be worth including a comment with a link to that.

2. Do we need to be looking at consentForStudyEnrollmentFirstYesAuthored to make sure they weren't originally marked as consented while still a minor?  Or do we only care if their most recent primary consent is "legal"?

dtharpeuno (approved):
Nice. Should we clean the `src_clean` db for the ones that have been saved there?

wangy70:
> Nice. Should we clean the `src_clean` db for the ones that have been saved there?

I think we can truncate it in the end of the ETL.

kenny-skaggs:
> > Nice. Should we clean the `src_clean` db for the ones that have been saved there?
> 
> I think we can truncate it in the end of the ETL.

The CDM database (where the table lives) is dropped and created from scratch each run of the ETL.

rdr_service/tools/tool_libs/curation.py
wangy70: Thanks for the comments :)
1. The key for this is https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_from-days
2. This is a good question, I will ask the requester about this.

rdr_service/tools/tool_libs/curation.py
```diff
@@ -372,13 +376,21 @@ def _get_base_src_clean_answers_select(session):
                 QuestionnaireResponseAnswer.valueString.isnot(None)
             ),
             QuestionnaireResponse.status != QuestionnaireResponseStatus.IN_PROGRESS,
-            QuestionnaireResponse.isDuplicate.is_(False)
+            QuestionnaireResponse.isDuplicate.is_(False),
+            or_(
+                ParticipantSummary.dateOfBirth.is_(None),
```
kenny-skaggs: Based on a communication to the curation team, we shouldn't be sending through participants that don't have a DOB entered https://precisionmedicineinitiative.atlassian.net/browse/ROC-1034?focusedCommentId=89587. Sorry that didn't make it all the way to the DA 😬

rdr_service/tools/tool_libs/curation.py
dtharpeuno: @wangy70 for #2 the original issue was flagged by curation as the actual DOB being < 18yrs, there was no correlation of consentForStudyEnrollmentFirstYesAuthored in the dataset.

wangy70: Maybe I can directly use TIMESTAMPDIFF(YEAR, date_of_birth,consent_date), let me try.

pegbertsch: @wangy70 It sounds from the Jira conversation as if this calculation should be based on the now() date (current age) at the time the code is run  rather than the age at consent.   But maybe I'm still not understanding if they also want to flag /exclude people who were too young to consent at the time they authored their consent.

dtharpeuno: Also if it helps, curation was using `where date_diff(DATE '2021-04-01', DATE(birth_datetime), YEAR) < 18`, guess the report was run in May for the April data.

wangy70: Thanks @dtharpeuno ! It's much clear now, I just updated the ticket description to reflect this. I will update the code.

wangy70: Done, please review.

rdr_service/tools/tool_libs/curation.py
wangy70: Done!

pegbertsch (approved):
Based on the offline conversation, I think the updated calculation is correct.  (age as of now >= 18). Would like @kenny-skaggs or @dtharpeuno to give this change a second thumbs up.

kenny-skaggs (approved):

PR #2536, opened 2021-07-23
[PDR-367] Implement retention eligible metrics data resource.

robabram:
## Implement Scott's retention eligible metrics data resource.
* Added code in fake participant generator to create retention eligible metric records.
* Defined resource schema, created generator
* Updated base generator code to remove non-schema fields from final resource.
* Created background task API for rebuilding batches of records
* Misc fixes

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/retention_eligible_metrics_dao.py
+++ b/rdr_service/dao/retention_eligible_metrics_dao.py
@@ -41,12 +41,11 @@ class RetentionEligibleMetricsDao(UpdatableDao):
         else:
             task = GCPCloudTask()
             params = {'batch': pids}
-            task.execute('batch_rebuild_retention_eligible_task', queue='resource-tasks', payload=params, in_seconds=15)
+            task.execute('batch_rebuild_retention_eligible_task', queue='resource-tasks', payload=params,
+                         in_seconds=15)
 
     def upsert_all_with_session(self, session, retention_eligible_metrics_records):
-        upsert_count = 0
-        rebuild_queue = list()
-
+        update_queue = list()
         for record in retention_eligible_metrics_records:
             dup_id, need_update = self._find_dup_with_session(session, record)
             if dup_id and not need_update:
@@ -54,15 +53,10 @@ class RetentionEligibleMetricsDao(UpdatableDao):
             elif dup_id and need_update:
                 record.id = dup_id
             session.merge(record)
-            upsert_count = upsert_count + 1
             # Batch up records for resource rebuilding.
-            rebuild_queue.append(int(record.participantId))
-            if len(rebuild_queue) >= 1000:
-                session.commit()
-                self._submit_rebuild_task(rebuild_queue)
-                rebuild_queue.clear()
-        # Submit any left over records for rebuilding.
-        if rebuild_queue:
-            session.commit()
-            self._submit_rebuild_task(rebuild_queue)
-        return upsert_count
+            update_queue.append(int(record.participantId))
+
+        session.commit()
+        if update_queue:
+            self._submit_rebuild_task(update_queue)
+        return len(update_queue)
--- a/rdr_service/dao/retention_eligible_metrics_dao.py
+++ b/rdr_service/dao/retention_eligible_metrics_dao.py
@@ -1,5 +1,8 @@
+from rdr_service.config import GAE_PROJECT
 from rdr_service.dao.base_dao import UpdatableDao
 from rdr_service.model.retention_eligible_metrics import RetentionEligibleMetrics
+from rdr_service.cloud_utils.gcp_cloud_tasks import GCPCloudTask
+from rdr_service.resource import generators
 
 
 class RetentionEligibleMetricsDao(UpdatableDao):
@@ -24,8 +27,26 @@ class RetentionEligibleMetricsDao(UpdatableDao):
         else:
             return None, None
 
+    def _submit_rebuild_task(self, pids):
+        """
+        Rebuild Retention Eligible Metrics resource records
+        :param pids: List of participant ids.
+        """
+        # Rebuild participant for BigQuery
+        if GAE_PROJECT == 'localhost':
+            res_gen = generators.RetentionEligibleMetricGenerator()
+            for pid in pids:
+                res = res_gen.make_resource(pid)
+                res.save()
+        else:
+            task = GCPCloudTask()
+            params = {'batch': pids}
+            task.execute('batch_rebuild_retention_eligible_task', queue='resource-tasks', payload=params, in_seconds=15)
+
     def upsert_all_with_session(self, session, retention_eligible_metrics_records):
         upsert_count = 0
+        rebuild_queue = list()
+
         for record in retention_eligible_metrics_records:
             dup_id, need_update = self._find_dup_with_session(session, record)
             if dup_id and not need_update:
@@ -34,4 +55,14 @@ class RetentionEligibleMetricsDao(UpdatableDao):
                 record.id = dup_id
             session.merge(record)
             upsert_count = upsert_count + 1
+            # Batch up records for resource rebuilding.
+            rebuild_queue.append(int(record.participantId))
+            if len(rebuild_queue) >= 1000:
+                session.commit()
+                self._submit_rebuild_task(rebuild_queue)
+                rebuild_queue.clear()
+        # Submit any left over records for rebuilding.
+        if rebuild_queue:
+            session.commit()
+            self._submit_rebuild_task(rebuild_queue)
         return upsert_count
--- a/rdr_service/resource/generators/_base.py
+++ b/rdr_service/resource/generators/_base.py
@@ -2,6 +2,7 @@
 # This file is subject to the terms and conditions defined in the
 # file 'LICENSE', which is part of this source code package.
 #
+import copy
 from dateutil import parser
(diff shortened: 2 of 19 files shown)


rdr_service/resource/generators/retention_metrics.py
```diff
@@ -0,0 +1,49 @@
...
+                logging.error(msg)
+                raise NotFound(msg)
+
+            # Populate Enum fields. Note: When Enums have a possible zero value, explicitly check for None.
+            if data['retention_eligible_status'] is not None:
+                data['retention_eligible_status'] = data['retention_eligible_status_id'] = \
+                       RetentionStatusEnum(data['retention_eligible_status'])
+            if data['retention_type'] is not None:
+                data['retention_type'] = data['retention_type_id'] = RetentionTypeEnum(data['retention_type'])
+            data['participant_id'] = 'P' + str(data['participant_id'])
```
kenny-skaggs, line 47: Should `retention_eligible_time` be set here as well?

rdr_service/dao/retention_eligible_metrics_dao.py
```diff
@@ -34,4 +55,14 @@ def upsert_all_with_session(self, session, retention_eligible_metrics_records):
...
+            # Batch up records for resource rebuilding.
+            rebuild_queue.append(int(record.participantId))
+            if len(rebuild_queue) >= 1000:
+                session.commit()
+                self._submit_rebuild_task(rebuild_queue)
+                rebuild_queue.clear()
+        # Submit any left over records for rebuilding.
+        if rebuild_queue:
+            session.commit()
+            self._submit_rebuild_task(rebuild_queue)
```
kenny-skaggs: Batching for storing the retention eligible already happens where the DAO code is called (https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/offline/retention_eligible_import.py#L35) and uses the same batch size. Would it potentially make sense to rebuild the batches there instead of in the DAO code?

That change would also align more with the biobank_samples_pipeline, which rebuilds outside of the DAO code rather than in it (https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/offline/biobank_samples_pipeline.py#L94)

rdr_service/resource/generators/retention_metrics.py
robabram: Date and Datetime values are automatically transformed to a string in the lower resource serialization code.  Code that transforms something here is because the underlying data does not exactly match the resource schema or there needs to be an explicit transformation, like the participant id integer to a string with a 'P' prefix.

kenny-skaggs: Oh, ok. Sorry for my misunderstanding.

rdr_service/tools/tool_libs/ptsc_retention_metrics.py
```diff
@@ -0,0 +1,71 @@
...
+        Main program process
+        :return: Exit code value
+        """
+        self.gcp_env.activate_sql_proxy()
+
+        # TODO: Glob the bucket and get the latest file or get file date from command line arg.
+        task_payload = {
+            'bucket': 'ptsc-metrics-all-of-us-rdr-prod',
+            'file_path': 'ptsc-metrics-all-of-us-rdr-prod/2021-07-26_retention_and_ubr_data.csv',
+            'upload_date': '2021-07-26 11:24:21'
```
kenny-skaggs, line 41: Can this find the latest file, or be left blank so that we know to fill it in with the latest file?

rdr_service/storage.py
```diff
@@ -370,3 +372,102 @@ def get_storage_provider():
...
+            provider = GoogleCloudStorageProvider()
+        with provider.open(cloud_csv_file, 'rt') as csv_file:
+            chunk_size = 1000 * 1024 * 10  # 10MB chunk.
+            while True:
+                chunk = csv_file.read(chunk_size)
+                tmp_file.write(chunk.encode('utf-8') if isinstance(chunk, str) else chunk)
+                if not chunk or len(chunk) < chunk_size:
+                    break
+        tmp_file.seek(0)
+        return tmp_file
```
kenny-skaggs, line 413: suggestion: what do you think of using the Blob's `download_to_file` function and use Google's implementation for getting a file downloaded to tmp?

rdr_service/dao/retention_eligible_metrics_dao.py
robabram: It's better leaving the rebuild there. The code there knows which records are going to be inserted/updated, so we don't have to rebuild all records.

rdr_service/storage.py
```diff
@@ -370,3 +372,102 @@ def get_storage_provider():
...
+            raise ValueError('Invalid cloud storage csv file name.')
+
+        self.cloud_csv_file = cloud_csv_file
+        self.column_names = column_names
+        self._delimiter = delimiter
+        self._tmp_file = None  # Temp file object handle.
+        self._strio = None  # StringIO object
+        self._header = None  # delimited string of column names.
+        self._reader = None  # csv.DictReader object.
+        self._batch_size = 1000  # number of rows from temp file to read.
```
kenny-skaggs, line 394: small concern: It still looks like a DictReader goes line by line on it's own without our intervention. This seems like it would increase memory compared to that, since we're reading 1000 lines into memory at once.

rdr_service/tools/tool_libs/ptsc_retention_metrics.py
robabram: I'm leaving the task of filling out features in this tool for later.

rdr_service/storage.py
robabram: I'm familiar with downloading the files this way, in the interest of time I want to go with something I know works correctly.

rdr_service/tools/tool_libs/ptsc_retention_metrics.py
kenny-skaggs: Ok, that's understandable. Is there any chance you can leave it blank then? Just to be sure we don't keep using the file from the 26th?

pegbertsch, line 41: Is this intended to still be hardcoded for now?

rdr_service/resource/generators/_base.py
```diff
@@ -66,6 +70,23 @@ def _load_data(self, schema, data):
...
+                except AttributeError:
+                    pass
+
+        # Remove any extra fields that are not declared in the schema.
+        schema_fields = schema._declared_fields.keys()
+        data_fields = data.keys()
+        remove = list()
+        for field in data_fields:
+            if field not in schema_fields:
+                remove.append(field)
```
pegbertsch, line 86: Is there a reason you have to build a remove list and loop through it again to pop the fields, instead of doing the data.pop(field) as you find them in the first loop?  (Not a functional issue and not a big deal for small lists, just wondering)

rdr_service/storage.py
robabram: We are never "comprehending" the entire 1000 records from the DictReader at one time in the class.  If the DictReader does what you said the other day, it should only be serving up one record at a time in this code.   But in a worse case scenario, the DictReader can only load 1,000 records into memory at any given time inside this class.

pegbertsch (commented):
See comments.  Will approve so long as the (temporary?) hardcoding of the file location info in the tool code was intentional and there is no concern about exposing the bucket name rather than grabbing from a config file.

I believe we should address finding the most recent file and/or allowing the user to input the file path as a tool parameter as soon as possible.

rdr_service/tools/tool_libs/ptsc_retention_metrics.py
robabram: Yes, for now.  It can be updated in the future to grab the latest bucket file or take one from a command line arg.

pegbertsch (approved):
Approved based on follow-up to Kenny on the hardcoded filepath question

rdr_service/resource/generators/_base.py
robabram: Removing records from the same list your iterating is a bad idea and in this specific case throws an error.

rdr_service/storage.py
kenny-skaggs: The DictReady would function in the way that it's been implemented and optimized. But I was specifically referring to this class. This implementation would hit the worst case on its own in the `_read_batch_from_file` by reading 1000 lines into memory at once.

rdr_service/tools/tool_libs/ptsc_retention_metrics.py
robabram: I changed my mind and removed the bucket and blob name.

pegbertsch (approved):

rdr_service/storage.py
robabram: In the StringIO object yes, but that is a byte array.  So lets say the average row is 50 bytes long in the csv file. The memory the StringIO would use can be calculated at 50 * 1,000.  So 50k + object overhead, that's not a lot of memory used there. If that is the worst case, I would take that any day.  What I am trying to work towards here is not to use the bare minimum memory, but rather to use a roughly fixed amount of memory that should never cause memory pressure issues like we have seen.

kenny-skaggs (approved):

PR #2539, opened 2021-07-27
[DA-2122] Fix AW2F Job for genome type

j-kanuch:
This PR removes the `genome_type` filter for the AW2F manifest. The filter was incorrectly set to only lookup Array samples. Since sample_ids are unique across genome types, there is no need for the filter. This was causing the genomic set member lookups to return None for all WGS samples, which caused an exception and therefore prevented any other AW2F files from being produced. A check was added in this PR to raise an exception if the member is not found. This PR should resolve the backlog.

## Tests
- [] unit tests

### DIFF
--- a/rdr_service/offline/genomic_pipeline.py
+++ b/rdr_service/offline/genomic_pipeline.py
@@ -327,7 +327,7 @@ def create_aw2f_manifest(feedback_record):
                               bucket_name=config.BIOBANK_SAMPLES_BUCKET_NAME,
                               ) as controller:
         controller.generate_manifest(GenomicManifestTypes.AW2F,
-                                     _genome_type=config.GENOME_TYPE_ARRAY,
+                                     _genome_type=None,
                                      feedback_record=feedback_record)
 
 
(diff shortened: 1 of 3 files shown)


rdr_service/offline/genomic_pipeline.py
```diff
@@ -327,7 +327,7 @@ def create_aw2f_manifest(feedback_record):
                               bucket_name=config.BIOBANK_SAMPLES_BUCKET_NAME,
                               ) as controller:
         controller.generate_manifest(GenomicManifestTypes.AW2F,
-                                     _genome_type=config.GENOME_TYPE_ARRAY,
+                                     _genome_type=None,
```
dtharpeuno, line 330: Should we just default the genome_type arg to being None in the corresponding methods and remove this declaration?

`def generate_manifest(self, manifest_type, _genome_type=None, **kwargs)` and in 
`def generate_and_transfer_manifest(self, manifest_type, genome_type=None, **kwargs):`

wangy70 (approved):

dtharpeuno (approved):
Nice catch! Left a small comment.

rdr_service/offline/genomic_pipeline.py
j-kanuch: It actually is defaulted to None in the lookup method, but we could default it in the referenced methods. I'll make a note to do this in a non-hotfix PR.

PR #2542, opened 2021-07-28
[PDR-367] Make sure we only return records of the correct schema.

robabram:
Tool should use participant id in query.

* Make sure we only return resource records of the correct schema.
* Use participant id in query.

## Tests
- [X] unit tests


pegbertsch (approved):
Approving given the PDR timeline.  But I assume we need to flesh out the unit test infrastructure to be able to catch these issues in unit testing.  Worth a backlog ticket?

robabram:
Agreed, at some point there needs to be work on creating unittests for the resource API.

pegbertsch (approved):

wangy70 (approved):

PR #2543, opened 2021-07-28
[PDR-364] Get cohort value from old participant summary table.

robabram:
* Switch to using participant cohort value from old participant summary for now.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -350,11 +350,18 @@ class ParticipantSummaryGenerator(generators.BaseGenerator):
         # credentials but are not correctly flagged in the participant table
         test_participant = p.isGhostId == 1 or p.isTestParticipant == 1 or (hpo and hpo.name == TEST_HPO_NAME)
 
+        # TODO: Workaround for PDR-364 is to pull cohort value from participant_summary. LIMITED USE CASE ONLY
+        cohort = ConsentCohortEnum.UNSET if not p or not p.participantSummary \
+                    or p.participantSummary.consentCohort is None \
+                    else ConsentCohortEnum(int(p.participantSummary.consentCohort))
+
         data = {
             'participant_id': f'P{p_id}',
             'biobank_id': p.biobankId,
             'research_id': p.researchId,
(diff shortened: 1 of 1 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -569,6 +569,15 @@ def _prep_modules(self, p_id, ro_session):
             filter(QuestionnaireResponse.questionnaireId ==
                    QuestionnaireConcept.questionnaireId).label('codeId')
 
+        # TODO: Workaround for PDR-364 is to pull cohort value from participant_summary. LIMITED USE CASE ONLY
```
pegbertsch: Can we move this into `_prep_participant_profile()` ?  We're already querying ParticipantSummary there;  just need to add the consentCohort column to the list of returned columns?

We already have some other spots where we query ParticipantSummary but it may help later on when we want to get rid of the ps dependencies to consolidate them as much as possible?

robabram: Sounds good, I'll do that.

rdr_service/resource/generators/participant.py
```diff
@@ -350,11 +350,18 @@ def _prep_participant(self, p_id, ro_session):
         # credentials but are not correctly flagged in the participant table
         test_participant = p.isGhostId == 1 or p.isTestParticipant == 1 or (hpo and hpo.name == TEST_HPO_NAME)
 
+        # TODO: Workaround for PDR-364 is to pull cohort value from participant_summary. LIMITED USE CASE ONLY
+        cohort = ConsentCohortEnum.UNSET if not p or not p.participantSummary \
```
pegbertsch, line 354: This may be fine as well.  I was thinking of adding this down around line 430 where we have other dependencies on ParticipantSummary for things that we ultimately thought might go in the proposed `participant_profile `table.

I'll leave it up to you if you want to do it here or there

pegbertsch (approved):
Approving as is but clarifying my prior suggestion in case it's worth an additional change.

dtharpeuno (approved):

PR #2545, opened 2021-07-28
[DA-2093] Add participant pairing history table to BQ and resource generator.

robabram:
* Add participant pairing history table to BQ and Resource data.

## Tests
- [X] unit tests

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -41,7 +41,7 @@ from rdr_service.model.ehr import ParticipantEhrReceipt
 from rdr_service.model.hpo import HPO
 from rdr_service.model.measurements import PhysicalMeasurements, PhysicalMeasurementsStatus
 from rdr_service.model.organization import Organization
-from rdr_service.model.participant import Participant
+from rdr_service.model.participant import Participant, ParticipantHistory
 from rdr_service.model.participant_cohort_pilot import ParticipantCohortPilot
 # TODO:  Using participant_summary as a workaround.  Replace with new participant_profile when it's available
 from rdr_service.model.participant_summary import ParticipantSummary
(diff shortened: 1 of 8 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -389,6 +389,30 @@ def _prep_participant(self, p_id, ro_session):
...
+        # Collect participant pairing history
+        pairing = ro_session.query(ParticipantHistory.lastModified, ParticipantHistory.hpoId,
+                                   ParticipantHistory.organizationId, ParticipantHistory.siteId). \
+            filter(ParticipantHistory.participantId == p_id).order_by(ParticipantHistory.lastModified).all()
+
+        pairing_history = None
+        if pairing:
+            pairing_history = list()
+            for item in pairing:
+                hpo = ro_session.query(HPO.name).filter(HPO.hpoId == item.hpoId).first()
```
pegbertsch: A few things:
Can we pull in the hpo / org ( / site ) details as outer joins on the `pairing` query above so we only have to hit the DB once?

Do we want the organization display string in this record or is the id sufficient?

And, do we also need to get the site details for each history record rather than using what's currently  in the `participant `table record?

rdr_service/resource/generators/participant.py
```diff
@@ -389,6 +389,30 @@ def _prep_participant(self, p_id, ro_session):
...
+                organization = ro_session.query(Organization.externalId). \
+                    filter(Organization.organizationId == item.organizationId).first()
+                pairing_history.append({
+                    'last_modified': item.lastModified,
+                    'hpo': hpo.name if hpo else None,
+                    'hpo_id': item.hpoId,
+                    'organization': organization.externalId if organization else None,
+                    'organization_id': item.organizationId,
+                    'site': self._lookup_site_name(p.siteId, ro_session),
```
pegbertsch: If there's a reason to get site values per history record instead the current `participant `table , we might be able to eliminate this if we join the info on the `pairing` query too?

rdr_service/resource/schemas/participant.py
```diff
@@ -243,6 +243,27 @@ class Meta:
         pii_filter = {}  # dict(field: lambda function).
 
 
+class ParticipantParingHistorySchema(Schema):
```
pegbertsch: if there's a need to validate resource vs. BQ schema, the unittest infrastructure is in 
tests/resource_tests/schema_tests/test_resource_schemas.py

pegbertsch (commented):
See comments on combining separate queries into one w/joins, and possible unittest for the new schema

rdr_service/resource/generators/participant.py
robabram: I can fix up the pairing history query to join the hpo, organization and site tables to get the descriptions, instead of calling for them in the loop.  The site ids will be different across the pairing history, so using the participant summary values doesn't make sense here.

rdr_service/resource/schemas/participant.py
robabram: I added this schema to the unittests.

rdr_service/resource/generators/participant.py
pegbertsch: Per above, shouldn' t this be looking up the `item.siteId` instead pf `p.siteId`?    That's why I thought instead of calling the lookup method every time,  you could just get the site name from a join in the pairing query.

robabram: Sorry, I misconstrued your comments.  This is fixed in the latest commits.

pegbertsch (approved):

rdr_service/resource/generators/participant.py
```diff
@@ -389,6 +390,32 @@ def _prep_participant(self, p_id, ro_session):
...
         }
 
+        # Collect participant pairing history
+        pairing_history = None
+        query = ro_session.query(ParticipantHistory.lastModified, ParticipantHistory.hpoId, HPO.name.label('hpo'),
+                                   ParticipantHistory.organizationId, Organization.externalId.label('organization'),
+                                   ParticipantHistory.siteId, Site.googleGroup.label('site')). \
+                join(HPO, HPO.hpoId == ParticipantHistory.hpoId).\
+                join(Organization, Organization.organizationId == ParticipantHistory.organizationId).\
+                join(Site, Site.siteId == ParticipantHistory.siteId).\
```
kenny-skaggs: Should these be `LEFT OUTER JOINS` rather than the default `INNER JOINS`?

robabram: Yes, they should be.  Good catch.

pegbertsch (approved):

PR #2549, opened 2021-07-29
[DA-2135] Detecting whether retrospectively validated files would have been synced already

kenny-skaggs:
## Resolves *[DA-2135](https://precisionmedicineinitiative.atlassian.net/browse/DA-2135)*
This adds the ability to determine whether files that have been retrospectively validated (files that pass validation and have been uploaded previous to when the new consent sync pipeline implementation started running in Prod) would have been copied to site buckets by the previous consent sync process. This will help us mark files that sites should already have, and keep us from unnecessarily sending them again.

This will only need to be run against the retrospectively validated files. We have them currently for a couple of the organizations and will need to run it against them and any newly validated files that may have already been copied. Once all of the older files have been validated this code can be deleted.

## Tests
- [x] unit tests


dtharpeuno (approved):

rdr_service/offline/sync_consent_files.py
```diff
@@ -42,6 +45,117 @@ class ParticipantPairingInfo:
     site_name: str
 
 
+@dataclass
```
pegbertsch, line 48: (Not an issue, just curious) In case I wasn't seeing it:  what features of `@dataclass` are required/wanted here?

rdr_service/offline/sync_consent_files.py
```diff
@@ -42,6 +45,117 @@ class ParticipantPairingInfo:
...
+
+        count = 0
+        for file in files:
+            pairing_info, summary = latest_participant_pairings[file.participant_id]
+            sync_date = self.get_sync_date(file=file, summary=summary, latest_pairing_info=pairing_info)
+            if sync_date is not None:
+                file.sync_time = sync_date
+                file.sync_status = ConsentSyncStatus.SYNC_COMPLETE
+
+            count += 1
```
pegbertsch: Need clarification to approve:
Is `count` supposed to track how many passes through the loop, or how many file records were actually updated to SYNC_COMPLETE?

Also:   what happens when the count is not a multiple of 500?  Do you need a commit() outside of the `for` loop if there is a remainder value?

rdr_service/offline/sync_consent_files.py
```diff
@@ -42,6 +45,117 @@ class ParticipantPairingInfo:
...
+        If any of them would have triggered a sync to the latest org's bucket, then the date is returned for the sync.
+        Otherwise None is returned.
+        """
+        if not cls._org_date_valid_for_sync(
+            upload_time=file.file_upload_time,
+            latest_pair_time=latest_pairing_info.start_date
+        ):
+            return None
+
+        primary_time = summary.consentForStudyEnrollmentTime
```
pegbertsch, line 115: So this tool only has to worry about the last received primary / EHR consents?    (For cases where there was also a previous "first yes" / multiple consent files of that type submitted)

rdr_service/offline/sync_consent_files.py
kenny-skaggs: Dataclass isn't exactly required, but just makes it easier to put information into an object. It takes the field definitions and makes them into instance variables for the objects. It also automatically provides a constructor so it's easy to instantiate the objects.

rdr_service/offline/sync_consent_files.py
kenny-skaggs: Count is tracking the number of files that have been iterated. I'm not entirely sure why I did that, and it probably could be the number of files that were modified. I'll update it and give it a better name.

kenny-skaggs: If there's a remainder value then that will still commit when the session's context closes.

rdr_service/offline/sync_consent_files.py
kenny-skaggs: We only receive the Primary consent once. For EHR: I've tried to take the simple route and have it detect whether files would have transferred based on EHR time for the latest submission. This may miss detecting that some participants have had files copied over for times that another EHR consent submission would have caused it, but in those scenarios we would simply copy their files over again. I think that's an acceptable outcome.

pegbertsch: Can you clarify?  We certainly have participants with multiple ConsentPII payloads (although unclear if they would reference the same pdf?), and we have the consent updates for earlier cohorts for the primary consent.   I wasn't sure if those cases were part of what's being checked here?

kenny-skaggs: Yeah, absolutely. Most participants that have multiple ConsentPII payloads only have one full response. Subsequent responses are usually updates to phone number or address and aren't new consents. I'd be curious about it if we had participants with multiple full consent responses, but even that case wouldn't have triggered the RDR to sync files again since the `consentForStudyEnrollmentTime` is only set once.

We do have PrimaryUpdate consents, but that doesn't change the `consentForStudyEnrollmentTime` field and wouldn't have triggered consent file syncing.

rdr_service/offline/sync_consent_files.py
pegbertsch: Maybe I'm missing something very simple/implied, but I could not find confirmation of this when reading through the SQLAlchemy Session documentation.  Most of the examples referenced explicit commit() calls.

kenny-skaggs: I think you're right that sqlalchemy sessions need an explicit commit call. I'm not sure what sqlalchemy would do on it's own, but this session is created by the generator function at https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/model/database.py#L98. That function handles calling commit at the end, when the context is closing.

pegbertsch (approved):

PR #2552, opened 2021-08-03
[PDR-367] Fix minor issues with 1.00.3 hotfix.

robabram:
* Fix batch cloud task submission in Resource tool.
* Change string replacement style to fix error caused by Google SDK's older python version.

### DIFF
--- a/rdr_service/resource/__init__.py
+++ b/rdr_service/resource/__init__.py
@@ -32,7 +32,10 @@ class SchemaMeta(object):
         if not isinstance(kwargs['resource_uri'], str):
             raise ValueError('Schema resource URI must be a string')
         if not isinstance(kwargs['schema_id'], SchemaID):
-            raise ValueError(f'Schema ID has not been set on Schema object {kwargs["resource_uri"]}.')
+            msg = 'unknown'
+            if kwargs and 'resource_uri' in kwargs:
+                msg = kwargs["resource_uri"]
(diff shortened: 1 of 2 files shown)


rdr_service/resource/__init__.py
```diff
@@ -32,7 +32,10 @@ def __init__(self, **kwargs):
         if not isinstance(kwargs['resource_uri'], str):
             raise ValueError('Schema resource URI must be a string')
         if not isinstance(kwargs['schema_id'], SchemaID):
-            raise ValueError(f'Schema ID has not been set on Schema object {kwargs["resource_uri"]}.')
+            msg = 'unknown'
+            if kwargs and 'resource_uri' in kwargs:
+                msg = kwargs["resource_uri"]
+            raise ValueError('Schema ID has not been set on Schema object {0}.'.format(msg))
```
dtharpeuno, line 38: Can we use an f-string here?

dtharpeuno (approved):

rdr_service/resource/__init__.py
robabram: No, I had to change it to format() from f-string.  The reason is the Google SDK (at least the version I am using) doesn't have support for f-string.  In a terminal, if you are in the 'rdr_service' directory and you run a SDK tool, this '__init__.py' file gets run  and throws an exception here when it is a f-string.  I do prefer f-string usage overall.

wangy70 (approved):

PR #2555, opened 2021-08-03
[ROC-1112] Add AW3 to process runner tool; Speed up WGS AW3 query

j-kanuch:
This PR adds the AW3 manifests to the process runner tool. The AW3 WGS query was updated to use a subquery instead of a self-join since the parent sample ID is not indexed.

## Tests
- [x] unit tests


rdr_service/genomic/genomic_data.py
```diff
@@ -21,6 +21,15 @@ def __init__(self, input_manifest=None):
...
 
+        self.subqueries = {
+            'aw3_wgs_parent_sample_id': (
+                sqlalchemy.select([GenomicSetMember.gcManifestParentSampleId]).select_from(GenomicSetMember).where(
+                    (GenomicSetMember.genomeType == "aou_array") &
+                    (GenomicSetMember.aw3ManifestJobRunID.isnot(None))
+                )
+            )
+        }
+
```
dtharpeuno, line 32: :nit would you mind running the code format tool on this or breaking the lines for it.

dtharpeuno (approved):
Nice, I like the idea of the subquery.

rdr_service/genomic/genomic_data.py
j-kanuch: Reformatted.

wangy70 (approved):

PR #2558, opened 2021-08-05
[DA-2090] Re-syncing consent files for repaired participants

kenny-skaggs:
## Resolves *[DA-2090](https://precisionmedicineinitiative.atlassian.net/browse/DA-2090)*
For any participants that have been paired to any organization and then get re-paired, we need to make sure their consent files get sent to the new organization's bucket.

Additionally, this will retrospectively validate any participants that have not yet had all their consent files validated. That functionality will be removed when all previously received consents have been validated for existing participants.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/consent_dao.py
+++ b/rdr_service/dao/consent_dao.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Collection
 
 from sqlalchemy import or_
+from sqlalchemy.orm import Session
 
 from rdr_service.dao.base_dao import BaseDao
 from rdr_service.model.consent_file import ConsentFile, ConsentSyncStatus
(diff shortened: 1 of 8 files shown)


rdr_service/dao/consent_dao.py
```diff
@@ -92,3 +92,15 @@ def get_files_ready_to_sync(self, org_names=None, session: Session = None) -> Co
...
+    def set_previously_synced_files_as_ready(cls, session: Session, participant_id: int):
+        session.query(
+            ConsentFile
+        ).filter(
+            ConsentFile.sync_status == ConsentSyncStatus.SYNC_COMPLETE,
+            ConsentFile.participant_id == participant_id
+        ).update({
+            ConsentFile.sync_status: ConsentSyncStatus.READY_FOR_SYNC,
+            ConsentFile.sync_time: None
+        })
```
dtharpeuno, line 106: Nice, will use this in the future.

rdr_service/dao/consent_dao.py
```diff
@@ -92,3 +92,15 @@ def get_files_ready_to_sync(self, org_names=None, session: Session = None) -> Co
                 return self._get_ready_to_sync_with_session(session=dao_session, org_names=org_names)
         else:
             return self._get_ready_to_sync_with_session(session=session, org_names=org_names)
+
+    @classmethod
```
dtharpeuno, line 96: Is there a reason to utilize the `@classmethod` on these methods and passing session as an arg instead of using the context manager for setting the session from self?

dtharpeuno (approved):

rdr_service/dao/consent_dao.py
kenny-skaggs: For how this method is being used currently, no not really. But I've been finding that the DAOs will create a session connected to the local database even when the rest of the environment is targeting an environment (like Prod), so I've been getting in the habit of passing a session in to where it would be used rather than creating one from scratch.

rdr_service/dao/participant_dao.py
```diff
@@ -268,8 +269,23 @@ def _do_update(self, session, obj, existing_obj):
                     obj.organizationId = None
                     need_new_summary = True
 
-        # No pairing updates sent, keep existing values.
-        if update_pairing == False:
+        if update_pairing:
+            if obj.organizationId != existing_obj.organizationId and existing_obj.participantSummary is not None:
+                # Get valid files ready for sync when a participant is paired to an organization
+                ConsentDao.set_previously_synced_files_as_ready(session, obj.participantId)
```
pegbertsch, line 275: Do we want to be able to quickly confirm which organizations should have gotten the files in their bucket?  I'm wondering if we need to implement a `sync_history` that we add to every time we move a file to SYNC_COMPLETE.  (I was thinking of the BYOD status JSON struct we have now in `participant_summary`)

If so, I don't know if there is a need to implement it in conjunction with this change, so as not to lose prior sync date(/organization?)  details that would be updated/overwritten.

kenny-skaggs: I don't think there's a pressing need for knowing where files had previously been copied to. Any usual investigation is for trying to determine if it copied to the bucket for the organization that needs to have it now. And the usual case this code will be handling is for participants that consented before the validation pipeline started and are being paired now.

pegbertsch: Per offline conversation, do not expect enough re-syncs (to repaired orgs) such that losing the previous `sync_time` value is an issue.   Per Kenny, only really interested in the most recent sync operation?

pegbertsch (approved):

PR #2560, opened 2021-08-05
[DA-2132] Turn back on notifications for missing files

dtharpeuno:
Slack alerts and genomic incidents are temporarily disabled for missing data files. These need to pull from the missing file table for missing data files.

## Tests
- [x] unit tests


wangy70 (approved):

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1944,52 +1939,46 @@ def update_reconciled_metric(self, _obj, missing_data_files, _gc_site_id):
                         gc_validation_metric_id=_obj.id
                     )
                     self.data_file_missing_dao.insert(missing_file_object)
+                self.process_missing_data(_obj)
```
j-kanuch: Does this run on every record? It looks like `process_missing_data()` runs a query against the missing data table to see if there are missing data files for every record individually. Previously, this was ran on all records at once, the iteration happened inside of `process_missing_data()`. I'm worried that the reconciliation performance may take too long and get killed.

j-kanuch (approved):

PR #2566, opened 2021-08-11
[DA-2006] workbench_resource_generator

wangy70:
## Resolves *[DA-2006](https://precisionmedicineinitiative.atlassian.net/browse/DA-2006)*
Create resource generator for workbench data.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/constants.py
+++ b/rdr_service/resource/constants.py
@@ -46,6 +46,12 @@ class SchemaID(IntEnum):
     workbench_institutional_affiliation = 4010
     workbench_workspace = 4020
     workbench_workspace_users = 4030
+    workbench_researcher_race = 4040
+    workbench_researcher_gender = 4050
+    workbench_researcher_sex_at_birth = 4060
+    workbench_researcher_degree = 4070
+    workbench_workspace_age = 4080
+    workbench_workspace_ethnicity = 4090
 
     # Covid study
     biobank_covid_antibody_sample = 5000
--- a/rdr_service/resource/generators/workbench.py
+++ b/rdr_service/resource/generators/workbench.py
@@ -1,0 +1,265 @@
+#
+# This file is subject to the terms and conditions defined in the
+# file 'LICENSE', which is part of this source code package.
+#
+import logging
+import json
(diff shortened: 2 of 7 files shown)


rdr_service/resource/constants.py
```diff
@@ -46,6 +46,12 @@ class SchemaID(IntEnum):
     workbench_institutional_affiliation = 4010
     workbench_workspace = 4020
     workbench_workspace_users = 4030
+    workbench_researcher_race = 4040
```
pegbertsch, line 49: questions for @robabram :   What is the trigger for needing a distinct SchemaID value when it comes to "sub"-schemas?  Is the expectation that any nested field in  a defined resource Schema will always need its own resource_type / SchemaID and should be retrievable  via the resource API?

Also:  where do we have the documentation for populating newly defined content in the RDR  `resource_type` and `resource_schema` tables when we  update these lists?

pegbertsch (approved):

rdr_service/resource/constants.py
robabram: Defining a schema id value is a bit arbitrary, but I wanted to be able to identify a schema id group by the left most digit(s) of the schema id.

rdr_service/resource/generators/workbench.py
```diff
@@ -0,0 +1,265 @@
...
+        :return: resource object
+        """
+        if not self.ro_dao:
+            self.ro_dao = ResourceDataDao(backup=backup)
+
+        with self.ro_dao.session() as ro_session:
+            row = ro_session.execute(
+                text('select * from rdr.workbench_workspace_user_history where id = :id order by modified desc'),
+                {'id': pk_id}).first()
+            # Fall back to workbench_workspace_user, it looks like there are missing records in the history table.
```
robabram, line 123: We could add a 'limit 1' to the SQL statements to ensure only one record is returned to SQLAlchemy instead of returning all matching rows and then only using the first row.  It's not big enough enough performance problem to worry about though.

robabram (approved):

PR #2568, opened 2021-08-11
[PDR-365] Add exluded survey fields and set boolean value

robabram:
* Add previously excluded questionnaire fields and set boolean value based on if there is a answer value.
* Validate that 'participant_id' resource schema field is defined correctly.
* Force correctly setting the 'participant_id' value if value is int.

## Tests
- [x] unit tests


rdr_service/model/bq_questionnaires.py
```diff
@@ -133,9 +132,12 @@ def get_fields(self):
                             continue
 
                         if bq_field_name.lower() not in skip_fieldnames_lower:
+                            # If field is in force boolean list, set to integer.
+                            fld_type = BQFieldTypeEnum.INTEGER.name if bq_field_name.lower() in force_boolean_lower \
+                                        else BQFieldTypeEnum.STRING.name
```
kenny-skaggs, line 137: I'm trying to understand the `force_boolean_lower` logic. Is this casting all the values for specified codes to integers? What would it do for dates in these answers?

pegbertsch: All of the question codes/ fields that are in the `_force_boolean_fields` list for the impacted modules (e.g., see line 215 or 549 as examples) are free text fields.   The idea is to just confirm if the free text field has any content or not, and set it to 1 if it does.   This was a specific request from the dashboard team to allow them to see which (free text) questions participants answered, since the text itself is not in PDR due to PII concerns.

pegbertsch: If it would help to talk offline about the history of why I used lowercase string comparisons in the existing code, let me know. But it had to do with "not production quality" code table data in our lower environments, where we could have two question codes with nearly identical value strings except for matching on case or because one had a space where the other had an underscore.  It caused some unexpected "collisions" on results from `self.make_bq_field_name` when trying to generate a BigQuery table schema/column list.   

I don't know if converting the `_force_boolean_fields` values to lowercase is needed here but I don't think it has any negative side effects when testing for inclusion.  We're still assigning the unchanged `bq_field_name` value on line 139.

robabram: It's been long enough I can't remember why I used the "lower()" when checking code values, but it doesn't hurt to have this extra step to ensure the equality check is successful.

rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -192,23 +192,29 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
+                        setattr(bqr, fld_name, 0)
+                    if fld_value is None:
                         continue
                     # question responses values need to be coerced to a String type.
                     if isinstance(fld_value, (datetime.date, datetime.datetime)):
                         setattr(bqr, fld_name, fld_value.isoformat())
-                    else:
-                        setattr(bqr, fld_name, str(fld_value))
-
+                    # Check to see if the field has been forced to boolean, set to 1 because we have a value.
```
pegbertsch, line 208: The over-commenter in me is wondering if it might be helpful to expand on this comment, i.e., that the only other BQFieldTypeEnum.INTEGER fields in a module schema are in the "ignore list" (`if fld_name in...`) above. Can safely assume all other INTEGER fields match the child module class `_force_boolean_fields` list

pegbertsch (approved):
Approve.  I did not go through and compare each `_force_boolean_field` list to the spreadsheet associated with this ticket (to make sure every free text field is accounted for), so if you need a second pair of eyes for that part let me know.

rdr_service/dao/bq_questionnaire_dao.py
robabram: Yes, all questionnaire module fields are forced to String because answers may be multiple data types.  IE: A free text can have a string answer or may be a code answer (code_id) of 'PMI_Skip', so it's easiest to force all answers to String.

pegbertsch (approved):

pegbertsch (approved):

PR #2572, opened 2021-08-12
[PDR-348] Convert and rename dv_order field in to enum for collection method.

robabram:
* Removed the 'dv_order' field from BQ and Resource biobank order models.
* Added 'collection_method' and 'collection_method_id' fields to models.
* Updated SQL in participant generator to set correct enum value for collection method.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -51,7 +51,7 @@ from rdr_service.model.questionnaire_response import QuestionnaireResponse, Ques
 from rdr_service.participant_enums import EnrollmentStatusV2, WithdrawalStatus, WithdrawalReason, SuspensionStatus, \
     SampleStatus, BiobankOrderStatus, PatientStatusFlag, ParticipantCohortPilotFlag, EhrStatus, DeceasedStatus, \
     DeceasedReportStatus, QuestionnaireResponseStatus, EnrollmentStatus, OrderStatus, WithdrawalAIANCeremonyStatus, \
-    TEST_HPO_NAME, TEST_LOGIN_PHONE_NUMBER_PREFIX
+    TEST_HPO_NAME, TEST_LOGIN_PHONE_NUMBER_PREFIX, SampleCollectionMethod
 from rdr_service.resource import generators, schemas
 from rdr_service.resource.calculators import EnrollmentStatusCalculator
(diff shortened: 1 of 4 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -928,15 +928,9 @@ def _make_sample_dict_from_row(bss=None, bos=None):
...
-                   case when exists (
-                     select bmko.participant_id
-                     from biobank_mail_kit_order bmko
-                     where bmko.biobank_order_id = bo.biobank_order_id
-                        and bo.participant_id = bmko.participant_id
-                        and bmko.associated_hpo_id is null
-                   )
-                   then 1 else 0 end as dv_order
-             from biobank_order bo where participant_id = :p_id
+                   case when bmko.id is not null then 1 else 2 end as collection_method
```
pegbertsch, line 931: Based on this SQL in `participant_summary_dao.py` are we folding a potential SampleCollectionMethod.UNSET scenario into one of the other collection method enum values here?   Could we be introducing deltas in the QC counts for RDR vs. PDR?

https://github.com/all-of-us/raw-data-repository/blob/964f9f2930eb38309ee57453a6036da4a55521c7/rdr_service/dao/participant_summary_dao.py#L198

pegbertsch: Never mind, I think it's taken care of by the "orderless sample" logic below...

pegbertsch (commented):
I think there may be an opportunity to update the `test_biobank_samples_pipeline.test_end_to_end` unittest case, could possibly add a call to the generator and verify the collection method values that get returned?  

If there's not time to do this before the release is cut, I can approve as is;  but I've been trying to take advantage of existing unittest data setups to inject more validations of PDR generator output when possible.

rdr_service/resource/generators/participant.py
robabram: As I see it, if there is a biobank order record the only two possible values are 1 or 2.  The exception is when we have orphaned sample records without a biobank order.  In the case of orphans, I defaulted to UNSET.  Based on the SQL in the 'participant_summary_dao.py' file, I can't ever see the UNSET else code path being used as the 'biobank_order' table is the left side of the join.

robabram:
I'm going to pass on updating the end-to-end test as we might be pulling out all of the BQ and Resource code from the RDR in the near future.

pegbertsch (approved):

wangy70 (approved):

PR #2573, opened 2021-08-12
[DA-2087] Creating endpoint for triggering validation on the server

kenny-skaggs:
## Partially Resolves *[DA-2087](https://precisionmedicineinitiative.atlassian.net/browse/DA-2087)*
This creates an endpoint on the server that allows for validating all consent files for specific participant ids. This will be used to perform the retro-validation for existing consents. It will likely be kept afterwards as a utility for correcting any issues in the pipeline when needed.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/offline/main.py
+++ b/rdr_service/offline/main.py
@@ -39,7 +39,8 @@ from rdr_service.offline.requests_log_migrator import RequestsLogMigrator
 from rdr_service.offline.service_accounts import ServiceAccountKeyManager
 from rdr_service.offline.sync_consent_files import ConsentSyncController
 from rdr_service.offline.table_exporter import TableExporter
-from rdr_service.services.consent.validation import ConsentValidationController, StoreResultStrategy
+from rdr_service.services.consent.validation import ConsentValidationController, ReplacementStoringStrategy,\
+    StoreResultStrategy
 from rdr_service.services.data_quality import DataQualityChecker
 from rdr_service.services.flask import OFFLINE_PREFIX, flask_start, flask_stop
 from rdr_service.services.gcp_logging import begin_request_logging, end_request_logging,\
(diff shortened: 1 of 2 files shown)


pegbertsch (approved):
Wondering about the projected use cases for this (e.g., how many participant ids per request) and the pros/cons of the /offline service option  vs. the /resource service & cloud task option.  

Just something to keep in mind, given how we've started moving certain chunks of work for genomics ingestions and PDR builds from the /offline service over to /resource service endpoints.  This may not have the usage characteristics where there's anything to gain, especially if it's intended to be a relatively short-lived feature while the bulk of the initial/retrospective validations are going on.

kenny-skaggs:
> Wondering about the projected use cases for this (e.g., how many participant ids per request) and the pros/cons of the /offline service option vs. the /resource service & cloud task option.
> 
> Just something to keep in mind, given how we've started moving certain chunks of work for genomics ingestions and PDR builds from the /offline service over to /resource service endpoints. This may not have the usage characteristics where there's anything to gain, especially if it's intended to be a relatively short-lived feature while the bulk of the initial/retrospective validations are going on.

I'm not sure how many ids would be sent yet. I imagined that would be something fine-tuned as it was used. This is mostly to set up the capability of the server validating as needed, where the client would be responsible for not overloading it. I don't think there are any significant pros or cons to which service is running it (offline vs resource). And I don't foresee any reasons to having it able to run as a cloud task (as I'd be manually setting up one set of participants at a time, and there's no need to have multiple validation tasks queued or running concurrently). But please let me know if there are benefits I'm not considering.

dtharpeuno (approved):
Nice, are you manually sending POSTs to this with a list of ids instead of hooking it into a cron job?

kenny-skaggs:
> Nice, are you manually sending POSTs to this with a list of ids instead of hooking it into a cron job?

Right, I'd use postman or set up a tool to make a call to it. This wouldn't be a scheduled script.

PR #2579, opened 2021-08-17
[DA-2117] retention_cronjob_for_stable

wangy70:
## Resolves *[DA-2117](https://precisionmedicineinitiative.atlassian.net/browse/DA-2117)*
Create a cronjob to calculate retention metrics for lower envs only

Instead of import from bucket file, use SQL to calculate retention metrics for lower envs for test purpose.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/offline/retention_eligible_import.py
+++ b/rdr_service/offline/retention_eligible_import.py
@@ -1,6 +1,8 @@
 import logging
+import datetime
 
 from dateutil.parser import parse
+from rdr_service.clock import CLOCK
 from rdr_service.storage import GoogleCloudStorageCSVReader
 from rdr_service.dao.retention_eligible_metrics_dao import RetentionEligibleMetricsDao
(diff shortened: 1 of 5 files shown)


rdr_service/offline/retention_eligible_import.py
```diff
@@ -108,6 +111,152 @@ def import_retention_eligible_metrics_file(task_data):
     logging.info(f"Import and update completed for gs://{csv_file_cloud_path}")
 
 
+def calculate_retention_eligible_metrics():
```
kenny-skaggs, line 116: Can you add `app_util`'s `nonprod` decorator to this method, just to make certain it doesn't get triggered in Prod on accident?

wangy70: Good idea, done!

kenny-skaggs (approved):

robabram (approved):

PR #2575, opened 2021-08-13
[DA-2148] Moving Genomic job run update to cloud task

dtharpeuno:
## Resolves *[ticket DA-2148]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2148

## Description of changes/additions
Moving GenomicSetMember job run an update to cloud task for mitigating memory usage in-app as well as sending ids instead of objects for more peformance.

## Tests
- [x] unit tests

PR #2583, opened 2021-08-18
[DA-2150] Adding validation for manifest generation

dtharpeuno:
## Resolves *[ticket DA-2150]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2150

The AW3 manifest should be validated before sending to the Broad DRC to ensure the correct formatting. This will prevent errors during their ingestion and reduce the need for RDR intervention going forward.

## Tests
- [x] unit tests


rdr_service/genomic/genomic_job_components.py
```diff
@@ -3070,65 +3081,67 @@ def generate_and_transfer_manifest(self, manifest_type, genome_type, **kwargs):
...
+        results = []
+        for row in source_data:
+            member = self.member_dao.get_member_from_sample_id(row.sample_id, genome_type)
+            if member is None:
+                raise NotFound(f"Cannot find genomic set member with sample ID {row.sample_id}")
+
+            if self.manifest_def.job_run_field is not None:
+                results.append(
+                    self.member_dao.update_member_job_run_id(
+                        member,
```
j-kanuch: Does this need to be `member.id` now?

rdr_service/genomic/genomic_job_components.py
```diff
@@ -3139,6 +3152,53 @@ def _pull_source_data(self):
         with self.member_dao.session() as session:
             return session.execute(self.manifest_def.source_data).fetchall()
 
+    def _validate_source_data(self, data, manifest_type):
+        invalid = False
```
j-kanuch, line 3149: This method works for the immediate need, so no need to change here, but we may want to revisit how this is done in the future. I'm specifically thinking about abstracting the rules, similar to how we validate the files we ingest using the `GenomicFileValidator` class, and the validate_filename() method. That way we can better control the complexity if more files or rules are added.

rdr_service/genomic/genomic_job_components.py
dtharpeuno: Haha, that is from another PR that proceeds this one, https://github.com/all-of-us/raw-data-repository/pull/2575, that hasn't been merged yet.

j-kanuch: Makes sense.

j-kanuch (approved):

rdr_service/genomic/genomic_job_components.py
dtharpeuno: 1. Should I add this to the GenomicFileValidator class now, the reason I didn't use it is that seems more catered to files, but I see where we can pass data to it to validate?
2. Tried to make this as general as I could with the idea of it only being targeted towards the AW3s for now as denoted as the priority for this. Plan on extending this later as need be.

j-kanuch: No, you're right. The AW3s are the priority here and this is needed as soon as possible. I don't think we'd want to use the `GenomicFileValidator`, as you said it is more tailored to file validation rather than data validation. What you have here is good and I've approved it. 
I'm sort of brainstorming that in the future we'd want to validate all output data, and so I am thinking of a new component class. This could be something like `GenomicOutputDataValidator` and a related class for `ValidationRule`. This is a lower priority and will also probably require more fleshed-out requirements, a tech design, etc.

wangy70 (approved):

PR #2586, opened 2021-08-19
Daily consent report rework

pegbertsch:
Per some follow-on discussions about what should be in the daily (and forthcoming weekly) consent validation reports, it was clarified that the daily report should only report on "newly received" consents.   We don't want to pick up any error counts related to retrospective validations that are also being performed and creating new `consent_file` table entries for old consent payloads/files, with timestamps that would have gotten included by the old report logic.

The main change is to revise the SQL query filter to use authored dates from participant summary instead of ` consent_file` table entry date values to look for consents to report on for a given date.

This includes mapping the appropriate authored field from `participant_summary` to the consent type being processed (e.g., PRIMARY -> `consent_for_study_enrollment_first_yes_authored`, CABOR -> `consent_for_cabor_authored`).  

The consent validation records are now retrieved in chunks by consent types and then merged/appended together into a single pandas dataframe, so the remaining report generation logic remains largely the same).

Also did some minor clean-up and tweaking of the query content (removing unused values) and the report text fields.

## Tests
- [x] unit tests


rdr_service/tools/tool_libs/consent_validation_report.py
```diff
@@ -99,63 +107,51 @@
...
                    (cf.file_exists AND cf.other_errors LIKE '%veteran consent for non-veteran participant%')
                       AS va_consent_for_non_va
-            FROM consent_file cf
-            JOIN participant_summary p on p.participant_id = cf.participant_id
-            LEFT OUTER JOIN hpo h on p.hpo_id = h.hpo_id
-            LEFT OUTER JOIN organization o on p.organization_id = o.organization_id
+            FROM participant_summary ps
+            -- Eliminate test/ghost pids from the result
+            JOIN participant p on p.participant_id = ps.participant_id
+                 AND p.is_test_participant = 0 and (p.is_ghost_id is null or not p.is_ghost_id) and p.hpo_id != 21
```
kenny-skaggs, line 126: Should the email and phone number also be looked at?

robabram (approved):

rdr_service/tools/tool_libs/consent_validation_report.py
pegbertsch: Personally, I'd like to get away from that because I think it masks new cases where PTSC should have flagged test participants for us and didn't.  Would like for them to be forced to look at them, where the resolution would be for them to properly flag them via the process they agreed to follow?

kenny-skaggs: Sounds good to me.

kenny-skaggs (approved):

PR #2589, opened 2021-08-23
Fix CircleCI MySQL install issues.

robabram:
Oracle MySQL 5.7 will not install on newer Ubuntu versions than 18.04.  (See [MySQL Support Matrix](https://www.mysql.com/support/supportedplatforms/database.html) )
Select Ubuntu 18.04 for image so we can use MySQL 5.7.
Install python 3.7 virtual environment package manually and force python 3.7 usage.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ version: 2.1
 
 job_defaults: &job_defaults
   docker:
-    - image: circleci/python:3.7.11
+    - image: cimg/base:2021.05-18.04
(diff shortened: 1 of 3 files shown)


pegbertsch (approved):
Approving but taking some questions (esp. regarding what triggered need for changing file permissions) offline.

Also:  is there a link for where you found the details on MySQL 5.7 not installing on newer Ubuntu versions, that you could add to the PR description?

dtharpeuno (approved):
👍

PR #2591, opened 2021-08-24
[DA-2087] Script for sending batches of ids to server for validation

kenny-skaggs:
## Partially Resolves *[DA-2087](https://precisionmedicineinitiative.atlassian.net/browse/DA-2087)*
This adds functionality to the consents script to be able to send batches of participant ids to the offline server for validating their consent files. Each request to the server will contain 10 participants (each batch completes within 50 seconds), and each run of the script to perform this process will validate the files for 5000 participants that don't have any files validated (each run completes in about 2 hours).

This also adds a `BatchManager` class to help encapsulate some of the pattern that we're developing in processing batches of objects (collecting them in a list and comparing them to a specific threshold, and making sure to process any remaining items that didn't make a full batch at the end).

Running the script with a SA also causes an error (because the SA doesn't have permission to the config). So this retrieves the config before activating an environment with the SA.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/consent_dao.py
+++ b/rdr_service/dao/consent_dao.py
@@ -16,9 +16,8 @@ class ConsentDao(BaseDao):
         super(ConsentDao, self).__init__(ConsentFile)
 
     @classmethod
-    def get_participants_with_consents_in_range(cls, session, start_date,
-                                                end_date=None) -> Collection[ParticipantSummary]:
-        query = session.query(
+    def _get_query_non_test_summaries(cls, session):
+        return session.query(
             ParticipantSummary
         ).join(
             Participant,
@@ -32,6 +31,11 @@ class ConsentDao(BaseDao):
                 ParticipantSummary.email.notlike('%@example.com')
             )
         )
+
+    @classmethod
+    def get_participants_with_consents_in_range(cls, session, start_date,
+                                                end_date=None) -> Collection[ParticipantSummary]:
+        query = cls._get_query_non_test_summaries(session)
         if end_date is None:
             query = query.filter(
                 or_(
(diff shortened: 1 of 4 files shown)


dtharpeuno (approved):

rdr_service/dao/consent_dao.py
```diff
@@ -32,6 +31,11 @@ def get_participants_with_consents_in_range(cls, session, start_date,
                 ParticipantSummary.email.notlike('%@example.com')
```
pegbertsch, line 31: For future consideration:  I've been thinking that we might be better off not doing additional screening to identify test participants that PTSC hasn't told us about, so we don't mask the fact that they haven't told us.  That's assuming we have an agreement that we will no longer have to check multiple conditions on our end (e.g., HPO pairing, too?) in addition to the `is_test_participant` flag in the `participant` table.   Would really like to rely on that one field if we can.

rdr_service/tools/tool_libs/consents.py
```diff
@@ -44,6 +61,31 @@ def run(self):
...
             self.check_retro_sync()
+        elif self.args.command == 'retro-validation':
+            self.retro_validate()
+
+    def _call_server_for_retro_validation(self, participant_ids):
+        if len(participant_ids) > 0:
+            print(f'Processing batch that starts with {participant_ids[0]}')
+
+        response = requests.post(
+            'https://offline-dot-all-of-us-rdr-prod.appspot.com/offline/ManuallyValidateFiles',
```
pegbertsch, line 72: What determines if we define/read endpoints from the system config vs. hardcoding?

pegbertsch (approved):
Minor comments/questions, but nothing to impact approval.

rdr_service/tools/tool_libs/consents.py
kenny-skaggs: I think it would be a case-by-case decision. This is specifically built to validate past consent files for participants that exist in Prod, and is only meant to be a temporary part of the script (it has no use case after the retrospective validation is complete and will be deleted). So it makes sense to me that we wouldn't be reading from the config or tailoring it to other environments (since we won't be running in there). But I'm open to other ideas.

PR #2592, opened 2021-08-25
[PDR-413] Reverting fields that should not be forced to boolean.

robabram:
Multiple COPE survey date fields were converted to boolean values, this PR reverts those fields back to default.

## Tests
- [x] unit tests


pegbertsch (approved):
Approving.   So we think these are the only fields that got converted to booleans which the dashboard team (and Scott?) are complaining about, and don't expect any more requests for reversions?

wangy70 (approved):

robabram:
Jonathan has confirmed that these are the only fields causing problems in the dashboards.

PR #2594, opened 2021-08-26
MIGRATION [DA-2168] Add manifest file ID for AW3 runs

dtharpeuno:
## Resolves *[ticket DA-2168]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2168

Investigating issues with the AW3 manifest is difficult because there is no direct way to determine the manifest file path from the `genomic_set_member` table. Adding the field `aw3_manifest_file_id` to `genomic_set_member` will solve this problem.
** Deleted unused methods in the pipeline for general cleanup
** Added update statement  in migration for backfill

## Tests
- [x] unit tests


wangy70 (approved):

rdr_service/genomic/genomic_job_components.py
```diff
@@ -2013,90 +2011,6 @@ def reconcile_gem_report_states(self, _last_run_time=None):
...
-        with sequencing reconciliation data
-        :param member: the GenomicSetMember to update
-        :param seq_file_name:
-        :param job_run_id:
-        :return: query result
-        """
-        return self.member_dao.update_member_sequencing_file(member,
-                                                             job_run_id,
-                                                             seq_file_name)
-
```
j-kanuch, line 2099: Appreciate the cleanup!

rdr_service/alembic/versions/377e8ced743f_adding_aw3_manifest_id.py
```diff
@@ -0,0 +1,66 @@
...
+    drop_table_history_table("genomic_set_member", op)
+    op.add_column('genomic_set_member', sa.Column('aw3_manifest_file_id', sa.Integer(), nullable=True))
+    op.create_foreign_key(None, 'genomic_set_member', 'genomic_manifest_file', ['aw3_manifest_file_id'], ['id'])
+
+    op.execute(
+        "Update genomic_set_member gsm"
+        "Inner Join genomic_job_run gjr"
+        "On gsm.aw3_manifest_job_run_id = gjr.id"
+        "Inner join genomic_file_processed gfp"
+        "On gjr.id = gfp.run_id"
```
j-kanuch: We're going to need to do this in a script, this join isn't a 1-to-1 anymore due to the 4k cutoff. I know you spent a lot of time getting this to work, but keep the query as we may be able to use parts of it in the script. We can create a new ticket for the back-fill and I can fill in the details.

j-kanuch: On second thought, hold off on removing and let's discuss alternatives. Maybe we can add a date filter and save us some work.

j-kanuch (approved):

PR #2598, opened 2021-08-31
[DA-2167] Genomic Cron Jobs based on config value

dtharpeuno:
## Resolves *[ticket DA-2167]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2167

An entry should be added to the configs for each genomic cron job. This entry should contain a boolean flag that can be set to enable or disable a cron job.

## Tests
- [] unit tests
** No current way to test decorator added, so deployed branch to sandbox for testing to verify log output


rdr_service/offline/genomic_pipeline.py
```diff
@@ -1,39 +1,19 @@
...
-        else:
-            logging.info("Validation failed.")
-        genomic_set_file_handler.create_genomic_set_status_result_file(genomic_set_id)
-    else:
-        logging.info("No file found or nothing read from genomic set file")
-
-    genomic_biobank_manifest_handler.process_genomic_manifest_result_file_from_bucket()
-    genomic_center_manifest_handler.process_genotyping_manifest_files()
+def run_genomic_cron_job(val):
+    def inner_decorator(f):
```
j-kanuch, line 9: do you need both `inner_decorator()` and `wrapped()`?

dtharpeuno: Yea for some reason with the decorator pattern the outer function will be given any arguments you pass explicitly, and should return the inner function. The inner function will be passed the function to decorate and return the modified function. We have the same pattern with other decorators in the code like `restrict_to_gae_project`.

j-kanuch: OK, thanks!

j-kanuch (approved):

wangy70 (approved):

PR #2599, opened 2021-09-01
[DA-1013] update_stored_procedure

wangy70:
## Resolves *[DA-1013](https://precisionmedicineinitiative.atlassian.net/browse/DA-1013)*
We don't use `parent_id` to track the survey hierarchy for the code book any more, update the stored procedure `sp_get_code_module_items` and `sp_get_questionnaire_answers` to use `survey` tables.

## Tests
No test case

### DIFF
--- a/rdr_service/alembic/versions/7c0585790bc8_update_sp_get_code_module.py
+++ b/rdr_service/alembic/versions/7c0585790bc8_update_sp_get_code_module.py
@@ -75,26 +75,16 @@ sp_get_questionnaire_answers = ReplaceableObject(
              SELECT DISTINCT CONCAT('GROUP_CONCAT(IF(code_id = ', code_id, ', answer, NULL) SEPARATOR ",") AS ',
                                     `value`) as output
              FROM (
-                    select code_id, value, display, code_type, parent_id, '0' AS sort_id
+                    select code_id, value
                     from code where code_id in
                     (
-                        select distinct s.code_id
-                        from survey s
-                        where s.code_id=@code_id
-                        UNION
                         select distinct sq.code_id
                         from survey s, survey_question sq
                         where s.id=sq.survey_id
                         and s.code_id=@code_id
-                        UNION
-                        select distinct sqo.code_id
-                        from survey s, survey_question sq, survey_question_option sqo
-                        where s.id=sq.survey_id and sq.id=sqo.question_id
-                        and s.code_id=@code_id
                     )
                     order by code_id
                   ) b
-             WHERE b.code_type = 3
          ) AS temp;
 
     SET @sql = CONCAT('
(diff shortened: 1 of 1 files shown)


rdr_service/alembic/versions/7c0585790bc8_update_sp_get_code_module.py
```diff
@@ -0,0 +1,188 @@
...
+                        and s.code_id=@code_id
+                        UNION
+                        select distinct sqo.code_id
+                        from survey s, survey_question sq, survey_question_option sqo
+                        where s.id=sq.survey_id and sq.id=sqo.question_id
+                        and s.code_id=@code_id
+                    )
+                    order by code_id
+                  ) b
+             WHERE b.code_type = 3
```
kenny-skaggs: If this subquery is only looking for question codes, it may be better to look at just the codes from `survey_question` table (rather than getting all of the codes for a survey and filtering by the type on the code table. The `code_type` column isn't very reliable anymore either, since a given code can be reused as different types.

pegbertsch: @kenny-skaggs Can a code be reused as a different type, or just be reused as the same type but across different surveys?  Do you have an example of it being used as a different type?  (Afraid the PDR generators wouldn't handle that)

@wangy70 FYI:   I'm trying to reconcile this logic with two places in the PDR-related code, since we basically had to try and recreate these stored procedures in Python code.  Maybe they can act as a comparison point to see if we're on the same page/getting consistent results.  First, in the BQ models, when we try to find all the question codes associated with a module:
https://github.com/all-of-us/raw-data-repository/blob/810fd28ef2a5df2e45712edae4e36d520c105c15/rdr_service/model/bq_questionnaires.py#L70

...and then when we are looking for the answers from a specific response payload (where we will "roll up" answers from the previous response if the more recent response did not have that question/answer code in the partial payload):
https://github.com/all-of-us/raw-data-repository/blob/810fd28ef2a5df2e45712edae4e36d520c105c15/rdr_service/resource/generators/participant.py#L1462

kenny-skaggs: @pegbertsch, it doesn't look like any codes are reused as different types yet. And maybe that's not a coincidence, but I know the possibility has been brought up before. It seems like it could be a good way of 'future proofing' if we already have the capability of recognizing them as questions through the `survey_question` table rather than reading from the `code_type` column.

wangy70: @kenny-skaggs Good catch, done!

kenny-skaggs (approved):

pegbertsch (approved):

PR #2603, opened 2021-09-03
[DA-2126] MIGRATION - Modify AW2F workflow

j-kanuch:
The previous design for the AW2F workflow used a 2/3rds feedback threshold to trigger the AW2F. This required that for each AW1 that the RDR ingested, the RDR had to ingest AW2 manifests until 2/3rds of the records from the AW1 had been ingested. It was at that point the AW2F manifest was sent. 
The RDR did not send any additional versions of the AW2F manifest after the first AW2F manifest. Because of the 2/3rds threshold constraint, a periodic backlog was created that required manual intervention to remedy every 2-3 months. This PR solves the backlog issue.

    1. The 2/3rds threshold and associated logic to send the AW2F manifest was removed.

    2. The AW2F workflow was modified be sent after the AW1 upload date + 60 days, regardless of the feedback record count.

    3. A monthly job was created to alleviate the AW2F backlog for any data ingested after the initial AW2F manifest was sent.

    4. An AW2F versioning system was implemented to allow for multiple AW2Fs to be generated against a single AW1.

## Tests
- [x] unit tests


dtharpeuno (approved):
Nice job!

Would you mind deleting this instance var, https://github.com/all-of-us/raw-data-repository/blob/49661521de0f6172ea707f3bf6cdc6142ea71a7c/rdr_service/genomic/genomic_job_controller.py#L92, since we are not utilizing it in the controller anymore?

wangy70 (approved):

PR #2606, opened 2021-09-08
[DA-2152] Generating Biobank cumulative received report based on schedule

kenny-skaggs:
## Resolves *[DA-2152](https://precisionmedicineinitiative.atlassian.net/browse/DA-2152)*
The Biobank would like to receive a cumulative received report quarterly. The cumulative report will contain the same information as the normal report, but will have a larger time window.

This changes the logic so that each time the received report is generated, a check is made for whether a cumulative report should be generated as well.

## Tests
- [x] unit tests


rdr_service/offline/biobank_samples_pipeline.py
```diff
@@ -354,7 +358,45 @@ def _query_and_write_withdrawal_report(exporter, file_path, report_cover_range,
...
+        "kit_id_system": _KIT_ID_SYSTEM,
+        "tracking_number_system": _TRACKING_NUMBER_SYSTEM,
+        "n_days_ago": start_date,
+        "dv_order_filter": 0
+    }
+
+
+def _query_and_write_received_report(exporter, report_path, query_params, report_predicate):
+    received_report_select = _RECONCILIATION_REPORT_SELECTS_SQL
+    if config.getSettingJson('enable_biobank_manifest_received_flags', default=False):
```
dtharpeuno: I don't see an entry for this on the prod config? 
Can we also make an entry for this on `rdr_service/config.py` as to conform to how we pull the other config data?
`if config.getSettingJson(config.ENABLE_BIOBANK_RECIEVED_FLAGS, default=False):`

dtharpeuno (approved):
Nice! Just left the one comment.

wangy70 (approved):

rdr_service/offline/biobank_samples_pipeline.py
kenny-skaggs: Sorry for the delayed response. There's no entry for it yet, since we're not sure yet when they would want it to run, but I'll be updating the config when we know.

And thanks for catching the config strings. Updated the code to read from the module.

dtharpeuno: Nice! Thanks!

PR #2614, opened 2021-09-14
[DA-2160] Adding ability to sync consents at HPO level

kenny-skaggs:
## Resolves *[DA-2160](https://precisionmedicineinitiative.atlassian.net/browse/DA-2160)*
For some awardees we need to sync their consents at the awardee level. So even if some of their participants are not paired to an organization, we should still copy the consent files of those participants to the awardee's bucket.

Previously the consent sync code would look only for participants that were paired at the organizational level, this updates the code to not require the organization (and sets the code up to be able to sync to a 'no organization' bucket in the awardee's bucket.

The consent sync configuration structure will also be updated to have two sections: one for organization setting (similar to how it works now), and an hpo section that would sync all participants for that hpo/awardee.

## Tests
- [x] unit tests


dtharpeuno (approved):
👍

wangy70 (approved):

PR #2618, opened 2021-09-16
[DA-2154] Implement Weekly Consent Issues report (manual tool)

pegbertsch:
This is a major refactoring of the `consent-report` tool to add the new weekly report.  Because the daily and weekly report share a considerable amount of logic, things have been reorganized into a ConsentReport base class along with DailyConsentReport and WeeklyConsentReport child classes.

The report generation has also been converted into a batch mode, so that all the report data is populated and then sent to the googlesheets API via a single gspread `batch_update()` call.  This avoids rate limit errors writing data to the sheet.

A `gspread-formatting` package is used to facilitate applying different formats to the report, with are also populated with the report data and then written with a single `format_cell_ranges()` batch call.   Because we no longer intend to integrate these spreadhseet report generators into a cron job and they are expected to be short-lived, the `gspread-formatting` package just needs to be installed on the local machine where the report is run (will not be added to RDR app requirements)

Some minor content changes and fixes were also included for the DailyConsentReport

Due to the wholesale reorganization of the code, the diff view may not be overly helpful in reviewing specific code changes.  The main areas of focus for reviewing the new version should be:

- SQL statements used to pull the data for the reports
- The WeeklyConsentReport class (new)

## Tests
Output from manual testing of the report generation can be viewed [here](https://docs.google.com/spreadsheets/d/17Ik-HulERnB7KKugQg9rB6BPLn5J8j_GXTFOmHSeCQw/edit?usp=sharing) 

Tabs with single dates are the DailyConsentReport consent, tabs with a date range are the WeeklyConsentReport


kenny-skaggs (approved):

rdr_service/tools/tool_libs/consent_validation_report.py
```diff
@@ -131,171 +134,152 @@
...
+    See: https://stackoverflow.com/questions/17215400/format-string-unused-named-arguments
+    Used with str.format_map() to allow partial formatting/replacement of placeholder values in a string
+    without incurring KeyError. E.g.: '{lastname}, {firstname} {lastname}'.format_map(SafeDict(lastname='Bond'))
+    yields the partially formatted result: 'Bond, {firstname} Bond'
+
+     This helper class will be used when formatting SQL template strings that need to have some of their placeholders
+     filled in before the values for the others can be determined.
+    """
+    def __missing__(self, key):
+        return '{' + key + '}'
```
dtharpeuno: Can we use an f-string here instead of concatenation and anywhere else where we use it or `format()` method?

dtharpeuno (approved):

rdr_service/tools/tool_libs/consent_validation_report.py
pegbertsch: I can change this here (grabbed this from stackoverflow).  For me, using `format()` is sometimes a compromise when the length of the f-string is getting problematic.  But I'll look for cases where that wasn't a factor.

PR #2622, opened 2021-09-20
[DA-2203] Remove consent status for draft participants

kenny-skaggs:
## Resolves *[DA-2203](https://precisionmedicineinitiative.atlassian.net/browse/DA-2203)*
This adds a script that will remove participant summaries, consent file records, and questionnaire responses for participants that have been identified by Vibrent as only having a draft version of the primary consent on file (and not having actually signed it). This will reset the participants back to an unconsented state.

The script uses a file listing the participant ids. I've attached the file provided by Vibrent to DA-2203 and will use those participant ids. The code also checks that the participants haven't yet sent a consent in case they consented fully after our checks and before running this script. The code also makes sure no incoming data would be deleted if they consent while the script is running.

## Tests
- [ ] unit tests
No unit tests were added, as this code will be removed once the participants are unconsented.

### DIFF
--- a/rdr_service/tools/tool_libs/unconsent.py
+++ b/rdr_service/tools/tool_libs/unconsent.py
@@ -4,12 +4,12 @@ from sqlalchemy import and_
 from sqlalchemy.orm import Session
 
 from rdr_service.model.code import Code
-from rdr_service.model.utils import from_client_participant_id
 from rdr_service.model.consent_file import ConsentFile
 from rdr_service.model.participant_summary import ParticipantSummary
 from rdr_service.model.questionnaire import QuestionnaireQuestion
 from rdr_service.model.questionnaire_response import QuestionnaireResponse, QuestionnaireResponseAnswer,\
     QuestionnaireResponseExtension
+from rdr_service.model.utils import from_client_participant_id
 from rdr_service.tools.tool_libs.tool_base import cli_run, logger, ToolBase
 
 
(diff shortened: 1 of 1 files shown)


dtharpeuno (approved):

rdr_service/tools/tool_libs/unconsent.py
```diff
@@ -0,0 +1,119 @@
...
+        with self.get_session() as session:
+            for participant_id in self._load_participant_ids():
+                # Retrieve the participant summary
+                summary_query = session.query(ParticipantSummary).filter(
+                    ParticipantSummary.participantId == participant_id
+                )
+                if not self.args.dry_run:
+                    # Obtain a lock on the participant summary to prevent possible
+                    # race conditions with incoming responses
+                    summary_query = summary_query.with_for_update()
```
pegbertsch, line 35: This may be a non-issue or may expose another race condition if it's possible for a consent to arrive during the unconsent check, but is it preferable to confirm if the participant has a signed consent first and needs the crud operation (continue to next pid if not) before obtaining a lock?

rdr_service/tools/tool_libs/unconsent.py
```diff
@@ -0,0 +1,119 @@
...
+                    QuestionnaireResponse.participantId == participant_id,
+                    QuestionnaireResponse.questionnaireResponseId == QuestionnaireResponseAnswer.questionnaireResponseId
+                )
+            ).join(
+                QuestionnaireQuestion,
+                QuestionnaireQuestion.questionnaireQuestionId == QuestionnaireResponseAnswer.questionId
+            ).join(
+                Code,
+                and_(
+                    Code.value == 'ExtraConsent_Signature',
```
pegbertsch, line 77: Can you clarify, is this always supposed to be present in a valid/signed Primary Consent?  (I've been looking at several PrimaryConsent payloads today where I haven't seen this in the payload)

rdr_service/tools/tool_libs/unconsent.py
kenny-skaggs: We'd need the lock before the check. If we do the check and then lock, then we'd potentially have received (or be in the middle of receiving) a consent after the check but before the lock was obtained. This way we get the lock first and there isn't any chance for anything to be saved between the check and the removal of the data.

This strategy locks participants that wouldn't need to be locked, but the check is quick so wouldn't hold the lock for very long.

kenny-skaggs: After some offline discussion updated to lock Participant record instead, matching what the QuestionnaireResponseDao does. Thanks for helping find that 😄

pegbertsch (approved):

PR #2623, opened 2021-09-21
[PDR-441] Investigate why PDR enrollment status calculations are off from RDR

robabram:
A boolean json value was used for a newly added field to PDR, BigQuery does not accept boolean values from JSON data. 
Because of the weekend rebuild of all PDR data, there were 305k records that could not be sent to BigQuery.
This PR fixes that issue and another issue that was allowing questionnaire responses with authored timestamps from before the program started with the new enrollment status calculation code.

There is a new tool that is used to convert the boolean value in the participant records in the 'bigquery_sync' table to 0 or 1, which allows the records to be sent to BigQuery.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/tools/tool_libs/bq_participant_fix.py
+++ b/rdr_service/tools/tool_libs/bq_participant_fix.py
@@ -1,0 +1,116 @@
+#! /bin/env python
+#
+# Template for RDR tool python program.
+#
+
+import argparse
+
+# pylint: disable=superfluous-parens
+# pylint: disable=broad-except
+import datetime
+import json
+import logging
+import sys
+
+from rdr_service.services.system_utils import setup_logging, setup_i18n, print_progress_bar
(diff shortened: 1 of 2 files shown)


rdr_service/tools/tool_libs/bq_participant_fix.py
```diff
@@ -0,0 +1,116 @@
...
+
+            for pm in resource['pm']:
+                pm['pm_final'] = 1 if pm['pm_final'] else 0
+
+            ins_args = [json.dumps(resource), datetime.datetime.utcnow(), id_]
+            cursor.execute(ins_sql, ins_args)
+            db_conn.commit()
+
+            pos += 1
+            print_progress_bar(
```
dtharpeuno, line 83: NIce

dtharpeuno (approved):

pegbertsch (approved):

PR #2624, opened 2021-09-21
[DA-2156] HealthPro consents 

dtharpeuno:
## Resolves *[ticket DA-2156]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2156

Implementation for copying physical consents paths to healthpro designated buckets for ingestion by HealthPro application via Participant Summary API
https://precisionmedicineinitiative.atlassian.net/browse/DA-2105

## Tests
- [x] unit tests


rdr_service/model/database.py
```diff
@@ -24,6 +24,7 @@
 from rdr_service.model.deceased_report_import_record import DeceasedReportImportRecord
 from rdr_service.model.ehr import EhrReceipt, ParticipantEhrReceipt
 from rdr_service.model.hpo import HPO
+from rdr_service.model.hpro_consent_files import HealthProConsentFile
```
dtharpeuno, line 27: Codacy complaining about this with no way to rectify or ignore the error.

j-kanuch (approved):

rdr_service/dao/hpro_consent_dao.py
```diff
@@ -0,0 +1,44 @@
...
+            needed_consents = session.query(
+                ConsentFile
+            ).outerjoin(
+                HealthProConsentFile,
+                HealthProConsentFile.consent_file_id == ConsentFile.id
+            ).filter(
+                ConsentFile.file_exists == 1,
+                ConsentFile.file_path.isnot(None),
+                HealthProConsentFile.id.is_(None)
+            )
```
kenny-skaggs, line 29: This would transfer consents that have been marked as invalid (missing signature, incorrect version, ...). Is that something we'd want to avoid?

rdr_service/dao/participant_summary_dao.py
```diff
@@ -966,8 +971,30 @@ def get_record_from_attr(self, *, attr, value):
...
+        consents_map = {
+            ConsentType.PRIMARY: 'consentForStudyEnrollment',
+            ConsentType.CABOR: 'consentForCABoR',
+            ConsentType.EHR: 'consentForElectronicHealthRecords',
+            ConsentType.GROR: 'consentForGenomicsROR'
+        }
+        participant = result['participantId']
+        records = self.hpro_consent_dao.get_by_participant(participant)
+
+        for key, value in consents_map.items():
```
kenny-skaggs: suggestion: Does it make sense to name these something that describe what they are in terms of what data they hold rather than `key` and `value`?

rdr_service/tools/tool_libs/hpro_consent.py
```diff
@@ -0,0 +1,99 @@
...
+
+# Tool_cmd and tool_desc name are required.
+# Remember to add/update bash completion in 'tool_lib/tools.bash'
+tool_cmd = "hpro-consents"
+tool_desc = "Tool for initial and subsequent large backfills of consent files to healthpro-consent bucket"
+
+
+class HealthProConsentTool(ToolBase):
+    def __init__(self, args, gcp_env: GCPEnvConfigObject):
+        super().__init__(args, gcp_env)
```
kenny-skaggs: Does having this `__init__` override do anything? Can it be removed (to have the super class be called by default)?

rdr_service/tools/tool_libs/hpro_consent.py
```diff
@@ -0,0 +1,99 @@
...
+
+            if not transfer:
+                _logger.warning(f'Healthpro consent {src} failed to transfer to {dest}')
+                transfer_failures.append({
+                    'original': consent.file_path,
+                    'destination': dest.split('gs://')[1],
+                })
+                continue
+
+            hpro_consents.dao.insert(obj)
```
kenny-skaggs: The logic here for transferring a file (constructing a source and destination, calling gcp, and recording the result) is similar between the service class and here. Do you think it makes sense to encapsulate it in a method in the service class and call it from here?

rdr_service/dao/participant_summary_dao.py
```diff
@@ -966,8 +971,30 @@ def get_record_from_attr(self, *, attr, value):
...
+            if has_consent_path:
+                result[value_path_key] = has_consent_path[0].file_path
+
+        return result
+
     def to_client_json(self, model: ParticipantSummary):
         result = model.asdict()
+        if self.user_roles == ['healthpro']:
+            result = self.get_hpro_consent_paths(result)
+
```
kenny-skaggs, line 995: I'm a little worried about how this will affect performance of the participant summary API for HPRO. They sometimes request summaries in batches of 1k, and for each summary this would make a separate query to the database. It seems like it would be good to fetch all of them upfront in a single query.

rdr_service/dao/hpro_consent_dao.py
pegbertsch: Was part of the use case for this effort that the end users could visually inspect/see issues with the consent files for themselves? But, unsure if they would also want to flag the file in the front end as having an issue and if we need a way to communicate that status to HPRO

dtharpeuno: @kenny-skaggs yea this is something that we would want to avoid, basically I was going off the initial tech design and was under the impression that if the file exists and the file path is not null then consent has been validated and is ready for transfer. 

@pegbertsch that is a great question but in talks with the HPRO team they are just wanting to see if the file path exists for this first pass. 

@kenny-skaggs  looking at the consent file model, should the only other values to consider in the query for transfer be:`is_signature_valid` and `is_signing_date_valid`?

rdr_service/tools/tool_libs/hpro_consent.py
dtharpeuno: No problem I went ahead and made the change to _lean_ into service more from what I was utilizing in the tool.

rdr_service/tools/tool_libs/hpro_consent.py
dtharpeuno: Done.

rdr_service/dao/participant_summary_dao.py
dtharpeuno: Made the change for readability.

rdr_service/dao/hpro_consent_dao.py
kenny-skaggs: I would look at the `sync_status` value. If it's `READY_FOR_SYNC` or `SYNC_COMPLETE` then it's valid and should be sent. Otherwise it should be left alone until it changes (if it does).

rdr_service/dao/participant_summary_dao.py
dtharpeuno: Understood, added batch fetching.
https://github.com/all-of-us/raw-data-repository/pull/2624/commits/1601bff476fd2f8cb7a70f9eba51e175ff226b03

rdr_service/api/participant_summary_api.py
```diff
@@ -35,6 +39,8 @@ def get(self, p_id=None):
         auth_awardee = None
         user_email, user_info = get_validated_user_info()
         self.user_info = user_info
+        self.dao.user_roles = self.user_info.get('roles')
```
kenny-skaggs: Is `user_roles` used in the DAO anymore?

rdr_service/dao/hpro_consent_dao.py
dtharpeuno: Sync status updated, thanks.

kenny-skaggs (approved):

PR #2644, opened 2021-10-04
Script for data-dictionary update

kenny-skaggs:
## Resolves *no ticket*
Deploying the app has a step for automatically updating the data-dictionary. But unfortunately the google API doesn't always respond well and errors out when trying to update the data-dictionary during the deploy process. This PR adds a script for separately running the update in case it fails during the deploy.

The logic for the interactive process of updating the data-dictionary is moved into a reusable function in the updater class. This requires context switching (between different accounts for different access), so requires access to the tool running it. The tool passes itself as a parameter to the updater so that context switching can happen.

## Tests
- [x] unit tests

Dictionary updating should be covered by existing tests.


dtharpeuno (approved):
Neat!

wangy70 (approved):

PR #2645, opened 2021-10-04
[DA-2213] Allowing for codes longer than 50 chars in curation ETL

kenny-skaggs:
## Resolves *[DA-2213](https://precisionmedicineinitiative.atlassian.net/browse/DA-2213)*
When exporting to curation, RDR has been truncating code values to 50 characters. But according to OHDSI, `VARCHAR(50)` in their documentation means that the minimum number of characters that should be in a code is 50 and that they can be longer than 50 characters.

This PR uses the full code value in the curation ETL, rather than the truncated version of the string. The truncating logic has been left in place, in case other downstream systems use it.

This also updates the data-dictionary to use the full code value.

## Tests
- [x] unit tests

Existing unit tests have been modified to no longer use the short code value.

### DIFF
--- a/rdr_service/model/code.py
+++ b/rdr_service/model/code.py
@@ -49,9 +49,12 @@ class _CodeBase(object):
     @rdr_dictionary_show_unique_values
     """
     value = Column("value", String(80), nullable=False)
-    # OMOP codes are supposed to be at most 50 characters long; for legacy codes that exceeded this
-    # limit, we populate a shortened version for use in OMOP here. Otherwise, shortValue is identical
-    # to value.
+    # OMOP codes can be any length according to OHDSI:
+    # https://ohdsi.github.io/CommonDataModel/dataModelConventions.html#general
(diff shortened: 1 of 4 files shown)


rdr_service/model/code.py
```diff
@@ -49,9 +49,12 @@ class _CodeBase(object):
     @rdr_dictionary_show_unique_values
     """
     value = Column("value", String(80), nullable=False)
```
j-kanuch, line 51: This isn't a big issue, but is a limit of 80 characters a safe-enough buffer for longer codes?

kenny-skaggs: I believe so. I'm not entirely sure about specs in other systems, but we haven't hit codes that long yet and we've rarely gone over 50 chars. I imagine we can handle that scenario and investigation later if we need to.

j-kanuch: Sounds good.

j-kanuch (approved):

pegbertsch (approved):

PR #2648, opened 2021-10-06
Filter CE participants from HPro consents

dtharpeuno:
Need to filter CE participants from sending consents to Healthpro in lieu of validation process for CE participants is not active. Filtering is done based on file path in consent files model.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/hpro_consent_dao.py
+++ b/rdr_service/dao/hpro_consent_dao.py
@@ -25,6 +25,7 @@ class HealthProConsentDao(UpdatableDao):
                     ConsentSyncStatus.READY_FOR_SYNC,
                     ConsentSyncStatus.SYNC_COMPLETE
                 ]),
+                ConsentFile.file_path.notlike('%ce-uploads%'),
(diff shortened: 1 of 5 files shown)


j-kanuch (approved):

rdr_service/dao/hpro_consent_dao.py
```diff
@@ -25,6 +25,7 @@ def get_needed_consents_for_transfer(self, limit=None):
                     ConsentSyncStatus.READY_FOR_SYNC,
                     ConsentSyncStatus.SYNC_COMPLETE
                 ]),
+                ConsentFile.file_path.notlike('%ce-uploads%'),
```
pegbertsch, line 28: It's probably not a significant performance difference (does add a join), but you can also join ConsentFile to Participant on `participant_id` and filter on the `participant.participant_origin` field (comparison of specific values vs. dealing with a `%match%` search.   Vibrent participants have `vibrent` as their origin, CE is `careevolution`

kenny-skaggs (approved):

PR #2650, opened 2021-10-07
[PDR-435] Add winter cope survey to PDR.

robabram:
Add support for winter COPE survey for PDR.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/model/bq_questionnaires.py
+++ b/rdr_service/model/bq_questionnaires.py
@@ -910,6 +910,80 @@ class BQPDRSDOHView(BQModuleView):
     _show_created = True
 
 
+class BQPDRCOPEVaccine3Schema(_BQModuleSchema):
+    """ COPE Vaccine Survey 3 (Winter 2021) """
+    _module = 'cope_vaccine3'
+    _force_boolean_fields = (
+        'cdc_covid_xx_b_firstdose_other',
+        'cdc_covid_xx_symptom_cope_350',
+        'cdc_covid_xx_b_seconddose_other',
+        'cdc_covid_xx_symptom_seconddose_cope_350',
+        'dmfs_29a',
+        'dmfs_29_seconddose_other',
+        'cdc_covid_xx_b_dose3_other',
+        'cdc_covid_xx_symptom_cope_350_dose3',
+        'cdc_covid_xx_type_dose3_other',
+        'dmfs_29_additionaldose_other',
+        'cdc_covid_xx_b_dose4_other',
+        'cdc_covid_xx_symptom_cope_350_dose4',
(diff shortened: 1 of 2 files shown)


rdr_service/model/bq_questionnaires.py
```diff
@@ -910,6 +910,80 @@ class BQPDRSDOHView(BQModuleView):
     _show_created = True
 
 
+class BQPDRCOPEVaccine3Schema(_BQModuleSchema):
+    """ COPE Vaccine Survey 3 (Winter 2021) """
+    _module = 'cope_vaccine3'
+    _force_boolean_fields = (
```
dtharpeuno, line 916: Sheesh...

dtharpeuno (approved):

rdr_service/model/bq_questionnaires.py
robabram: There are 6 or 7 distinct covid surveys, their code id naming hasn't been very consistent.

rdr_service/model/bq_questionnaires.py
```diff
@@ -910,6 +910,80 @@ class BQPDRSDOHView(BQModuleView):
...
+        'cdc_covid_xx_b_firstdose_other',
```
pegbertsch, line 917: You have more visibility than I do into the codebook extract, but are all of these codes (say, dose5 fields and up) expected to have response values in this next Winter Survey?   Or are they trying to create a generic set of codes codebook will also apply to future COPE Vaccine modules should there also be additional boosters?

Do we know which codes the dashboard team is expecting to extract into the `v_pdr_mod_cope_tableau` view?  would have to update all the SELECTs that are part of the UNION ALL logic

pegbertsch (commented):
Do you need to add the new module to the model/__init__.py file?

robabram:
Yes, good catch.

rdr_service/model/bq_questionnaires.py
robabram: I don't have visibility in to that.

rdr_service/model/bq_questionnaires.py
pegbertsch: There are some shared "forced boolean" fields between all the surveys that included vaccine questions (Feb Cope + the minute surveys?)

But I don't understand why this survey has questions about so many doses.   (Haven't seen the codebook)

rdr_service/model/bq_questionnaires.py
pegbertsch: Can you tell what the display values are for all those question codes, or do we have to wait until Kenny has run his import to look at the display strings in RDR?

pegbertsch (approved):

PR #2652, opened 2021-10-07
[DA-2220] Data quality checks for questionnaire responses

kenny-skaggs:
## Resolves *[DA-2220](https://precisionmedicineinitiative.atlassian.net/browse/DA-2220)*
This PR adds additional checks for questionnaire responses to make sure we're receiving valid data from PTSC. The following four checks are made, and if they fail validation we will get a log message printed out:
- authored date after participant's suspension or withdrawal time
- missing or malformed authored date
- authored date before the survey launch date (approximated by using the questionnaire's created date)
- authored date any date in the future (checked using the datetime that the RDR receives the response payload)

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/services/data_quality.py
+++ b/rdr_service/services/data_quality.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from datetime import datetime
+from datetime import datetime, timedelta
 import logging
 from sqlalchemy import func, or_
 from typing import List, Type
@@ -20,6 +20,13 @@ class _ModelQualityChecker(ABC):
     @abstractmethod
     def run_data_quality_checks(self, for_data_since: datetime = None):
         ...
+
+    @classmethod
+    def _date_less_than_or_equal(cls, earlier_date: datetime, later_date: datetime, tolerance : timedelta = None):
+        if tolerance is None:
+            tolerance = timedelta(seconds=0)
+
+        return earlier_date <= (later_date + tolerance)
 
 
 class DeceasedReportQualityChecker(_ModelQualityChecker):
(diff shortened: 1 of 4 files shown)


rdr_service/services/data_quality.py
```diff
@@ -111,26 +118,43 @@ def run_data_quality_checks(self, for_data_since: datetime = None):
...
+                        f'Response {response_id} authored with future date '
+                        f'of {authored_time} (received at {created_time})'
+                    )
+                if authored_time < participant_signup_time:
+                    logging.error(
+                        f'Response {response_id} authored at {authored_time} '
+                        f'but participant signed up at {participant_signup_time}'
+                    )
+                if suspension_datetime and authored_time > suspension_datetime:
+                    logging.error(f'Response {response_id} authored for suspended participant')
```
j-kanuch, line 152: Is there a plan to do something with these validation errors in addition logging them? For example, will there be alerts set up in GCP, or will the invalid requests be stored in the database so we can provide PTSC a list of records to re-send? I see the importance of logging them, but I may be missing context as to what else is happening to handle these errors.

kenny-skaggs: Yeah, sorry I hadn't described that. The idea will be to monitor the logs for them and notify PTSC if any are seen. A possible next step could be to have them set up as Slack alerts, but for now it seems to make the most sense to view them through GCP's logs.

j-kanuch: Yes, GCP alerts the will be helpful. You may want to add a common string to all of the log messages, like "Questionnaire Validation Error: Response ..." or something similar. Then you can build a log-based metrics query in GCP based on that string. That may save you time in the future.

j-kanuch (approved):

dtharpeuno (approved):

PR #2674, opened 2021-10-19
Fixing email module import issues

kenny-skaggs:
## Resolves *no ticket*
Adding the `email` module caused an error with starting the gcp_daemon. This changes the name of the module to prevent it from conflicting with an existing module that SendGrid uses.

This renames the module to `email_service` to prevent the name collision with the external module.

## Tests
- [ ] unit tests

Should be covered by the existing unit tests.


dtharpeuno (approved):
Thanks for fixing this

j-kanuch (approved):

PR #2678, opened 2021-10-20
[DA-2241] Renaming 'gender' to 'sex' for MayoLINK API

kenny-skaggs:
## Resolves *[DA-2241](https://precisionmedicineinitiative.atlassian.net/browse/DA-2241)*
MayoLINK is changing a field name from `gender` to `sex` in their `create` endpoint. This updates the BiobankOrderDao and MailkitOrderDao to reflect that change.

The PR also encapsulates the logic for building the request structure for MayoLINK out into the same class for making the request to the server.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/api/mayolink_api.py
+++ b/rdr_service/api/mayolink_api.py
@@ -123,7 +123,7 @@ class MayoLinkApi:
                     'last_name': order.last_name,
                     'middle_name': '',
                     'birth_date': '3/3/1933',
-                    'gender': order.sex,
+                    'sex': order.sex,
                     'address1': order.address1,
                     'address2': order.address2,
                     'city': order.city,
--- a/rdr_service/api/mayolink_api.py
+++ b/rdr_service/api/mayolink_api.py
@@ -1,13 +1,49 @@
+from dataclasses import dataclass
 import json
 import httplib2
 import logging
+from werkzeug.exceptions import ServiceUnavailable
 import xml.etree.ElementTree as ET
 import xmltodict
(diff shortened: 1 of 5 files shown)


wangy70 (approved):

rdr_service/api/mayolink_api.py
```diff
@@ -32,9 +77,9 @@ def _get_credentials(cls, credentials_key):
 
         return credentials.get('username'), credentials.get('password'), credentials.get('account')
 
-    @auth_required(RDR_AND_PTC)
-    def post(self, order):
-        xml = self.__dict_to_mayo_xml__(order)
+    def post(self, order: MayoLinkOrder):
+        check_auth(RDR_AND_PTC)
```
dtharpeuno, line 81: Is there a reason to move away from the decorator?

dtharpeuno (approved):
Nice job on the refactor, dig the dataclasses usage. Just left a question about the decorator.

rdr_service/api/mayolink_api.py
kenny-skaggs: The decorator complicates things in a few ways when the only purpose seems to be to make sure we're only getting RDR and PTC users in this code. It was frustrating getting a test around it here since the `auth_required` method would require the use of the database for saving a request log. But I think that also was causing some strangeness in the request log table when something went wrong with the requests: we would see two records being logged for the same request. I think this fixes it, but even if it doesn't it definitely simplifies the test.

PR #2684, opened 2021-10-21
Improvements for data quality checks

kenny-skaggs:
## Resolves *no ticket*
The data quality checks are still flagging some responses that aren't really errors. Vibrent sends responses to later versions of the Questionnaires but for the same concept. The way the quality checks were set up they all look like they're authored earlier than they should have been.

This also removes the check for responses with no answers, since the Health Data Consent modules don't have answers within them.

This looks for the earliest version of a questionnaire with the same module code. I think this is the best approximation of the release date that we have for now. If the authored date is earlier than that, then the response is flagged.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/services/data_quality.py
+++ b/rdr_service/services/data_quality.py
@@ -1,13 +1,16 @@
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
 import logging
-from sqlalchemy import func, or_
+from sqlalchemy import and_, func, or_
+from sqlalchemy.orm import aliased
 from typing import List, Type
 
+from rdr_service.model.code import Code
 from rdr_service.model.deceased_report import DeceasedReport
 from rdr_service.model.participant import Participant
 from rdr_service.model.patient_status import PatientStatus
-from rdr_service.model.questionnaire import Questionnaire, QuestionnaireHistory, QuestionnaireQuestion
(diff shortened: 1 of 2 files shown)


wangy70 (approved):

rdr_service/services/data_quality.py
```diff
@@ -120,19 +127,39 @@ def run_data_quality_checks(self, for_data_since: datetime = None):
...
+                )
+            )
             .group_by(QuestionnaireResponse.questionnaireResponseId)
         )
         if for_data_since is not None:
             query = query.filter(QuestionnaireResponse.created >= for_data_since)
 
         for response_id, created_time, authored_time, participant_signup_time, suspension_datetime,\
-                withdrawal_datetime, answer_count, questionnaire_created_time in query.all():
+                withdrawal_datetime, min_questionnaire_created_time, module_code in query.all():
```
dtharpeuno, line 162: Not in this PR, but should we write these logs to a table for record-keeping and analyses in the future?

dtharpeuno (approved):

rdr_service/services/data_quality.py
kenny-skaggs: I think that's a good idea, but maybe we can wait to see how often errors come up and how they are handled so we'd know how much value there would be in it.

PR #2686, opened 2021-10-22
[PDR-473] Move UBR calculation code to a resource calculator.

robabram:
Move UBR calculation code to a resource calculator class.
Change code to only pull the first 'TheBasics' response when calculating UBR.
Prepare for next UBR update to switch from boolean values to 3 possible UBR values.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/calculators/participant_ubr.py
+++ b/rdr_service/resource/calculators/participant_ubr.py
@@ -7,7 +7,6 @@
 #
 from dateutil.relativedelta import relativedelta
 from dateutil.parser import parse
-from typing import List
 
 from rdr_service.resource.helpers import RURAL_ZIPCODES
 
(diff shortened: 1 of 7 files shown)


rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -0,0 +1,198 @@
...
+            if consent_date:
+                break
+            for consent in consents:
+                if consent['consent'] == consent_type and \
+                    consent['consent_value'] in ['ConsentPermission_Yes', 'DVEHRSharing_Yes']:
+                    consent_date = consent['consent_date']
+                    break
+
+        if consent_date:
+            rd = relativedelta(consent_date, answer)
```
pegbertsch: Is this where you were going to use enrl_participant_time so you didn't unintentionally pick up the wrong/later ConsentPII authored date for participants who had multiple ConsentPII payloads sent? (E.g., updating name/address/email and stuff vs. when they signed the consent?)

rdr_service/resource/generators/participant.py
```diff
@@ -1437,111 +1430,71 @@ def is_replay(prev_data_dict, data_dict, ignore_keys=None):
 
         return True
 
-    def _calculate_ubr(self, summary):
+    def _calculate_ubr(self, p_id, summary, ro_session):
```
pegbertsch, line 1433: Just a heads up:   I still need to confirm if, since they enabled using "partial" Basics payloads to transmit secondary contact details only in July, we have any pids where their first TheBasics response is not actually a full survey.   Ticket is DA-2192.   It could be none or a small number of pids.  Worst case may be in the hundreds/low thousands.

wangy70 (approved):

pegbertsch (approved):

rdr_service/resource/calculators/participant_ubr.py
robabram: Yes

PR #2687, opened 2021-10-25
Weekly consent report fixes

pegbertsch:

- Added a filter to exclude records present in the `consent_file` table for validations/consents not yet meant to be reported  in the burndown "resolved files" count (e.g., records resulting from in-progress development for primary consent update validation)
- Corrected logic in the weekly report that was applying an `authored` date range to find validation records created in the report range; use a separate `validation_end_date` to account for records added by the validation cron job that runs the following morning
- Clean up `datetime.timedelta()` calls to use consistent syntax/keyword parameter
- Added `participant_origin` to consent report SQL query (may be used in PDR data generation)

## Tests
Manually tested weekly report results against the daily reports for the same authored date range


rdr_service/tools/tool_libs/consent_validation_report.py
```diff
@@ -150,19 +151,21 @@
...
-# resolutions up to the specified end date for this report
+# resolutions up to the specified end date for this report.
 ALL_RESOLVED_SQL = """
            SELECT cf.participant_id,
                   cf.type,
                   DATE(cf.modified) AS resolved_date
            FROM consent_file cf
-           WHERE cf.sync_status = 3 AND DATE(cf.modified) <= "{end_date}"
+           -- Alias to ps to be consistent with other queries that use common filters
+           JOIN participant ps on ps.participant_id = cf.participant_id
```
dtharpeuno, line 169: Do you need this `JOIN` since you are not utilizing any logic or attributes from the `participant` table?

rdr_service/tools/tool_libs/consent_validation_report.py
```diff
@@ -150,19 +151,21 @@
...
+           WHERE cf.sync_status = 3 AND DATE(cf.modified) <= "{validation_end_date}"
```
dtharpeuno: Judging by the var name here, shouldn't this be `cf.syns_status in (2, 4)` since that enum is denoted as `OBSOLETE`?

rdr_service/tools/tool_libs/consent_validation_report.py
pegbertsch: I've corrected the SQL (applied the wrong stash).   This separate SELECT/query is being updated to be able to filter out test/ghost participant status and also apply the VIBRENT_SQL_FILTER correctly

rdr_service/tools/tool_libs/consent_validation_report.py
pegbertsch: All reports that were in NEEDS_CORRECTING will transition to an end state of OBSOLETE when a newer, successfully validated copy of the consent PDF is uploaded (that newer file gets a separate `consent_file` table entry which starts in READY_TO_SYNC and transitions to SYNC_COMPLETE after the monthly sync).  This query wants to find the reports that transitioned out of NEEDS_CORRECTING to OBSOLETE.

This is per Kenny's validation design and the need to maintain a history of the older consent PDFs that contained errors but are still archived in the upload buckets.  Each upload/retransmission of a consent generates a new PDF filename in the bucket (old PDF files that didn't pass validation are not deleted)

dtharpeuno: Understood, thanks for clarifying.

rdr_service/tools/tool_libs/consent_validation_report.py
dtharpeuno: 👍

kenny-skaggs (approved):

dtharpeuno (approved):

PR #2689, opened 2021-10-26
[DA-2246] Batch Manifest Ingestion

dtharpeuno:
## Resolves *[ticket DA-2246]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2246

Batch manifest ingestions in groups of 1000

## Tests
- [x] unit tests


j-kanuch (approved):
I'm going to comment here just to help clarify what is going on: Let's say you have 1000 rows to ingest, and `max_num` is 100. You will end up with `all_ingestions = [[row1, row2...row100], [row101,...,row200],[row201,...,row300]...]`. This is a list of 10 lists with 100 rows in each sub-list.
When `_ingest_genomic_file()` runs, it will call the specific ingestion method 10 times, once for each list in `all_ingestions`. These will run the ingestion method once on each set of 100 rows. This makes sense, because the memory issue isn't in the amount of data in the manifest being held in memory (you're still loading that in memory in `all_ingestions`). The memory issue is when we have to load the `genomic_set_member` and `gc_validation_metrics` records. Here, we'll never load more than `max_num` at a time. OK, I'm good with that. Nice work!

wangy70 (approved):

PR #2697, opened 2021-11-02
[PDR-473] Put state and zip code back in.

robabram:
Code to set State and Zip code values was accidentally removed.  This PR restores the code.

## Tests
- [x] unit tests


kenny-skaggs:
Should unit tests be added to make sure it can't be accidentally removed again?

pegbertsch (approved):

pegbertsch (approved):
I realize this fix is time-sensitive, but moving forward, I propose any time we make changes to the generators, we try to add a "sanity check" unit test and build up the unit test coverage.  Some tests of generated values are baked in to other RDR unit test cases to take advantage of their test data setups, but we probably didn't have one specifically for address details.

dtharpeuno (approved):

robabram:
It is very likely that this code will be removed from the RDR project soon and we will be refactoring all this code to work in the PDR code base.

PR #2698, opened 2021-11-02
[DA-2089] CE consent file parsing

kenny-skaggs:
## Resolves *[DA-2089](https://precisionmedicineinitiative.atlassian.net/browse/DA-2089)*
Adds parsing for CE consent files.

CE files all have the signature on the last page, and they all have the signature and date above specific text. This PR adds different classes for each consent type, but they all utilize a common wrapper object for parsing the signature data from the PDF.

Most of CE's files are structured in a way that doesn't lend itself easily to file parsing. PDFMiner's `extract_text` is used to get the raw text of the PDF to search for identifying strings to determine which consent type the PDF is for. Then each character is analyzed to find the location of the signature (as the exact positioning changes between different versions of the file). 

## Tests
- [x] unit tests


rdr_service/services/consent/files.py
```diff
@@ -476,16 +502,176 @@ def is_agreement_selected(self):
...
+
+        return None
+
+    def _get_last_page(self):
+        last_page_index = len(self.pdf.pages) - 1
+        return self.pdf.pages[last_page_index]
+
+    @classmethod
+    def _rect_shifted_up_from_footer(cls, footer_rect: Rect) -> Rect:
+        return Rect.from_edges(
```
pegbertsch, line 546: Is your process for determining these offset values documented somewhere?

rdr_service/services/consent/files.py
```diff
@@ -200,35 +206,51 @@ def _get_source_prefix(self) -> str:
...
 class CeConsentFactory(ConsentFileAbstractFactory):
     def _is_primary_consent(self, blob_wrapper: '_ConsentBlobWrapper') -> bool:
-        pass
+        pdf = blob_wrapper.get_parsed_pdf()
+        return pdf.has_text([(
+            'Consent to Join the All of Us Research Program',
+            'Consentimiento para Participar en el Programa Científico'
+        )])
 
     def _is_cabor_consent(self, blob_wrapper: '_ConsentBlobWrapper') -> bool:
```
pegbertsch, line 215: Similar to previous ask, can we refer to the expected version strings at the top of the PDF that we expect to have the matching text?

rdr_service/services/consent/files.py
```diff
@@ -200,35 +206,51 @@ def _get_source_prefix(self) -> str:
 
 class CeConsentFactory(ConsentFileAbstractFactory):
     def _is_primary_consent(self, blob_wrapper: '_ConsentBlobWrapper') -> bool:
-        pass
+        pdf = blob_wrapper.get_parsed_pdf()
+        return pdf.has_text([(
```
pegbertsch, line 213: Similar to previous ask, can we refer to the expected version strings at the top of the PDF that we expect to have the matching text for each of these `has_text` calls?

pegbertsch (approved):
Comments/questions but do not impact functionality

dtharpeuno (approved):

rdr_service/services/consent/files.py
kenny-skaggs: No, this was just from working with the files and finding the relative positioning from the labels. How do you think this should be documented?

pegbertsch: I think it would be helpful for maintenance if there was a (informal) tech doc of some kind that captured how you figured out the offsets, or any other "trial and error"(?)-type processes on how you worked with the `pdf-miner` library In case someone else is trying to do the same kind of validation development for a different consent down the line and is leveraging from existing code.  (I know there's limited bandwidth for going the extra mile for more documentation right now, so just something to keep in mind)

kenny-skaggs: Thanks, I'll keep it in mind.

PR #2703, opened 2021-11-04
[PDR-526] Update geography ubr to use both rural zipcode versions.

robabram:
Calculating the Geography UBR value requires using both the 2014 and 2020 versions of the rural zipcode list.  The code has been updated to load both lists and the UBR calculator now chooses which list to use based on the first ConsentPII submission.

## Tests
- [x] unit tests


dtharpeuno (approved):

rdr_service/resource/helpers.py
```diff
@@ -115,20 +116,30 @@ def any(self):
...
+        if os.path.exists(os.path.join(path, files[0])):
+            for file in files:
+                codes = list()
+                with open(os.path.join(path, file)) as handle:
+                    # pylint: disable=unused-variable
+                    for count, line in enumerate(handle):
+                        # If 5-digit zipcodes starting with 0 had the leading zero dropped in the source file,
+                        # add it back.
+                        codes.append(line.split(',')[1].strip().zfill(5))
+                data[re.sub("[^0-9]", "", file)] = codes
```
pegbertsch, line 139: Are there any ZIP+4 formats (containing a dash) in the newer rural zip code file?  Or are they all 5-digit codes?  If there were ZIP+4 in the file, not sure we want to remove the '-'?  Unless PDR needs a numeric field unlike RDR `participant_summary.zip_code` which is varchar?

pegbertsch: (Got confirmation they are 5-digit only)

pegbertsch (approved):

rdr_service/resource/helpers.py
robabram: No, the zipcodes are all 5-digit codes.

PR #2716, opened 2021-11-16
Minor changes to GEM to GP tool

j-kanuch:
This PR updates the GEM to GP export tool to allow a CSV of PIDs to be provided. Bulk inserts using `bulk_insert_mappings()` require a dict, not an object, so this was updated as well.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/genomics_dao.py
+++ b/rdr_service/dao/genomics_dao.py
@@ -2446,18 +2446,19 @@ class GemToGpMigrationDao(BaseDao):
         super(GemToGpMigrationDao, self).__init__(
             GemToGpMigration, order_by_ending=['id'])
 
-    @staticmethod
-    def prepare_obj(row, job_run, file_path):
-        return GemToGpMigration(
+    def prepare_obj(self, row, job_run, file_path):
+        return self.to_dict(GemToGpMigration(
             file_path=file_path,
             run_id=job_run,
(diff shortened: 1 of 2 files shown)


rdr_service/dao/genomics_dao.py
```diff
@@ -2446,18 +2446,19 @@ def __init__(self):
         super(GemToGpMigrationDao, self).__init__(
             GemToGpMigration, order_by_ending=['id'])
 
-    @staticmethod
-    def prepare_obj(row, job_run, file_path):
-        return GemToGpMigration(
+    def prepare_obj(self, row, job_run, file_path):
+        return self.to_dict(GemToGpMigration(
```
dtharpeuno, line 2450: If you wanted to keep it as an object and not have to convert it to a `dict`, you can do `bulk_save_objects(objs)` in `self.gem_gp_dao.insert_bulk` and you don't have to pass the `model_type` as well.

dtharpeuno (approved):

kenny-skaggs (approved):

rdr_service/dao/genomics_dao.py
j-kanuch: Thanks, I did not know that was in there. I'm going to merge this as is for now, but maybe next week or some point in the near future I will go back and refactor it to use the other method.

PR #2721, opened 2021-11-18
[ROC-1237] Don't process if missing contamination

j-kanuch:
The Raw AW2 Reconciliation process is failing on the records that are missing contamination. This PR modifies this process to skip these raw records. Further investigation will need to occur as to how to handle the skipped raw records. 

## Tests
- [x] unit tests


dtharpeuno (approved):
Thanks

kenny-skaggs (approved):

PR #2729, opened 2021-11-19
[PDR-533] Create sexual orientation table in PDR for multiple choice answers

robabram:
* Change base class functions to look up values in code table one at a time to a single fetch all, to eliminate a large number of single record database queries.
* Create a new PDR table to hold individual sexual orientation question answer records.

## Tests
- [x] unit tests


rdr_service/resource/generators/_base.py
```diff
@@ -286,26 +289,44 @@ def _lookup_code_value(self, code_id, session):
...
-        return result.value
 
-    def _lookup_code_id(self, code, ro_session):
+        if _CODEBOOK_CODE_VALUES is None:
+            results = session.query(Code.codeId, Code.value).all()
+            if not results:
+                return None
+            _CODEBOOK_CODE_VALUES = dict()
+            for row in results:
+                _CODEBOOK_CODE_VALUES[row.codeId] = row.value
```
kenny-skaggs, line 306: It may be a rare issue, if it ever happens at all. But it looks like this isn't thread-safe and two threads could attempt to generate the `_CODEBOOK_CODE_VALUES` dict at the same time.

robabram: That is a possibility, but even if two threads entered the block to set a global value at the same time, there would be no deadlock because of the Python Global Interpreter Lock and which ever one finished last would set the var to a dict successfully and the first dict would be discarded.  This change broke the unittests, so I'll look in to making some changes though.  A deadlock could be an issue in a different programming language that has true multi-threading.

robabram: Additionally, both threads would create and set the exact same data here, which is different than say the logging code which needs to capture logging data per thread.

rdr_service/resource/generators/_base.py
```diff
@@ -286,26 +290,68 @@ def _lookup_code_value(self, code_id, session):
...
+
+        # Code id to value mappings change from one unittest to another, we can't create a global
+        # cache of data while running unittests.
+        if not os.environ.get('UNITTEST_FLAG', 0) == 1 and _CODEBOOK_CODE_IDS is None:
+            results = session.query(Code.codeId, Code.value).all()
+            if not results:
+                return None
+            _CODEBOOK_CODE_IDS = dict()
+            for row in results:
+                _CODEBOOK_CODE_IDS[row.value] = row.codeId
```
pegbertsch, line 342: FYI: I don't know if this can somehow impact testing on stable,  but not all `code.value` fields are unique on stable.   It seems like this could possibly be an indeterminate result on which `code_id` gets stored in the `_CODEBOOK_CODE_IDS` hashed dict.   I would not expect this to be an issue with production data.

rdr_service/resource/schemas/participant.py
```diff
@@ -14,6 +14,14 @@
 from rdr_service.resource.constants import SchemaID
 
 
+class SexualOrientationEnum(Enum):
```
pegbertsch, line 17: Where are these options derived from?  Codebook or data glossary?   Hoping to include links to relevant source documentation in our code.

rdr_service/resource/schemas/participant.py
```diff
@@ -122,6 +130,19 @@ class Meta:
         pii_filter = {}  # dict(field: lambda function).
 
 
+class SexualOrientationSchema(Schema):
```
pegbertsch, line 135: Do you want to add a test case to the `tests/resource_tests/schema_tests/test_resource_schemas.py` file, since we're creating both BQ and the resource schemas for this?

pegbertsch (approved):
Approved;  see note about stable environment data impacting the lookup of code id by code value in case it could impact QA testing, but this has probably been a pre-existing issue due to duplicate code values in the stable code table.

rdr_service/resource/generators/_base.py
robabram: Thanks Peggy, I forgot about the duplicate values created by CE.  I'll add a filter to only pull the correct ones.

rdr_service/resource/schemas/participant.py
robabram: There is no prior source for this mapping for the Sexual Orientation values, I created the mapping here.

rdr_service/resource/schemas/participant.py
robabram: Done

pegbertsch (approved):

rdr_service/resource/generators/_base.py
```diff
@@ -286,26 +290,70 @@ def _lookup_code_value(self, code_id, session):
...
-        result = ro_session.query(Code.codeId).filter(Code.value == code).first()
-        if not result:
-            return None
-        return result.codeId
+
+        # Code id to value mappings change from one unittest to another, we can't create a global
+        # cache of data while running unittests.
+        if not os.environ.get('UNITTEST_FLAG', 0) == 1 and _CODEBOOK_CODE_IDS is None:
+            results = session.query(Code.codeId, Code.value).\
+                        filter(Code.system=='http://terminology.pmi-ops.org/CodeSystem/ppi').all()
```
wangy70, line 337: nit: We have a constant for this PPI: https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/code_constants.py#L4

wangy70 (approved):

rdr_service/resource/generators/_base.py
robabram: Cool.  I forgot about that constant.

PR #2749, opened 2021-12-07
[PDR-575] Correctly calulate UBR_Overall value.

robabram:
The UBR Overall value calculation should only be testing for a UBR value across all UBR values.
Change enum name to represent only ‘PMI_PreferNotToAnswer’ and ‘PMI_Skip’.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/calculators/participant_ubr.py
+++ b/rdr_service/resource/calculators/participant_ubr.py
@@ -247,5 +247,5 @@ class ParticipantUBRCalculator:
         # Test each UBR value for a UBR status.
         for k, v in data.items():
             if k.startswith('ubr_') and v == UBRValueEnum.UBR:
-                result = 1
+                return UBRValueEnum.UBR
         return result
--- a/rdr_service/resource/calculators/participant_ubr.py
+++ b/rdr_service/resource/calculators/participant_ubr.py
@@ -18,7 +18,7 @@ class UBRValueEnum(IntEnum):
 
     RBR = 0
     UBR = 1
-    NullSkip = 2
+    NotAnswer_Skip = 2  # PMI_PreferNotToAnswer or PMI_Skip value.
 
 
 class ParticipantUBRCalculator:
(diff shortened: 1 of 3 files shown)


rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -233,3 +233,19 @@ def ubr_age_at_consent(consent_time, answer: (str, None)):
...
+        :param data:
+        :return: 0 or 1
+        """
+        result = UBRValueEnum.RBR
+        if data is None:
+            return result
+        # Test each UBR value for a UBR status.
+        for k, v in data.items():
+            if k.startswith('ubr_') and v == UBRValueEnum.UBR:
+                result = 1
```
kenny-skaggs: :nit: Further up on line 244 `UBRValueEnum.RBR` is used, but here an integer value is used instead. Would it make sense to use `UBRValueEnum.UBR`?

rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -120,13 +120,13 @@ def ubr_ethnicity(answers):
         :return: UBRValueEnum
         """
         if answers is None:
-            return UBRValueEnum.NullSkip
+            return UBRValueEnum.NotAnswer_Skip
```
pegbertsch, line 123: To make sure I understand, a "missing" answer is supposed to be treated the same as a PMI_PreferNotToAnswer or PMI_Skip answer code?

rdr_service/resource/calculators/participant_ubr.py
robabram: Yes, I thought I got all those.  I'll make another commit.

rdr_service/resource/calculators/participant_ubr.py
robabram: The details of that are to be worked out in the AST meeting, I wanted to change the enum name to something that represents the current state of the code.

pegbertsch: Ok, but the "current state of the code" is also treating null/missing answer codes the same as the legit PMI_ codes that can be present?  Not asking for an enum change, just wanted to be sure there wasn't supposed to be any distinct logic for null/missing answers in this PR.  Sounds like that may change pending outcome of AST meeting?

pegbertsch (approved):

kenny-skaggs (approved):

PR #2758, opened 2021-12-14
[DA-2360] Integration for Ghost API

kenny-skaggs:
## Partially Resolves *[DA-2360](https://precisionmedicineinitiative.atlassian.net/browse/DA-2360)*
This adds code for integrating with the Vibrent API for checking for ghost participants. The business logic lives in a service class for checking for ghosts, a class for interacting with the PTSC API, and a model for recording the checks in the database.

Later PRs will add a cron job that will check for ghost participants weekly, as well as a tool for running the code manually (particularly for backfilling).

Each time a participant is checked for being a ghost or not, a record is saved for the check and whether a change was made because of the check. This makes it easier to determine when participants have been checked, as well as allows for viewing a history of the changes made by this script.

## Tests
- [x] unit tests


rdr_service/dao/ghost_check_dao.py
```diff
@@ -0,0 +1,54 @@
...
+from sqlalchemy.orm import Session
+
+from rdr_service.clock import CLOCK
+from rdr_service.dao.base_dao import BaseDao
+from rdr_service.model.ghost_api_check import GhostApiCheck, GhostFlagModification
+from rdr_service.model.participant import Participant
+
+
+class GhostCheckDao(BaseDao):
+    @classmethod
```
dtharpeuno, line 13: I am all for the use of classmethods in the logic, but in keeping with the rest of the design patterns for the daos, is there a reason why we are deviating from that? As well as passing Session as a param instead of it being initialized with the dao classes?

kenny-skaggs: We've seen strange cases with the DAOs creating sessions that connect to the local database while running tools against prod or stable. Passing in the session helps make sure we don't run into that.

The DAOs require a model class to be specified when initializing an instance, but I really only needed a place for having a query (with no instance level state needed). This is a deviation from the usual patterns, but I imagine it could be a good step towards making them more flexible.

dtharpeuno: 👍 Understood!

dtharpeuno (approved):

j-kanuch (approved):

PR #2763, opened 2021-12-16
[DA-2310] schedule_cronjob_for_ce_health_data_reconciliation

wangy70:
## Resolves *[DA-2310](https://precisionmedicineinitiative.atlassian.net/browse/DA-2310)*

1. Schedule the cronjob for CE health data reconciliation for lower envs for testing
2. Disable it on Prod


rdr_service/cron_prod.yaml
```diff
@@ -1,4 +1,5 @@
 cron:
+- description: Ce Health Data Reconciliation Pipeline
```
kenny-skaggs, line 2: We've found previously that this doesn't disable cron jobs for environments. For example sandbox has this same setup for various cron jobs and they're still scheduled and running there.

You can use the `app_util.nonprod` decorator on the cron view function instead.

wangy70: Thanks, done!

kenny-skaggs (approved):

j-kanuch (approved):

PR #2765, opened 2021-12-21
[ROC-1254] Fixing survey code export

kenny-skaggs:
## Resolves *[ROC-1254](https://precisionmedicineinitiative.atlassian.net/browse/ROC-1254)*
The codebook sync script has a process for exporting a file of the codes in RDR to Google Drive, but this process has been broken for the last few codebook imports.

This PR fixes the error that occurs when trying to export the file. `CodesExportClass.initialize_process_context` needed to be updated to match some previous changes (`initialize_process_context` is no longer a static method).

This PR also speeds up the export process by retrieving the parent and module code data for all the codes and processes it in memory, rather than making a database call for each code individually.

## Tests
- [ ] unit tests
Existing unit tests cover the code for finding module and parent codes.

### DIFF
--- a/rdr_service/tools/tool_libs/codes_management.py
+++ b/rdr_service/tools/tool_libs/codes_management.py
@@ -1,19 +1,16 @@
-from collections import defaultdict
 import csv
 from datetime import datetime
 from oauth2client.service_account import ServiceAccountCredentials
 import os
-from typing import Dict, List, Set
 
 from googleapiclient.discovery import build
 from googleapiclient.http import MediaFileUpload
-from sqlalchemy.orm import Session, joinedload
 
-from rdr_service.services.gcp_utils import gcp_get_iam_service_key_info
+from rdr_service.dao.code_dao import CodeDao
 from rdr_service.model.code import Code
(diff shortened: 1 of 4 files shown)


rdr_service/tools/tool_libs/codes_management.py
```diff
@@ -216,12 +221,64 @@ def run(self, skip_file_export_write=False):
...
 
         return exit_code
 
 
+class CodeHierarchyMap:
+    """Constructs a data structure for accessing the parent and module code values for a given code id"""
+
+    def __init__(self, session: Session):
+        # Load all surveys with the question and option codes used
+        query = session.query(
```
dtharpeuno: :nit should we move this to the CodeDao?

dtharpeuno (approved):

j-kanuch (approved):

rdr_service/tools/tool_libs/codes_management.py
kenny-skaggs: Good idea, that would make it more reusable. I'll work on that.

kenny-skaggs: Ok, it's moved over

PR #2769, opened 2021-12-21
Updating Google SDK version for CircleCI

kenny-skaggs:
## Resolves *no ticket*
Updates the Google SDK version that CircleCI uses. This will hopefully resolve some deployment issues we're seeing there. It seems like the SDK is expecting something to exist in setuptools that isn't there in newer versions.

## Tests
- [ ] unit tests

### DIFF
--- a/ci/install_google_sdk.sh
+++ b/ci/install_google_sdk.sh
@@ -3,7 +3,7 @@
 #
 # Install the Google Cloud SDK
 #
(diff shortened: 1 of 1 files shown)


j-kanuch (approved):

pegbertsch (approved):
Has this version been verified in local dev environments yet?

kenny-skaggs:
> Has this version been verified in local dev environments yet?

No, it hasn’t. But if we see issues with the deploy we can revert it.

PR #2774, opened 2021-12-29
[DA-2327] Updating validation to check for any unvalidated files

kenny-skaggs:
## Resolves *[DA-2327](https://precisionmedicineinitiative.atlassian.net/browse/DA-2327)*
The nightly consent file validation script is set up to check for any consent files that have been authored within the last day or so. This is missing any consents that are received that have older authored dates.

This PR updates the participant summaries that are checked nightly. Instead of any consents authored within the last day, the script will now validate any consents that haven't yet been validated.

The script will now load any participant summaries that have expected consents (such as a SUBMITTED gror, or a questionnaire response that indicates they should have a Primary Update consent) but don't yet have validation records for that consent type.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/consent_dao.py
+++ b/rdr_service/dao/consent_dao.py
@@ -1,15 +1,20 @@
 from datetime import datetime
 from typing import Collection
 
-from sqlalchemy import or_
-from sqlalchemy.orm import Session
+from sqlalchemy import and_, or_
+from sqlalchemy.orm import aliased, Session
 
+from rdr_service.code_constants import PRIMARY_CONSENT_UPDATE_QUESTION_CODE
 from rdr_service.dao.base_dao import BaseDao
-from rdr_service.model.consent_file import ConsentFile, ConsentSyncStatus
+from rdr_service.model.code import Code
+from rdr_service.model.consent_file import ConsentFile, ConsentSyncStatus, ConsentType
 from rdr_service.model.hpo import HPO
 from rdr_service.model.organization import Organization
 from rdr_service.model.participant import Participant
 from rdr_service.model.participant_summary import ParticipantSummary
+from rdr_service.model.questionnaire import QuestionnaireQuestion
+from rdr_service.model.questionnaire_response import QuestionnaireResponse, QuestionnaireResponseAnswer
+from rdr_service.participant_enums import QuestionnaireStatus
 
 
 class ConsentDao(BaseDao):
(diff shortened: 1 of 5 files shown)


rdr_service/dao/consent_dao.py
```diff
@@ -33,29 +38,81 @@ def _get_query_non_test_summaries(cls, session):
...
+                QuestionnaireResponseAnswer.questionnaireResponseId == QuestionnaireResponse.questionnaireResponseId
+            ).outerjoin(
+                QuestionnaireQuestion,
+                QuestionnaireQuestion.questionnaireQuestionId == QuestionnaireResponseAnswer.questionId
+            ).outerjoin(
+                Code,
+                and_(
+                    Code.codeId == QuestionnaireQuestion.codeId,
+                    Code.value == PRIMARY_CONSENT_UPDATE_QUESTION_CODE
+                )
```
j-kanuch, line 92: Should these joins from `QuestionnaireResponse`, `QuestionnaireResponseAnswer`, `QuestionnaireQuestion`, and `Code` be inner joins instead of outer joins? The filter condition `Code.codeId.isnot(None)` requires that records exist through this join path.

kenny-skaggs: They'd need to be outer joins. The query should return any participant summaries that have submitted a consent but don't have the file for that consent validated yet. So a participant that hasn't responded to the PrimaryUpdate question but needs an EHR validated should be returned.

The check for whether a code was matched from this join is within an `or` clause, so the check for a code can fail but another condition could still cause the participant to be returned.

j-kanuch: I see, makes sense, thanks.

j-kanuch (approved):

rdr_service/dao/consent_dao.py
```diff
@@ -33,29 +38,81 @@ def _get_query_non_test_summaries(cls, session):
...
+            ).filter(
                 or_(
-                    ParticipantSummary.consentForStudyEnrollmentFirstYesAuthored.between(start_date, end_date),
-                    ParticipantSummary.consentForCABoRAuthored.between(start_date, end_date),
-                    ParticipantSummary.consentForElectronicHealthRecordsAuthored.between(start_date, end_date),
-                    ParticipantSummary.consentForGenomicsRORAuthored.between(start_date, end_date)
+                    primary_consent.id.is_(None),
+                    and_(
+                        ParticipantSummary.consentForCABoR == QuestionnaireStatus.SUBMITTED,
```
pegbertsch, line 97: I assume CABoR is always an implicit yes/SUBMITTED similar to primary consent if a consent response exists, but the EHR and GROR statuses may be something else (SUBMITTED_NO_CONSENT/SUBMITTED_NOT_SURE)?

pegbertsch: Per offline conversation, apparently don't need to worry about cases where current consentForElectronicHealthRecords  or consentForGenomicsROR  is not / is no longer SUBMITTED.

pegbertsch (approved):

rdr_service/dao/consent_dao.py
kenny-skaggs: To clarify: the cases where a file is missed when a participant submitted consent and then immediately revoked it may be important and should be brought up in a discussion. But that will require larger changes than to these filters here, and this change doesn't introduce that scenario (it is currently happening).

PR #2775, opened 2021-12-29
Adding module code to log for when a participant responds when suspended

kenny-skaggs:
## Resolves *no ticket*
Suspended participants are able to withdraw, so it's valid to receive a survey response from them when they do (when they're responding to the withdrawal screens). This adds the module they responded to so it's easy to see what the response was for and know if there's something wrong or if it was just for the withdrawal screens.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/services/data_quality.py
+++ b/rdr_service/services/data_quality.py
@@ -195,7 +195,7 @@ class ResponseQualityChecker(_ModelQualityChecker):
                         f'but participant signed up at {participant_signup_time}'
                     )
                 if suspension_datetime and authored_time > suspension_datetime:
-                    logging.error(f'Response {response_id} authored for suspended participant')
+                    logging.error(f'Response {response_id} authored for suspended participant (module: {module_code})')
                 if withdrawal_datetime and authored_time > withdrawal_datetime:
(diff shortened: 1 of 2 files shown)


rdr_service/services/data_quality.py
```diff
@@ -195,7 +195,7 @@ def run_data_quality_checks(self, for_data_since: datetime = None):
                         f'but participant signed up at {participant_signup_time}'
                     )
                 if suspension_datetime and authored_time > suspension_datetime:
-                    logging.error(f'Response {response_id} authored for suspended participant')
+                    logging.error(f'Response {response_id} authored for suspended participant (module: {module_code})')
```
pegbertsch, line 198: If a withdrawal is a valid case, should there be a check that if the `module_code` is a withdrawal module (values `withdrawal_intro` or `StopParticipating`) then there is no error logged?

kenny-skaggs: I think that would be a good next step, but for now I was just looking to get the logs to tell me what was responded to (so I didn't have an extra step of looking it up).

wangy70 (approved):

pegbertsch (approved):

PR #2785, opened 2022-01-05
[DA-2030] Add PDR support for color user metrics file import.

robabram:
Create PDR resource model and generator for Color user metrics file import.
Integrate generator calls with file loading and later table updates.

## Tests
- [X] unit tests

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -26,7 +26,8 @@ from rdr_service.model.code import Code
 from rdr_service.model.participant_summary import ParticipantRaceAnswers, ParticipantSummary
 from rdr_service.model.config_utils import get_biobank_id_prefix
 from rdr_service.resource.generators.genomics import genomic_set_member_update, genomic_gc_validation_metrics_update, \
-    genomic_set_update, genomic_file_processed_update, genomic_manifest_file_update, genomic_set_member_batch_update
+    genomic_set_update, genomic_file_processed_update, genomic_manifest_file_update, genomic_set_member_batch_update, \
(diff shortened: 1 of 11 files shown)


j-kanuch (approved):

j-kanuch (approved):

rdr_service/genomic/genomic_job_components.py
```diff
@@ -952,14 +953,22 @@ def ingest_metrics_file_from_filepath(self, metric_type, file_path):
 
             if item_count == batch_size:
                 with metric_dao.session() as session:
-                    session.bulk_save_objects(batch)
+                    # Use session add_all() so we can get the newly created primary key id values back.
```
pegbertsch, line 956: @j-kanuch and @robabram :  How many rows are expected to be impacted here and are there negative implications to changing from bulk operation to `add_all()`?  What was the motivation for using the bulk operation originally?

pegbertsch: Resolved offline, Robert relayed this was already discussed with Josh and it should not impact performance/load.

pegbertsch (approved):

rdr_service/genomic/genomic_job_components.py
j-kanuch: @pegbertsch The `batch_size` is set to 100. The `bulk_save_objects()` method was used because we only needed to insert and didn't need any of the object data. The `add_all()` method returns the object data, and because this is needed for PDR, this is now being used. The `bulk_save_objects` is more performative, but since these objects are not very big, the performance cost of using `add_all` shouldn't be too noticeable. We probably could have used `add_all` to begin with for this object.

PR #2787, opened 2022-01-06
enable_fitbit_reconciliation_on_prod

wangy70:
## Resolves

Enable CE Fitbit data reconciliation pipeline on Prod env

### DIFF
--- a/rdr_service/cron_prod.yaml
+++ b/rdr_service/cron_prod.yaml
@@ -1,5 +1,4 @@
 cron:
-- description: Ce Health Data Reconciliation Pipeline
 - description: Validate the previous day's consent files
   url: /offline/ValidateConsentFiles
   schedule: every day 02:00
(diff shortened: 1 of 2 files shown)


rdr_service/cron_prod.yaml
```diff
@@ -80,3 +79,8 @@ cron:
   timezone: America/New_York
   schedule: every day 08:15
   target: offline
+- description: Ce Health Data Reconciliation Pipeline
+  url: /offline/CeHealthDataReconciliation
+  timezone: America/New_York
+  schedule: every day 3:45
+  target: offline
```
kenny-skaggs: This cron job would already be scheduled by an entry in the default file (https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/cron_default.yaml#L177). Does it need to be redefined here?

wangy70: It could be removed.

dtharpeuno (approved):

kenny-skaggs (approved):

PR #2788, opened 2022-01-06
[DA-2398] Adding way to ignore specific questionnaire response answers

kenny-skaggs:
## Resolves *[DA-2398](https://precisionmedicineinitiative.atlassian.net/browse/DA-2398)*
There are some answers that were given for the SDOH survey that need to be ignored from exports to downstream systems. This PR adds the ability to mark them as ignored in SQL, and clears out there answers in the curation ETL.

## Tests
- [x] unit tests


wangy70 (approved):

rdr_service/alembic/versions/afb0333cb471_adding_ignore_field_to_questionnaire_.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('questionnaire_response_answer', sa.Column('ignore', sa.Boolean(), nullable=True))
```
j-kanuch, line 40: Just FYI, `ignore` is a reserved word in MySQL. It should be fine here, but you'll need to wrap it in back-ticks if you use raw SQL.

j-kanuch (approved):

rdr_service/alembic/versions/afb0333cb471_adding_ignore_field_to_questionnaire_.py
kenny-skaggs: Ah, ok. Thanks. I'm sure escaping it will be fine if we need to reference it in raw sql.

PR #2790, opened 2022-01-07
[DA-2326] Fix for consent sync memory issues

kenny-skaggs:
## Resolves *[DA-2326](https://precisionmedicineinitiative.atlassian.net/browse/DA-2326)*
For the past few months the monthly consent sync process has been crashing because of the memory limit. I haven't yet been able to debug the memory usage to find if there is more that could help, but the changes in this PR should be a good step forward in alleviating memory usage. And will hopefully get the process to successfully run as a cron job.

The most memory intensive part is likely zipping the consent files. To be able to zip them, they need to be locally stored on the machine. In a GAE instance we don't have access to disk storage, so anything in the `/tmp` directory is held in memory instead.

Previously the sync cron job would work through all of the consent files for that month before zipping and uploading any that needed to be zipped. So while working through the list it would download any into memory that needed to be zipped, and then go through all of them at the end. We don't really need all files at once, we just need all the files that would go into a single zip.

These changes reduce the total number of files we have in memory at one time, collecting all the files for a given HPO, zipping them, and then removing them from memory (freeing up space before moving on to the next HPO). Zips are created for each site, so later this can be further reduced to only collecting the files needed for a specific site, but this is a start.

## Tests
- [ ] unit tests
Existing unit tests cover the consent sync functionality.

### DIFF
--- a/rdr_service/offline/sync_consent_files.py
+++ b/rdr_service/offline/sync_consent_files.py
@@ -51,12 +51,6 @@ class ParticipantPairingInfo:
 
 
 @dataclass
-class ConsentFileAndPairingInfo:
-    file: ConsentFile
-    pairing_info: ParticipantPairingInfo
-
-
-@dataclass
 class PairingHistoryRecord:
     org_name: str
     start_date: datetime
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):
Nice!!

j-kanuch (approved):

PR #2794, opened 2022-01-10
[DA-2195 partial] Automate consent error reporting to PTSC

pegbertsch:
When the consent validation process runs, generally as a daily cron job, we want to automatically generate any error reports for newly validated consents that we can transmit to PTSC.  The intended design is to email reports to a Jira-linked address that will automatically convert the email into a new PTSC Service Desk ticket.

Transmitting these emails programmatically is on hold while the use of Sendgrid for these specific reports is under review.  In the interim, a manual tool can be used to direct the generated report content to a file.

Functional changes:
-  At the end of the `validate_consent_uploads` method that runs as part of the daily cron job, dispatch a task to review newly created validation records for errors.   Since we generally expect to only have to process at most a few error records per day,  this is being done in a single task which is separate from/not dependent on the completion of the consent metric resource data generation (the `batch_rebuild_consent_metrics` tasks)
-   Task will look for newly created NEEDS_CORRECTING records, as well as reviewing new primary consent data to confirm validity of the DOB value and the participant's age  (not part of the consent PDF validation)
-  Until sending of generated emails is enabled, a warning will be logged if error reports would have been generated
-  The manual `consent-error-report` tool can be used to generate the report details and save to a file instead

The content/format of the email reports was decided on based on input from PTSC, so they can potentially write scripts to process the new error reports.

## Tests
- [x] unit tests
ConsentMetricsGenerator tests have been updated to also verify error report generation


rdr_service/tools/tool_libs/consent_validation_report.py
```diff
@@ -171,7 +171,7 @@
                  AND ps.is_test_participant = 0 and (ps.is_ghost_id is null or ps.is_ghost_id = 0) and ps.hpo_id != 21
 """
 
-ORIGIN_SQL_FILTER = ' AND ps.participant_origin = "{origin_filter}"'
+ORIGIN_SQL_FILTER = ' AND ps.participant_origin = "{origin}"'
```
kenny-skaggs: Does nothing else need to be updated to match this change (from `origin_filter` to `origin`)?

pegbertsch: I think that was an unintentionally greedy refactor.    I can revert it in the original file, but also that's the deprecated reporting tool that isn't being used currently.

pegbertsch: Fixed in latest commit

kenny-skaggs (approved):

robabram (approved):

PR #2798, opened 2022-01-11
fix_fitbit_reconciliation_time_format_parse_issue

wangy70:
## Resolves
Fix the CE Fitbit reconciliation timestamp parsing error.
The fractional seconds are not padded. If the value is under 1,000,000 it will be less than 7 digits. The possible range is 0-9,999,999. Example: both "2021-11-18T17:24:27.3Z" and "2021-11-18T17:24:27.3252561Z" are valid

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/offline/ce_health_data_reconciliation_pipeline.py
+++ b/rdr_service/offline/ce_health_data_reconciliation_pipeline.py
@@ -119,7 +119,12 @@ class CeHealthDataReconciliationPipeline:
                 for item in content['FileKeys']:
                     health_file_path_list.append(self.bucket_name + '/' + item)
             if 'TransferTime' in content:
-                file_transferred_time = parse_datetime_from_iso_format(content['TransferTime'][:26] + 'Z')
+                # The fractional seconds are not padded. If the value is under 1,000,000 it will be less than 7 digits.
+                # The possible range is 0-9,999,999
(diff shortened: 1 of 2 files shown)


kenny-skaggs:
Do we need to make sure TransferTime follows a certain pattern, or would it be better to be flexible for any other variations with something like https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse?

wangy70:
> Do we need to make sure TransferTime follows a certain pattern, or would it be better to be flexible for any other variations with something like https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse?

wangy70:
> Do we need to make sure TransferTime follows a certain pattern, or would it be better to be flexible for any other variations with something like https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse?

Based on CE team, the current pattern is ISO-8601, the only difference is that Python only support up to 6 sub-second digits, but their language support 7.

wangy70:
> Do we need to make sure TransferTime follows a certain pattern, or would it be better to be flexible for any other variations with something like https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse?

Sorry, closed by accident ...

kenny-skaggs (approved):

j-kanuch (approved):

PR #2809, opened 2022-01-17
Updates for manual consent file checking process

kenny-skaggs:
## Resolves *no ticket*
This makes some updates to the consent utility script. A file can be created that contains links for checking on files. That file can be modified and then another script command can be used to upload the modifications to the database (and rebuild them for PDR).

## Tests
- [] unit tests


rdr_service/tools/tool_libs/consents.py
```diff
@@ -3,8 +3,10 @@
 from datetime import datetime, timedelta
 from dateutil.parser import parse
 from itertools import islice
+from typing import Collection
 
 import requests
+from sqlalchemy import Boolean, Integer
```
dtharpeuno, line 9: Instead of importing these, and leveraging them from sqlalchemy, would it make more sense to use the more pythonic way of testing for this?
`isinstance(value, bool)`
`isinstance(value, int)`

dtharpeuno (approved):

rdr_service/tools/tool_libs/consents.py
kenny-skaggs: SqlAlchemy's `Boolean` and `Integer` are classes of their own rather than labels for the primitive data types. So later in the code we'd need to check if the fields are instances of these rather than the primitive types.

`isinstance(ConsentFile.is_signature_valid.expression.type, bool)` returns false, but `isinstance(ConsentFile.is_signature_valid.expression.type, Boolean)` returns true.

rdr_service/tools/tool_libs/consents.py
```diff
@@ -80,6 +86,60 @@ def _call_server_for_retro_validation(self, participant_ids):
...
+        with open('output.csv', 'w') as output_file, self.get_session() as session:
+            # Updatable query to get the files to verify and modify
+            files: Collection[ConsentFile] = session.query(ConsentFile).join(
+                Participant, Participant.participantId == ConsentFile.participant_id
+            ).filter(
+                ConsentFile.type == ConsentType.EHR,
+                ConsentFile.sync_status == ConsentSyncStatus.NEEDS_CORRECTING
+            ).order_by(ConsentFile.participant_id, ConsentFile.file_upload_time).limit(500).all()
+
+            fields_to_output = [
```
pegbertsch, line 101: All of these fields come from the ConsentFile row?  Is the join to the Participant table intended to filter out "problem" ConsentFile data?

pegbertsch (approved):
Curious about  the `files_for_update` query construction (in part to ensure similar query logic  isn't needed in consent metric-related tools).

rdr_service/tools/tool_libs/consents.py
kenny-skaggs: Yes, all the fields come from the ConsentFile. No, the join there is an artifact of an earlier version of the query. This query is going to be a point of change and will likely be modified for each run of the script to get different sets of files that need to be looked into further.

PR #2810, opened 2022-01-17
Using CircleCI docker image for Mysql

kenny-skaggs:
## Resolves *no ticket*
This updates the CircleCI config to use the pre-configured image for Mysql, rather than installing it manually in the base image. This resolves the current issue we're having with the mysql install, and will hopefully bypass any future issues that come up with it too.

## Tests
- [ ] unit tests


dtharpeuno (approved):
Nice thanks for fixing this!

j-kanuch (approved):

PR #2812, opened 2022-01-18
Renaming MayoLink Api class as MayoLink Client and moving to services directory

kenny-skaggs:
## Resolves *no ticket*
It's helpful to have a naming convention to distinguish between classes responsible for acting as an API that we provide to downstream users, and classes that contain logic for interacting with an external API. I've been naming classes that interact with external APIs as 'Client' to help better articulate their purpose. This updates the MayoLinkAPI class to MayoLinkClient so that it matches that convention. These changes also move the file out of the `api` directory and into the `services` directory.

Follow up PRs will further clean up the logic for sending orders to MayoLink so that a utility can be built that can reuse the same code.

## Tests
Existing unit tests updated.

### DIFF
--- a/rdr_service/dao/biobank_order_dao.py
+++ b/rdr_service/dao/biobank_order_dao.py
@@ -11,7 +11,7 @@ from rdr_service.lib_fhir.fhirclient_1_0_6.models.address import Address
 from sqlalchemy import or_, cast, Date, and_
 from sqlalchemy.orm import subqueryload, joinedload
 from werkzeug.exceptions import BadRequest, Conflict, PreconditionFailed, ServiceUnavailable
-from rdr_service.api.mayolink_api import MayoLinkApi, MayoLinkOrder, MayolinkQuestion, MayoLinkTest
+from rdr_service.services.mayolink_client import MayoLinkClient, MayoLinkOrder, MayolinkQuestion, MayoLinkTest
 from rdr_service import clock
 from rdr_service.api_util import get_site_id_by_site_value as get_site, format_json_code
 from rdr_service.app_util import get_account_origin_id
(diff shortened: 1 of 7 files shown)


dtharpeuno (approved):
Agreed on the file naming convention.

For the file move/rename did/will you use `git` move command, so that we can maintain the history on that file?

kenny-skaggs:
> For the file move/rename did/will you use `git` move command, so that we can maintain the history on that file?

I didn't, I didn't know that was a thing. I thought Git's algorithm would detect it on it's own, but maybe it would be better for helping maintain the history. Do you know if there's a way to check if the history isn't transferring?

In the list of changed files for this PR it looks like it's aware that the file is being moved (rather than deleted and a new one added), and in pycharm I'm able to see the whole history for the moved file (rather than it starting from when I moved it, and I'm assuming pycharm is using git commands underneath rather than checking for history data in other ways). It seems like the history is being maintained, but I could be missing something.

dtharpeuno:
> > For the file move/rename did/will you use `git` move command, so that we can maintain the history on that file?
> 
> I didn't, I didn't know that was a thing. I thought Git's algorithm would detect it on it's own, but maybe it would be better for helping maintain the history. Do you know if there's a way to check if the history isn't transferring?
> 
> In the list of changed files for this PR it looks like it's aware that the file is being moved (rather than deleted and a new one added), and in pycharm I'm able to see the whole history for the moved file (rather than it starting from when I moved it, and I'm assuming pycharm is using git commands underneath rather than checking for history data in other ways). It seems like the history is being maintained, but I could be missing something.

Nice as long as we can see the history. Thanks for checking.

wangy70 (approved):

PR #2820, opened 2022-01-25
[DA-2419] Detect and flag PROFILE_UPDATE TheBasics payloads on receipt

pegbertsch:
Update POST QuestionnaireResponse handler to detect and flag TheBasics payloads which are just profile updates and not full surveys.

This will add a list of specific code id values to the app singleton cache to compare against the payload as the `questionnaire_response_answer` data is being generated.   If the payload is determined to not be a full survey, the `participant_summary` is not updated.

The tool to analyze TheBasics response data was refactored so that the TSV file is appended to as each participant's data is analyzed (vs. doing one big write at the end),  to cut down on memory usage and execution time.
## Tests
- [x] unit tests
Updated existing unit test data files to generate participants whose TheBasics data only contains secondary contact info 
Added unit tests related to handling the new cache data


rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -403,13 +436,26 @@ def insert_with_session(self, session, questionnaire_response):
...
+
         # IMPORTANT: update the participant summary first to grab an exclusive lock on the participant
         # row. If you instead do this after the insert of the questionnaire response, MySQL will get a
         # shared lock on the participant row due the foreign key, and potentially deadlock later trying
         # to get the exclusive lock if another thread is updating the participant. See DA-269.
         # (We need to lock both participant and participant summary because the summary row may not
         # exist yet.)
-        if questionnaire_response.status == QuestionnaireResponseStatus.COMPLETED:
+        if (questionnaire_response.status == QuestionnaireResponseStatus.COMPLETED and
+               questionnaire_response.classificationType == QuestionnaireResponseClassificationType.COMPLETE):
```
kenny-skaggs, line 462: Wouldn't we still need to update the participant summary even if it wasn't a completed response (like if they just change their phone number or address)?

pegbertsch: I'm not aware of there being any secondary contact fields/ details in the `participant_summary` .   The participant's own details are updated via the ConsentPII module?

I've added a TODO comment in case we ever bother to classify subsequent ConsentPII payloads as PROFILE_UPDATE;  we may have a longer-term solution for updating profile details via an API vs. QuestionnaireResponse payloads before it gets to that

kenny-skaggs: Ah, you're right. Sorry, I forgot which survey had which data.

rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -349,6 +368,18 @@ def _validate_link_ids_from_resource_json_group(self, resource, link_ids):
...
+    def _get_module_from_questionnaire_history(questionnaire_history):
+        """ Extract the module code value from a questionnaire_history result JSON data (resource field) """
+
+        # Unittest/lower environments may not have expected questionnaire_history content, so allow for missing data
+        result = None
+        history_data = json.loads(questionnaire_history.resource) if questionnaire_history.resource else {}
+        if 'group' in history_data.keys() and 'concept' in history_data['group'].keys():
+            concepts = history_data['group']['concept']
+            result = concepts[0]['code'] if len(concepts) else None
+        return result
```
kenny-skaggs, line 381: Is there a reason to have this parse through the json again, or can it use `questionnaire_history.concepts`. I'm worried about impacts on maintainability from having the logic for parsing the concepts here and when we first read the questionnaire json.

pegbertsch: I'll look again, maybe I was missing a different location of the content.  I was having issues with errors in local database/unittest database setups which I thought did not have conformant `questionnaire_history` data and I was trying to get it working as fast as possible because of the time pressure on getting this into production.

robabram (approved):

rdr_service/dao/questionnaire_response_dao.py
pegbertsch: @kenny-skaggs I am working on a refactor.   Can we move forward with this PR as is if there are no other issues, and I will make sure there's a backlog ticket to adapt the way the code value is retrieved?  I think I will also look at optimizing the way the profile update code list is generated via the code cache.

kenny-skaggs (approved):

PR #2823, opened 2022-01-26
MIGRATIONS - [DA-2422] Genomic => PDR updates

dtharpeuno:
## Resolves *[ticket DA-2422]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2422

When data is modified it needs to be sent to PDR for querying by the analysts. This runs a scheduled job to be able to push all modified records so that the PDR datasets are in sync with RDR.
** Part 2 of this workflow will be removing all of the PDR method calls in the various parts of the pipeline and just relying on this scheduled job for pushing the updates

## Tests
- [x] unit tests


kenny-skaggs (approved):

rdr_service/genomic_enums.py
```diff
@@ -107,6 +107,7 @@ class GenomicJob(messages.Enum):
     METRICS_FILE_INGEST = 50
     RECONCILE_INFORMING_LOOP_RESPONSES = 51
     INGEST_RESULT_VIEWED = 52
+    RECONCILE_PDR_DATA = 53
```
j-kanuch: I just merged an enum with value `53`.

j-kanuch: And `54` too.

dtharpeuno: Updated https://github.com/all-of-us/raw-data-repository/pull/2823/commits/6c058d65e12e047ba74bc02b0d66c88cab9737f8

j-kanuch (approved):

pegbertsch:
@robabram @dtharpeuno Do we need to make any adjustments to handle new `created` and `modified` column names so there are no conflicts with metadata added to the PDR data records?

robabram:
@pegbertsch No, the code is fine.  The Resource models don't have conflicts over existing 'created' or 'modified' fields, the older BigQuery models do have that problem.  Since the Genomics tables are only implemented for the Resource API, there is no problem.

PR #2824, opened 2022-01-26
[DA-2329] Validating reconsents from VA participants

kenny-skaggs:
## Partially Resolves DA-2329
This links consent validation results to the QuestionnaireResponse that provided the consent. This helps determine the correct signing time if the new consent wouldn't change the participant summary authored date (such as when validating reconsented files for VA participants).

This also updates the validation process to load any ConsentResponses that don't yet have validation results.

## Tests
- [x] unit tests


rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -808,6 +829,50 @@ def _update_participant_summary(
...
+            # If the participant summary is missing, this would be a new consent to create one and no dates
+            # would need to be checked
+            for consent_type in self.consents_provided:
+                session.add(ConsentResponse(response=questionnaire_response, type=consent_type))
+        else:
+            # Check authored dates to see if it's a new consent response,
+            # or if it's potentially just a replay of a previous questionnaire response
+            consent_type_authored_time_map = {
+                ConsentType.PRIMARY: summary_before_response.consentForStudyEnrollmentAuthored,
+                ConsentType.CABOR: summary_before_response.consentForStudyEnrollmentAuthored,
```
pegbertsch: Intentional not to use consentForCABoRAuthored here?

rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -808,6 +829,50 @@ def _update_participant_summary(
...
+        if not summary_before_response:
+            # If the participant summary is missing, this would be a new consent to create one and no dates
+            # would need to be checked
+            for consent_type in self.consents_provided:
+                session.add(ConsentResponse(response=questionnaire_response, type=consent_type))
+        else:
+            # Check authored dates to see if it's a new consent response,
+            # or if it's potentially just a replay of a previous questionnaire response
+            consent_type_authored_time_map = {
+                ConsentType.PRIMARY: summary_before_response.consentForStudyEnrollmentAuthored,
```
pegbertsch: Intentional not to use consentForStudyEnrollmentFirstYesAuthored here?

rdr_service/dao/questionnaire_response_dao.py
kenny-skaggs: Nope, that would likely be better. Thanks.

rdr_service/dao/questionnaire_response_dao.py
kenny-skaggs: Yes. In either case we wouldn't want to revalidate a replay of a consent. We only want to validate them if they're a consent we haven't seen before. But If they've re-consented (or had a Primary Update) they're going to have two different dates and we wouldn't want to validate if the authored date matches either of them. I'm going to have to think of a better solution for this. Thanks for noticing it.

kenny-skaggs:
@pegbertsch, I've updated the logic to use previous ConsentResponse records rather than rely on the dates in the participant summary. Let me know what you think.

rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -808,6 +823,49 @@ def _update_participant_summary(
...
+                session.add(ConsentResponse(response=questionnaire_response, type=consent_type))
+
+    @classmethod
+    def _authored_times_match(cls, new_authored_time: datetime, current_authored_item: datetime):
+        if new_authored_time.tzinfo is None:
+            new_authored_time = new_authored_time.replace(tzinfo=pytz.utc)
+        if current_authored_item.tzinfo is None:
+            current_authored_item = current_authored_item.replace(tzinfo=pytz.utc)
+        difference_in_seconds = abs((new_authored_time - current_authored_item).total_seconds())
+        return difference_in_seconds < 300  # Allowing 5 minutes of difference between authored dates
```
pegbertsch, line 922: So I understand, this 5 minute window would allow for potentially multiple QuestionnaireResponse records that may not be/are not yet flagged as duplicates;  but for consent validation purposes, it's assumed they relate to the same consent PDF file if the authored times are so close?

kenny-skaggs: Right, it doesn't make sense to have consents (like Primary or Cabor) that have been authored in such a short time span. But even if a participant has gone back and forth on a consent that quickly (which could potentially happen with EHR), we'd only need one of the PDFs validated anyway (since at the moment they would look identical).

dtharpeuno (approved):

rdr_service/dao/questionnaire_response_dao.py
pegbertsch: I'm wondering more about PTSC's potential ability to send extraneous QuestionnaireResponse payloads (symptomatic of a process error) within a 5-minute window.   (Not necessarily duplicates and/or potentially referring to multiple PDFs or multiple authored timestamps that may have been improperly generated).

Would such a process error be easily detectable on our end if we only record one of them as a new consent response?  Is it worth logging a warning if we do detect "matching" authored times within the 5-minute window?

kenny-skaggs: If I'm understanding correctly, you're suggesting that this be used to also detect if a QuestionnaireResponse is being received multiple times when it shouldn't be. I think that's a good idea to implement, but that will take some careful thought and design to implement well, and that isn't the purpose of the changes I'm working on here.

pegbertsch (approved):

PR #2825, opened 2022-01-26
[DA-2300] Script for manually pushing mayolink orders

kenny-skaggs:
## Resolves DA-2300
Adds a script for posting a mail kit order to the MayoLINK API and creating the biobank order.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/tools/tool_libs/biobank_orders.py
+++ b/rdr_service/tools/tool_libs/biobank_orders.py
@@ -1,0 +1,53 @@
+from datetime import datetime
+
+import argparse
+
+from rdr_service.dao.mail_kit_order_dao import MailKitOrderDao
+from rdr_service.model.participant import Participant
+from rdr_service.services.biobank_order import BiobankOrderService
+from rdr_service.tools.tool_libs.tool_base import cli_run, ToolBase, logger
+
+tool_cmd = 'biobank_orders'
+tool_desc = 'Utility script for managing biobank orders'
+
+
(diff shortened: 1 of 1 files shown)


rdr_service/tools/tool_libs/biobank_orders.py
```diff
@@ -0,0 +1,53 @@
...
+    def upload_order_to_mayolink(self, barcode):
+        dao = MailKitOrderDao()
+        with self.get_session() as session:
+            mail_kit_order = dao.get_with_barcode(barcode=barcode, session=session)
+            if mail_kit_order is None:
+                logger.error(f'Unable to find order with barcode "{barcode}"')
+
+            participant_origin = session.query(Participant.participantOrigin).filter(
+                Participant.participantId == mail_kit_order.participantId
+            ).scalar()
```
dtharpeuno, line 30: Same as `one_or_none`?

dtharpeuno (approved):

wangy70 (approved):

rdr_service/tools/tool_libs/biobank_orders.py
kenny-skaggs: Not exactly. This will return None if the participant wasn't found, but if they are found then it would return the string for their origin. `one_or_none` would return a result object and I would then need to access the value using `participant_origin.participantOrigin`. This stores the string value directly in the variable so I can access it as `participant_origin` (without needing to use any other attributes).

PR #2832, opened 2022-01-28
[PDR-400] Handle less ordered samples than stored samples case.

robabram:
There are about 20 participants that have less Biobank ordered samples than the Biobank reports in the stored sample table. In this case, we want to make sure that we report all of the stored sample records.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -1005,6 +1005,12 @@ class ParticipantSummaryGenerator(generators.BaseGenerator):
                     if bss.biobank_order_id == row.biobank_order_id:
                         bbo_samples.append(_make_sample_dict_from_row(bss=bss, bos=None))
                         stored_count += 1
+            # PDR-400: There are about 20 participants that have less ordered samples than stored samples.
+            elif len(bss_results) > 0 and len(bos_results) < len(bss_results):
+                for bss in bss_results:
+                    if bss.biobank_order_id == row.biobank_order_id:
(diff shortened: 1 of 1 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -1005,6 +1005,12 @@ def _make_sample_dict_from_row(bss=None, bos=None):
                     if bss.biobank_order_id == row.biobank_order_id:
                         bbo_samples.append(_make_sample_dict_from_row(bss=bss, bos=None))
                         stored_count += 1
+            # PDR-400: There are about 20 participants that have less ordered samples than stored samples.
+            elif len(bss_results) > 0 and len(bos_results) < len(bss_results):
```
Seeker1911, line 1009:
```suggestion
            elif (len(bss_results) > 0) and (len(bos_results) < len(bss_results)):
```

dtharpeuno (approved):

rdr_service/resource/generators/participant.py
robabram: Thanks Michael for the suggestion!  I think I'll pass on this suggestion though. :)

Seeker1911: 😆 I just love commenting on RDR PR's. Trying to get that commit count back up.

robabram: You could always come back...

pegbertsch (approved):

PR #2834, opened 2022-01-31
[DA-2419 refactor] Use CodeDao methods for code and value lookups

pegbertsch:
Refactoring changes suggested in [PR #2820](https://github.com/all-of-us/raw-data-repository/pull/2820)
Uses cached codes / code maps to get the module name associated with a QuestionnaireResponse payload and to populate the list of TheBasics profile update codes used in detecting partial TheBasics responses.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/questionnaire_response_dao.py
+++ b/rdr_service/dao/questionnaire_response_dao.py
@@ -299,16 +299,20 @@ class QuestionnaireResponseDao(BaseDao):
                                                              cache_ttl_seconds=86400
                                                              )
 
-    def _load_thebasics_profile_update_codes(self):
+    @staticmethod
+    def _load_thebasics_profile_update_codes():
         """
         Invoked when the singleton cache needs to load the list of TheBasics profile update codes
         :return:  List of code id values for the profile update / secondary contact questions
         """
-        with self.session() as session:
-            profile_codes = session.query(Code.codeId
-                                          ).filter(Code.value.in_(BASICS_PROFILE_UPDATE_QUESTION_CODES)).all()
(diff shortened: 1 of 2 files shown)


rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -369,15 +373,19 @@ def _validate_link_ids_from_resource_json_group(self, resource, link_ids):
...
-            concepts = history_data['group']['concept']
-            result = concepts[0]['code'] if len(concepts) else None
+        if isinstance(questionnaire_history, QuestionnaireHistory):
+            concepts = questionnaire_history.concepts
+            if concepts:
+                concept_code = concepts[0].codeId
+                code_obj = CodeDao().get(concept_code)
+                result = code_obj.value if code_obj else None
+        else:
+            logging.debug(f'Unexpected questionnaire_history parameter type {type(questionnaire_history)}')
```
kenny-skaggs:
```suggestion
            logging.error(f'Unexpected questionnaire_history parameter type {type(questionnaire_history)}')
```

What do you think of bumping this up to the error level? If this method gets reused and gets passed something else that isn't a QuestionnaireHistory then it might be tough to track down the issue if this is only getting flagged at the debug level.

pegbertsch: I put this in as defensive code in case the unit test / lower environment QuestionnaireHistory data is not fully populated (to avoid triggering repeated error messages for this), but I raised the level in the most recent commit.

kenny-skaggs (approved):

robabram (approved):

PR #2837, opened 2022-02-01
Fixing circle issues with NumPy

dtharpeuno:
Since there is a current issue with the latest Numpy release and safety checks are also failing in relation, commenting out Pandas req for now since the package is not used in any app engine instances and has an internal dependency on NumPy. 

## Tests
- [] unit tests
** All current tests should pass


j-kanuch (approved):

kenny-skaggs (approved):

pegbertsch:
FYI:  If anyone else uses the consent-validation,  consent-error-report, or thebasics-analyzer tools locally, they will need to install pandas into their local virtual environment.   I believe at this time, I'm the only one impacted.

PR #2846, opened 2022-02-03
[DA-2321] Validating mail kit order before uploading to Mayolink

kenny-skaggs:
## Resolves *[DA-2321](https://precisionmedicineinitiative.atlassian.net/browse/DA-2321)*
We send a request to Mayolink when we get a SupplyDelivery request for an order. After the order is sent to Mayolink we make some validation checks and can potentially fail the request (return a 400) if there's something wrong. We've seen that we can get caught in retry loops that cause the RDR to send the same data to Mayolink multiple times.

This updates the process to move some of the validation checks in front of our interaction with Mayolink. This way if any of the validations fail we don't send the data to Mayolink before sending a BadRequest response.

## Tests
- [ ] unit tests


rdr_service/tools/tool_libs/biobank_orders.py
```diff
@@ -18,12 +21,23 @@ def run(self):
         if self.args.command == 'send-to-mayolink':
             self.upload_order_to_mayolink(barcode=self.args.barcode)
 
+    def _load_mail_kit(self, barcode: str, session: Session) -> Optional[BiobankMailKitOrder]:
+        mail_kit_orders = MailKitOrderDao.get_with_barcode(barcode=barcode, session=session)
+        if len(mail_kit_orders) == 0:
```
dtharpeuno: :nit Since you are returning a list, you could say `if not mail_kit_orders:`

dtharpeuno (approved):

rdr_service/tools/tool_libs/biobank_orders.py
kenny-skaggs: 👍 I can't figure out where I land on that. I like the explicit check in some scenarios, but `if not mail_kit_orders` does feel like it fits well here.

wangy70 (approved):

PR #2847, opened 2022-02-04
Edits to PDR reconcile

dtharpeuno:
In the genomic to PDR rebuild jobs, there might be memory limitations in requesting the entire object, especially in dealing with the user metric files daily drops, since there will be > 400k records created daily. Removed PDR metric rebuild job from cloud task workflow, and added method back into the genomic pipeline as it was initially in this PR https://github.com/all-of-us/raw-data-repository/pull/2785. Also edited `get_last_updated_records` to only supply ids in return list as to mitigate memory size for large record sets.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/genomic/genomic_job_controller.py
+++ b/rdr_service/genomic/genomic_job_controller.py
@@ -751,8 +751,7 @@ class GenomicJobController:
             self.file_processed_dao,
             self.metrics_dao,
             self.manifest_file_dao,
-            self.manifest_feedback_dao,
-            self.event_dao
+            self.manifest_feedback_dao
         ]
 
         for dao in reconcile_daos:
--- a/rdr_service/genomic/genomic_job_controller.py
+++ b/rdr_service/genomic/genomic_job_controller.py
@@ -760,11 +760,11 @@ class GenomicJobController:
             if not hasattr(dao, 'get_last_updated_records'):
                 continue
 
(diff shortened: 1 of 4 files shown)


kenny-skaggs (approved):

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -751,20 +751,19 @@ def reconcile_pdr_data(self):
...
                 continue
 
-            records = dao.get_last_updated_records(from_date=last_job_run)
-            if not records:
+            record_ids = dao.get_last_updated_records(from_date=last_job_run)
+            if not record_ids:
                 continue
 
-            batch_ids = [obj.id for obj in records]
+            batch_ids = [obj.id for obj in record_ids]
```
j-kanuch, line 766: Is `record_ids` a list of ids or objects?

dtharpeuno: So in SQLAlchemy when using an  `.all()` even though you specified wanting only one field it still returns a list of objects with one key/value. Still have to do the list comprehension on the records, but the initial record load is significantly smaller this way.

j-kanuch: Ah, gotcha. Sounds good.

j-kanuch (approved):

PR #2851, opened 2022-02-08
[DA-2404] add_cati_to_curation_etl

wangy70:
## Resolves *[DA-2404](https://precisionmedicineinitiative.atlassian.net/browse/DA-2404)*

Provide NON_PARTICIPANT_AUTHOR_INDICATOR, LANGUAGE and SURVEY CODE information of a questionnaire response to the Curation team in the ETL process.

Curation ETL SQL updates, no test case.

### DIFF
--- a/rdr_service/etl/raw_sql/finalize_cdm_data.sql
+++ b/rdr_service/etl/raw_sql/finalize_cdm_data.sql
@@ -1521,6 +1521,33 @@ CREATE TABLE cdm.pid_rid_mapping (
 );
 INSERT INTO cdm.pid_rid_mapping SELECT DISTINCT participant_id, research_id, external_id FROM cdm.src_clean;
 
+
+-- ---------------------------------------------------------------------
+-- Insert into questionnaire_response_additional_info
+-- ---------------------------------------------------------------------
+DROP TABLE IF EXISTS cdm.questionnaire_response_additional_info;
+CREATE TABLE cdm.questionnaire_response_additional_info (
+    questionnaire_response_id       bigint,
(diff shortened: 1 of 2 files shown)


rdr_service/etl/raw_sql/finalize_cdm_data.sql
```diff
@@ -1521,6 +1521,33 @@ CREATE TABLE cdm.pid_rid_mapping (
...
+INSERT INTO cdm.questionnaire_response_additional_info SELECT DISTINCT
+questionnaire_response_id, 'LANGUAGE' as type, language as value
+from rdr.questionnaire_response
+where language is not null;
+
+INSERT INTO cdm.questionnaire_response_additional_info SELECT DISTINCT
+qr.questionnaire_response_id, 'CODE' as type, c.value as value
+from rdr.questionnaire_response qr,  rdr.questionnaire_concept qc, rdr.code c
+where qr.questionnaire_id=qc.questionnaire_id
+and qc.code_id=c.code_id;
```
kenny-skaggs: The `src_clean` table filters out some questionnaires based on duplication and otherwise having multiple responses for a participant. Pulling ids directly from the questionnaire response table will mean this table will include the ids for some responses that were filtered out. Is that intentional, or should these three queries also join to the `src_clean` table to only send response ids that also appear in the other tables?

wangy70: Yes, good catch, updated.

kenny-skaggs (approved):

j-kanuch (approved):

PR #2858, opened 2022-02-10
Adding retry to message broker auth token refresh

kenny-skaggs:
## Resolves *no ticket*
We occasionally see a 403 response from Vibrent when trying to refresh an oauth token, but we'll be able to continue successfully later using the same credentials. This adds a retry loop for refreshing the token. This should keep us from missing data while we continue to investigate.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/message_broker/message_broker.py
+++ b/rdr_service/message_broker/message_broker.py
@@ -1,8 +1,12 @@
-from werkzeug.exceptions import BadRequest, BadGateway
-import requests
-from rdr_service import clock
+import logging
 from datetime import timedelta
 
+import backoff
+import requests
+from werkzeug.exceptions import BadRequest, BadGateway, HTTPException
+
+from rdr_service import clock
+from rdr_service.model.message_broker import MessageBrokerDestAuthInfo
 from rdr_service.model.utils import to_client_participant_id
 from rdr_service.dao.database_utils import format_datetime
 from rdr_service.dao.message_broker_metadata_dao import MessageBrokerMetadataDao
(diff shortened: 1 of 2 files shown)


rdr_service/message_broker/message_broker.py
```diff
@@ -31,22 +35,29 @@ def get_access_token(self):
...
-                expired_at = now + timedelta(seconds=r_json['expires_in'])
-                auth_info.expiredAt = expired_at
-                self.dest_auth_dao.update(auth_info)
-                return r_json['access_token']
-            else:
-                raise BadGateway(f'can not get access token for dest: {self.message.messageDest}, '
-                                 f'response error: {str(response.status_code)}')
+            return self._request_new_token(auth_info=auth_info)
+
+    @backoff.on_exception(backoff.constant, HTTPException, max_tries=3)
```
dtharpeuno, line 40: Nice!

dtharpeuno (approved):

wangy70 (approved):

rdr_service/message_broker/message_broker.py
wangy70: Yeah, nice job!

wangy70: Should this be `HTTPException` or `BadGateway` ?

kenny-skaggs: `BadGateway` is a subclass of `HTTPException`, it could be `BadGateway` but I thought it could help to be a little more general.

PR #2864, opened 2022-02-15
[DA-2461] Checking a subset of validation errors based on last time they were checked

kenny-skaggs:
There's a process in place for revalidating consents that may have been corrected. But currently it's crashing because it tries to process too much and runs out of memory. This updates the code to only check a subset of files. Additionally, when a file is checked a timestamp is recorded. This way we can re-validate the batch of files that haven't yet been revalidated, or haven't been re-validated recently.

## Tests
- [x] unit tests


rdr_service/dao/consent_dao.py
```diff
@@ -171,13 +171,16 @@ def get_participants_needing_validation(cls, session) -> Collection[ParticipantS
         return query.all()
 
     @classmethod
-    def get_files_needing_correction(cls, session, min_modified_datetime: datetime = None) -> Collection[ConsentFile]:
-        query = session.query(ConsentFile).filter(
-            ConsentFile.sync_status == ConsentSyncStatus.NEEDS_CORRECTING
+    def get_next_revalidate_batch(cls, session, limit=1000) -> Collection[ConsentFile]:
+        query = (
+            session.query(ConsentFile.participant_id, ConsentFile.type)
```
dtharpeuno, line 176: Nice, reducing the volume of the objects and attributes called should definitely help.

dtharpeuno (approved):

rdr_service/alembic/versions/48b5635ab9cd_adding_last_checked_time_to_consent_.py
```diff
@@ -0,0 +1,60 @@
...
+from rdr_service.participant_enums import EnrollmentStatus, Race, SampleStatus, OrganizationType, BiobankOrderStatus
+from rdr_service.participant_enums import OrderShipmentTrackingStatus, OrderShipmentStatus
+from rdr_service.participant_enums import MetricSetType, MetricsKey, GenderIdentity
+from rdr_service.model.base import add_table_history_table, drop_table_history_table
+from rdr_service.model.code import CodeType
+from rdr_service.model.site_enums import SiteStatus, EnrollingStatus, DigitalSchedulingStatus, ObsoleteStatus
+
+# revision identifiers, used by Alembic.
+revision = '48b5635ab9cd'
+down_revision = ('bd90a19851f8', '24f6c59ecd4c')
```
pegbertsch, line 24: Do you expect two down_revision values here?

kenny-skaggs: I wasn't expecting it, but alembic is set up to generate new migration files as dependent on any existing 'leaf' migrations. These two migrations must have been created and added in parallel. So my new migration would require that they're both run against the database first.

pegbertsch (approved):

PR #2868, opened 2022-02-17
[DA-2447] improve_reserch_directory_api_performance

wangy70:
## Resolves *[DA-2447](https://precisionmedicineinitiative.atlassian.net/browse/DA-2447)*
Improve the research directory API performance

Remove the unnecessary `joinedload` to speed up the query

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/workbench_dao.py
+++ b/rdr_service/dao/workbench_dao.py
@@ -4,7 +4,7 @@ from werkzeug.exceptions import BadRequest
 from dateutil.parser import parse
 import pytz
 from sqlalchemy import desc, or_, and_, func, distinct, case
-from sqlalchemy.orm import subqueryload, joinedload, aliased
+from sqlalchemy.orm import subqueryload, aliased
 from rdr_service.dao.base_dao import UpdatableDao
 from rdr_service import clock
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


dtharpeuno (approved):

rdr_service/dao/workbench_dao.py
```diff
@@ -518,8 +515,6 @@ def get_workspaces_with_user_detail(self, **kwargs):
             query = (
                 session.query(WorkbenchWorkspaceApproved, WorkbenchResearcher, WorkbenchWorkspaceUser.role,
                               snapshot_subquery.c.snapshot_id)
-                    .options(joinedload(WorkbenchWorkspaceApproved.workbenchWorkspaceUser),
-                             joinedload(WorkbenchResearcher.workbenchInstitutionalAffiliations))
```
kenny-skaggs, line 522: Maybe I'm misunderstanding something, but it seems like at least this one is still needed. Down at line 552 (or 556 before the changes) the access to the researcher's `workbenchInstitutionalAffiliations` would cause another query to lazy-load them.

wangy70: Yea, that was what I thought before. But after I compared the queries and the results, I found we don't need those `joinedload`, I removed these two lines, the WorkbenchResearcher.workbenchInstitutionalAffiliations are still there, it seems SQLalchemy lazy load them by default.

kenny-skaggs (approved):

PR #2874, opened 2022-02-23
[DA-2427] Update report state on consent removal

j-kanuch:
This PR creates a cron job for the updating the participant report state when gRoR and primary consent is removed. This PR modifies the query in `get_unconsented_gror_since_date()` to remove the date constraint. Methods that were previously specified for GEM only were renamed to be more generic. Logic was updated to check for the existence of consents to prevent value errors.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/genomics_dao.py
+++ b/rdr_service/dao/genomics_dao.py
@@ -2024,13 +2024,17 @@ class GenomicOutreachDaoV2(BaseDao):
             'pgx': {
                 GenomicReportState.PGX_RPT_READY: "ready",
                 GenomicReportState.PGX_RPT_PENDING_DELETE: "pending_delete",
-                GenomicReportState.PGX_RPT_DELETED: "deleted"
+                GenomicReportState.PGX_RPT_DELETED: "deleted",
+                GenomicReportState.CVL_RPT_PENDING_DELETE: "pending_delete",
+                GenomicReportState.CVL_RPT_DELETED: "deleted"
             },
             'hdr': {
                 GenomicReportState.HDR_RPT_UNINFORMATIVE: "uninformative",
                 GenomicReportState.HDR_RPT_POSITIVE: "positive",
                 GenomicReportState.HDR_RPT_PENDING_DELETE: "pending_delete",
-                GenomicReportState.HDR_RPT_DELETED: "deleted"
+                GenomicReportState.HDR_RPT_DELETED: "deleted",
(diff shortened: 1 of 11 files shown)


dtharpeuno (approved):

rdr_service/dao/genomics_dao.py
```diff
@@ -580,13 +580,18 @@ def get_gem_consent_removal_date(self, member):
...
         # Earliest date between GROR or Primary if both,
         withdraw_dates = []
-        if consent_status.consentForGenomicsROR != QuestionnaireStatus.SUBMITTED:
+        if consent_status.consentForGenomicsROR != QuestionnaireStatus.SUBMITTED and \
+            consent_status.consentForGenomicsRORAuthored:
             withdraw_dates.append(consent_status.consentForGenomicsRORAuthored)
 
-        if consent_status.consentForStudyEnrollment != QuestionnaireStatus.SUBMITTED:
+        if consent_status.consentForStudyEnrollment != QuestionnaireStatus.SUBMITTED and \
+            consent_status.consentForStudyEnrollmentAuthored:
```
kenny-skaggs, line 590: Is `consentForStudyEnrollment` ever anything but `SUBMITTED`? I think what we've been needing to do is remove the participant summary altogether if we need to remove their Primary consent status.

rdr_service/offline/main.py
```diff
@@ -938,6 +945,12 @@ def _build_pipeline_app():
         view_func=genomic_gem_a3_workflow,
         methods=["GET"]
     )
+    offline_app.add_url_rule(
+        OFFLINE_PREFIX + "GenomicUpdateReportStateForConsentRemoval",
+        endpoint="update_report_state_for_consent_removal",
+        view_func=update_report_state_for_consent_removal,
+        methods=["GET"]
+    )
```
kenny-skaggs, line 953: Should the cron job be scheduled (like in cron_default.yaml), or is it intentional to leave it unscheduled for now?

j-kanuch: Yes it should be scheduled. I completely forgot to add it to the cron schedule, thanks!

rdr_service/dao/genomics_dao.py
j-kanuch: I didn't realize that we remove the records. If that is the case, then this is really just a safety precaution. I will add a check for the withdrawal status as well.

dtharpeuno (approved):

kenny-skaggs (approved):

PR #2877, opened 2022-02-24
[DA-2438] Script for invalidating repeated COPE vaccination questions

kenny-skaggs:
## Resolves *[DA-2438](https://precisionmedicineinitiative.atlassian.net/browse/DA-2438)*
First off... sorry this is as big of a PR as it is. I'd break it apart if there was a bit more time 😅 

The February and all Minute COPE surveys ask about whether or not the participant got the covid vaccine (as well as when and what manufacturer if they did say they got it). Once we have an answer for the first or second dose, the participant isn't supposed to see that question again. There are multiple instances of participants seeing and responding to the questions about each dose of the vaccine several times when they shouldn't have been able to. This PR sets up a script for detecting those invalid answers and marking them as such.

As part of this effort, I added a method to the dao for loading a batch of responses. It filters them by survey code and participant id. I also added a domain model layer on top of questionnaire responses. This makes the data of a response a little easier to work with (atleast for me).

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/code_constants.py
+++ b/rdr_service/code_constants.py
@@ -125,6 +125,25 @@ CONSENT_COPE_YES_CODE = "COPE_A_44"
 CONSENT_COPE_NO_CODE = "COPE_A_13"
 CONSENT_COPE_DEFERRED_CODE = "COPE_A_231"
 
+COPE_DOSE_RECEIVED_QUESTION = 'cdc_covid_xx'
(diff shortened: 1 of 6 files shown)


dtharpeuno (approved):
Wow, nice job on this.

j-kanuch (approved):

PR #2882, opened 2022-03-01
Preventing locks and fixing map for invalid response codes

kenny-skaggs:
## Resolves *no ticket*
When running the ETL process the questionnaire response table was locked up while pulling all the metadata into the new tables. This sets the isolation level to prevent the locks when that's happening.

Also, for questions that needed to be marked as invalid, the ETL process wasn't sending the invalid concept code when the answer was anything other than a code (from radio or checkbox question). This updates the input to the ETL to trigger the concept codes to be mapped correctly. There were also issues with the unit being set to null in the observation table, so this uses a new unit for invalid answers.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/etl/raw_sql/finalize_cdm_data.sql
+++ b/rdr_service/etl/raw_sql/finalize_cdm_data.sql
@@ -106,7 +106,8 @@ CREATE TABLE cdm.src_mapped
     value_date                  datetime,
     value_string                varchar(1024),
     questionnaire_response_id   bigint,
-    unit_id                     varchar(50)
+    unit_id                     varchar(50),
+    is_invalid                  tinyint(1)
 );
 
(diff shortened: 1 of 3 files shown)


rdr_service/etl/raw_sql/finalize_cdm_data.sql
```diff
@@ -1535,6 +1538,9 @@ CREATE TABLE cdm.questionnaire_response_additional_info (
     value                           varchar(255)
 );
 
+-- Preventing locks on questionnaire_response table when reading data
+SET SESSION TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
```
dtharpeuno, line 1542: Neat!

dtharpeuno (approved):

wangy70 (approved):

PR #2887, opened 2022-03-08
Fixing bug in consent validation process

kenny-skaggs:
## Resolves *no ticket*
The consent validation process primarily uses `ConsentResponse` records to know what consents need validation, but the previous algorithm is still in place to catch any that may have been missed. That is reporting some that haven't yet been processed, but I believe all of those are still queued in the output strategy and haven't yet been taken care of because the context hasn't closed.

This PR adds a line to have the last of the files queued in the output strategy get flushed to the database before checking to see if there are any others that need to be validated.

## Tests
- [ ] unit tests

### DIFF
--- a/tests/service_tests/consent_tests/test_consent_controller.py
+++ b/tests/service_tests/consent_tests/test_consent_controller.py
@@ -109,7 +109,7 @@ class ConsentControllerTest(BaseTestCase):
 
         # Confirm a call to the dispatcher to rebuild the consent metrics resource data, with the ConsentFile.id
         # values from the expected_updates list
-        self.assertDispatchRebuildConsentMetricsCalled([2, 5, 6, 4])
+        self.assertDispatchRebuildConsentMetricsCalled([2, 5, 6, 4], call_count=2)
 
     def test_validating_specific_consents(self):
         """Make sure only the provided consent types are validated when specified"""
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):

tests/service_tests/consent_tests/test_consent_controller.py
```diff
@@ -109,7 +109,7 @@ def test_new_consent_validation(self):
 
         # Confirm a call to the dispatcher to rebuild the consent metrics resource data, with the ConsentFile.id
         # values from the expected_updates list
-        self.assertDispatchRebuildConsentMetricsCalled([2, 5, 6, 4])
+        self.assertDispatchRebuildConsentMetricsCalled([2, 5, 6, 4], call_count=2)
```
pegbertsch, line 112: if this is the only change in the unit test, is it expected behavior now for the dispatch task to be triggered twice?  What is happening on the extra call and what are the expected arguments?

kenny-skaggs: Yeah, it's expected to be called twice for the moment because the output strategy would call it twice (once each time it's `process_results` method is called, first time explicitly called to finalize any validations that were part of the first query and then again when the context ends).

In this case nothing is happening on the second call because all the expected consent files are handled by the first call. After being sure the ConsentResponse algorithm catches everything, the old algorithm (being used as a fallback right now) will be taken out and the explicit `process_results` call won't be needed (so the call count will go back to 1 again).

pegbertsch (approved):

wangy70 (approved):

PR #2890, opened 2022-03-09
[PDR-805] 'Prefer not to answer' should not be RBR.

robabram:
Due to changes in the Program Data Glossary, when a participant answers with 'PMI_PreferNotToAnswer', these should not be counted as RBR.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/calculators/participant_ubr.py
+++ b/rdr_service/resource/calculators/participant_ubr.py
@@ -57,10 +57,9 @@ class ParticipantUBRCalculator:
         if answer in [None, 'PMI_Skip', 'PMI_PreferNotToAnswer']:
             return UBRValueEnum.NotAnswer_Skip
 
-        if answer not in ['SexualOrientation_Straight', 'PMI_PreferNotToAnswer']:
-            return UBRValueEnum.UBR
-
-        return UBRValueEnum.RBR
+        if answer == 'SexualOrientation_Straight':
+            return UBRValueEnum.RBR
+        return UBRValueEnum.UBR
 
     @staticmethod
     def ubr_gender_identity(birth_sex, gender_ident, gender_ident_closer):
--- a/rdr_service/resource/calculators/participant_ubr.py
+++ b/rdr_service/resource/calculators/participant_ubr.py
@@ -18,7 +18,8 @@ class UBRValueEnum(IntEnum):
 
     RBR = 0
     UBR = 1
-    NotAnswer_Skip = 2  # PMI_PreferNotToAnswer or PMI_Skip value.
+    # NotAnswer_Skip: Answer is Null (only if a TheBasics has been submitted), PMI_PreferNotToAnswer or PMI_Skip value.
+    NotAnswer_Skip = 2
 
 
 class ParticipantUBRCalculator:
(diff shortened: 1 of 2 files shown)


rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -53,8 +54,7 @@ def ubr_sexual_orientation(answer: (str, None)):
         # they did not answer the expected SexualOrientation_None response to the parent TheBasics_SexualOrientation
         # question first.  This is not  consistent with the documented survey branching logic.  Therefore, UBR/RBR
         # calculations are still based only on the response to the TheBasics_SexualOrientation question alone
-
-        if answer is None or answer == 'PMI_Skip':
+        if answer in [None, 'PMI_Skip', 'PMI_PreferNotToAnswer']:
             return UBRValueEnum.NotAnswer_Skip
 
         if answer not in ['SexualOrientation_Straight', 'PMI_PreferNotToAnswer']:
```
pegbertsch: Can we just change this to `if answer != 'SexualOrientation_Straight` now, for clarity?

rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -93,14 +93,13 @@ def ubr_gender_identity(birth_sex, gender_ident, gender_ident_closer):
```
pegbertsch, line 92: Above on line 86/87:  Do we also need to account for potential unexpected/non-conformant response data where they could have skipped (`PMI_Skip`) the `gender_ident` question but provided a non-skip answer to `gender_ident_closer`?

rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -132,13 +131,14 @@ def ubr_ethnicity(answers):
         """
         if answers is None:
             return UBRValueEnum.NotAnswer_Skip
+
         # Note: assume UBR by default and check for exclusion criteria.
         answers = answers.split(',')
         if len(answers) == 1:
```
pegbertsch, line 136: Do we need to confirm there is no unexpected/non-conformant response data ingested by the RDR where there are multiple answer codes but one of them is `PMI_Skip` or `PMI_PreferNotToAnswer`?

pegbertsch (changes requested):
Asking for the update in `ubr_sexual_orientation()` line 60, since if `PMI_PreferNotToAnswer` was the answer code passed in, we already would have returned a value on line 58

rdr_service/resource/calculators/participant_ubr.py
robabram: Yes, I'll make that change.

rdr_service/resource/calculators/participant_ubr.py
robabram: I looked at the possible answers for 'Race_WhatRaceEthnicity' and there are no cases where a Skip or Prefer Not To Answer are mixed with other answers.  I think this logic is fine in this method.

rdr_service/resource/calculators/participant_ubr.py
robabram: I checked, there are no cases where Gender_GenderIdentity has PMI_Skip and Gender_CloserGenderDescription has a non-skip answer.

pegbertsch (approved):

dtharpeuno (approved):

PR #2891, opened 2022-03-09
Fix for duplicate/simultaneous requests on Biobank Specimen API

kenny-skaggs:
## Resolves *no ticket*
The RDR occasionally receives duplicate requests from the Biobank to create a specimen or aliquot. Currently the API will  start processing both requests and then error out on one of them when attempting the `insert` (because of a unique constraint on the rlims id).

To avoid further errors with these duplicate requests, this PR sets up named locks around processing the requests. When we receive two simultaneous requests to insert the same data, the first request processed will take an exclusive lock on the database and prevent the second request from being processed in parallel. When the insert for the first request completes, the second request will then be able take the lock (as long as the 30 second timeout hasn't occurred). The second request will then perform an update instead of another insert.

The DeceasedReportDao already has functionality for named locks. That code was refactored into a reusable class.

## Tests
- [x] unit tests
Existing unit tests cover expected behavior (tests for named locks were moved in the refactor)

### DIFF
--- a/rdr_service/dao/database_utils.py
+++ b/rdr_service/dao/database_utils.py
@@ -1,8 +1,10 @@
 """Helpers for querying the SQL database."""
-import re
+import logging
 from datetime import datetime
+import re
 
 import pytz
+from sqlalchemy.orm import Session
 
 from rdr_service.dao.database_factory import get_database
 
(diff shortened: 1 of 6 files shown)


rdr_service/dao/database_utils.py
```diff
@@ -62,3 +64,51 @@ def replace_null_safe_equals(sql):
     if _is_sqlite():
         return re.sub(_NULL_SAFE_PATTERN, r"is", sql)
     return sql
+
+
+class NamedLock:
+    """
+    Retrieve an explicit and mutually exclusive database lock to ensure a specific piece of code is the only instance
+    of that code running at the time that the lock is held.
+    """
```
j-kanuch, line 73: I like this idea a lot!

j-kanuch (approved):

dtharpeuno (approved):
👍

PR #2895, opened 2022-03-14
[PDR-819] Fix handling of PMI_Skip answer to EmploymentWorkAddress_Zi…

pegbertsch:
…pCode

We have ~75K TheBasics responses where the answer to the EmploymentWorkAddress_ZipCode question was `PMI_Skip`
This was getting treated like a zipcode and truncated to three characters so it shows up as `PMI` in the PDR data.

Adds a sanity check to the generator so that it will only do the zipcode truncation if the answer starts with a digit.  This allows `PMI_Skip` responses to show up in PDR unaltered

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/bq_questionnaire_dao.py
+++ b/rdr_service/dao/bq_questionnaire_dao.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+from re import match as re_match
 
 from sqlalchemy import text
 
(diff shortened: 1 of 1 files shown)


rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -185,9 +185,10 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
                     elif field['type'] == BQFieldTypeEnum.INTEGER.name:
                         if fld_value != PMI_SKIP_CODE:
                             setattr(bqr, fld_name, 1)
-                    # Truncate zip codes to 3 digits
+                    # Truncate zip codes to 3 digits.  Sanity check that the answer starts with a digit (consistent
+                    # with a 5-digit or Zip+4 format zipcode answer), and is not a code string like PMI_Skip
                     elif fld_name in ('StreetAddress_PIIZIP', 'EmploymentWorkAddress_ZipCode') and \
-                            len(fld_value) > 2:
+                            len(fld_value) > 2 and fld_value[0].isdigit():
```
kenny-skaggs: Has the answer been trimmed of potentially leading whitespace before this check?

I'm wondering if there might be some zip codes that would go through if only the first digit of a string is checked.
Should this look for PMI_SKIP specifically instead? Or maybe parsing the entire string to see if it's an integer value?

pegbertsch: There are no occurrences of leading whitespace in the RDR `questionnaire_response_answer.value_string` fields for the EmploymentWorkAddress_ZipCode question.   I didn't want to deviate from the existing logic which is not limited to only allowing  `PMI_Skip` through as a non-numeric answer.   I don't think the intention is to silently "scrub" bad data/other codes (e.g., `PMI_PreferNotToAnswer` or `PMI_DontKnow`) if they show up unexpectedly.

I'm not using `isnumeric()` in case the zipcode was in zip+4 format.

pegbertsch: Added logic to handle edge case if leading whitespace exists.  First non-whitespace char must be a digit for truncation to occur, and length must be at least three chars after stripping leading whitespace.

robabram (approved):

kenny-skaggs (approved):

PR #2899, opened 2022-03-15
[DA-2489] Sending notification if date of birth validation fails

kenny-skaggs:
## Resolves *[DA-2489](https://precisionmedicineinitiative.atlassian.net/browse/DA-2489)*
We need the RDR to flag any dates of birth that would mean a participant is currently, or at the time of their consent, less than 18 or more than 125 years old.

This adds code to have dates of birth checked when we receive them. When the QuestionnaireResponseDao detects a response to the date of birth question, it will send the date and participant id to a task to have it checked in 30 seconds. If the date of birth is outside the expected range, then a slack notification will be sent to notify us that a validation error occurred and needs to be investigated.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/api/cloud_tasks_api.py
+++ b/rdr_service/api/cloud_tasks_api.py
@@ -5,6 +5,7 @@ from flask_restful import Resource
 from werkzeug.exceptions import NotFound
 
 from rdr_service.api.data_gen_api import generate_samples_task
+from rdr_service.api_util import returns_success, parse_date
 from rdr_service.app_util import task_auth_required
 from rdr_service.dao.bq_code_dao import rebuild_bq_codebook_task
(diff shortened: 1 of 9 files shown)


wangy70 (approved):

j-kanuch (approved):

pegbertsch (commented):
@kenny-skaggs Just confirming:  The daily CheckConsentErrors task that runs after the daily consent validation cron job will still send emails (to PTSC, but not CE) for invalid dob/invalid age at consent errors.   Should I assume we still want that to happen in addition to the automated slack notification?

kenny-skaggs:
I'm not sure, but I think it's fine if those emails still go out.

PR #2905, opened 2022-03-21
RIP initial cvl workflows

dtharpeuno:
Initially, for GHR3 I was hesitant to delete these workflows to see if they could be leveraged somehow, but since the complexity of the cvl pipeline has changed seems better to clean these up. RIP 

## Tests
- [] unit tests
** Only deletes


kenny-skaggs (approved):

j-kanuch (approved):
RIP. What is dead may never die.

dtharpeuno:
> RIP. What is dead may never die.

Do I have repeat this back to you per the Iron Islands custom?

PR #2906, opened 2022-03-22
[DA-2476] Adding W1IL PGX and HDR job ids

kenny-skaggs:
## Partially Resolves *[DA-2476](https://precisionmedicineinitiative.atlassian.net/browse/DA-2476)*
Adding separate job ids for PGX and HDR W1IL jobs to genomic set member.

## Tests
- [ ] unit tests


pegbertsch (commented):
@kenny-skaggs @j-kanuch Do we need to add these columns to the PDR data?

kenny-skaggs:
> @kenny-skaggs @j-kanuch Do we need to add these columns to the PDR data?

I'm not sure. I'll defer to Josh or to you @pegbertsch  (as representation of what data PDR needs to make available to your stakeholders). As far as I understand, these and other job ids are used for tracking when and where data was interacted with so that we can do some investigation and debugging if we need to later, so maybe they wouldn't need to go to PDR. But I may be missing something about what consumers of PDR data would do with them.

pegbertsch:
@kenny-skaggs Consider my question resolved for now.  I realized this is part of the larger multi-ticket effort to add the new GHR3 content and does not need to be addressed in the individual PRs.

dtharpeuno (approved):

j-kanuch (approved):

PR #2908, opened 2022-03-22
[DA-2515] genomic_set_member history table

kenny-skaggs:
## Resolves *[DA-2515](https://precisionmedicineinitiative.atlassian.net/browse/DA-2515)*
This adds a history table and triggers for the genomic_set_member table. Hooks were also created for alembic so that we can automatically detect changes being made to the genomic_set_member table and also apply them to the history table (as well as update the triggers).

The alembic changes allow for defining which columns shouldn't be tracked as part of the history table. And any columns that are added or removed from the base table should have revisions generated to keep the history table up to date.

## Tests
- [ ] unit tests


dtharpeuno (approved):
👍

rdr_service/alembic/env.py
```diff
@@ -159,6 +159,7 @@ def run_migrations_online():
                 target_metadata=target_metadata.get(name),
                 include_schemas=True,
                 include_object=include_object_fn(name),
+                process_revision_directives=database.AutoHistoryRevisionGenerator.process_revision_directives
```
j-kanuch, line 162: Does this process every migration, or just the migrations that are greater than the current revision head?

rdr_service/model/genomics.py
```diff
@@ -54,6 +54,31 @@ class GenomicSetMember(Base):
...
+        'cvl_w2_manifest_job_run_id',
+        'cvl_w3_manifest_job_run_id',
+        'cvl_w4_manifest_job_run_id',
+        'cvl_w4f_manifest_job_run_id',
+        'cvl_aw1c_manifest_job_run_id',
+        'cvl_aw1cf_manifest_job_run_id',
+        'cvl_w3sr_manifest_job_run_id',
+        'cvl_w2sc_manifest_job_run_id',
+        'color_metrics_job_run_id'
+    ]
```
j-kanuch, line 87: If we want to add a column to this exclusion list after it was added in a previous release, do we need to create a new migration or will adding it to this list be picked up in the `AutoHistoryRevisionGenerator` run?

rdr_service/alembic/env.py
kenny-skaggs: This only processes operations in the migration file being created, it wouldn't look at existing migrations (whatever they may be relative to the current revision head of the database). It essentially lets us look at (and modify) what will be in a migration file that is being autogenerated before the file is written.

rdr_service/model/genomics.py
kenny-skaggs: It wouldn't be automatically detected, we would need to create a new migration for it. Do you think that's something that has to be in now, or can that be something we're conscious of and maybe add later if it happens often enough?

rdr_service/alembic/env.py
j-kanuch: Sounds good.

rdr_service/model/genomics.py
j-kanuch: That should be OK to defer for now. Going forward, if we create a migration for a new field and we want it excluded, we'll add it to the list at that time.

j-kanuch (approved):

j-kanuch (approved):

PR #2913, opened 2022-03-29
[DA-2558] Update GP event reconciliation query

j-kanuch:
The RDR receives a superset of the events that require reconciliation against the genomic_informing_loop table. This PR updates the GP user events reconciliation query to only compare against events found in `informing_loop_event_mapping`. Additionally, the process to remove GP user events after 8 days was removed. This was an artifact from when the user events file was cumulative instead of incremental.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/config/base_config.json
+++ b/rdr_service/config/base_config.json
@@ -98,7 +98,6 @@
     "reconcile_gc_data_file_to_table_workflow": 1,
     "reconcile_pdr_data": 1,
     "retry_manifest_ingestion_failures": 1,
-    "genomic_delete_old_gp_user_events": 0,
     "reconcile_raw_to_aw1_ingested_workflow": 1,
     "reconcile_raw_to_aw2_ingested_workflow": 1,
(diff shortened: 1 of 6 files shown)


rdr_service/config/base_config.json
```diff
@@ -98,7 +98,7 @@
     "reconcile_gc_data_file_to_table_workflow": 1,
     "reconcile_pdr_data": 1,
     "retry_manifest_ingestion_failures": 1,
-    "genomic_delete_old_gp_user_events": 1,
+    "genomic_delete_old_gp_user_events": 0,
```
dtharpeuno: Can we just delete this as well?

dtharpeuno (approved):

rdr_service/config/base_config.json
j-kanuch: yes, I'll remove it.

kenny-skaggs (approved):

PR #2916, opened 2022-03-31
[DA-2476] W1IL Manifest generation

kenny-skaggs:
## Resolves *[DA-2476](https://precisionmedicineinitiative.atlassian.net/browse/DA-2476)*
Generates W1IL manifest for each informing loop decision type (pgx or hdr) for each CVL. And uploads it to the CVLs bucket (read from the config).

Todos:
- The `GenomicResultWorkflowState` still needs to be initialized when the W1IL is generated, let me know if that should be a part of this PR
- The CVL buckets still need to be set up in the config

Minor changes:
- Another query (`get_latest_state_for_pid`) was already set up to find the latest informing-loop decision, I refactored that a bit so the query could be reused as a subquery for finding set members for the W1IL

## Tests
- [] unit tests


rdr_service/dao/genomics_dao.py
```diff
@@ -3245,3 +3256,120 @@ def get_w3sr_records(self, **kwargs):
...
+                informing_loop_subquery,
+                informing_loop_subquery.participant_id == GenomicSetMember.participantId
+            ).outerjoin(
+                GenomicResultWorkflowState,
+                and_(
+                    GenomicResultWorkflowState.genomic_set_member_id == GenomicSetMember.id,
+                    GenomicResultWorkflowState.results_module == result_module_type
+                )
+            ).filter(
+                GenomicGCValidationMetrics.processingStatus.like('pass'),
```
j-kanuch: For this and the other filters using `like`, we don't transform the values to `lower` when we ingest, so these filters will need to be case insensitive. Though `LIKE` is case insensitive in MySql, if I'm not mistaken, SqlAlchemy's `like()` is case sensitive. You'll probably want to use `ilike()` here, which is insensitive, or wrap it in a func.lower() (which we do elsewhere): `func.lower(GenomicGCValidationMetrics.processingStatus) == "pass",`. 
This is also true for `sexConcordance` ("TRUE"), `drcSexConcordance` ("PASS"), and `gcManifestSampleSource` ("Whole Blood").

rdr_service/offline/genomic_pipeline.py
```diff
@@ -157,6 +157,19 @@ def aw3_wgs_manifest_workflow():
...
+    for site_id in config.GENOMIC_CVL_SITES:
+        cvl_bucket = cvl_site_bucket_map[site_id]
+
+        with GenomicJobController(
+            GenomicJob.CVL_W1IL_WORKFLOW,
+            bucket_name=cvl_bucket,
+            cvl_site_id=site_id
+        ) as controller:
+            controller.generate_manifest(GenomicManifestTypes.CVL_W1IL_PGX, _genome_type=config.GENOME_TYPE_WGS)
+            controller.generate_manifest(GenomicManifestTypes.CVL_W1IL_HDR, _genome_type=config.GENOME_TYPE_WGS)
```
j-kanuch: Can we split these into two separate jobs so they can be run separately? I can see a case where we get a request to run only a PGx or HDR manifest, or perhaps change the job frequency for one and not the other since PGx is much faster for the CVLs to process.

rdr_service/dao/genomics_dao.py
```diff
@@ -3245,3 +3256,120 @@ def get_w3sr_records(self, **kwargs):
...
+
+                ParticipantSummary.consentForGenomicsROR == QuestionnaireStatus.SUBMITTED,
+                GenomicGCValidationMetrics.drcFpConcordance.like('pass'),
+                sample_collected_site.siteType != 'diversion pouch',
+                GenomicSetMember.gcSiteId.like(gc_site_id),
+                ParticipantSummary.participantOrigin != 'careevolution',
+
+                GenomicSetMember.ignoreFlag != 1,
+                GenomicSetMember.genomicWorkflowState == GenomicWorkflowState.CVL_READY,
+                GenomicResultWorkflowState.id.is_(None)
```
j-kanuch: Can you change this to look for the PGx or HDR W1IL job run ID instead? I know it wasn't clear from the requirements, but I can foresee a future request that may create a case where we have a `GenomicResultWorkflowState` record for a reason upstream and independent of inclusion in the W1IL manifest.

tests/genomics_tests/test_genomic_cvl_pipeline.py
```diff
@@ -144,6 +148,120 @@ def test_w2sc_manifest_to_raw_ingestion(self):
...
+                self.expected_w1il_row(hdr_and_pgx_set_member, hdr_and_pgx_validation_metrics, hdr_and_pgx_summary)
+            ]
+        )
+        self.assert_manifest_has_rows(
+            manifest_cloud_writer_mock=co_w1il_manifest,
+            expected_rows=[
+                self.expected_w1il_row(co_set_member, co_validation_metrics, co_summary)
+            ]
+        )
+
```
j-kanuch: Nice use of mocks to test the row values. Can we also test the following? 
- The column headers in the CSV match what is expected by the CVLs. You may be doing this and I'm missing it.
- The `genomic_set_member` records are updated with the correct values for the HDR and PGx job run IDs. 
Note that the new job run IDs need to be added to `GenomicSetMemberDao.valid_job_id_fields` for this to work.
- That subsequent runs of the jobs don't include samples already sent in the W1IL for that module.
- The CSV files are being pushed to the correct bucket, i.e. Color isn't going to Broad's data bucket. This test is probably redundant, but it ensures the job is looking at the correct bucket-CVL mapping.

rdr_service/dao/genomics_dao.py
kenny-skaggs: Good catch. I didn't know there was an `ilike`. I guess other platforms can have a case-sensitive LIKE. I'll switch this to make sure it stays case-insensitive even if we switch to something else (like postgres).

rdr_service/offline/genomic_pipeline.py
kenny-skaggs: Will do 😄

rdr_service/dao/genomics_dao.py
kenny-skaggs: Yeah, definitely!

tests/genomics_tests/test_genomic_cvl_pipeline.py
kenny-skaggs: > The column headers in the CSV match what is expected by the CVLs. You may be doing this and I'm missing it.

You're right, that's not part of the test right now. I'll add that in.

> The genomic_set_member records are updated with the correct values for the HDR and PGx job run IDs.
Note that the new job run IDs need to be added to GenomicSetMemberDao.valid_job_id_fields for this to work.

I will look into that, and may have other questions later. It seems like tasks are generated to do that and wouldn't actually execute while the test is running, but I might be able to figure something out. We may also need to add some of the other CVL jobs too, but I may be missing how they function.

> That subsequent runs of the jobs don't include samples already sent in the W1IL for that module.

Good catch, thanks. I thought I had one, but it looks like I modified it at some point. I'll add one in.

> The CSV files are being pushed to the correct bucket, i.e. Color isn't going to Broad's data bucket. This test is probably redundant, but it ensures the job is looking at the correct bucket-CVL mapping.

Will do 👍

rdr_service/dao/genomics_dao.py
j-kanuch: Yea, neither did I. Apparently PosgreSQL's `LIKE` is case sensitive, and so it has the operator `ILIKE`. For systems that don't have `ILIKE`, like MySQL, SqlAlchemy wraps it in a `lower()` under the hood. The reason we don't mutate the values to lower (on ingestion) is because if we have to pass these through, we have to follow the manifest specs, as un-standardized as they may be :)

tests/genomics_tests/test_genomic_cvl_pipeline.py
j-kanuch: Thanks. I know a lot of this would have been clearer if we'd given you a tech design. for the job ID fields on the set members, you correctly added the field string to the manifest configs in the definition provider, i.e. `'job_run_field': 'cvlW1ilPgxJobRunId',`, but that string value needs to be added to the list here or else it won't actually write to the tables:`GenomicSetMemberDao.valid_job_id_fields`.

j-kanuch: `/rdr_service/dao/genomics_dao.py:222`. It looks like other CVL jobs need to be added to it too. I probably missed these in other reviews.

j-kanuch: OK, per our discussion, please remove the validation check on `/rdr_service/dao/genomics_dao.py:835`. That way the job IDs get written and we don't need to worry about this step for now.

kenny-skaggs: The ManifestCompiler is set up to deploy a task that updates the job run ids on the GenomicSetMember. @j-kanuch, what do you think of having the test check that the task was created for setting the HDR and PGx job run ids?

j-kanuch: That should cover it since the job run field validation was removed in #2919.

j-kanuch (approved):
Looks good. Thanks for making those changes, especially splitting the two jobs up. It will help keep it flexible in the future.

dtharpeuno (approved):

PR #2918, opened 2022-04-01
MIGRATIONS [DA-2480] GHR3 W3SS workflow

dtharpeuno:
## Resolves *[ticket DA-2480]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2480

GHR3 project is now underway and there needs to be a system in place to move participant samples to the CVLs and then RHP for results to be generated and returned to participants. In order to move the genomics process along, a new W pipeline needs to be created at the DRC.

** This PR is for the W3SS workflow

## Tests
- [x] unit tests


rdr_service/model/genomics.py
```diff
@@ -982,6 +1038,56 @@ class GenomicGCValidationMetrics(Base):
...
+
+    biobank_id = Column(String(255), nullable=True)
+    sample_id = Column(String(255), nullable=True)
+
+    packageId = Column(String(250), nullable=True)
+    version = Column(String(255), nullable=False, default=0)
+    box_storageunit_id = Column(String(255), nullable=True)
+    box_id_plate_id = Column(String(255), nullable=True)
+    well_position = Column(String(255), nullable=True)
+    cvl_sample_id = Column(String(255), nullable=True)
```
j-kanuch: Can you add an index here?

rdr_service/model/genomics.py
```diff
@@ -982,6 +1038,56 @@ class GenomicGCValidationMetrics(Base):
...
+
+    __tablename__ = 'genomic_cvl_second_sample'
+
+    id = Column('id', Integer, primary_key=True, autoincrement=True, nullable=False)
+    created = Column(DateTime, nullable=True)
+    modified = Column(DateTime, nullable=True)
+    genomic_set_member_id = Column(ForeignKey('genomic_set_member.id'), nullable=False, index=True)
+
+    biobank_id = Column(String(255), nullable=True)
+    sample_id = Column(String(255), nullable=True)
```
j-kanuch: Can you add an index here?

rdr_service/model/genomics.py
```diff
@@ -982,6 +1038,56 @@ class GenomicGCValidationMetrics(Base):
...
+    total_concentration = Column(String(255), nullable=True)
+    total_dna = Column(String(255), nullable=True)
+    visit_description = Column(String(255), nullable=True)
+    sample_source = Column(String(255), nullable=True)
+    study = Column(String(255), nullable=True)
+    tracking_number = Column(String(255), nullable=True)
+    contact = Column(String(255), nullable=True)
+    email = Column(String(255), nullable=True)
+    study_pi = Column(String(255), nullable=True)
+    site_name = Column(String(255), nullable=True)
```
j-kanuch: Can you also add one here?

rdr_service/model/genomics.py
dtharpeuno: Added

rdr_service/model/genomics.py
dtharpeuno: Added

rdr_service/model/genomics.py
dtharpeuno: Added

rdr_service/genomic/genomic_state_handler.py
```diff
@@ -247,6 +247,12 @@ def transition_function(self, signal):
             return ResultsWorkflowState.CVL_W3SR
 
 
+class W3NSState(GenomicStateBase):
```
j-kanuch, line 250: I'm good with this for now, just noting that we'll probably want to revisit the Secondary Confirmation loop when we're designing the reconciliation process. Looking at the workflow diagram, the state transitions are a little messy.

j-kanuch (approved):

rdr_service/genomic/genomic_state_handler.py
dtharpeuno: Yea the transitions for the whole workflow have been based on the directional arrows in that diagram, so some of them will need to be tweaked maybe. Think its better at this point to build the base and reconcile it later as well.

kenny-skaggs (approved):

PR #2928, opened 2022-04-14
[DA-2612] Add PDR data rebuild tasks to ghost check cron job

pegbertsch:
The Ghost API check cron job has been running for a few months and can make updates to the RDR participant records, but the PDR participant data is not being rebuilt with the RDR data changes.  In between full PDR rebuilds, this causes the PDR participant data/ghost id counts to become increasingly out of sync with RDR.

Add logic to fire off PDR data rebuild tasks from the cron job

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/services/ghost_check_service.py
+++ b/rdr_service/services/ghost_check_service.py
@@ -3,6 +3,7 @@ from logging import Logger
 
 from sqlalchemy.orm import Session
 
+from rdr_service.config import GAE_PROJECT
 from rdr_service.dao.ghost_check_dao import GhostCheckDao, GhostFlagModification
 from rdr_service.dao.participant_dao import ParticipantDao
 from rdr_service.model.participant import Participant
@@ -17,7 +18,7 @@ class GhostCheckService:
         self._config = ptsc_config
         self._participant_dao = ParticipantDao()
 
-    def run_ghost_check(self, start_date: date, end_date: date = None):
+    def run_ghost_check(self, start_date: date, end_date: date = None, project=GAE_PROJECT):
         """
         Finds all the participants that need to be checked to see if they're ghosts and calls out to Vibrent's API
         to check them, recording the result.
@@ -72,7 +73,7 @@ class GhostCheckService:
 
         if len(pdr_rebuild_list):
             # PDR BQ module views select the test/ghost flag from participant data; don't need to rebuild module data
-            dispatch_participant_rebuild_tasks(pdr_rebuild_list, build_modules=False)
+            dispatch_participant_rebuild_tasks(pdr_rebuild_list, project=project, build_modules=False)
 
     def _record_ghost_result(self, is_ghost_response: bool, participant: Participant) -> bool:
         """ Update the participant isGhostId status if needed.  Returns true if update was performed """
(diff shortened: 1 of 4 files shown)


rdr_service/services/ghost_check_service.py
```diff
@@ -59,12 +67,25 @@ def run_ghost_check(self, start_date: date, end_date: date = None):
...
-    def _record_ghost_result(self, is_ghost_response: bool, participant: Participant):
+        task = GCPCloudTask()
+        for batch in list_chunks(pdr_rebuild_list, 100):
+            # PDR/BQ module views select the test(/ghost) setting from the participant data; don't need module rebuild
+            payload = {'batch': batch, 'build_modules': False}
+            if GAE_PROJECT == 'localhost':  # e.g., unittest case
+                batch_rebuild_participants_task(payload)
+            else:
+                task.execute('rebuild_participants_task', payload=payload, project_id=GAE_PROJECT,
+                             queue='resource-rebuild', in_seconds=15, quiet=True)
```
kenny-skaggs: Can we use the `dispatch_participant_rebuild_tasks` method instead, maybe by expanding it to include an option for defining the "build_modules" option? That way the logic for how to create a task for rebuilding a participant is better encapsulated.

rdr_service/services/ghost_check_service.py
```diff
@@ -3,11 +3,15 @@
 
 from sqlalchemy.orm import Session
 
+from rdr_service.config import GAE_PROJECT
```
kenny-skaggs, line 6: The GhostCheckService class is used by a script as well as the cron job. Can you have them pass in the project id, rather than have the class try to determine which one it is on its own? That way the service class can still be used by the script and have the rebuild tasks created on the target environment.

pegbertsch: No longer needed after the other change.   Project is checked in `dispath_participant_rebuild_tasks`

rdr_service/services/ghost_check_service.py
pegbertsch: Updated.   FYI:   We're going to be deprecating all of this rebuild task dispatching when the new RDR-PDR pipeline is implemented

rdr_service/services/ghost_check_service.py
kenny-skaggs: I could be mistaken, but I think it's still needed. If `dispatch_participant_rebuild_tasks` is called without the `project_id` parameter it will use this same concept and try to determine the target environment itself. So in the same way it would find the project as localhost when the ghost check script is being run against something like prod. What I've seen before is that letting it find localhost will have it try to rebuild everything into the local database. And as far as I could tell at the time, that ends up not making it to the server for PDR.

The `build_locally` parameter would need to be set to False to avoid having that happen too.

pegbertsch: Updated

pegbertsch: Final (hopefully) update:   revised the logic of how the `build_locally` flag is handled in the dispatch tasks method so that it's not required unless you really want to override the cloud task and force a local build.

kenny-skaggs (approved):

robabram (approved):

PR #2931, opened 2022-04-18
[DA-2573] Updating validation of GROR PDF

kenny-skaggs:
## Resolves *[DA-2573](https://precisionmedicineinitiative.atlassian.net/browse/DA-2573)*
The GROR consent process is being updated. This PR is to update the validation process to be sure it continues to function after the changes take affect. The only change that will affect the validation is that the wording of title of the file is changing from "Consent to Receive DNA Results" to "Consent to Get DNA Results". We use file names when detecting consent types for Vibrent, but the CE code needs to use the title to determine what the consent type is (the Spanish translation hasn't changed).

## Tests
- [ ] unit tests


dtharpeuno (approved):
That was easy!

wangy70 (approved):

PR #2932, opened 2022-04-19
MIGRATIONS [DA-2534] Calc informing loop ready status

dtharpeuno:
## Resolves *[ticket DA-2534]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2534

Adding multiple jobs to handle the assigning of informing loop ready status.

** Tech design stated that the history needed to be tracked so let these attributes off the exclusion list.

## Tests
- [x] unit tests


wangy70 (approved):

rdr_service/dao/genomics_dao.py
```diff
@@ -1045,6 +1050,66 @@ def get_member_from_raw_aw1_record(self, record):
                 GenomicSetMember.sampleId.is_(None)
             ).first()
 
+    def get_members_for_informing_loop_ready(self, limit=None):
```
j-kanuch, line 1053: Can you set a default limit=0? This way it won't pick anybody up unless there is a value set in the config.

dtharpeuno: Yea if I set the value to 0, I have to evaluate still in the logic since the value of 0 is still False. I know in the tech design it was stated to default this to 0 until ready and adding the value in the config, but if we never add the cron config item for letting the job run then that would suffice as well?

j-kanuch: Ahh, I see, 0 would still be false on line 1108.  It should work if you switch line 1108 to 
```if limit is not None:```

dtharpeuno: Yea that would work, but do we need to add this check if we never enable the cron job?

j-kanuch: That should be fine.

j-kanuch (approved):

j-kanuch (approved):

PR #2938, opened 2022-04-20
Fix pylintrc in Python 3.8

brianmkimmel:

Pylintrc was unable to run in a Python 3.8 environment because of a problem with line continuation.
This shortens the string to exclude the comments and changes the sys.path list to work with multiple versions of Python 3.

## Tests
- [] unit tests


kenny-skaggs (approved):

j-kanuch (approved):
Thanks for making this a separate PR.

PR #2943, opened 2022-04-22
[DA-2611 partial] MIGRATION Add consent_error_report table

pegbertsch:
In order to resolve DA-2611, there needs to be a way to track which `consent_file` records that still have a NEEDS_CORRECTING status have already been reported to PTSC.   

This table just links the `consent_file` records to a `consent_error_report` record and provides a note field that could be used to include info such as the PTSC SD ticket that corresponds to a validation error.   

The error reporting task code is being refactored to insert records  into this table and use it for determining which errors are still unreported,  and will be in a follow-on PR.  There may also be some manual backfilling of this table once the RDR release makes production.

## Tests
Manual tests with local DB


kenny-skaggs (approved):

pegbertsch:
I think the Codacy issue can be ignored here.

brianmkimmel (approved):

PR #2947, opened 2022-04-25
Add site ID to external query; add other models.

j-kanuch:
This PR modifies the CVL test manifest generator tool. This PR adds the gc_site_id filter and other possible CVL models to the external manifests query.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/genomic_datagen_dao.py
+++ b/rdr_service/dao/genomic_datagen_dao.py
@@ -5,7 +5,9 @@ from sqlalchemy.orm import Query, aliased
 from sqlalchemy.sql import functions
 
 from rdr_service.dao.base_dao import BaseDao
-from rdr_service.model.genomics import GenomicInformingLoop, GenomicSetMember
+from rdr_service.dao.genomics_dao import GenomicQueriesDao
+from rdr_service.model.genomics import GenomicInformingLoop, GenomicSetMember, GenomicCVLSecondSample, \
+    GenomicGCValidationMetrics, GenomicCVLAnalysis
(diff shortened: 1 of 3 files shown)


rdr_service/dao/genomic_datagen_dao.py
```diff
@@ -195,11 +197,28 @@ def get_template_by_name(self, project_name, template_name):
...
+                GenomicCVLSecondSample.genomic_set_member_id == GenomicSetMember.id
+            ).outerjoin(
+                GenomicCVLAnalysis,
+                GenomicCVLAnalysis.genomic_set_member_id == GenomicSetMember.id
+            ).filter(GenomicSetMember.sampleId.in_(sample_ids))
+
+            if cvl_id:
+                query_dao = GenomicQueriesDao()
+                gc_site_id = query_dao.transform_cvl_site_id(cvl_id)
+                query = query.filter(GenomicSetMember.gcSiteId == gc_site_id)
```
dtharpeuno, line 222: Thanks for doing this quickly.

We can convert the `transform_cvl_site_id` to a class method and then interact with it as `GenomicQueriesDao.transform_cvl_site_id` instead of having to instantiate the entire `GenomicQueriesDao` class.

dtharpeuno (approved):

brianmkimmel (approved):

rdr_service/dao/genomic_datagen_dao.py
j-kanuch: True that, I'll make that change in a separate PR so that we're not holding up testing.

PR #2953, opened 2022-04-27
[DA-2611] Refactor consent error reporting automation to not rely on validation job timestamps

pegbertsch:
The consent error report automation tasks are triggered at the end of each nightly consent validation job.  They were passed a timestamp from the start of that nightly validation job, used to  filter for `consent_file` records created since that timestamp.  Because the validation job has been regularly failing (OOM killer) before triggering the task,  that's no longer sufficient since there may be error records from previous partial/failed validation runs that are still  not processed.

This refactors the `check_consent_errors` task  to use the ConsentErrorReport table added in PR #2943, for tracking error reports and filtering to find unreported errors from the `consent_file` table.

Main changes:

- Deprecate `created_since` timestamp filters and logic
- Add a `get_unreported_error_ids()` method to find unreported NEEDS_CORRECTING `consent_file` records
- Add records to the ConsentErrorReport tracking table based on the new email reports that get sent
-  Add regeneration of the primary consent metrics records to the PDR  participant data rebuild, since those metrics records can be impacted by the receipt of new ConsentPII `questionnaire_response` payloads (unrelated to PDF validation)
- Enable reports for CareEvolution consent errors; they will only be sent to DRC/RDR recipients and do not have a Jira target email 

Side effect of these changes is the invalid DOB or invalid age at consent errors cannot be reliably discovered/reported every day since they don't have tracking in the `consent_file` table and may not be associated with a NEEDS_CORRECTING record.  DOB/age errors that were associated with the same ConsentPII payload that had a failed primary consent PDF validation may still be discovered when the related PDF error record is processed.

## Tests
- [x] unit tests (updated)

### DIFF
--- a/rdr_service/resource/tasks.py
+++ b/rdr_service/resource/tasks.py
@@ -110,7 +110,7 @@ def check_consent_errors_task(payload):
     origin = payload.get('participant_origin', 'vibrent')
     # DA-2611: Generate a list of all previously unreported errors, based on ConsentErrorReport table content
     gen = ConsentErrorReportGenerator()
-    id_list = gen.get_unreported_error_ids(origin=origin)
+    id_list = gen.get_unreported_error_ids()
     if len(id_list):
         gen.create_error_reports(participant_origin=origin, id_list=id_list)
     else:
--- a/rdr_service/resource/tasks.py
+++ b/rdr_service/resource/tasks.py
@@ -111,7 +111,7 @@ def check_consent_errors_task(payload):
     # DA-2611: Generate a list of all previously unreported errors, based on ConsentErrorReport table content
     gen = ConsentErrorReportGenerator()
     id_list = gen.get_unreported_error_ids()
-    if len(id_list):
+    if id_list and len(id_list):
         gen.create_error_reports(participant_origin=origin, id_list=id_list)
     else:
         logging.info(f'No unreported consent errors found for participants with origin {origin}')
--- a/rdr_service/resource/tasks.py
+++ b/rdr_service/resource/tasks.py
@@ -105,19 +105,16 @@ def batch_rebuild_retention_metrics_task(payload):
 
 def check_consent_errors_task(payload):
     """
-    Review new records from the daily consent validation and generate an automated error report
+    Review previously unreported consent errors and generate an automated error report
     """
-    created_since = payload.get('errors_created_since')
     origin = payload.get('participant_origin', 'vibrent')
-    if isinstance(created_since, datetime):
-        date_filter=created_since
-    elif isinstance(created_since, str):
-        date_filter = datetime.strptime(created_since, "%Y-%m-%dT%H:%M:%S")
-    else:
-        raise ValueError('Invalid errors_created_since value in payload for check_consent_errors_task')
-
+    # DA-2611: Generate a list of all previously unreported errors, based on ConsentErrorReport table content
     gen = ConsentErrorReportGenerator()
-    gen.create_error_reports(errors_created_since=date_filter, participant_origin=origin)
+    id_list = gen.get_unreported_error_ids(origin=origin)
+    if len(id_list):
+        gen.create_error_reports(participant_origin=origin, id_list=id_list)
+    else:
+        logging.info(f'No unreported consent errors found for participants with origin {origin}')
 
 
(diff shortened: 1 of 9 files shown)


pegbertsch:
I've reviewed the Codacy issues and fixed the necessary/most helpful ones.  Others I'm leaving as is because they were either inconsequential (e.g., test case names too long) or the code conforms to the same style/patterns used by other developers in recently approved PRs.

rdr_service/resource/tasks.py
```diff
@@ -105,19 +105,16 @@ def batch_rebuild_retention_metrics_task(payload):
...
-    elif isinstance(created_since, str):
-        date_filter = datetime.strptime(created_since, "%Y-%m-%dT%H:%M:%S")
-    else:
-        raise ValueError('Invalid errors_created_since value in payload for check_consent_errors_task')
-
+    # DA-2611: Generate a list of all previously unreported errors, based on ConsentErrorReport table content
     gen = ConsentErrorReportGenerator()
-    gen.create_error_reports(errors_created_since=date_filter, participant_origin=origin)
+    id_list = gen.get_unreported_error_ids()
+    if len(id_list):
```
kenny-skaggs: If there are no validation errors that need to be reported `get_unreported_error_ids` returns None and the length check would raise an error.

rdr_service/tools/tool_libs/consent_error_report.py
```diff
@@ -80,17 +80,19 @@ def execute(self):
...
 
-        errors_since = self.args.errors_since
         report = ConsentErrorReportGenerator()
+        id_list = self.id_list
         # Specific ids will override any date filter
-        if self.id_list:
-            errors_since = None
-        report.create_error_reports(errors_created_since=errors_since,
+        if not id_list:
+            report.get_unreported_error_ids()
```
kenny-skaggs: Why is `get_unreported_error_ids` called here?

pegbertsch: Because a specific id list wasn't provided (--id or --from-file options) so it will default to generating the report for outstanding unreported errors.    Fixed the related issue with `get_unreported_error_ids()` possibly returning None, and made some other changes for clarity

kenny-skaggs: I may still be missing something, but it seems like `get_unreported_error_ids` doesn't store the list of ids on it's own (just returns them). Should the returned list be stored in something like `id_list`?

rdr_service/tools/tool_libs/consent_error_report.py
```diff
@@ -80,17 +83,21 @@ def execute(self):
...
-                                    recipients=self.recipients,
-                                    cc_list=self.cc_recipients,
-                                    participant_origin=self.args.origin,
-                                    to_file=self.args.to_file)
+        # If no user-provided ids were specified (--id or --from-file), default to all unreported errors
+        id_list = self.id_list or report.get_unreported_error_ids()
+
+        if id_list and len(id_list):
+            report.create_error_reports(
+                                        id_list=self.id_list,
```
kenny-skaggs: I see the new changes and that helps clear things up for me, thanks. One last thing, should this be using the `id_list` (that is potentially retrieving the unreported errors) rather than `self.id_list`?

pegbertsch: Had to fix another typo from my last commit (sorry).

Per using `or` with objects:  the assignment statement on line 88 will assign the first "truthy" object.  `self.id_list` will be assigned if it is not None (i.e., user specified an `--id `or `--from-file` list).   So the intended behavior is that user-specified ids from the command line will  override the default (`get_unreported_error_ids()` )

kenny-skaggs: So should the call to `create_error_reports` be receiving `id_list` (which could be the result of the `get_unreported_error_ids` or user input) instead of `self.id_list` (which, from what I understand could only be user input)?

pegbertsch: Thanks, Kenny.  Misunderstood which line you were referring to.  Just pushed the change to line 92

kenny-skaggs (approved):

robabram (approved):

PR #2963, opened 2022-05-05
[PDR-842] Allow participant research_id to populate in PDR.

robabram:
Allow the participant's 'research_id' value to be in PDR now.

## Tests
- [x] unit tests


dtharpeuno (approved):

pegbertsch (approved):
Reminder since BQ isn't quite dead yet:  is anyone (incl. NiFi) still using the `v_pdr_participant_withdrawn` custom BQ view?  That could require a manual update to the SQL definition  to add columns since that was converted to a custom view (see the commented out view definition on line 278 in `bq_pdr_participant_summary.py` )

PR #2969, opened 2022-05-06
Daily Drop of VA Work Queue Participants to Bucket

brianmkimmel:

Creates two jobs for handling daily drop of VA work queue participants.
ExportVaWorkQueue creates the csv file and saves it in the VA's bucket.
DeleteOldVaWorkQueueReports deletes files with a timestamp in the file name that is more than 7 days old.

## Tests
- [x] unit tests


rdr_service/offline/export_va_workqueue.py
```diff
@@ -0,0 +1,326 @@
...
+def get_workqueue_participants():
+    with ParticipantSummaryDao().session() as session:
+        return session.query(ParticipantSummary
+                            ).join(HPO, HPO.hpoId == ParticipantSummary.hpoId
+                            ).join(Participant, Participant.participantId == ParticipantSummary.participantId
+                            ).filter(HPO.name == 'VA',
+                                     Participant.isTestParticipant != 1,
+                                     # Just filtering on isGhostId != 1 will return no results
+                                     or_(Participant.isGhostId != 1, Participant.isGhostId == None)
+                            ).all()
```
dtharpeuno: We generally follow the practice of the dao's being the database layer for queries. 
Can we move this to the `ParticipantSummaryDao`?

brianmkimmel: I've moved the query into ParticipantSummaryDao

rdr_service/offline/export_va_workqueue.py
```diff
@@ -0,0 +1,223 @@
...
+from rdr_service.dao.participant_summary_dao import ParticipantSummaryDao
+
+CSV_HEADER = ['PMI ID',
+              'Biobank ID',
+              'Last Name',
+              'First Name',
+              'Middle Initial',
+              'Age Range',
+              'Participant Status',
+              'Primary Consent Status',
```
kenny-skaggs, line 18: Should the export have the authored dates for the consents and survey completion dates?

brianmkimmel: The dictionary created by ParticipantSummaryDao.to_client_json doesn't include the datetime fields if they're null even when the line to strip None values is skipped. I could include those values by adding lines like participant_filtered.setdefault('consentForStudyEnrollmentAuthored','UNSET') before creating participant_row

kenny-skaggs: I think that will be good. We'll want to make sure the export has the same data as the work queue.

rdr_service/offline/export_va_workqueue.py
```diff
@@ -0,0 +1,369 @@
...
+from rdr_service.offline.sql_exporter import SqlExporter
+from rdr_service.dao.hpo_dao import HPODao
+from rdr_service.dao.participant_summary_dao import ParticipantSummaryDao
+
+CSV_HEADER = ['PMI ID',
+              'Biobank ID',
+              'Last Name',
+              'First Name',
+              'Middle Initial',
+              'Age Range',
```
kenny-skaggs, line 16: Thanks for getting that unit test set up. The only other thing I'm seeing that might need to be updated is that the date of birth should be included in the file output (both here in the headers, and down where the rows are written to the file).

brianmkimmel: It should be. I was looking at to_client_json and thought dateOfBirth was being replaced with ageRange because of the way it handles all of the datetime fields. I've added it in now.

dtharpeuno (approved):

kenny-skaggs (approved):

PR #2971, opened 2022-05-06
[PDR-908] Create a repeatable unique id for all biobank order and sample records.

robabram:
Create a repeatable unique record primary key for every biobank order and sample record.

## Tests
- [x] unit tests


tests/helpers/unittest_base.py
```diff
@@ -502,7 +502,7 @@ def setUp(self, with_data=True, with_consent_codes=False) -> None:
         self.addCleanup(logger.removeHandler, stream_handler)
 
         # Change this to logging.ERROR when you want to see API server errors.
-        logger.setLevel(logging.CRITICAL)
+        logger.setLevel(logging.ERROR)
```
pegbertsch: need to revert this to CRITICAL?

wangy70: Do you intend to submit this change? Or by accident?

rdr_service/resource/generators/participant.py
```diff
@@ -947,7 +948,18 @@ def _make_sample_dict_from_row(bss=None, bos=None):
...
 
+            # Create a unique repeatable primary key value for each biobank sample record.
+            test_id = BIOBANK_UNIQUE_TEST_IDS[test] if test in BIOBANK_UNIQUE_TEST_IDS else '99'
+            if bo_pk:
+                # From known biobank orders
+                id_ = int(f'{bo_pk}{test_id}')
+            else:
+                # For unknown biobank orders, use participant_id + 99 + test id.
+                # All tests here will be grouped together under this 99 id.
+                id_ = int(f'{bss.participant_id}99{test_id}')
```
pegbertsch: Have we verified if there are any cases where a participant has more than one `biobank_stored_sample` row for the same test type, neither of which has a `biobank_order_id`?  Would that break the key uniqueness?

pegbertsch (commented):
Believe the unittest logging level change needs reverting before this is merged.

tests/helpers/unittest_base.py
robabram: It's by accident.  Thank you for catching it.

wangy70 (approved):

pegbertsch (approved):

PR #2976, opened 2022-05-09
Updates to survey response validation

kenny-skaggs:
## Resolves *[DA-2595](https://precisionmedicineinitiative.atlassian.net/browse/DA-2595)*
Making some adjustments to the response validation script:
- Ignoring data for test participants
- Enabling check for requirements specified in recap (previously just the branching logic was checked, now things like min/max of the answer and the data type are checked)
- Using logger set up for the scripts so the log lines get printed to the console
- Fixing how the rebuild tasks were sent

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/dao/participant_summary_dao.py
+++ b/rdr_service/dao/participant_summary_dao.py
@@ -39,6 +39,7 @@ from rdr_service.dao.site_dao import SiteDao
 from rdr_service.model.config_utils import from_client_biobank_id, to_client_biobank_id
 from rdr_service.model.consent_file import ConsentType
 from rdr_service.model.retention_eligible_metrics import RetentionEligibleMetrics
+from rdr_service.model.participant import Participant
 from rdr_service.model.participant_summary import (
     ParticipantGenderAnswers,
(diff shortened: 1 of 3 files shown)


rdr_service/dao/participant_summary_dao.py
```diff
@@ -1118,7 +1119,14 @@ def update_ehr_status(summary, update_time):
 
     @classmethod
     def get_all_consented_participant_ids(cls, session):
-        db_results = session.query(ParticipantSummary.participantId).all()
+        db_results = session.query(
```
pegbertsch, line 1122: Would there be value in excluding ghost ids, or do they have to be treated as potentially valid?

kenny-skaggs: If they're marked as ghosts and we're getting survey responses from them, then there's likely some other investigation that needs to be happening and filtering them out here would cause us to miss that.

dtharpeuno (approved):

wangy70 (approved):

PR #2983, opened 2022-05-13
Adding more checks in CVL/Manifest ingestion

dtharpeuno:
CVLs are sending data in some files that are not expected in general development, also removing manifest from the queue if an exception occurs.

## Tests
- [] unit tests
** current tests should still pass


j-kanuch (approved):
Thanks for finding and fixing this!

wangy70 (approved):

PR #2984, opened 2022-05-13
[PDR-852] updated UBR calculation from new data glossary.

robabram:
Version 3.1 of the Data Glossary has been approved.  This PR updates the UBR calculation code to be inline with the new version.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/calculators/participant_ubr.py
+++ b/rdr_service/resource/calculators/participant_ubr.py
@@ -91,6 +91,9 @@ class ParticipantUBRCalculator:
         answers = [i for i in tmp_list if i]
         if not answers:
             return UBRValueEnum.RBR
+        # If the gender_ident_closer list is empty, remove the 'GenderIdentity_AdditionalOptions' answer if needed.
+        if not gender_ident_closer and 'GenderIdentity_AdditionalOptions' in answers:
+            answers.remove('GenderIdentity_AdditionalOptions')
 
         # Assume ubr by default and check for exclusion criteria. If there is more than one answer in the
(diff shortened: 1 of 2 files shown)


rdr_service/resource/calculators/participant_ubr.py
```diff
@@ -90,7 +92,8 @@ def ubr_gender_identity(birth_sex, gender_ident, gender_ident_closer):
         if not answers:
             return UBRValueEnum.RBR
 
-        # Note: assume ubr by default and check for exclusion criteria.
+        # Assume ubr by default and check for exclusion criteria. If there is more than one answer in the
+        # list, assume UBR.
```
pegbertsch, line 99: To confirm:   if someone has `GenderIdentity_AdditionalOptions` as an extra answer in their `gender_ident` answer list, but then did not provide any answer  (including `PMI_Skip`) to `gender_ident_closer`, they should be UBR?  E.g., we have two pids with this for `birth_sex / gender_ident / gender_closer`:

`SexAtBirth_Female / GenderIdentity_Woman,GenderIdentity_AdditionalOptions / None`

pegbertsch (approved):
Approved, unless we get clarification for the corner case mentioned.

rdr_service/resource/calculators/participant_ubr.py
robabram: I have added a check for this edge case and now remove the 'GenderIdentity_AdditionalOptions' answer if needed.

dtharpeuno (approved):

PR #2988, opened 2022-05-17
[DA-2660] Fixes for CE consent file validation

kenny-skaggs:
## Resolves *[DA-2660](https://precisionmedicineinitiative.atlassian.net/browse/DA-2660)*
This updates the validation of CE files with new versions of the Spanish files. This will help keep us from flagging files that are valid but the code is finding as invalid.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/services/consent/files.py
+++ b/rdr_service/services/consent/files.py
@@ -252,7 +252,8 @@ class CeConsentFactory(ConsentFileAbstractFactory):
         pdf = blob_wrapper.get_parsed_pdf()
         return pdf.has_text([(
             'Consent to Join the All of Us Research Program',
-            'Consentimiento para Participar en el Programa Científico'
+            'Consentimiento para Participar en el Programa Científico',
+            'Consentimiento para Participar en el\nPrograma Cientíﬁco'
         )])
 
     def _is_cabor_consent(self, blob_wrapper: '_ConsentBlobWrapper') -> bool:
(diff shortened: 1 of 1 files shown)


dtharpeuno (approved):

wangy70 (approved):

rdr_service/services/consent/files.py
```diff
@@ -623,39 +624,31 @@ def _try_to_parse_signature(self, footer_text):
...
 
-        self.shift_params = (8, 10)
-        signature_str = self._try_to_parse_signature("Name (printed)")
-        if signature_str:
-            return signature_str
+        signature_label_strings = [
+            "Participant's Name (printed)",
+            "'s Name (printed)",
+            "Name (printed)",
+            'Nombre del Participant',
```
pegbertsch, line 636: @kenny-skaggs FYI: I was looking at a CE "missing signature" error report from this morning and the file in question (that had a printed name on the signature line) had the text "Nombre de Participant"  (instead of "del").  I'm assuming this list needs to contain all possible exact string matches?

pegbertsch, line 636: @kenny-skaggs FYI: I was looking at a CE "missing signature" error report from this morning and the file in question (that had a printed name on the signature line) had the text "Nombre de Participant"  (instead of "del").  I'm assuming this list needs to contain all possible exact string matches?

kenny-skaggs: Yes, I'll get the new one added.

PR #2989, opened 2022-05-17
[DA-2648] Have CircleCI check that the app is successfully deploy to the test environment

kenny-skaggs:
## Resolves *[DA-2648](https://precisionmedicineinitiative.atlassian.net/browse/DA-2648)*
We've seen that everything can successfully run for the unit tests in CircleCI, but then when deployed to the Test environment the instance can fail to start. This adds a step to CircleCI that will check to be sure that the instance is running and responding to requests.

As of the current state of configuration, if an instance receives a GET request at the root url it will respond with the version number. This request fails if there is no instance running to process it.

## Tests
- [ ] unit tests

### DIFF
--- a/ci/verify_instance.py
+++ b/ci/verify_instance.py
@@ -1,0 +1,18 @@
+import logging
+import sys
+
+import requests
+
+
+logger = logging.getLogger()
+stream_handler = logging.StreamHandler(sys.stdout)
+logger.addHandler(stream_handler)
+logger.setLevel(logging.INFO)
+
+response = requests.get('https://ks-check-instance-running-1-dot-all-of-us-rdr-sandbox.appspot.com/')
+
+if response.status_code != 200:
+    logger.error(f'ERROR: server responded with {response.status_code} status code')
+    exit(1)
+else:
+    logger.info('Instance successfully responded!')
--- a/ci/verify_instance.py
+++ b/ci/verify_instance.py
@@ -9,7 +9,12 @@ stream_handler = logging.StreamHandler(sys.stdout)
 logger.addHandler(stream_handler)
 logger.setLevel(logging.INFO)
 
-response = requests.get('https://ks-check-instance-running-1-dot-all-of-us-rdr-sandbox.appspot.com/')
+project_id = sys.argv[1]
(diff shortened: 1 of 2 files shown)


ci/verify_instance.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+logger = logging.getLogger()
+stream_handler = logging.StreamHandler(sys.stdout)
+logger.addHandler(stream_handler)
+logger.setLevel(logging.INFO)
+
+project_id = sys.argv[1]
+version_id = sys.argv[2]
+app_engine_url = f'https://{version_id}-dot-{project_id}.appspot.com/'
```
dtharpeuno, line 14: What route is this? Never know this existed. Does it need to have the `offline` prefix?

ci/verify_instance.py
```diff
@@ -0,0 +1,23 @@
...
+
+logger.info(f'Checking "{app_engine_url}" for a running instance')
+response = requests.get(app_engine_url)
+
+if response.status_code != 200:
+    logger.error(f'ERROR: server responded with {response.status_code} status code')
+    exit(1)
```
dtharpeuno, line 21: Will this throw a "failure" on the CircleCI job?

ci/verify_instance.py
kenny-skaggs: It doesn't seem to need to have a service prefix (like `offline`). I've tried a few different versions in sandbox and they spin up an instance (of that version) on the `default` service. It's documented here https://cloud.google.com/appengine/docs/standard/python/how-requests-are-routed

ci/verify_instance.py
kenny-skaggs: Yes. If any command run by a CircleCI step exits with any code other than 0 then the step is marked as a failure and the workflow fails.

dtharpeuno: Nice, thanks for clarifying

ci/verify_instance.py
dtharpeuno: 👍

dtharpeuno (approved):

j-kanuch (approved):

PR #2991, opened 2022-05-18
More fixes for processing new layouts of CE consent PDFs

kenny-skaggs:
## Resolves *no ticket*
This updates the validation of CE files to help the code find the signature and signing date in new layouts of the CE consent PDFs.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/services/consent/files.py
+++ b/rdr_service/services/consent/files.py
@@ -625,16 +625,15 @@ class CeFileWrapper:
 
     def get_signature_on_file(self):
         signature_offsets = [
-            (8, 10),
-            (73, 75)
+            (8, 15),
+            (73, 80)
         ]
 
         signature_label_strings = [
             "Participant's Name (printed)",
             "'s Name (printed)",
             "Name (printed)",
-            'Nombre del Participant',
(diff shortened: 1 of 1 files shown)


rdr_service/services/consent/files.py
```diff
@@ -625,16 +625,15 @@ def _try_to_parse_signature(self, footer_text):
...
+            (73, 80)
         ]
 
         signature_label_strings = [
             "Participant's Name (printed)",
             "'s Name (printed)",
             "Name (printed)",
-            'Nombre del Participant',
-            'Nombre del'
+            'Nombre'
```
pegbertsch, line 636: Use/matching of these strings is always case sensitive, correct?

pegbertsch (approved):
Approved assuming that the `signature_label_strings` are always treated as case sensitive.

rdr_service/services/consent/files.py
kenny-skaggs: Yeah, but I don't think it would need to be case sensitive. Do you think it would need to be?

dtharpeuno (approved):

PR #3014, opened 2022-06-02
[DA-2518] Unblocking Tool

brianmkimmel:
Creates tool to unblock participants from a file containing a list of sample ids or biobank ids.
Will run AW1 and AW2 ingestion for records with missing data matching a raw manifest.
Will run data file reconciliation for records with missing data files.n 

## Tests
- [x] unit tests


rdr_service/dao/genomics_dao.py
```diff
@@ -1239,6 +1239,25 @@ def update(self, obj):
...
+        """
+        Returns genomicSetMember objects for list of Biobank IDs.
+        :param biobank_ids:
+        :param genome_type:
+        :return:
+        """
+        with self.session() as session:
+            members = session.query(GenomicSetMember).filter(
+                GenomicSetMember.biobankId.in_(biobank_ids),
+                GenomicSetMember.genomicWorkflowState != GenomicWorkflowState.IGNORE,
```
j-kanuch, line 1252: You'll want to check `GenomicSetMember.ignore_flag == 0` here too.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+class UnblockSamples(ToolBase):
+    """
+    Unblocks a set of samples.
+    """
 
+    def __init__(self, args, gcp_env: GCPEnvConfigObject):
+        super().__init__(args, gcp_env)
+
+    def run(self):
+        self.gcp_env.activate_sql_proxy()
```
j-kanuch, line 2172: In general, the readability of this method could be improved by breaking it into smaller chunks. I would split it into AW1, AW2, and Reconciliation methods and call them from the `run` method. There is cyclomatic complexity in this method that could be simplified. I've pointed out where I would simplify some of the if statements.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+        member_dao = GenomicSetMemberDao()
+        metrics_dao = GenomicGCValidationMetricsDao()
+        aw1_raw_dao = GenomicAW1RawDao()
+        aw2_raw_dao = GenomicAW2RawDao()
+
+        no_aw1_data = []
+        no_aw2_data = []
+        needs_data_file_reconciliation = []
+
+        if self.args.file_path:
```
j-kanuch: You have it marked as a required argument, so this condition isn't needed.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+            if not os.path.exists(self.args.file_path):
+                _logger.error(f'File {self.args.file_path} was not found.')
+                return 1
+        else:
+            _logger.error('Need --file-path')
+            return 1
+        if self.args.results_only is True and self.args.research_only is True:
```
j-kanuch: You don't need `is True`. It can be `if self.args.results_only and self.args.research_only:` Same goes for the other conditions below where you have `is True`.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+            _logger.error('Can only set one of --research-only or --results-only')
+            return 1
+
+        unblock_results = True
+        unblock_research = True
+        if self.args.results_only is True:
+            unblock_research = False
+        elif self.args.research_only is True:
+            unblock_results = False
```
j-kanuch: You can get rid of 2190 - 2195. See comment below on line 2217.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+
+            for row in csv_reader:
+                id_list.append(row[0])
+
+            if id_type == 'sample_id':
+                set_members = member_dao.get_members_from_sample_ids(id_list)  # will not return results in IGNORE state
+            else:
+                set_members = member_dao.get_members_from_biobank_ids(id_list)
+            for set_member in set_members:
+                if set_member.ignoreFlag == 1:
```
j-kanuch: Implementing my comment on the `get_members_from_biobank_ids` method will render this condition unnecessary.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+                    continue
+                if unblock_results:
```
j-kanuch: You can get rid of the conditions 2190-2195 by changing 2217-2220 to the following:
`set_member.blockResults = 0
set_member.blockResearch = 0
if self.args.results_only: 
     set_member.blockResearch = 1
 if self.args.research_only: 
     set_member.blockResults = 1`

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+                        # Perform AW2 ingestion
+                        if not self.args.dryrun:
+                            self._ingest_member('METRICS_INGESTION', server_config, set_member)
+                        else:
+                            _logger.info(f"Will try AW2 ingestion for {set_member.id}")
+                    else:
+                        no_aw2_data.append(set_member.sampleId)
+                else:
+                    if set_member.genomeType == 'aou_array':
+                        if vmetric.idatRedReceived != 1 or vmetric.idatGreenReceived != 1 \
```
j-kanuch: You can use `if any()` here instead of a series of `or` clauses. Also, check out `rdr_service/genomic/genomic_mappings.py:137-210` and `rdr_service/genomic/genomic_job_components.py:2460` you can use that pattern to check for required files.

rdr_service/dao/genomics_dao.py
```diff
@@ -3024,16 +3043,22 @@ def truncate(self):
...
+            query = session.query(
                 GcDataFileStaging
             ).outerjoin(
                 GenomicGcDataFile,
                 GcDataFileStaging.file_path == GenomicGcDataFile.file_path
             ).filter(
                 GenomicGcDataFile.id.is_(None)
-            ).all()
+            )
+            if sample_ids is not None:
```
dtharpeuno: :nit Since sample_ids will be a list probably be better to check like this:
`if sample_ids`

rdr_service/dao/genomics_dao.py
```diff
@@ -3024,16 +3043,22 @@ def truncate(self):
...
+                query.filter(
```
dtharpeuno: Does this need to be `query = query.filter`?

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+        if self.args.results_only is True:
+            unblock_research = False
+        elif self.args.research_only is True:
+            unblock_results = False
+
+        with open(self.args.file_path, encoding='utf-8-sig') as file:
+            csv_reader = csv.reader(file)
+            id_list = []
+            header = next(csv_reader)[0]
+            if header == 'sample_id' or header == 'biobank_id':
```
dtharpeuno: Probably could simplify this with `if header in <tuple of strings>`

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2148,8 +2155,136 @@ def make_new_set():
...
+                    continue
+                if unblock_results:
+                    set_member.blockResults = 0
+                if unblock_research:
+                    set_member.blockResearch = 0
+                if not self.args.dryrun:
+                    member_dao.update(set_member)
+                else:
+                    _logger.info(f"Will unblock genomic_set_member {set_member.id}")
+                if id_type == 'biobank_id' and set_member.sampleId is None:
```
dtharpeuno: :nit `and not set_member.sampleId`

rdr_service/tools/tool_libs/genomic_utils.py
brianmkimmel: There are records that have block_research = 1 and block_results = 0. If those were unblocked with research_only applied this logic would change block_results to 1.

j-kanuch: I see, that makes sense. I guess we do need it in the else then:
```if self.args.results_only: 
  set_member.blockResults = 0
elif self.args.research_only: 
  set_member.blockResearch = 0
else:
  set_member.blockResearch = 0
  set_member.blockResults = 0```

rdr_service/dao/genomics_dao.py
brianmkimmel: Does GenomicSetMemberDao.get_members_from_sample_ids need this added to the filter as well?

rdr_service/dao/genomics_dao.py
brianmkimmel: Yes, fixed this in the new commit.

dtharpeuno (approved):

rdr_service/dao/genomics_dao.py
j-kanuch: Yea, we probably should add it. If it doesn't affect this PR, we can have it as a todo.

j-kanuch (approved):

PR #3020, opened 2022-06-08
[PDR-863] Filter pending duplicate responses from PDR data

pegbertsch:

Currently,  the PDR participant generator for the  `participant_module` survey response summary history has an `is_replay()` method (existed before `answer_hash` values were added to RDR `questionnaire_response` records) to recognize and exclude replays/duplicate responses.

The PDR module data generator had no corresponding filtering.  This adds the logic to look for duplicate responses which may not have been flagged yet in RDR, and proactively exclude them from PDR module data builds.

Also fixed:

- corrected placement of the `order by` clause when retrieving participant module responses in participant generator
- Made the participant generator `is_replay()` logic more consistent with the RDR `ResponseDuplicationDetector ` class logic so that potential duplicates are not missed.

## Tests
- [x] unit tests
- Manual testing of participants with unflagged duplicates to make sure PDR participant data and module data included the same `questionnaire_response_id` records

### DIFF
--- a/rdr_service/dao/bq_questionnaire_dao.py
+++ b/rdr_service/dao/bq_questionnaire_dao.py
@@ -52,7 +52,7 @@ class BQPDRQuestionnaireResponseGenerator(BigQueryGenerator):
         # Intended to make filtering module response data by test_participant status more efficient in BigQuery
         _participant_module_responses_sql = """
             select qr.questionnaire_id, qr.questionnaire_response_id, qr.created, qr.authored, qr.language,
-                   qr.participant_id, qh2.external_id, qr.status,
+                   qr.participant_id, qh2.external_id, qr.status, qr.answer_hash,
                    CASE
                        WHEN p.is_test_participant = 1  or p.is_ghost_id = 1 or h.name = :test_hpo THEN 1
                        ELSE 0
@@ -71,8 +71,8 @@ class BQPDRQuestionnaireResponseGenerator(BigQueryGenerator):
                       and qc.questionnaire_version = qh.version
                 inner join code c on c.code_id = qc.code_id
                 where c.value = :module_id and c.system = :system
-                order by qr.created DESC
-            );
+            )
+            order by qr.created DESC;
         """
 
         # Query to get questionnaire answers for a specific questionnaire_response_id (no ordering needed)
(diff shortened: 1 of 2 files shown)


rdr_service/dao/bq_questionnaire_dao.py
```diff
@@ -113,9 +113,20 @@ def make_bqrecord(self, p_id, module_id, latest=False, convert_to_enum=False):
...
                                                                             'test_hpo': TEST_HPO_NAME,
                                                                             'system': PPI_SYSTEM})
+            answer_hashes_processed = []
             for qr in responses:
+                if qr.answer_hash in answer_hashes_processed:
+                    # PDR-863:  This response is pending being flagged by RDR as a duplicate (same answer hash as
+                    # a more recently created response).  Proactively exclude from PDR data to avoid orphaned records
+                    continue
+
+                answer_hashes_processed.append(qr.answer_hash)
```
kenny-skaggs, line 123: Does this only compare answer hashes? We could receive some responses multiple times that have the same answer hash (such as responding multiple times to EHR or GROR consent). Using external id as part of the comparison could help make sure distinct responses that have the same answers are accounted for.

kenny-skaggs, line 123: Does this only compare answer hashes? We could receive some responses multiple times that have the same answer hash (such as responding multiple times to EHR or GROR consent). Using external id as part of the comparison could help make sure distinct responses that have the same answers are accounted for.

pegbertsch: My thinking is for PDR, if they supplied the same answers to the same questions for the same module, what we want is just to keep the most recently received response of any group that has identical question/answer content, regardless of what form entry id / `external_id` the sender put on it.  I may try to set up a huddle with @robabram  so we can talk this through offline.  

How often does PTSC (or CE) resubmit the exact same content with a different form entry id?

kenny-skaggs: I'm not sure how often it happens, but there are some surveys that the meta-data and timing relative to other responses matters as much (if not more) than the answer content. Such as if a participant goes back and forth on providing Yes and No responses to EHR/GROR. Or responses to digital health sharing data that don't have answers, but instead depend on the survey code being used (like `participantintendstosharefitbit` and `participantintendstostopsharingfitbit`).

Will PDR users only ever need the the latest response for a given survey, or is it possible that they'd need time ranges for things like the consents above?

pegbertsch: I think the authored timestamp is the more important differentiator.  E.g, if it's a different authored time but everything else  (regardless of `external_id`) matches then we should keep both.   if it's an identical authored time and answered hash, we just want the most recently received "copy".

I'll put this on hold to revise.

kenny-skaggs: Sounds good. I agree that authored time would be ideal, and I think we would have gone with that for the duplicate detection in RDR. But we've seen that the authored times can shift by a small amount between replays of responses, so for something we know to be a replay of a response we've seen differing authored times but the same external id.

pegbertsch:
Closing this pull request as the work keeps getting bumped and it may be superseded by new RDR-PDR data pipeline

PR #3025, opened 2022-06-10
[DA-2736] Recognition and validation of sensitive EHR consent

kenny-skaggs:
## Resolves *[DA-2736](https://precisionmedicineinitiative.atlassian.net/browse/DA-2736)*
This adds validation of the new sensitive EHR consent files (validation that the files have initials where required, and that the sensitive form was provided to participants in the correct states). This also adds to the QuestionnaireResponseDao to recognize when EHR consent was provided through the sensitive form.

## Tests
- [x] unit tests


dtharpeuno (approved):

rdr_service/code_constants.py
```diff
@@ -31,6 +31,7 @@
 EDUCATION_QUESTION_CODE = "EducationLevel_HighestGrade"
 INCOME_QUESTION_CODE = "Income_AnnualIncome"
 EHR_CONSENT_QUESTION_CODE = "EHRConsentPII_ConsentPermission"
+EHR_SENSITIVE_CONSENT_QUESTION_CODE = "ehrconsentpii_sensitivetype2"
```
pegbertsch, line 34: I'm wondering if we need to recognize when a participant responds to the `decisionoptions`  question code with a `removeauthorization` response and process that as if it were a "no consent" (`sensitivetype2__donotagree`)

kenny-skaggs: Agreed, we should make sure we're aware of how that will work. In the meeting this morning it sounds like the decsionoptions flow exists in the screens now, but is mapped to a No answer before getting to us. Do you think this PR should wait until that is resolved, or can the work of validation and happy-path consent recognition be merged (with a follow up PR if needed)?

pegbertsch: I got the impression from our Monday standup that Megan wasn't sure yet what the UI flow is or what if any mapping of answers exists? We have the follow-up discussion on Tuesday.   

I would be concerned if the "happy path" has gone live, but `removeauthorization` codes ultimately show up,  (including possibly due to erroneous payload construction on the portal side).  RDR would just store the data without updating consent status/POST would still be successful, and participant would think they successfully revoked consent?

pegbertsch: Per follow-up, this needs to be merged so we can test on stable

pegbertsch (approved):

PR #3026, opened 2022-06-10
Add investigation genome types to reconciliation query

brianmkimmel:
Updates the query used by the RECONCILE_ARRAY_DATA and RECONCILE_WGS_DATA genomic jobs so they will find genomic set members with investigation genome types.  

## Tests
- [] unit tests


dtharpeuno (commented):
Can you add this change to the unit tests as well to make sure the files get picked up? We can tell by the logic that this will work but still like to encompass unit test additions/updates with logic changes.

dtharpeuno (approved):

kenny-skaggs (approved):

PR #3027, opened 2022-06-13
[PDR-912[ Update/add genomic tables for PDR.

robabram:
* Update existing PDR models with removed and new genomics table fields 
* Add 'genomic_informing_loop' table to resources.
* Added initial support for sending Pub/Sub events for new RDR to PDR data pipeline. (non-prod for now).

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/resource/schemas/genomics.py
+++ b/rdr_service/resource/schemas/genomics.py
@@ -93,12 +93,11 @@ class GenomicSetMemberSchema(Schema):
     cvl_w4f_manifest_job_run_id = fields.Int32()
     genomic_workflow_state = fields.EnumString(enum=GenomicWorkflowState)
     genomic_workflow_state_id = fields.EnumInteger(enum=GenomicWorkflowState)
-
-    genomic_workflow_state_history = fields.JSON()
+    # genomic_workflow_state_history = fields.JSON()
     collection_tube_id = fields.String(validate=validate.Length(max=80))
     gc_site_id = fields.String(validate=validate.Length(max=11))
-    arr_aw3_manifest_job_run_id = fields.Int32()
-    wgs_aw3_manifest_job_run_id = fields.Int32()
+    # arr_aw3_manifest_job_run_id = fields.Int32()
+    # wgs_aw3_manifest_job_run_id = fields.Int32()
     genomic_workflow_state_modified_time = fields.DateTime()
     report_consent_removal_date = fields.DateTime()
     qc_status = fields.EnumString(enum=GenomicQcStatus)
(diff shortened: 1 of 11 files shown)


dtharpeuno (approved):

requirements_base.txt
```diff
@@ -73,7 +73,7 @@ parameterized==0.7.4      # via -r requirements.in
 pbr==5.4.5                # via jira
 pdfminer.six==20201018
 # pip-tools==5.1.2          # via -r requirements.in
-protobuf==3.15.0          # via -r requirements.in, google-api-core, google-cloud-bigquery, googleapis-common-protos
+protobuf==3.20.1          # via -r requirements.in, google-api-core, google-cloud-bigquery, googleapis-common-protos
```
pegbertsch, line 76: What triggered this requirements update?

rdr_service/resource/schemas/genomics.py
```diff
@@ -177,6 +216,7 @@ class GenomicManifestFileSchema(Schema):
     rdr_processing_complete_date = fields.DateTime()
     ignore_flag = fields.Int16()
     file_name = fields.String(validate=validate.Length(max=255))
+    manifest_type_id_str = fields.String(validate=validate.Length(max=64))
```
pegbertsch: Maybe this isn't the convention for the genomics tables, but there is an enum class for these manifest types (`rdr_service/genomic_enums.py GenomicManifestTypes` .  Do we want to have the `manifest_type` / `manifest_type_id` paired fields like we do for other PDR schemas?

rdr_service/resource/schemas/genomics.py
```diff
@@ -93,12 +93,11 @@ class GenomicSetMemberSchema(Schema):
     cvl_w4f_manifest_job_run_id = fields.Int32()
     genomic_workflow_state = fields.EnumString(enum=GenomicWorkflowState)
     genomic_workflow_state_id = fields.EnumInteger(enum=GenomicWorkflowState)
-
-    genomic_workflow_state_history = fields.JSON()
+    # genomic_workflow_state_history = fields.JSON()
```
pegbertsch: What triggered the commenting out of the fields in this table?

requirements_base.txt
robabram: If you jump to the github home page for the RDR project, it is displaying a library version vulnerability alert. https://github.com/all-of-us/raw-data-repository.
This takes care of that alert.

rdr_service/resource/schemas/genomics.py
robabram: Good catch, I didn't pay attention to this field.  The RDR devs have added an enum string field for the 'manifest_type_id' field, but we already handled having both the id and str fields previously.  I have removed the new field as it is not needed, in my next commit.

rdr_service/resource/schemas/genomics.py
robabram: I commented out those fields when I found they had been removed from the RDR table.  I have removed them in my next PR.

rdr_service/resource/schemas/genomics.py
robabram: Turns out I was adding a few unneeded enum string fields.  I should have looked harder at those new fields.  I'm glad you caught this.

pegbertsch (approved):
Newline/codacy behavior is just FYI.  The other Codacy issues are based on var/function naming that we don't want to change so there won't be a clean Codacy run regardless.

PR #3031, opened 2022-06-15
[DA-2626] Checking for W1IL participants that have revoked and then re-submitted GROR consent

kenny-skaggs:
## Resolves *[DA-2626](https://precisionmedicineinitiative.atlassian.net/browse/DA-2626)*
When a participant has appeared in the W1IL and then later revokes their GROR consent their results are hidden from them. But if they go back and give GROR consent again (respond with a Yes) they should be able to view their results again.

This adds a cron job that runs weekly that detects participants that fall into that category and sends an email to notify that they should be able to view their results again.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/offline/genomic_pipeline.py
+++ b/rdr_service/offline/genomic_pipeline.py
@@ -1,14 +1,14 @@
 import logging
 
-from rdr_service import clock
-from rdr_service.dao.genomics_dao import GenomicAW1RawDao, GenomicAW2RawDao, GenomicAW3RawDao, GenomicAW4RawDao, \
-    GenomicJobRunDao, GenomicW2SCRawDao, GenomicW3SRRawDao, GenomicW4WRRawDao, GenomicW3SCRawDao, GenomicW3NSRawDao, \
-    GenomicW5NFRawDao, GenomicW3SSRawDao, GenomicW2WRawDao, GenomicW1ILRawDao
+from rdr_service import clock, config
+from rdr_service.dao.genomics_dao import GenomicQueriesDao, GenomicAW1RawDao, GenomicAW2RawDao, GenomicAW3RawDao, \
+    GenomicAW4RawDao, GenomicJobRunDao, GenomicW2SCRawDao, GenomicW3SRRawDao, GenomicW4WRRawDao, GenomicW3SCRawDao, \
+    GenomicW3NSRawDao, GenomicW5NFRawDao, GenomicW3SSRawDao, GenomicW2WRawDao, GenomicW1ILRawDao
 from rdr_service.genomic.genomic_cvl_reconciliation import GenomicCVLReconcile
-from rdr_service.services.system_utils import JSONObject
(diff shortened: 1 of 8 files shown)


rdr_service/offline/genomic_pipeline.py
```diff
@@ -604,3 +604,28 @@ def load_awn_manifest_into_raw_table(
...
+                  '(after having appeared in a W1IL manifest and then revoking their GROR consent):\n\n'
+        message += '\n'.join([f'P{participant.participantId}' for participant in participant_list])
+
+        EmailService.send_email(
+            Email(
+                recipients=[notification_email_address],
+                subject='GHR3 participants recently re-submitting GROR consent',
+                plain_text_content=message
+            )
+        )
```
dtharpeuno: Would you move this into a method attached to the GenomicJobController? The reason is that we use the context manager in the controller to create different job run records, so we know when the last time a specific job run was for overall tracking and daily reporting.

rdr_service/offline/genomic_pipeline.py
```diff
@@ -604,3 +604,28 @@ def load_awn_manifest_into_raw_table(
...
+def notify_email_group_of_yes_no_yes_w1il_participants(since_datetime):
+    dao = GenomicQueriesDao()
+    participant_list = dao.get_w1il_yes_no_yes_participants(start_datetime=since_datetime)
+
+    logging.warning(
+        f'The following {len(participant_list)} participants '
+        f'switched back to a Yes consent for GROR after being in W1IL: ' +
+        ','.join([f'{participant.participantId}' for participant in participant_list])
+    )
+
```
dtharpeuno: Should we be checking for the `participant_list` list before further executing the email?

kenny-skaggs: Good idea, I'll add that.

rdr_service/offline/genomic_pipeline.py
kenny-skaggs: I thought the jobs were things that would move members between states. So should this be added as another job run id on genomic_set_member?

dtharpeuno: So in the pipeline we store records for when jobs run, this is outside of the `genomic_set_member` idea, this pertains to the pipeline overall. For ex in another method in `rdr_service/offline/genomic_pipeline`:

`def update_members_blocklists():
    with GenomicJobController(GenomicJob.UPDATE_MEMBERS_BLOCKLISTS) as controller:
        controller.update_members_blocklists()`
       
  We use the enum field in `GenomicJob` denoting when a certain job started/stopped/finished execution on the `genomic_job_run` table, and on the added context methods we write and update those records.

kenny-skaggs: Ah, ok. Cool. I'll get this PR updated to match.

j-kanuch (approved):

dtharpeuno (approved):

PR #3033, opened 2022-06-15
Validating EHR form based on latest ReceiveCare_PIIState answer

kenny-skaggs:
## Resolves *no ticket*
Responses to the ConsentPII module can be received multiple times (when participants update their address for example). So the current code for validating EHR forms may miss if the state the participant receives care hasn't been updated (so wouldn't be sent in a new payload).

This updates the validation code to use the latest answer for where a participant receives care, even if that answer isn't in the latest ConsentPII response received.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/dao/questionnaire_response_dao.py
+++ b/rdr_service/dao/questionnaire_response_dao.py
@@ -23,7 +23,7 @@ from rdr_service.participant_enums import QuestionnaireResponseStatus, PARTICIPA
     PARTICIPANT_COHORT_3_START_TIME
 from rdr_service.app_util import get_account_origin_id, is_self_request
 from rdr_service import storage
-from rdr_service import clock, config
+from rdr_service import clock, code_constants, config
 from rdr_service.code_constants import (
     CABOR_SIGNATURE_QUESTION_CODE,
     CONSENT_COHORT_GROUP_CODE,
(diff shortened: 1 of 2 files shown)


rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -1399,6 +1399,32 @@ def get_responses_to_surveys(
...
+                and_(
+                    question_code.codeId == QuestionnaireQuestion.codeId,
+                    question_code.value == code_constants.RECEIVE_CARE_STATE
+                )
+            )
+            .join(
+                answer_code,
+                answer_code.codeId == QuestionnaireResponseAnswer.valueCodeId
+            )
+            .order_by(QuestionnaireResponse.authored.desc())
```
pegbertsch, line 1422: This is an admittedly small risk, but:  we have gotten two payloads for a module (TheBasics) with an identical authored time but conflicting non-null answer codes to the same question, so using only authored for the `order_by` can have indeterminate results.

For PDR we often use `QuestionnaireResponse.created.desc() ` as a secondary `order_by ` value just for predictable ordering results.

pegbertsch (approved):

wangy70 (approved):

PR #3043, opened 2022-06-21
[DA-2750] Update Format on VA HealthPro Workqueue Drop

brianmkimmel:
Updates the file structure to match columns from example file.
Exports the integer values for enums instead of the text value.
Converts the data stored in JSON format for digitalHealthSharingStatus and patientStatus into separate columns.
Removes prefixes from values in state, sex, and gender identity columns.

## Tests
- [x] unit tests


rdr_service/offline/export_va_workqueue.py
```diff
@@ -286,135 +333,164 @@ def generate_workqueue_report():
...
+                           SAMPLE_STATUS.get(participant_filtered.get("sampleStatus1ED10", "UNSET")),
+                           _export_datetime(participant_filtered.get("sampleStatus1ED10Time")),
+                           SAMPLE_STATUS.get(participant_filtered.get("sampleStatus2ED10", "UNSET")),
+                           _export_datetime(participant_filtered.get("sampleStatus2ED10Time")),
+                           SAMPLE_STATUS.get(participant_filtered.get("sampleStatus1CFD9", "UNSET")),
+                           _export_datetime(participant_filtered.get("sampleStatus1CFD9Time")),
+                           SAMPLE_STATUS.get(participant_filtered.get("sampleStatus1PXR2", "UNSET")),
+                           _export_datetime(participant_filtered.get("sampleStatus1PXR2Time")),
+                           SAMPLE_STATUS.get(participant_filtered.get("sampleStatus1UR10", "UNSET")),
+                           _export_datetime(participant_filtered.get("sampleStatus1UR10Time")),
```
kenny-skaggs: It looks like the documentation from HPRO defines these as dates rather than datetimes. `sampleStatus1SALTime` below looks good, but should the other status time fields be updated to match?

j-kanuch (approved):

kenny-skaggs (approved):

rdr_service/offline/export_va_workqueue.py
brianmkimmel: Good catch, I've updated so the fields that should exported as dates will be

kenny-skaggs (approved):

PR #3051, opened 2022-06-28
[PDR-826] Support the new researcher DSV2 survey in PDR

robabram:
* Add new Research Workbench Researcher survey questions into PDR.
* Remove PII fields.
* Switch to using only one data generator for both the Resource and BQ data.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/model/bq_participant_summary.py
+++ b/rdr_service/model/bq_participant_summary.py
@@ -241,9 +241,9 @@ class BQPairingHistorySchema(BQSchema):
     """
     Participant pairing history
     """
-    last_modified = BQField('last_modified', BQFieldTypeEnum.DATETIME, BQFieldModeEnum.REQUIRED)
-    hpo = BQField('hpo', BQFieldTypeEnum.STRING, BQFieldModeEnum.REQUIRED)
-    hpo_id = BQField('hpo_id', BQFieldTypeEnum.INTEGER, BQFieldModeEnum.REQUIRED)
+    last_modified = BQField('last_modified', BQFieldTypeEnum.DATETIME, BQFieldModeEnum.NULLABLE)
+    hpo = BQField('hpo', BQFieldTypeEnum.STRING, BQFieldModeEnum.NULLABLE)
+    hpo_id = BQField('hpo_id', BQFieldTypeEnum.INTEGER, BQFieldModeEnum.NULLABLE)
     organization = BQField('organization', BQFieldTypeEnum.STRING, BQFieldModeEnum.NULLABLE)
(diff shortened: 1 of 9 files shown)


wangy70 (approved):

rdr_service/model/bq_participant_summary.py
```diff
@@ -241,9 +241,9 @@ class BQPairingHistorySchema(BQSchema):
     """
     Participant pairing history
     """
-    last_modified = BQField('last_modified', BQFieldTypeEnum.DATETIME, BQFieldModeEnum.REQUIRED)
-    hpo = BQField('hpo', BQFieldTypeEnum.STRING, BQFieldModeEnum.REQUIRED)
-    hpo_id = BQField('hpo_id', BQFieldTypeEnum.INTEGER, BQFieldModeEnum.REQUIRED)
+    last_modified = BQField('last_modified', BQFieldTypeEnum.DATETIME, BQFieldModeEnum.NULLABLE)
```
pegbertsch, line 244: Are we failing to populate any records currently because these fields were set to REQUIRED but were null in the RDR data?   It seems like this will need a full truncate of the `pdr_participant` table in BigQuery before we can apply this change and the migration may need to be applied separately from the RWB survey schema/migration changes based on the timing of when we can rebuild the participant data.

Should we do these changes as a separate PR?

rdr_service/resource/generators/workbench.py
```diff
@@ -250,57 +250,89 @@ def make_resource(self, src_pk_id, backup=False):
             data['degrees'] = [{'degree': str(WorkbenchResearcherDegree(v)),
                                 'degree_id': int(WorkbenchResearcherDegree(v))} for v in degrees]
 
+            data['disability'] = str(WorkbenchResearcherDisability(row.disability))
```
pegbertsch, line 253: Is this a state flag like ubr_disability is for participants?  Meaning they indicated one or more of the disabilities from the researcher V2 survey?  It comes to RDR via the old API?

pegbertsch (approved):
Approving, but believe we need to be careful about applying the BQ migration for the existing `pairing_history` subschema, since it will force truncation (drop/recreate) of the pdr_participant data.

rdr_service/resource/generators/workbench.py
robabram: I assume so, this field is from the original version of the workbench Research table.

rdr_service/model/bq_participant_summary.py
robabram: I believe we can do a Alter Table statement to change these fields in BigQuery to Nullable.

PR #3054, opened 2022-06-30
Limiting Safety package to 1.x

kenny-skaggs:
## Resolves
The Safety check is failing after 2.0 got released, it seems to be requiring that a config yml file exists (https://app.circleci.com/pipelines/github/all-of-us/raw-data-repository/6012/workflows/a7d4db49-2849-4a3e-b794-ce30164d0b83/jobs/14920).

Running Safety locally doesn't seem to require the file in the same way (but I also wasn't able to get it to actually check the packages, so I may have been doing something wrong). There's a default file in the Safety code base, but including that file seems to bring up another error when trying to parse it (https://app.circleci.com/pipelines/github/all-of-us/raw-data-repository/6015/workflows/5f1720c3-8ce3-4c02-a1e9-ea16b7653aa9/jobs/14923). That may potentially be a bug in Safety, or at least a bug that occurs when running it in our CI environment.

Freezing Safety at 1.x for now, until we can look further into the cause and/or future versions of Safety resolve the issue.

## Tests
- [ ] unit tests


dtharpeuno (approved):
Never seen the string denotation for less than, TIL.
You can also use `safety==1.*`.

Thanks for fixing this

pegbertsch (approved):

PR #3064, opened 2022-07-07
[DA-2793] Genomics Unblocking Tool: Update block reason when unblocking

brianmkimmel:

Updates block_research_reason and block_results_reason when the record is unblocked.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/tools/tool_libs/genomic_utils.py
+++ b/rdr_service/tools/tool_libs/genomic_utils.py
@@ -2211,11 +2211,15 @@ class UnblockSamples(ToolBase):
         for set_member in set_members:
             if self.args.results_only:
                 set_member.blockResults = 0
+                set_member.blockResultsReason = f"Formerly blocked due to '{set_member.blockResultsReason}'"
             elif self.args.research_only:
                 set_member.blockResearch = 0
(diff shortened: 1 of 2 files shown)


rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2211,11 +2211,15 @@ def run(self):
...
                 set_member.blockResults = 0
+                set_member.blockResultsReason = f"Formerly blocked due to '{set_member.blockResultsReason}'"
             elif self.args.research_only:
                 set_member.blockResearch = 0
+                set_member.blockResearchReason = f"Formerly blocked due to '{set_member.blockResearchReason}'"
             else:
                 set_member.blockResults = 0
                 set_member.blockResearch = 0
+                set_member.blockResultsReason = f"Formerly blocked due to '{set_member.blockResultsReason}'"
+                set_member.blockResearchReason = f"Formerly blocked due to '{set_member.blockResearchReason}'"
```
dtharpeuno: Should we just be checking the type of "blocked" it is @ execution and then update the reason? Or if you are passing the type in the args can we have a check to see if it was actually blocked so we won't have discrepancies in the data such as "Formerly blocked due to None".

j-kanuch (approved):

dtharpeuno (approved):

PR #3081, opened 2022-07-21
[DA-2614] ptsc_health_data_transfer_result_notification

wangy70:
## Resolves *[DA-2614](https://precisionmedicineinitiative.atlassian.net/browse/DA-2614)*
Send slack alert if PTSC health data transfer failed.

1. update data transfer job configuration to send event to topic
2. subscribe topic with cloud function
3. use cloud task to get the event payload and send slack alert

## Tests
- manual test on sandbox

### DIFF
--- a/rdr_service/api/cloud_tasks_api.py
+++ b/rdr_service/api/cloud_tasks_api.py
@@ -22,6 +22,9 @@ from rdr_service.resource.generators.workbench import res_workspace_batch_update
 from rdr_service.resource.tasks import batch_rebuild_participants_task, batch_rebuild_retention_metrics_task, \
     batch_rebuild_consent_metrics_task, batch_rebuild_user_event_metrics_task, check_consent_errors_task
 from rdr_service.services.participant_data_validation import ParticipantDataValidation
+from rdr_service.services.slack_utils import SlackMessageHandler
+from rdr_service import config
+from rdr_service.config import RDR_SLACK_WEBHOOKS
 
 
 def log_task_headers():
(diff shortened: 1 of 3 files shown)


j-kanuch (approved):

rdr_service/api/cloud_tasks_api.py
```diff
@@ -182,6 +185,30 @@ def post(self):
...
+    Cloud Task endpoint: Ptsc Health Data Transfer Result
+    """
+    @task_auth_required
+    def post(self):
+        log_task_headers()
+        data = request.get_json(force=True)
+        logging.info(f'Ptsc Health Data Transfer Result: {data.get("attributes").get("eventType")}')
+        # possible event types: TRANSFER_OPERATION_SUCCESS, TRANSFER_OPERATION_FAILED, TRANSFER_OPERATION_ABORTED
+        event_type = data.get("attributes").get("eventType")
+        if event_type in ['TRANSFER_OPERATION_FAILED', 'TRANSFER_OPERATION_ABORTED']:
```
kenny-skaggs, line 199: This receives the status that is displayed in the UI for the data transfer jobs? Even when it successfully runs it still gets marked as a failure because of a few files that it doesn't have access to (temp files that don't appear to need transferred anyway). How should we handle those so we don't see failure messages for them?

wangy70: I saw those two files are failed due to no permission. I think we can either check PTSC for permission or request them to remove those two files if possible.

kenny-skaggs (approved):

PR #3084, opened 2022-07-25
Update to consent validation trigger

kenny-skaggs:
## Resolves *no ticket*
We're seeing that some responses are being sent for consents we already have validated consents for. The consent process is currently only looking for one valid response to a given consent type. So this updates the check for what needs to be validated. With these changes we'll only validate consent response types that haven't yet been validated.

Currently we're seeing a growing number of ConsentResponses that are being validated each day, but no ConsentFiles are being stored for them since we already have ConsentFiles of the same type for these participants.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/consent_dao.py
+++ b/rdr_service/dao/consent_dao.py
@@ -47,13 +47,22 @@ class ConsentDao(BaseDao):
         """
         consent_batch_limit = 500
 
-        # A ConsentResponse hasn't been validated yet if there aren't any ConsentFiles that link to the response
-        consent_responses = session.query(ConsentResponse).outerjoin(
-            ConsentFile
-        ).filter(
-            ConsentFile.id.is_(None)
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):

pegbertsch:
@kenny-skaggs Is this something we need to raise with PTSC (/CE) to try and avoid sending the extraneous payloads/responses in the first place, or is it an expected side effect of planned participant data replays?

pegbertsch (approved):

kenny-skaggs:
> @kenny-skaggs Is this something we need to raise with PTSC (/CE) to try and avoid sending the extraneous payloads/responses in the first place, or is it an expected side effect of planned participant data replays?

It's a side-effect of our own processing more than anything else. I don't think we need to try to push for changes for it.

PR #3100, opened 2022-08-02
Match manually validated wear consents to their questionnaire responses

kenny-skaggs:
## Partially Resolves *[DA-2889](https://precisionmedicineinitiative.atlassian.net/browse/DA-2889)*
Manually validating consents doesn't currently set the matching consent_response record for the validation result. This PR creates a script for matching up the consent validation records to a questionnaire response. At the moment, all the WEAR validation results that are missing a link to a questionnaire response only have one response that they could link to.

If, in the future, this script needs to be applied to other consent types or needs to handle a case where multiple responses are found, then the functionality will be expanded at that time.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/tools/tool_libs/consent_response_matching.py
+++ b/rdr_service/tools/tool_libs/consent_response_matching.py
@@ -1,0 +1,40 @@
+import logging
+
+from rdr_service.dao.questionnaire_response_dao import QuestionnaireResponseDao
+from rdr_service.model.consent_file import ConsentFile, ConsentType
+from rdr_service.tools.tool_libs.tool_base import cli_run, ToolBase
+
+tool_cmd = 'consent-match'
+tool_desc = 'find the response for consents that do not have a response set'
+
+
+class ConsentMatchScript(ToolBase):
+    def run(self):
+        super(ConsentMatchScript, self).run()
+
+        with self.get_session() as session:
+            unmatched_wear_consents = session.query(ConsentFile).filter(
+                ConsentFile.type == ConsentType.WEAR,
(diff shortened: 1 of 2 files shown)


rdr_service/tools/tool_libs/consent_response_matching.py
```diff
@@ -0,0 +1,40 @@
...
+                participant_ids=[file.participant_id for file in unmatched_wear_consents]
+            )
+
+            for file in unmatched_wear_consents:
+                if file.participant_id not in wear_responses:
+                    logging.error(f'response for {file.participant_id} not found')
+                else:
+                    response_list = wear_responses[file.participant_id].responses
+                    if len(response_list) != 1:
+                        logging.warning(f'{len(response_list)} responses found for {file.participant_id}')
```
pegbertsch: FYI:  I think there will be some cases where we see this warning and cannot definitively match. There appear to be 40 participants who had multiple `questionnaire_response` payloads for WEAR consent and do not have a `consent_response_id` in their `consent_file` validation record for their WEAR consent.

dtharpeuno (approved):

rdr_service/tools/tool_libs/consent_response_matching.py
kenny-skaggs: Good catch. I'd made a mistake with the logger and didn't get them printed out. I'll update the code.

kenny-skaggs:
Updated the algorithm for matching ConsentFiles to ConsentResponses (and actually have it setting the link correctly).

pegbertsch (approved):

PR #3101, opened 2022-08-02
[DA-2886] Update to Participant Summary API/Ops Data API for Physical Measurements Fields

brianmkimmel:
Creates physicalMeasurementsStatus, physicalMeasurementsTime, physicalMeasurementsFinalizedTime, physicalMeasurementsCreatedSite, physicalMeasurementsFinalizedSite, physicalMeasurementsCollectType as API fields from the clinic and self reported physical measurements columns in ParticipantSummary

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/model/participant_summary.py
+++ b/rdr_service/model/participant_summary.py
@@ -1386,6 +1386,14 @@ class ParticipantSummary(Base):
     lastModified = Column("last_modified", UTCDateTime6)
     """UTC timestamp of the last time the participant summary was modified"""
 
+    #Placeholder fields for deprecated columns
+    physicalMeasurementsStatus = None
+    physicalMeasurementsTime = None
+    physicalMeasurementsFinalizedTime = None
+    physicalMeasurementsCreatedSite = None
+    physicalMeasurementsFinalizedSite = None
+    physicalMeasurementsCollectType = None
+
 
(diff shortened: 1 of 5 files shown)


rdr_service/model/participant_summary.py
```diff
@@ -1386,6 +1386,14 @@ def enrollmentSiteId(cls):
...
     """UTC timestamp of the last time the participant summary was modified"""
 
+    #Placeholder fields for deprecated columns
+    physicalMeasurementsStatus = None
+    physicalMeasurementsTime = None
+    physicalMeasurementsFinalizedTime = None
+    physicalMeasurementsCreatedSite = None
+    physicalMeasurementsFinalizedSite = None
+    physicalMeasurementsCollectType = None
+
```
dtharpeuno: Not sure if we need these here, in the dao since you are appending the key/values onto the `result` dict anyway.

kenny-skaggs: I agree, I think they could be taken out. I'm going ahead to approve the PR, but taking them out may be a good idea.

kenny-skaggs (approved):

rdr_service/model/participant_summary.py
brianmkimmel: I took them out locally and didn't see any issues while I was testing so I'm updating the PR.

dtharpeuno (approved):

PR #3102, opened 2022-08-03
MIGRATIONS [DA-2864] - Results Withdrawal Workflow

dtharpeuno:
## Resolves *[ticket DA-2864]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2864

We need a way to know when a participant who has previously entered a results pipeline (GEM or GHR3) has withdrawn, so a program decision on how to proceed can be made.

## Tests
- [x] unit tests


rdr_service/dao/genomics_dao.py
```diff
@@ -3813,6 +3813,54 @@ def get_w1il_yes_no_yes_participants(self, start_datetime):
...
+                ).label('cvl_results')
+            ).join(
+                ParticipantSummary,
+                ParticipantSummary.participantId == GenomicSetMember.participantId
+            ).outerjoin(
+                array_results,
+                and_(
+                    array_results.id == GenomicSetMember.id,
+                    array_results.gemA1ManifestJobRunId.isnot(None)
+                )
```
kenny-skaggs: For my own understanding: would joining the GenomicSetMember table to itself based on the id column essentially join each row to itself?

dtharpeuno: Is there another way I should be doing it? Should I use another attribute such as `participant_id` instead?

dtharpeuno: Edited query, good catch.

https://github.com/all-of-us/raw-data-repository/pull/3102/commits/c8282c04ea8cd5a96191f16cf988f415fa9baf70

kenny-skaggs (approved):

j-kanuch (approved):

PR #3103, opened 2022-08-04
[PDR-956] Update and add new workbench tables in PDR

robabram:
* Added workbench Audit table to PDR
* Switched to using Researcher and Institution Affiliation history tables
* Update bad BigQuery view definitions

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/api/redcap_workbench_audit_api.py
+++ b/rdr_service/api/redcap_workbench_audit_api.py
@@ -3,6 +3,8 @@ from rdr_service.api.base_api import BaseApi
 from werkzeug.exceptions import BadRequest
 from rdr_service.api_util import REDCAP_AND_RDR
 from rdr_service.app_util import auth_required
+from rdr_service.config import GAE_PROJECT
+from rdr_service.dao.bq_workbench_dao import rebuild_bq_audit
 from rdr_service.dao.workbench_dao import WorkbenchResearcherDao, WorkbenchWorkspaceAuditDao
 
 
@@ -57,6 +59,20 @@ class RedcapWorkbenchAuditApi(BaseRedcapApi):
         super(RedcapWorkbenchAuditApi, self).get()
         return self.dao.workspace_dao.get_redcap_audit_workspaces(**self.get_filters)
 
+    def _do_insert(self, m):
+        audit_records = super()._do_insert(m)
+        # Generate tasks to build PDR records.
+        if GAE_PROJECT == 'localhost':
+            rebuild_bq_audit(audit_records)
+        else:
+            payload = {'table': 'audit', 'ids': []}
+            for obj in audit_records:
+                payload['ids'].append(obj.id)
+
+            if len(payload['ids']) > 0:
+                self._task.execute('rebuild_research_workbench_table_records_task', payload=payload,
+                             in_seconds=15, queue='resource-rebuild')
+        return audit_records
 
 class RedcapResearcherAuditApi(BaseRedcapApi):
     def __init__(self):
--- a/rdr_service/api/redcap_workbench_audit_api.py
+++ b/rdr_service/api/redcap_workbench_audit_api.py
@@ -6,6 +6,7 @@ from rdr_service.app_util import auth_required
 from rdr_service.config import GAE_PROJECT
 from rdr_service.dao.bq_workbench_dao import rebuild_bq_audit
 from rdr_service.dao.workbench_dao import WorkbenchResearcherDao, WorkbenchWorkspaceAuditDao
+from rdr_service.services.system_utils import list_chunks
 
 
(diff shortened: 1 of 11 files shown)


rdr_service/api/redcap_workbench_audit_api.py
```diff
@@ -57,6 +59,20 @@ def get(self):
...
         return self.dao.workspace_dao.get_redcap_audit_workspaces(**self.get_filters)
 
+    def _do_insert(self, m):
+        audit_records = super()._do_insert(m)
+        # Generate tasks to build PDR records.
+        if GAE_PROJECT == 'localhost':
+            rebuild_bq_audit(audit_records)
+        else:
+            payload = {'table': 'audit', 'ids': []}
+            for obj in audit_records:
```
pegbertsch, line 70: Do we need to limit the number of ids being rebuilt per task?

rdr_service/dao/bq_workbench_dao.py
```diff
@@ -72,8 +72,9 @@ def make_bqrecord(self, src_pk_id, convert_to_enum=False):
             data['income_level'] = str(WorkbenchWorkspaceIncomeLevel(row.income_level))
             data['income_level_id'] = int(WorkbenchWorkspaceIncomeLevel(row.income_level))
 
-            data['access_tier'] = str(WorkbenchWorkspaceAccessTier(row.access_tier))
-            data['access_tier_id'] = int(WorkbenchWorkspaceAccessTier(row.access_tier))
+            if row.access_tier:
```
pegbertsch, line 75: Is there any need to check for the presence of the other `row.*` values above, before trying to cast  as enum values?

rdr_service/dao/workbench_dao.py
```diff
@@ -1169,6 +1172,30 @@ def _insert_history(session, researchers):
...
             history.workbenchInstitutionalAffiliations = affiliations_history
             session.add(history)
+            hist_researchers.append(history)
+        session.flush()
+        session.commit()
+
+        # Generate tasks to build PDR records.
+        if GAE_PROJECT == 'localhost':
+            rebuild_bq_wb_researchers(hist_researchers)
+        else:
```
pegbertsch, line 1182: Do we need to limit how many records are rebuilt per task?

rdr_service/model/bq_workbench_workspace.py
```diff
@@ -169,3 +172,31 @@ class BQRWBWorkspaceUsersView(BQView):
     __viewdescr__ = 'Research Workbench Workspace Users View'
     __pk_id__ = 'id'
     __table__ = BQRWBWorkspaceUsers
+
+
+class BQRWBAuditSchema(BQSchema):
```
pegbertsch, line 177: So we expect to bring over email and notes text unaltered, and move this to a more locked down PostgreSQL schema?

rdr_service/dao/workbench_dao.py
```diff
@@ -1169,6 +1172,30 @@ def _insert_history(session, researchers):
                 affiliations_history.append(affiliation_obj)
             history.workbenchInstitutionalAffiliations = affiliations_history
             session.add(history)
+            hist_researchers.append(history)
+        session.flush()
+        session.commit()
```
pegbertsch, line 1177: Is the flush() already an implicit part of the commit(), if you don't need to access anything in between the two calls?

pegbertsch (commented):
Main question is around lack of limits on ids built per task queued.  This may be deliberate if the expectation is these datasets will remain pretty small.

rdr_service/dao/workbench_dao.py
robabram: I have removed the Flush() call in the new commit.

rdr_service/dao/bq_workbench_dao.py
robabram: No, the other enum fields have a default value of zero.

rdr_service/dao/workbench_dao.py
robabram: I have added a 'list_hhunks()' for loop to limit batch size.

rdr_service/model/bq_workbench_workspace.py
robabram: The code to change the values to boolean for these fields is in the BQ generator.

rdr_service/api/redcap_workbench_audit_api.py
robabram: I have added a 'list_chunks()' for loop to limit batch size.

pegbertsch (approved):

dtharpeuno (approved):

PR #3110, opened 2022-08-15
[DA-2856] Genomic Scheduling API

dtharpeuno:
## Resolves *[ticket DA-2856]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2856

Adding genomic scheduling API for partner usage for sending notifications when appointments with genomic counselors have been scheduled

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/genomics_dao.py
+++ b/rdr_service/dao/genomics_dao.py
@@ -2216,8 +2216,6 @@ class GenomicOutreachDaoV2(BaseDao):
 class GenomicSchedulingDao(BaseDao):
     def __init__(self):
         super().__init__(GenomicSetMember, order_by_ending=['id'])
-        self.allowed_modules = ['hdr', 'pgx']
-        self.module = self.allowed_modules
 
     def get_id(self, obj):
         pass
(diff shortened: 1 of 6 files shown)


rdr_service/dao/genomics_dao.py
```diff
@@ -2213,6 +2213,106 @@ def get_report_state_query_config(self):
...
+                and_(
+                    note_alias.appointment_id == GenomicAppointmentEvent.appointment_id,
+                    note_alias.event_type.like('%note_available')
+                )
+            ).filter(
+                and_(
+                    GenomicAppointmentEvent.appointment_id == max_appointment_id_subquery.c.max_appointment_id,
+                    GenomicAppointmentEvent.event_authored_time ==
+                    max_event_authored_time_subquery.c.max_event_authored_time
+                )
```
j-kanuch, line 2296: Do these conditions work if more than one row is returned in the sub-query?  I'm more familiar with using raw SQL and doing an inner join to the sub-query, but it may be handled differently via the ORM.

dtharpeuno: Good question, yes they do I added unit tests to encompass all the usual edge cases to test the queries correctly. Also for each query change, I extracted and ran the corresponding RAW SQL output to make sure the query was being translated correctly from the ORM.

j-kanuch: I see you're testing multiple participants in `test_pass_start_date_params`.

j-kanuch (approved):

kenny-skaggs (approved):

PR #3121, opened 2022-08-22
[DA-2564] Logic for 3.0 and 3.1 enrollment status calculations (and refactor for legacy version)

kenny-skaggs:
## Resolves *[DA-2564](https://precisionmedicineinitiative.atlassian.net/browse/DA-2564)*
Implements algorithms for calculating the enrollment status defined by the 3.0 and 3.1 data glossaries. Also similarly implements the current enrollment status calculation performed in the ParticipantSummaryDao.

This is not yet integrated (submitting this PR now to reduce the amount of code that needs to be reviewed at one time). But the next PR will remove the enrollment status calculation from the ParticipantSummaryDao, replacing it with a call to this class. The overall goal is to have this code separated out from any particular function (such as Biobank pipeline, or responding to an API request) so it can be easier to reuse and maintain.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/logic/enrollment_info.py
+++ b/rdr_service/logic/enrollment_info.py
@@ -43,6 +43,8 @@ class EnrollmentDependencies:
     lifestyle_authored_time: datetime
 
     ehr_consent_date_range_list: List[DateRange]
+    """DateRanges of when the participant expressed interest in sharing EHR data. Must be in chronological order"""
+
     earliest_biobank_received_dna_time: datetime
     earliest_ehr_file_received_time: datetime
     earliest_physical_measurements_time: datetime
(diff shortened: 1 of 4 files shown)


rdr_service/logic/enrollment_info.py
```diff
@@ -0,0 +1,279 @@
...
+        return not any(
+            required_date is None
+            for required_date in cls._get_dates_needed_for_core(participant_info)
+        )
+
+    @classmethod
+    def _get_dates_needed_for_core(cls, participant_info: EnrollmentDependencies):
+        return [
+            participant_info.earliest_physical_measurements_time,
+            *cls._get_dates_needed_for_core_minus_pm(participant_info)
```
dtharpeuno, line 264: Is this needed here?

dtharpeuno (approved):
Nice!

brianmkimmel (approved):

PR #3123, opened 2022-08-24
Fix VA Workqueue Export Timestamp Parsing Error

brianmkimmel:
The VA workqueue export is unable to run because of a timestamp parsing error in digitalHealthSharingStatus. Updates datetime parsing to handle timestamps with 'Z' at the end.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/offline/export_va_workqueue.py
+++ b/rdr_service/offline/export_va_workqueue.py
@@ -182,7 +182,7 @@ def _get_json_fields(participant):
 
 def _export_datetime(api_datetime):
     if api_datetime:
-        return datetime.datetime.fromisoformat(api_datetime).strftime("%m/%d/%Y %H:%M:%S")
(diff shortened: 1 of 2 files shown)


dtharpeuno (approved):

rdr_service/offline/export_va_workqueue.py
```diff
@@ -182,7 +182,7 @@ def _get_json_fields(participant):
 
 def _export_datetime(api_datetime):
     if api_datetime:
-        return datetime.datetime.fromisoformat(api_datetime).strftime("%m/%d/%Y %H:%M:%S")
+        return datetime.datetime.fromisoformat(api_datetime.replace("Z", "")).strftime("%m/%d/%Y %H:%M:%S")
```
j-kanuch, line 185: This works, an alternative method is using the parse() or isoparse() methods from the `dateutil` library. Something like 
`from dateutil.parser import parse`
`return parse(api_datetime).strftime("%m/%d/%Y %H:%M:%S")`

j-kanuch (approved):

PR #3154, opened 2022-09-14
[MIGRATION][DA-2934] Switch File Ingestion Check to AW3 Level

brianmkimmel:

Updates genomic pipeline to populate data file paths in the metrics record when AW2 ingestion occurs. 
Creates tool to find and update samples with the workflow state GC_DATA_FILES_MISSING with available data files.
Creates weekly email report of any AW3 ready samples which are missing data files.
Disables existing data file reconciliation jobs.

## Tests
- [x] unit tests


rdr_service/cron_default.yaml
```diff
@@ -89,16 +89,6 @@ cron:
...
-- description: Genomic Reconciliation Array Data Workflow
-  url: /offline/GenomicArrayReconciliationWorkflow
-  timezone: America/New_York
-  schedule: every day 05:30
-  target: offline
-- description: Genomic Reconciliation WGS Data Workflow
-  url: /offline/GenomicWGSReconciliationWorkflow
-  timezone: America/New_York
-  schedule: every day 05:45
-  target: offline
```
dtharpeuno, line 101: If we are removing these jobs, we need to clean the repo of the logic that coincides with this also.

rdr_service/cron_default.yaml
```diff
@@ -169,16 +159,6 @@ cron:
...
-- description: Genomic Missing Files Resolve
-  url: /offline/GenomicMissingFilesResolve
-  timezone: America/New_York
-  schedule: every 6 hours
-  target: offline
-- description: Reconcile GC Data File to Table
-  url: /offline/ReconcileGCDataFileToTable
-  timezone: America/New_York
-  schedule: every tuesday 02:00
-  target: offline
```
dtharpeuno, line 181: Same here with cleaning corresponding logic.

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -1905,6 +1905,28 @@ def check_results_withdrawals(self):
...
+            EmailService.send_email(
+                Email(
+                    recipients=[notification_email_address],
+                    subject='AW3 ready samples with missing data files',
+                    plain_text_content=message
+                )
+            )
+        self.job_result = GenomicSubProcessResult.SUCCESS
+
+
```
dtharpeuno, line 1878: Sorry, why are we doing this? I thought the point of this feature set was to backfill the data files on the metrics records, not sure why we are concerned with missing files anymore.

brianmkimmel: The idea behind this report is to catch any samples that don't have all of the data files so we don't build up a backlog if a file doesn't get uploaded or fails to be added to the database.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1694,6 +1697,63 @@ def calculate_contamination_category(self, sample_id, raw_contamination, member:
...
+                    file_path = f'gs://{gc_bucket}/{prefix_map[site_id][file_def["file_type"]]}/{site_id.upper()}_' + \
+                                f'{row["biobankid"]}_{row["sampleid"]}_{row["limsid"]}_1.{file_def["file_type"]}'
+                    row[file_def['file_path_attribute']] = file_path
+                    row[file_def['file_received_attribute']] = 1
+            # Need to set array file received flags to 0 to prevent an error when ingesting updated AW2
+            for file_def in array_file_types_attributes:
+                if file_def['required']:
+                    row[file_def['file_received_attribute']] = 0
+
+        return row
```
dtharpeuno: Think it might be a good idea to refactor this to make it a little bit cleaner in the execution, seems like some of this code, mainly setting the attributes is being repeated multiple times.

rdr_service/genomic/genomic_job_controller.py
dtharpeuno: The idea from what I understand is to alleviate bottlenecks in this process of the pipeline. One of those bottlenecks is checking for the data files in the buckets. The idea should just be to build the paths to the files and store them on the metrics records and not keep track of anything missing cause we are trying to take that responsibility off of the pipeline.

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2344,6 +2345,71 @@ def _check_gc_data_files(self):
...
+        if not self.args.dryrun:
+            for member in members_to_update:
+                updated = self._update_member(member)
+                if updated:
+                    updated_count += 1
+        else:
+            _logger.info(f"Will update {len(members_to_update)} samples")
+        _logger.info(f"Found {len(members_to_update)} members to update. Updated {updated_count}.")
+
+    def _update_member(self, member: GenomicSetMember) -> bool:
```
dtharpeuno: This function updates the metric records, might be a good idea to rename is `_update_metrics`

rdr_service/tools/tool_libs/genomic_utils.py
```diff
@@ -2344,6 +2345,71 @@ def _check_gc_data_files(self):
...
+        files = None
+        file_types_attributes = None
+
+        metrics = self.metrics_dao.get_metrics_by_member_id(member.id)
+        if member.genomeType == 'aou_array':
+            file_types_attributes = array_file_types_attributes
+            files = self.data_file_dao.get_with_chipwellbarcode(metrics.chipwellbarcode)
+        elif member.genomeType == 'aou_wgs':
+            file_types_attributes = wgs_file_types_attributes
+            files = self.data_file_dao.get_with_sample_id(member.sampleId)
```
dtharpeuno, line 2397: @j-kanuch we are not worried about storing data files in our tables anymore right? So shouldn't we just be checking for the metrics records with missing file paths and "building"/inserting those?

tests/genomics_tests/test_genomic_pipeline.py
```diff
@@ -1107,1865 +1123,2476 @@ def test_aw2_wgs_gc_metrics_ingestion(self):
...
-                                         folder=config.getSetting(config.GENOMIC_AW2_SUBFOLDERS[1]))
-
-        self._update_test_sample_ids()
-
-        self._create_stored_samples([
-            (1, 1001),
-            (2, 1002),
-            (3, 1003)
-        ])
+    # DA-2934 Reconciliation process deprecated
```
dtharpeuno: Instead of commenting on the tests, we need to delete them as well as the corresponding logic.

tests/tool_tests/test_genomic_utils.py
```diff
@@ -437,3 +437,107 @@ def test_unblock_samples(self):
...
+            hfVcfTbiReceived=0,
+        )
+        array_files = [
+            f'test_data_folder/10001_R01C01.vcf.gz',
+            f'test_data_folder/10001_R01C01.vcf.gz.tbi',
+            f'test_data_folder/10001_R01C01.vcf.gz.md5sum',
+            f'test_data_folder/10001_R01C01_Red.idat',
+            f'test_data_folder/10001_R01C01_Grn.idat',
+            f'test_data_folder/10001_R01C01_Red.idat.md5sum',
+            f'test_data_folder/10001_R01C01_Grn.idat.md5sum',
```
dtharpeuno: No need to f-strings here if you are not inserting values

rdr_service/genomic/genomic_job_controller.py
j-kanuch: For reference, during AW3 generation, we still need to check that we don't send Broad records with missing files. This was Megan's requirement, so we're still using gc_data_file to check that there are at least the number of required files there. The weekly report is just to say "Here's records that should be in an AW3 but don't have enough data files." It won't be updating data or setting any flags.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1694,6 +1697,63 @@ def calculate_contamination_category(self, sample_id, raw_contamination, member:
 
         return contamination_category
 
+    def _set_metrics_data_file_paths(self, row: dict):
+        cvl_site_bucket_map = config.getSettingJson(config.GENOMIC_CVL_SITE_BUCKET_MAP)
```
j-kanuch: At this point in the pipeline, you'll want to use the GCs' buckets, not the CVLs'. CVLs are downstream of the AW2. The CVL site map replaces BI with CO, i.e. Broad is not a CVL and Color is not a GC.

rdr_service/genomic/genomic_job_components.py
j-kanuch: I concur. Since an AW2 will either be for Array or WGS, you can reduce the complexity by splitting these up into separate functions. 
You can also go ahead and make the gVCF "required" in the file_def dict, since we are now getting rid of the reconciliation process and only need these paths for the AW3. Doing so will reduce the need for a separate condition for the gVCF, since it is essentially required anyway.
All `file_received_attributes` for the AW2's genome type should be set when creating the gc_validation_metrics record. We wouldn't ever ingest an Array AW2 for a WGS sample, and vice-versa. They will always have different sample IDs, so you should be able to remove setting these attributes to 0 from `_set_metrics_data_file_paths()`. I'm not sure what the error was, but let me know if there are any questions about the above.

rdr_service/tools/tool_libs/genomic_utils.py
j-kanuch: This is for clearing up the bottle neck. We are still checking gc_data_file.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1694,6 +1697,63 @@ def calculate_contamination_category(self, sample_id, raw_contamination, member:
...
+                    row[file_def['file_path_attribute']] = file_path
+                    row[file_def['file_received_attribute']] = 1
+            # Need to set array file received flags to 0 to prevent an error when ingesting updated AW2
+            for file_def in array_file_types_attributes:
+                if file_def['required']:
+                    row[file_def['file_received_attribute']] = 0
+
+        return row
+
+    def _update_member_state_after_aw2(self, member: GenomicSetMember):
```
j-kanuch, line 1757: I appreciate the re-use of the state machine here, but in practice this part of our AW2 job takes forever. I know it wasn't part of the original request, but it would be super helpful if you could refactor this part of the job to update the workflow state on a set of member IDs in bulk. You can create a new member_dao method if needed, pass it either the GEM_READY or CVL_READY workflow state and the list of IDs and just bulk update them. Let me know if you have any questions about this.

rdr_service/cron_default.yaml
brianmkimmel: I've removed the code for these jobs.

rdr_service/cron_default.yaml
brianmkimmel: Removed the code for these as well

tests/genomics_tests/test_genomic_pipeline.py
brianmkimmel: The commented out tests have been deleted.

rdr_service/dao/genomics_dao.py
```diff
@@ -1217,6 +1218,83 @@ def get_members_from_biobank_ids(self, biobank_ids, genome_type=None):
...
+                GenomicGcDataFile.identifier_value == GenomicGCValidationMetrics.chipwellbarcode
+            ).filter(
+                GenomicSetMember.genomicWorkflowState == GenomicWorkflowState.GC_DATA_FILES_MISSING,
+                GenomicSetMember.genomeType == 'aou_array',
+                GenomicSetMember.ignoreFlag != 1,
+                GenomicGCValidationMetrics.ignoreFlag != 1,
+                GenomicGcDataFile.ignore_flag != 1,
+                GenomicGcDataFile.file_type.in_(
+                    [file_type['file_type'] for file_type in array_file_types_attributes
+                     if file_type['required']])
```
j-kanuch: It may help readability to set a variable higher up for `required_array_file_types`. You can then use it here and in calculating the len(). I have a hard time mentally computing list comprehensions :)

rdr_service/dao/genomics_dao.py
```diff
@@ -1217,6 +1218,83 @@ def get_members_from_biobank_ids(self, biobank_ids, genome_type=None):
...
+                GenomicGcDataFile.identifier_value == GenomicSetMember.sampleId
+            ).filter(
+                GenomicSetMember.genomicWorkflowState == GenomicWorkflowState.GC_DATA_FILES_MISSING,
+                GenomicSetMember.genomeType == 'aou_wgs',
+                GenomicSetMember.ignoreFlag != 1,
+                GenomicGCValidationMetrics.ignoreFlag != 1,
+                GenomicGcDataFile.ignore_flag != 1,
+                GenomicGcDataFile.file_type.in_(
+                    [file_type['file_type'] for file_type in wgs_file_types_attributes
+                     if file_type['required']])
```
j-kanuch: same here regarding a variable for the list comp.

rdr_service/dao/genomics_dao.py
```diff
@@ -3537,8 +3645,26 @@ def get_aw3_array_records(self, **kwargs):
...
+                GenomicGCValidationMetrics.genomicSetMemberId == GenomicSetMember.id
+            ).outerjoin(
+                GenomicGcDataFile,
+                GenomicGcDataFile.identifier_value == GenomicGCValidationMetrics.chipwellbarcode
+            ).filter(
+                GenomicSetMember.genomeType == genome_type,
+                GenomicGcDataFile.ignore_flag != 1,
+                GenomicGcDataFile.file_type.in_(
+                    [file_type['file_type'] for file_type in array_file_types_attributes
+                     if file_type['required']])
```
j-kanuch: maybe we make it an instance variable since this comprehension is used multiple times.

rdr_service/cron_default.yaml
```diff
@@ -249,5 +229,10 @@ cron:
   timezone: America/New_York
   schedule: every day 06:00
   target: offline
+- description: Report for AW3 Ready Samples Missing Data Files (Weekly)
```
dtharpeuno: Can we prefix this with "Genomic"

rdr_service/offline/main.py
```diff
@@ -795,12 +767,18 @@ def export_va_workqueue_report():
...
 
+
 @app_util.auth_required_cron
 def delete_old_va_workqueue_reports():
     export_va_workqueue.delete_old_reports()
     return '{"success": "true"}'
 
 
+@app_util.auth_required_cron
+def aw3_ready_missing_files_report():
```
dtharpeuno: Need to add the decorator here for enabling/disabling this job.

rdr_service/dao/genomics_dao.py
```diff
@@ -3602,15 +3731,44 @@ def get_aw3_array_records(self, **kwargs):
                 ParticipantSummary.withdrawalStatus == WithdrawalStatus.NOT_WITHDRAWN,
                 ParticipantSummary.suspensionStatus == SuspensionStatus.NOT_SUSPENDED,
                 GenomicAW3Raw.id.is_(None)
-            ).distinct().all()
+            )
```
j-kanuch, line 3607: The query looks good. I tested it on prod and it took 20 seconds, which is acceptable considering it needs to scan the gc_data_file table. As we discussed, please remove the file-received flags.

rdr_service/dao/genomics_dao.py
```diff
@@ -3677,7 +3840,20 @@ def get_aw3_wgs_records(self, **kwargs):
                 ParticipantSummary.withdrawalStatus == WithdrawalStatus.NOT_WITHDRAWN,
```
j-kanuch, line 3703: Same thing here with the file-received flags.

rdr_service/genomic/genomic_job_components.py
brianmkimmel: This has been revised to use a new GENOMIC_GC_SITE_BUCKET_MAP configuration setting

rdr_service/genomic/genomic_job_components.py
brianmkimmel: This is still used for the method GenomicFileIngester.ingest_single_aw2_row_for_member, but for GenomicFileIngester._process_gc_metrics_data_for_insert I added a member_dao method to perform a bulk update.

rdr_service/dao/genomics_dao.py
brianmkimmel: I've added a variable for the list comprehension.

rdr_service/dao/genomics_dao.py
brianmkimmel: same for this one

rdr_service/cron_default.yaml
brianmkimmel: Done

rdr_service/offline/main.py
brianmkimmel: The decorator has been added.

rdr_service/cron_default.yaml
dtharpeuno: Sorry, can we add the Genomic to the url route as well?

rdr_service/dao/genomics_dao.py
```diff
@@ -1217,10 +1205,118 @@ def get_members_from_biobank_ids(self, biobank_ids, genome_type=None):
...
+                members = members.filter(
+                    GenomicSetMember.sampleId.in_(sample_list)
+                )
+            return members.all()
+
     def bulk_update_members(self, members: List[Dict]):
         with self.session() as session:
             session.bulk_update_mappings(GenomicSetMember, members)
 
+    def bulk_update_genomics_workflow_state(self, members, new_state):
```
dtharpeuno: So instead of this, we already have bulk update methods on this class, instead of adding this to the dao, can we handle this in the components that are calling it maybe another method and send the objects to be updated to the method right above this? `bulk_update_members`

rdr_service/offline/main.py
```diff
@@ -1291,6 +1247,13 @@ def _build_pipeline_app():
         methods=['GET']
     )
 
+    offline_app.add_url_rule(
+        OFFLINE_PREFIX + 'AW3ReadyMissingDataFilesReport',
```
dtharpeuno: Add "Genomic" prefix here as well.

rdr_service/dao/genomics_dao.py
brianmkimmel: Yes, I've revised GenomicFileIngester._process_gc_metrics_data_for_insert so it uses bulk_update_members instead of a new dao method.

dtharpeuno: Thanks looks good

rdr_service/offline/main.py
brianmkimmel: I've updated the cron job to use Genomic in the route name

tests/tool_tests/test_genomic_utils.py
```diff
@@ -437,3 +436,96 @@ def test_unblock_samples(self):
...
+        for file_name in array_files:
+            # Set file type
+            file_type = file_name.split('/')[-1].split("_")[-1] if "idat" in file_name.lower() else \
+                '.'.join(file_name.split('.')[1:])
+
+            test_file_dict = {
+                'file_path': f'{bucket_name}/{file_name}',
+                'gc_site_id': 'rdr',
+                'bucket_name': bucket_name,
+                'file_prefix': f'test_data_folder',
```
dtharpeuno: :nit can we clean all the f-strings that we are not inserting values into.

tests/genomics_tests/test_genomic_pipeline.py
```diff
@@ -196,6 +201,19 @@ def setUp(self):
...
+        config.override_setting(config.GENOMIC_GC_SITE_PREFIX_MAP, {'rdr': {
+            'cram': 'Wgs_sample_raw_data/CRAMs_CRAIs',
+            'cram.crai': 'Wgs_sample_raw_data/CRAMs_CRAIs',
+            'cram.md5sum': 'Wgs_sample_raw_data/CRAMs_CRAIs',
+            'hard-filtered.vcf.gz': 'Wgs_sample_raw_data/SS_VCF_clinical',
+            'hard-filtered.vcf.gz.md5sum': 'Wgs_sample_raw_data/SS_VCF_clinical',
+            'hard-filtered.vcf.gz.tbi': 'Wgs_sample_raw_data/SS_VCF_clinical',
+            'hard-filtered.gvcf.gz': 'Wgs_sample_raw_data/SS_VCF_research',
+            'hard-filtered.gvcf.gz.md5sum': 'Wgs_sample_raw_data/SS_VCF_research'
+        }})
```
dtharpeuno: Arent these going to be the same for all the GCs?

Can we rename this variable to better encompass what its storing?

rdr_service/config.py
```diff
@@ -128,6 +128,10 @@
 GENOMIC_API_MODES = ('gem', 'rhp', 'pgx', 'hdr')
 GENOMIC_CVL_SITES = ('co', 'uw', 'bcm')
 GENOMIC_CVL_SITE_BUCKET_MAP = 'cvl_site_bucket_map'
+GENOMIC_GC_SITE_BUCKET_MAP = 'gc_site_bucket_map'
+JH_BUCKET_NAME = "jh_bucket_name"
+BI_BUCKET_NAME = "bi_bucket_name"
```
dtharpeuno: Are we missing some GC buckets here?

tests/genomics_tests/test_genomic_pipeline.py
brianmkimmel: No, there are differences in capitalization for the file paths depending on the GC so it needs to be set based on the site.

dtharpeuno: So then why are we setting this in the config then? Can you not reuse/utilize/add to the dicts in `genomic_mappings.py`

rdr_service/config.py
brianmkimmel: The existing buckets are futher up in the file on lines 108-110

tests/genomics_tests/test_genomic_pipeline.py
brianmkimmel: I set up this way so that if for some reason the file paths for the data files changed then the system could be updated without a code change.

rdr_service/config.py
dtharpeuno: Would you mind moving them to where we have the other buckets defined?

tests/genomics_tests/test_genomic_pipeline.py
dtharpeuno: It's very messy to maintain config file mappings like this, config changes should be very simple. We already have this existing type of functionality in the mappings file and would prefer to have them there instead of in the config. If the GC decides for some reason to update their file paths they will let us know.

rdr_service/genomic/genomic_job_components.py
```diff
@@ -1694,6 +1716,51 @@ def calculate_contamination_category(self, sample_id, raw_contamination, member:
 
         return contamination_category
 
+    def _set_metrics_array_data_file_paths(self, row: dict) -> dict:
+        gc_site_bucket_map = config.getSettingJson(config.GENOMIC_GC_SITE_BUCKET_MAP)
+        site_id = self.file_obj.fileName.split('_')[0].lower()
```
dtharpeuno: Can we add a fallback for these configs entries so if they are not present the code doesn't fail?

tests/genomics_tests/test_genomic_pipeline.py
brianmkimmel: I've moved this into genomic_mappings so it's not pulling the dictionary from config any more.

rdr_service/config.py
brianmkimmel: I have these lines with the other buckets now.

rdr_service/genomic/genomic_job_components.py
brianmkimmel: I've added defaults and a conditional return so we won't get a runtime error if a value can't be found in the config.

dtharpeuno (approved):

j-kanuch (approved):
Sorry, I thought I approved this already.

PR #3168, opened 2022-09-27
add distinct to IL ready query.

j-kanuch:
This PR removes duplicate records from the set of samples returned from the IL ready query. Multiple biobank_order_identifier records per biospecimen were causing duplicate sample_ids to be returned. The duplicate sample_ids were taking the place of unique IDs and therefore causing the job to update less genomic_set_member records than expected.

## Tests
- [x] unit tests


dtharpeuno (approved):
Nice, thanks for finding this

kenny-skaggs (approved):

PR #3170, opened 2022-09-28
[DA-2972] Core Minus PM Time Backfill

brianmkimmel:
Calculates the Core Minus PM Time for all particpants and updates the value when it is greater than the calculated time.

## Tests
- [] unit tests
Results of the calculated Core Minus PM Time were compared with the participant summaries for a subset to confirm that the correct time was being used.

### DIFF
--- a/rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
+++ b/rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
@@ -1,7 +1,6 @@
 import logging
 
 from rdr_service import clock
-from rdr_service.dao.participant_summary_dao import ParticipantSummaryDao
 from rdr_service.logic.enrollment_info import EnrollmentCalculation, EnrollmentDependencies
 from rdr_service.model.participant_summary import ParticipantSummary
 from rdr_service.participant_enums import EnrollmentStatus
@@ -26,7 +25,7 @@ class BackfillCoreMinusPMTime(ToolBase):
             ).order_by(
                 ParticipantSummary.participantId
             ).all()
-            summary_dao = ParticipantSummaryDao()
+
             logger.info(f'PIDs to validate Core Minus PM Time: {len(participant_ids)}')
             chunk_size = 50
             count = 0
@@ -45,7 +44,6 @@ class BackfillCoreMinusPMTime(ToolBase):
                         batch_count += 1
                         summary.enrollmentStatusCoreMinusPMTime = calculated_date
                         summary.modifiedTime = clock.CLOCK.now()
-                        summary_dao.update(summary)
                 session.commit()
                 logger.info(f'Updated {batch_count} of {chunk_size}')
                 count += 1
--- a/rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
+++ b/rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
@@ -1,0 +1,77 @@
+from rdr_service import clock
+from rdr_service.dao.participant_summary_dao import ParticipantSummaryDao
+from rdr_service.logic.enrollment_info import EnrollmentCalculation, EnrollmentDependencies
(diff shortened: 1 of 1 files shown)


rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
```diff
@@ -0,0 +1,77 @@
...
+
+tool_cmd = 'backfill-coreminuspmtime'
+tool_desc = 'Backfill CoreMinusPMTime values that had been incorrectly updated'
+
+
+class BackfillCoreMinusPMTime(ToolBase):
+    def run(self):
+        super().run()
+        with self.get_session() as session:
+            summary_list = session.query(ParticipantSummary).all()
```
kenny-skaggs: suggestion: I'm finding with the the enrollment status backfill that it takes a while and I sometimes lose connection with the database (whether it's my computer going to sleep or some other cause). Something I've done is ordered the participants by their id and logged what the last id was of each subset. That way if it fails because of a connection issue you can add a filter for the participant id and start from the same spot rather than from the beginning.

rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
```diff
@@ -0,0 +1,77 @@
...
+            summary_dao = ParticipantSummaryDao()
+
+            chunk_size = 50
+            for summary_list_subset in list_chunks(lst=summary_list, chunk_size=chunk_size):
+                for summary in summary_list_subset:
+                    calculated_date = self.calculate_core_minus_pm_time(summary, session)
+                    if calculated_date:
+                        if calculated_date < summary.enrollmentStatusCoreMinusPMTime:
```
kenny-skaggs: readability suggestion:
```suggestion
                    if calculated_date and calculated_date < summary.enrollmentStatusCoreMinusPMTime:
```

rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
kenny-skaggs: One thing that's known about the participants that need to be backfilled is that they have a core-minus-pm date set already, it's just that the date needs to be updated, right? Can we filter the number of participants loaded by this to ones that already have a date set?

kenny-skaggs (commented):
In reviewing this, I'm realizing that there's a bug with the enrollment backfill code as well. All participant summary data is loaded up front, and then calculations are made and stored based on that. By the time the code gets to participants further down the list, the data for those participants might be out of date and would need to be reloaded.

A better approach could be to get a list of all the participant ids initially, and then load each chunk of 50, with `for update` to lock them, so we're calculating based on the latest data and we're guaranteed that something won't change.

rdr_service/tools/tool_libs/backfill_coreminuspmtime.py
```diff
@@ -0,0 +1,99 @@
...
+                ).filter(
+                    ParticipantSummary.participantId.in_(participant_id_subset)
+                ).with_for_update().all()
+                for summary in summary_list:
+                    calculated_date = self.calculate_core_minus_pm_time(summary, session)
+                    if calculated_date and calculated_date < summary.enrollmentStatusCoreMinusPMTime:
+                        batch_count += 1
+                        summary.enrollmentStatusCoreMinusPMTime = calculated_date
+                        summary.modifiedTime = clock.CLOCK.now()
+                        summary_dao.update(summary)
```
kenny-skaggs: This line can be removed (and might need to be considering how the database locking would work). You're loading the summaries further up and then setting the values on them. Later when you commit the session those changes will be sent to the database.

brianmkimmel: I tested this out and that does conflict with the database lock so I've removed it now.

kenny-skaggs (approved):

dtharpeuno (approved):

PR #3178, opened 2022-10-05
[DA-2817] MIGRATION - Refactor informing loop reconciliation

j-kanuch:
This PR refactors the Informing Loop reconciliation process. Mismatches between the informing loop `decision_value` and the nightly Color file are now pulled from the database in a single query. The job no longer creates incidents, but instead  inserts records into `genomic_informing_loop` using the Color file as the source of truth. 
Additionally, this PR adds indices and a new column to `genomic_informing_loop`, as well as indices to `user_event_metrics`.

## Tests
- [x] unit tests


brianmkimmel (approved):

rdr_service/dao/genomics_dao.py
```diff
@@ -3047,21 +3047,22 @@ def build_latest_decision_query(cls, module: str) -> sqlalchemy.orm.Query:
             )
         )
 
-    def get_latest_state_for_pid(self, pid, module="gem", decision_values_only=True):
+    def get_latest_il_for_pids(self, pid_list, module="gem", decision_values_only=True):
```
dtharpeuno, line 3038: I am not seeing any reference in the code where this method is being invoked. Are you using this somewhere other than that unit tests?

j-kanuch: Not anymore, at the moment it is just in the unit tests.

PR #3185, opened 2022-10-06
MIGRATIONS [DA-2809] - Appointment metrics ingestion/reconciliation 

dtharpeuno:
## Resolves *[ticket DA-2809]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2809

Added processes for ingesting nightly drops of appointment metrics from partners for JSON formatted files as well as daily reconciliation from metrics data to appointment events.                                     

* started breaking logic apart from the large ingestion method on `GenomicJobController` into a reusable class called `GenomicMessageBroker`, which only includes appointment ingestion events. 

## Tests
- [x] unit tests


rdr_service/dao/genomics_dao.py
```diff
@@ -4250,3 +4251,39 @@ def from_client_json(self):
...
+                GenomicAppointmentEventMetrics.appointment_event,
+                GenomicAppointmentEventMetrics.reconcile_job_run_id,
+                GenomicAppointmentEvent.id.label('appointment_event_id')
+            ).outerjoin(
+                GenomicAppointmentEvent,
+                and_(
+                    GenomicAppointmentEvent.participant_id == GenomicAppointmentEventMetrics.participant_id,
+                    GenomicAppointmentEvent.module_type == GenomicAppointmentEventMetrics.module_type,
+                    GenomicAppointmentEvent.event_type == GenomicAppointmentEventMetrics.event_type,
+                    GenomicAppointmentEvent.event_authored_time == GenomicAppointmentEventMetrics.event_authored_time
```
j-kanuch, line 4283: Just curious, do you want to join on appointment ID here too, or is event_authored_time sufficient? Also, you may want to index it.

dtharpeuno: I think `event_authored_time` is sufficient tbh in this case. I am using it because its actually the most granular attribute that I can use for reconciliation since each one of the json objects lacks a unique id on them.

j-kanuch: Sounds good.

brianmkimmel (approved):

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -459,23 +473,58 @@ def gem_results_to_report_state(self):
...
+                        'created_from_metric_id': metric.id
+                    }
+
+                    message_body = json.loads(metric.appointment_event).get('messageBody')
+                    for k, value in message_body.items():
+                        key = 'appointment_id' if k == 'id' else k
+                        appointment_obj[key] = value
+
+                    appointment_obj = self.appointment_dao.get_model_obj_from_items(appointment_obj.items())
+                    self.appointment_dao.insert(appointment_obj)
```
j-kanuch, line 509: You may have a reason for it, but since you are inserting the appointments here, do you need to update the reconciliation job run ID? In your query can you just look for total missing appointments instead of filtering on reconciliation ID? Once they are inserted they won't be returned from the missing query.

dtharpeuno: Yea I was totally on the fence with that tbh. There really is no need for it I guess and looking at the logic for the other reconciles we arent using it. I was just wondering about how performant the query would be as we got so many of these events stored and figured this would be a way to mitigate the memory load caused by the query if I had another value to filter on.

j-kanuch: Sounds good, we'll monitor in case.

j-kanuch (approved):
Good call creating the message broker class!

PR #3196, opened 2022-10-19
MIGRATIONS [DA-2926] Update genomic storage process for archiving

dtharpeuno:
## Resolves *[ticket DA-2926]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-2926

Need the ability to update genomic data files to Cold storage to reduce costs program costs for storage.

** In genomics dao, added a base default dao to utilize instead of defining the same base dao for every model. Will add refactor soon to utilize this class more in the app logic.

## Tests
- [x] unit tests


rdr_service/model/genomics.py
```diff
@@ -1597,3 +1597,25 @@ class GenomicResultWithdrawals(Base):
...
+    __tablename__ = "genomic_storage_update"
+
+    id = Column(Integer,
+                primary_key=True, autoincrement=True, nullable=False)
+    created = Column(DateTime)
+    modified = Column(DateTime)
+    metrics_id = Column(Integer, ForeignKey("genomic_gc_validation_metrics.id"), nullable=False, index=True)
+    genome_type = Column(String(80), nullable=True)
+    storage_class = Column(String(250), nullable=False)
+    has_error = Column(SmallInteger, nullable=False, default=0)
```
j-kanuch, line 1617: Good call with adding this table. Can you add an `ignore_flag` field? I'm thinking that situations may arise where we need to reclassify files after metrics records have already been processed, like re-ingestions or reprocessing, but we still may want to see if they've been previously processed.

rdr_service/genomic/genomic_storage_class.py
```diff
@@ -0,0 +1,130 @@
...
+            return
+
+        self.update_file_paths(
+            file_dict=self.get_file_dict_from_metrics(
+                metrics=wgs_metrics,
+                metric_type=config.GENOME_TYPE_WGS
+            )
+        )
+
+    def update_file_paths(self, file_dict: List[dict]):
```
j-kanuch: Nitpicky, but can you name this something like "reclassify" or "update_storage_class_for_file_paths"? `update_file_paths` makes me think that we're changing the file paths themselves, not the storage class.

brianmkimmel (approved):

j-kanuch (approved):

PR #3220, opened 2022-11-08
[MIGRATIONS][DA-3111] Cron Job for 14 Day GCR Intervention

brianmkimmel:
Creates cron job to notify Color when an hdr positive result is more than 14 days old and no appointment is scheduled or completed.

## Tests
- [x] unit tests


rdr_service/dao/genomics_dao.py
```diff
@@ -3178,6 +3218,19 @@ def from_client_json(self):
...
+        super(GenomicGCROutreachEscalationNotifiedDao, self).__init__(
+            GenomicGCROutreachEscalationNotified, order_by_ending=['id'])
+
+    def get_id(self, obj):
+        pass
+
+    def from_client_json(self):
+        pass
+
+
```
dtharpeuno: Instead of us having to create a new one of these each time, I created a default dao called `GenomicDefaultBaseDao` so maybe we can utilize this?

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -1941,6 +1941,31 @@ def check_appointments_gror_changed(self):
...
+        if notification_email_addresses and escalate_participants:
+            notified_participants = []
+            for pid in escalate_participants:
+                EmailService.send_email(
+                    Email(
+                        recipients=notification_email_addresses,
+                        subject='GCR Outreach 14 Day Escalation',
+                        plain_text_content=str(pid)
+                    )
+                )
```
dtharpeuno: Are we sending one for each participant? What if there are hundreds? Might be better to double-check since we don't want to spam Color with this.

rdr_service/dao/genomics_dao.py
```diff
@@ -3040,6 +3040,46 @@ def get_report_state_from_wf_state(wf_state):
...
+                GenomicMemberReportState.genomic_report_state == GenomicReportState.HDR_RPT_POSITIVE,
+                GenomicMemberReportState.event_authored_time < result_before,
+                or_(GenomicAppointmentEvent.event_type.notin_(('appointment_completed', 'appointment_scheduled',
+                                                               'appointment_updated')),
+                    GenomicAppointmentEvent.event_type.is_(None)),
+                or_(GenomicAppointmentEvent.event_authored_time ==
+                    max_event_authored_time_subquery.c.max_event_authored_time,
+                    GenomicAppointmentEvent.event_authored_time.is_(None))
+            ).distinct().all()
+        return [pid[0] for pid in pids]
```
dtharpeuno: This is a little outside of the realm of what we utilize the daos for, can you do this list-casting in the method where you are utilizing this query?

rdr_service/dao/genomics_dao.py
```diff
@@ -3040,6 +3040,46 @@ def get_report_state_from_wf_state(wf_state):
                 return value
         return None
 
+    def get_hdr_result_positive_no_appointment(self) -> List[str]:
+        """Returns participants with hdr positive result over 14 days ago and no appointment scheduled/completed"""
+        result_before = clock.CLOCK.now() - timedelta(days=14)
```
dtharpeuno: Can we add this as a param with a default?
`get_hdr_result_positive_no_appointment(self, days=14)`
`result_before = clock.CLOCK.now() - timedelta(days)`

rdr_service/offline/genomic_pipeline.py
```diff
@@ -612,3 +612,8 @@ def notify_aw3_ready_missing_data_files():
 def notify_appointment_gror_changed():
     with GenomicJobController(GenomicJob.CHECK_APPOINTMENT_GROR_CHANGED) as controller:
         controller.check_appointments_gror_changed()
+
+
+def notify_gcr_outreach_escalation():
+    with GenomicJobController(GenomicJob.CHECK_GCR_OUTREACH_ESCALATION) as controller:
+        controller.check_gcr_14day_escalation()
```
dtharpeuno, line 619: Can we change this name?
`check_gcr_appointment_escalation()`

rdr_service/genomic/genomic_job_controller.py
brianmkimmel: Yes, sending one email for each PID is a requirement from the ticket.

rdr_service/dao/genomics_dao.py
brianmkimmel: Yes, I've updated the PR to use GenomicDefaultBaseDao.

rdr_service/dao/genomics_dao.py
brianmkimmel: This has been moved

dtharpeuno (approved):

j-kanuch (approved):

j-kanuch (approved):

PR #3234, opened 2022-12-01
MIGRATIONS [DA-3072] Genomic WGS Reprocessing Flows

dtharpeuno:
## Resolves *[ticket DA-3072]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-3072

The GC's want to reprocess WGS data files for a newer version of `dragen` this will enable that by separating the current WGS flows, maintaining current pipeline while adding more.

## Tests
- [x] unit tests


kenny-skaggs (approved):

rdr_service/dao/genomics_dao.py
```diff
@@ -3947,6 +3966,13 @@ def get_aw3_wgs_records(self, **kwargs):
         # should be only wgs genome but query also
         # used for wgs investigation workflow
         genome_type = kwargs.get('genome_type', config.GENOME_TYPE_WGS)
+        pipeline_id = kwargs.get('pipeline_id')
+
+        if not pipeline_id:
+            return []
+
+        # only updated dragen version files will be placed in 'dragen_v' subfolder
+        data_file_path = pipeline_id if pipeline_id == config.GENOMIC_UPDATED_WGS_DRAGEN else '/'
```
j-kanuch, line 3975: Nice way to get around the case statements.

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -1815,18 +1814,37 @@ def _get_blocklist_str_attributes(all_values, map_values):
...
+                    'aw3ManifestFileId': kwargs.get('manifest_id')
+                }
+                members_to_update.append(member_dict)
+            self.member_dao.bulk_update(members_to_update)
+
+            for metric in process_metrics:
+                metric_dict = {
+                    'id': metric.id,
+                    'aw3ReadyFlag': 0,
+                    'aw3ManifestFileId': kwargs.get('manifest_id'),
```
j-kanuch, line 1843: are you updating aw3ManifestJobRunID somewhere else?

dtharpeuno: Good catch
79fbf81639d23dd58acc4dd74812464493f82346

rdr_service/dao/genomics_dao.py
```diff
@@ -3992,7 +4018,9 @@ def get_aw3_wgs_records(self, **kwargs):
                     GenomicSetMember.blockResearch == 1,
                     sqlalchemy.sql.expression.literal("True"),
                     sqlalchemy.sql.expression.literal("False")),
-                GenomicSetMember.blockResearchReason
+                GenomicSetMember.blockResearchReason,
+                GenomicGCValidationMetrics.pipelineId,
+                GenomicGCValidationMetrics.processingCount
```
j-kanuch: Should we have processingCount + 1 here? I.e. if it is 0 we want to send 1.

dtharpeuno: Should we default it to 1 then in the model?

dtharpeuno: "Proposed new field, will start with 0 the first time an AW3 (WGS) is created and increment by 1 each time a new one is created."

j-kanuch: If we default to 0 and send it as 1, then we update it below to actually be 1. When we get AW4, it matches Broad. If we default to 1, we send as 1, but when we update it, it is 2, which is out of sync with Broad. I was thinking 0 means we haven't sent it, 1 means first time sending.

dtharpeuno: Understood, made the change
https://github.com/all-of-us/raw-data-repository/pull/3234/commits/fb082594f0cbd753210eb3bde2ce09355060eccf

j-kanuch (approved):
Nice job, this is a huge project. Also thanks for putting up with all the changes.

PR #3236, opened 2022-12-01
[DA-3151] Use ProfileUpdate API to modify participant summary

kenny-skaggs:
## Resolves *[DA-3151](https://precisionmedicineinitiative.atlassian.net/browse/DA-3151)*
This connects the ProfileUpdate API to the database by using some fields to alter the participant summary, and to store the JSON of the request to a new table used for archiving the requests.

## Tests
- [x] unit tests


dtharpeuno (approved):

rdr_service/alembic/versions/bd7e69e0e71c_profile_update_archive_table.py
```diff
@@ -0,0 +1,65 @@
...
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('profile_update',
+    sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
```
pegbertsch, line 41: Perhaps on a follow up PR, we can add a `modified` column as well?  There are reasons that having `id, created, modified `as standard columns is beneficial for the RDR-to-PDR pipeline, because any tables that have those three columns (especially modified) can be handled by a generic process that looks for recently modified records.

pegbertsch (approved):
I don't want to hold up this PR from release, but before this API goes "live" in production, can we  also add a non-nullable `modified` field?

rdr_service/alembic/versions/bd7e69e0e71c_profile_update_archive_table.py
kenny-skaggs: These would be raw records being stored and wouldn't be further modified. Does it make sense that they still have a modified time even if it will never be anything different than the created time?

PR #3241, opened 2022-12-07
[DA-3065] EtM Response Validation

kenny-skaggs:
## Resolves *[DA-3065](https://precisionmedicineinitiative.atlassian.net/browse/DA-3065)*
Adds validation for EtM responses. Whenever a response is submitted for EtM, the latest questionnaire definition will be loaded and used to validate the response. The questionnaire will be used to determine what outcomes and metadata should be expected on the response, as well as checking that each required question is answered. If validation fails, an error response will be returned.

## Tests
- [x] unit tests


brianmkimmel (approved):

dtharpeuno (approved):
Nice

PR #3244, opened 2022-12-08
MIGRATIONS [DA-3156] (Partial) Add fields for participant mediated EHR

pegbertsch:
This is the first part of the temporary implementation for participant mediated EHR tracking.  This is a separate category of EHR data that is linked to AoU by the participants themselves, as opposed to being provided via an HPO upload.

 This PR adds new fields to `participant_summary` and  corrects issues with the existing unit tests.   The new fields will eventually be backfilled using a file drop that has a snapshot of participants with mediated EHR, along with their first/last receipt times.  The fields will not be exposed via the API and the logic to hide the fields by default is based on  PR #3141  and the new Goal 1 metrics fields.

Fields added here:
wasParticipantMediatedEhrAvailable
firstParticipantMediatedEhrReceiptTime
latestParticipantMediatedEhrReceiptTime

There is no `isParticipantMediatedEhrAvailable` field being added at this time.  It may be added as part of the future permanent implementation that will track mediated EHR activity on an ongoing basis.

The next PR for this ticket will  update the Computed column logic for `health_data_stream_sharing_status_v_3_1[_time]` fields in a separate migration file (to be applied after the new columns are present).  New unit tests to test the business logic changes for enrollment status details will also be added.

## Tests
- [x] unit tests (pre-existing)


pegbertsch:
@kenny-skaggs @ayazamlani FYI:  I have reviewed the Codacy output but will not be making additional updates;  I'm choosing to keep the `format` (vs. f-string) syntax that matches the existing code for readability, and am not changing the variable names.

rdr_service/alembic/versions/c199139b10c2_participant_mediated_ehr.py
```diff
@@ -0,0 +1,69 @@
...
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute("""
+                  ALTER TABLE participant_summary
+                  ADD COLUMN was_participant_mediated_ehr_available boolean not null default 0,
+                  ADD COLUMN first_mediated_ehr_receipt_time datetime null default null,
```
pegbertsch: I applied the wrong stash before most recent commit.  Updating these references to be `first_participant_mediated_... / last_participant_mediated....` column names (here and in the downgrade section).  Will push commit after local unit tests complete

ayazamlani (approved):

rdr_service/model/participant_summary.py
```diff
@@ -446,6 +446,29 @@ def enrollmentSiteId(cls):
...
+
+    firstParticipantMediatedEhrReceiptTime = None
+    """
+    UTC timestamp indicating first reported occurrence of participant-mediated EHR content
+    """
+
+    latestParticipantMediatedEhrReceiptTime = None
+    """
+    UTC timestamp indicating the latest reported occurrence of participant-mediated EHR content
+    """
```
kenny-skaggs, line 470: Should `firstParticipantMediatedEhrReceiptTime` and `latestParticipantMediatedEhrReceiptTime` load the values from the table?

pegbertsch: I pushed some updates I was making in the next PR for when the columns will be referenced/used, that should probably just go in this PR.  Can you let me know if updating the `None` to the `Column` definitions addresses this?

kenny-skaggs: Yep, that's what I meant. Sorry it wasn't clear.

kenny-skaggs (approved):

PR #3249, opened 2022-12-09
MIGRATIONS [DA-3112] Adding AIAN to Participant Summary

dtharpeuno:
## Resolves *[ticket DA-3112]
https://precisionmedicineinitiative.atlassian.net/browse/DA-3112

Need to add AIAN flag to participant summary API, **a subsequent ticket to be included next sprint will encompass the backfilling of this flag**.

## Tests
- [x] unit tests


pegbertsch:
FYI:  If this only looks for the `WhatRaceEthnicity_AIAN` answer code, it probably won't catch the small number of pids we found in the recent CDR review that had managed to answer the branching logic question of AIAN subcategories (or free text for tribe affiliation) without answering `WhatRaceEthnicity_AIAN` to the main race/ethnicity question.    Just a heads up that a manual backfill of this field may be needed for those pids.

dtharpeuno:
> FYI: If this only looks for the `WhatRaceEthnicity_AIAN` answer code, it probably won't catch the small number of pids we found in the recent CDR review that had managed to answer the branching logic question of AIAN subcategories (or free text for tribe affiliation) without answering `WhatRaceEthnicity_AIAN` to the main race/ethnicity question. Just a heads up that a manual backfill of this field may be needed for those pids.

Understood, I added that in the initial description of the ticket.

j-kanuch (approved):

j-kanuch (approved):

j-kanuch (approved):

rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -922,6 +919,10 @@ def _update_participant_summary(
             if race != participant_summary.race:
                 participant_summary.race = race
                 something_changed = True
+            if not participant_summary.aian \
+                    and Race.AMERICAN_INDIAN_OR_ALASKA_NATIVE in get_all_races_from_codes(race_codes):
+                participant_summary.aian = True
+                something_changed = True
```
kenny-skaggs, line 925: So if a participant answers that they're AIAN and then later we get an updated answer where they don't say AIAN, then we'll still keep the AIAN flag set to True?

rdr_service/model/participant_summary.py
```diff
@@ -1555,7 +1555,8 @@ def enrollmentSiteId(cls):
     """User who recorded ID verification occurrence"""
     onsiteIdVerificationSite = Column("onsite_id_verification_site_id", Integer, ForeignKey("site.site_id"))
     """The site where ID verification took place"""
-
+    aian = Column(Boolean, default=0)
+    """Denotes if the participants race is AI_AN (AMERICAN_INDIAN_OR_ALASKA_NATIVE)"""
```
kenny-skaggs, line 1559: Would it be better to leave this defaulting as Null (which would show up as "UNSET" in the API). That way it doesn't look like we have a definite answer for whether a participant is AIAN when we haven't gotten their TheBasics response yet (we'd show "UNSET" until we get a response, and then we would show Yes or No). I'm realizing now that this would also need a change in how this field gets set when we get a response.

rdr_service/dao/questionnaire_response_dao.py
dtharpeuno: Correct, if we have any answers for AIAN it needs to be set to `true` and cannot change.

rdr_service/model/participant_summary.py
dtharpeuno: I thought of that originally, but in talking with Megan, we decided to default it to `false` and let the subsequent backfill handle the current participants that need to be changed to `true`, since they always want this flag present in the payload.

kenny-skaggs: I wasn't thinking for current participants, but instead participants that are in the process of signing up. But that might be irrelevant.

kenny-skaggs (approved):

rdr_service/model/participant_summary.py
dtharpeuno: Understood.

PR #3272, opened 2022-12-21
Handle Genomic Ingestion With Missing Raw Record

brianmkimmel:
When reingesting ingestion deltas the process will often stop because no raw record is found for the genomic set member.

GenomicAW1RawDao.get_raw_record_from_identifier_genome_type will not raise the errors caught in the try block. It returns None if no record is found.

Updates the method to only add a genomic set member for ingestion if the raw record is found.

## Tests
Existing tests should pass


j-kanuch (approved):
Thanks for spotting and fixing this!

ayazamlani (approved):

PR #3284, opened 2023-01-10
Dc/da 3026

xtoller:
Implement: Ancillary Studies Order API

## Tests
- [] unit tests


rdr_service/model/study_nph.py
```diff
@@ -12,12 +14,131 @@ class Participant(NphBase):
...
+event.listen(BiobankFileExport, "before_insert", model_insert_listener)
+
+
+class SampleExport(NphBase):
+    __tablename__ = "sample_export"
+
+    id = Column("id", BigInteger, autoincrement=True, primary_key=True)
+    ignore_flag = Column(TINYINT, default=0)
+    export_id = Column(BigInteger, ForeignKey("biobank_file_export.id"))
+    sample_update_id = Column(BigInteger, ForeignKey("sample_update.id"))
```
kenny-skaggs: Why are the classes for the file export to the Biobank added? Are they needed for the order API?

rdr_service/api/nph_participant_biobank_order_api.py
```diff
@@ -0,0 +1,72 @@
...
+        pass
+
+    def put(self, nph_participant_id, rdr_order_id):
+        with database_factory.get_database().session() as session:
+            self.dao.set_order_cls(request.get_data())
+            order = self.dao.order_cls
+            self.dao.validate(rdr_order_id, nph_participant_id, session)
+            self.dao.order_sample_dao.update_order_sample(order, rdr_order_id, session)
+            self.dao.update_order(rdr_order_id, nph_participant_id, session)
+            session.commit()
```
kenny-skaggs: The commit can be removed here. The context manager for the session handles that.

rdr_service/api/nph_participant_biobank_order_api.py
```diff
@@ -0,0 +1,72 @@
...
+                    logging.warning(f'Inserting new order to study_category table: module = {order.module}, '
+                                    f'visitType: {order.visitType}, timePoint: {order.timepoint}')
+                    time_point_id = self.dao.insert_study_category_with_session(order, session)
+                order_id = self.dao.insert_with_session(time_point_id, nph_participant_id, session)
+                self.dao.insert_ordered_sample_dao_with_session(order, order_id, session)
+                order.id = order_id
+                return construct_response(order), 201
+            except NotFound as ex:
+                logging.error(ex)
+                return {}, 400
```
kenny-skaggs: If we received a NotFound error, that means something (like a participant id) was referenced but didn't exist right? Should that be returned as a 404 instead? And we should return a message detailing what was wrong with the request.

rdr_service/api/nph_participant_biobank_order_api.py
```diff
@@ -0,0 +1,72 @@
...
+            try:
+                with database_factory.get_database().session() as session:
+                    self.dao.set_order_cls(request.get_data())
+                    order = self.dao.order_cls
+                    self.dao.patch_update(order, rdr_order_id, nph_participant_id, session)
+                    session.commit()
+                    return construct_response(order), 200
+            except BadRequest as ex:
+                logging.error(ex)
+                return {}, 400
```
kenny-skaggs: We should return a message of what went wrong with the request.

rdr_service/dao/participant_dao.py
```diff
@@ -204,6 +204,7 @@ def _do_update(self, session, obj, existing_obj):
         obj.lastModified = clock.CLOCK.now()
         obj.signUpTime = existing_obj.signUpTime
         obj.biobankId = existing_obj.biobankId
+        obj.biobankId = existing_obj.biobankId
```
kenny-skaggs: This line looks like a duplicate of the line above it. Is this addition needed?

rdr_service/dao/study_nph_dao.py
```diff
@@ -1,11 +1,493 @@
...
+        query.session = session
+        result = query.filter(Participant.id == participant_id).first()
+        if result:
+            return True
+        else:
+            return False
+
+    @staticmethod
+    def convert_id(nph_participant_id: str) -> int:
+        return int(nph_participant_id[3:])
```
kenny-skaggs: We're expecting NPH ids to be prefixed by "1000".
```suggestion
        return int(nph_participant_id[4:])
```

kenny-skaggs: Also, this method should ensure that the participant id isn't an empty string and has atleast the needed number of characters.

rdr_service/model/study_nph.py
```diff
@@ -12,12 +14,131 @@ class Participant(NphBase):
...
+    id = Column("id", BigInteger, autoincrement=True, primary_key=True)
+    created = Column(UTCDateTime)
+    name = Column(String(128))
+    type_label = Column(String(128))
+    parent_id = Column(BigInteger, ForeignKey("study_category.id"))
+    parent = relation("StudyCategory", remote_side=[id])
+    children = relation("StudyCategory", remote_side=[parent_id], uselist=True)
+
+
+event.listen(StudyCategory, "before_insert", model_insert_listener)
```
kenny-skaggs, line 40: `model_insert_listener` will fail since StudyCategory doesn't have a "modified" field. The "created" field should be set to a default instead.

rdr_service/api/nph_participant_biobank_order_api.py
```diff
@@ -0,0 +1,72 @@
...
+            session.commit()
+        return construct_response(order), 201
+
+    def post(self, nph_participant_id: str):
+        with database_factory.get_database().session() as session:
+            try:
+                self.dao.set_order_cls(request.get_data())
+                order = self.dao.order_cls
+                exist, time_point_id = self.dao.get_study_category_id(session)
+                if not exist:
```
kenny-skaggs, line 68: If I'm following this correctly `get_study_category_id` will try to find the timepoint's id for the order based on the names given for the module, visitType, and timepoint. And if they're not found, `insert_study_category_with_session` creates them all in the database.

So that works well if a new order comes in for `NPH Module 1`->`LMT`->`15 min` (all StudyCategories would be created). But then if we get another order for `NPH Module 1`->`LMT`->`30 min` they would all be created again, right? We want to be sure to add `30 min` to the `LMT` category we already have.

kenny-skaggs (commented):
There are several unit tests commented out. Do they need to be left in, or can they be removed?

rdr_service/api/nph_participant_biobank_order_api.py
xtoller: this has been replace

rdr_service/api/nph_participant_biobank_order_api.py
xtoller: Updated to use Status 404 instead of 400

rdr_service/dao/study_nph_dao.py
xtoller: updated to slice to index to 4
including the participant ID validation to make sure the length is at least 5.

rdr_service/api/nph_participant_biobank_order_api.py
xtoller: commit removed

rdr_service/api/nph_participant_biobank_order_api.py
xtoller: Made the validation check against the StudyCategory table and only insert item(s) if it misses

rdr_service/model/study_nph.py
jmtrujillo75: @kenny-skaggs , I ran these unittests and they PASSED. Unless I am testing it wrong.

```python
class NphStudyCategoryTest(BaseTestCase):

    def setUp(self):
        super().setUp()
        self.nph_study_category_dao = NphStudyCategoryDao()

    def test_get_before_insert(self):
        self.assertIsNone(self.nph_study_category_dao.get(1))

    def _create_study_category(self, study_category_obj: Dict[str, Any]) -> StudyCategory:
        nph_study_category = StudyCategory(**study_category_obj)
        with FakeClock(TIME):
            return self.nph_study_category_dao.insert(nph_study_category)

    def test_insert_parent_study_category(self):
        parent_study_category = {
            "name": "Parent Study Category",
            "type_label": "PARENT",
            "parent_id": None
        }
        _parent_study_category = self._create_study_category(parent_study_category)
        expected_parent_study_category = {
            "id": 1,
            "created": TIME,
            "name": "Parent Study Category",
            "type_label": "PARENT",
            "parent_id": None
        }
        expected_parent_study_category_ = StudyCategory(**expected_parent_study_category)
        self.assertEqual(
            expected_parent_study_category_.asdict(),
            _parent_study_category.asdict()
        )

    def test_insert_child_study_category(self):
        parent_study_category_obj = {
            "name": "Parent Study Category",
            "type_label": "PARENT",
            "parent_id": None
        }
        parent_study_category = self._create_study_category(parent_study_category_obj)

        child_study_category_obj = {
            "name": "Child Study Category",
            "type_label": "CHILD",
            "parent_id": parent_study_category.id
        }
        child_study_category = self._create_study_category(child_study_category_obj)
        expected_child_study_category = {
            "id": 2,
            "created": TIME,
            "name": "Child Study Category",
            "type_label": "CHILD",
            "parent_id": parent_study_category.id
        }
        expected_child_study_category_ = StudyCategory(**expected_child_study_category)
        self.assertEqual(
            expected_child_study_category_.asdict(),
            child_study_category.asdict()
        )

    def tearDown(self):
        self.clear_table_after_test("nph.study_category")
```

kenny-skaggs: @jmtrujillo75, you're right. We're already doing the same thing here that we're doing with ApiUser (calling model_insert_listener without a modified field on the object). I'm not sure what I was conflating this with, but I was wrong to think it'd be an issue. Sorry about that!

rdr_service/api/nph_participant_biobank_order_api.py
```diff
@@ -0,0 +1,103 @@
...
+
+class NphOrderApi(UpdatableApi):
+
+    def __init__(self):
+        super(NphOrderApi, self).__init__(NphOrderDao())
+
+    def update_with_patch(self, id_, resource, expected_version):
+        pass
+
+    def put(self, nph_participant_id, rdr_order_id):
```
kenny-skaggs, line 34: I know you're working on some more unit tests, but I noticed something else we'll want to make sure to update. Only Hpro should be able to make requests to these endpoints, so we'll want to use the `@app_util.auth_required(RTI_AND_HEALTHPRO)` decorator. This should be added to `post` and `patch` as well.

xtoller: Adding the auth_required for RTI and HealthPro

kenny-skaggs (approved):

brianmkimmel (approved):

PR #3302, opened 2023-01-23
DA-3280 NPH Participant Variable are expected to include the prefix 

xtoller:
DA-3280 NPH Participant Variable are expected to include the prefix (1000)
NPH Participant API Query variable nph_Id expected to include the prefix(1000)

## Tests
- [X] unit tests

### DIFF
--- a/rdr_service/dao/study_nph_dao.py
+++ b/rdr_service/dao/study_nph_dao.py
@@ -32,38 +32,39 @@ class NphParticipantDao(BaseDao):
         return obj.id
 
     def get_id(self, session, nph_participant_id: str) -> int:
-        participant_id = self.convert_id(nph_participant_id)
+        nph_participant_id = self.convert_id(nph_participant_id)
         query = Query(Participant)
         query.session = session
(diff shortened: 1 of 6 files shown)


rdr_service/dao/study_nph_dao.py
```diff
@@ -166,12 +167,12 @@ def __init__(self):
...
     @staticmethod
-    def _fetch_site_id(session, site_name) -> int:
+    def _fetch_site_id(session, external_id) -> int:
         query = Query(Site)
         query.session = session
-        result = query.filter(Site.name == site_name).first()
+        result = query.filter(Site.external_id == external_id).first()
         if result is None:
-            raise NotFound(f"Site is not found -- {site_name}")
+            raise NotFound(f"Site is not found -- {external_id}")
```
j-kanuch, line 175: Can you move this check to the API layer? We may need to lookup sites for other purposes in the system and may not want to raise an exception on no results.

j-kanuch (approved):

kenny-skaggs (approved):

PR #3338, opened 2023-02-10
[DA-3238] Reset Replated Samples Backfill Tool

brianmkimmel:
While investigating DA-3238 442 replated samples with the same sample id as the sample they were replated from were identified. This likely occured during AW1 reingestion.

This tool will be used to reset the identified genomic set members to the original state.

## Tests
- [x] unit tests

[DA-3238]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


j-kanuch (approved):

rdr_service/tools/tool_libs/backfill_reset_replated_samples.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, args, gcp_env=None, tool_name=None):
+        super().__init__(args, gcp_env, tool_name)
+        self.member_dao = GenomicSetMemberDao()
+
+    def run(self):
+        super(ResetMembersBackfillTool, self).run()
+        # Get list of paths
+        id_list = self.get_ids_from_file()
+
+        members = self.member_dao.get_members_from_member_ids(id_list)
```
dtharpeuno, line 21: Sorry was there suppose to be some sort of lookup for the last created timestamp for these? Or was there a change in that? If not should we just add this to the Frequently used queries doc then?

brianmkimmel: No, this PR is just for reseting the members which were incorrectly updated from a reingestion. Adding a timestamp comparison to the ingestion process is a different ticket.

dtharpeuno: Understood, so can we just add this to the Frequently Used Queries as Update query instead of building a tool for this?

PR #3355, opened 2023-02-16
[DA-3315] Rebuilding PDR data when EHR status gets updated

kenny-skaggs:
## Resolves *[DA-3315](https://precisionmedicineinitiative.atlassian.net/browse/DA-3315)*
This rebuilds the participant's PDR data when the consent validation would update their EHR status.

## Tests
- [ ] unit tests

[DA-3315]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/services/consent/validation.py
+++ b/rdr_service/services/consent/validation.py
@@ -92,11 +92,14 @@ class EhrStatusUpdater(ConsentMetadataUpdater):
                 session=self._session
             )
             # Rebuild for PDR
+            additional_args = {}
+            if self._project_name is not None:
+                additional_args['project_id'] = self._project_name
             self._task.execute(
                 'rebuild_one_participant_task',
                 payload={'p_id': participant_id},
                 in_seconds=30,
-                project_id=self._project_name
+                **additional_args
             )
 
         self._session.commit()  # release the for_update lock obtained on the participant_summary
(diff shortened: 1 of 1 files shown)


rdr_service/services/consent/validation.py
```diff
@@ -88,6 +91,14 @@ def _update_status(self, participant_id, has_valid_file):
                 summary=participant_summary,
                 session=self._session
             )
+            # Rebuild for PDR
+            self._task.execute(
+                'rebuild_one_participant_task',
+                payload={'p_id': participant_id},
+                in_seconds=30,
+                project_id=self._project_name
```
pegbertsch: Are there use cases where the `self._project_name` attribute here can be set to None (e.g., if StorageResultsStrategy instance was not initialized with a specific `project_id` )?   Wondering if there are cases where the call to `execute()` here would end up overriding the GCPCloudTask `project_id` GAE_PROJECT default and lead to an exception.

dtharpeuno (approved):

rdr_service/services/consent/validation.py
kenny-skaggs: You're right, good catch.

kenny-skaggs: Should be fixed now.

pegbertsch (approved):

PR #3362, opened 2023-02-21
removing old docs

ayazamlani:
Removing old documentation from docs folder 

## Tests
- [] unit tests


pegbertsch (commented):
Was this just a simple `git rm` on the local branch? If so:
There may be more required to permanently delete it from the repository history,  but I'm not clear on whether/where we will be archiving this info instead first, before it's nuked.

See: https://stackoverflow.com/questions/872565/remove-sensitive-files-and-their-commits-from-git-history

ayazamlani:
NOTE from the BFG TOOL: 

I need to remove the disaster_recovery folder first and then I can run the tool. 

I will re request approval now

____
By default the BFG doesn't modify the contents of your latest commit on your master (or 'HEAD') branch, even though it will clean all the commits before it.

That's because your latest commit is likely to be the one that you deploy to production, and a simple deletion of a private credential or a big file is quite likely to result in broken code that no longer has the hard-coded data it expects - you need to fix that, the BFG can't do it for you. Once you've committed your changes- and your latest commit is clean with none of the undesired data in it - you can run the BFG to perform it's simple deletion operations over all your historical commits.

dtharpeuno (approved):

brianmkimmel (approved):

PR #3374, opened 2023-03-01
Updates to the NPH export tracking

kenny-skaggs:
## Resolves *no ticket*
The NPH file export is occasionally missing sample updates in the file we send to the Biobank. The most common (if not the only) cause for this seems to be that the updates that get missed are arriving while the file is being generated. The process for generating the file is this: the script 1) pulls all the updates that have happened since the last file was created, 2) pulls together all the json data for the new file, and then 3) records the timestamp for the new file. The issue is when some samples get updated between when step 1 pulled the updates and step 3 recorded the time.

This updates the code to look for any SampleUpdate records that haven't been in an export, regardless of when they were created.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/offline/study_nph_biobank_file_export.py
+++ b/rdr_service/offline/study_nph_biobank_file_export.py
@@ -55,24 +55,18 @@ def _get_nph_participant(participant_id: int) -> NphParticipant:
         return session.query(NphParticipant).get(participant_id)
 
 
-def _get_latest_biobank_file_export() -> BiobankFileExport:
-    nph_biobank_file_export = NphBiobankFileExportDao()
-    with nph_biobank_file_export.session() as session:
-        return session.query(BiobankFileExport).order_by(
-            BiobankFileExport.created.desc()
-        ).first()
-
-
-def _get_sample_updates_since_last_file_export() -> Iterable[SampleUpdate]:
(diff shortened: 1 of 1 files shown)


rdr_service/offline/study_nph_biobank_file_export.py
```diff
@@ -55,24 +55,18 @@ def _get_nph_participant(participant_id: int) -> NphParticipant:
...
+    Return any SampleUpdates that haven't already been included in an export
+    """
     sample_update_dao = NphSampleUpdateDao()
     with sample_update_dao.session() as session:
-        return session.query(SampleUpdate).filter(
-            SampleUpdate.created >= last_successful_biobank_file_export_date
+        return session.query(
+            SampleUpdate
+        ).outerjoin(
+            SampleExport
```
dtharpeuno, line 67: SQLALchemy infers what to join to for this? TIL

dtharpeuno (approved):

rdr_service/offline/study_nph_biobank_file_export.py
kenny-skaggs: Yeah, as long as there's only one other model that relates to what's being joined, and it only relates through one foreign key, then SQLAlchemy will essentially assume you meant the one and only possible thing.

j-kanuch (approved):

PR #3384, opened 2023-03-07
[MIGRATION][DA-3406] Implement NPH fields in ParticipantSummary

brianmkimmel:
Adds 6 new fields to ParticipantSummary for NPH values:

* consentForNphModule1
* consentForNphModule1Authored
* nphWithdrawal
* nphWithdrawalAuthored
* nphDeactivation
* nphDeactivationAuthored
The fields are updated from the NPH Intake API using a cloud task.

## Tests
- [x] unit tests

[DA-3406]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/api/nph_intake_api.py
```diff
@@ -248,6 +250,18 @@ def post(self):
...
+                    data = {
+                        'event_type': activity_data.name,
+                        'participant_id': participant_id,
+                        'event_authored_time': event_objs[0]['event_authored_time']
+                    }
+                    if GAE_PROJECT == 'localhost':
+                        pass
+                    else:
+                        _task = GCPCloudTask()
+                        _task.execute(endpoint='update_participant_summary_task', payload=data, queue='nph')
```
dtharpeuno: This needs to be `update_participant_summary_for_nph_task` based on the route implementation.

rdr_service/api/nph_intake_api.py
```diff
@@ -248,6 +250,18 @@ def post(self):
...
+
```
dtharpeuno: @brianmkimmel for this logic, I think it needs to be moved into some sort of data structure and executed **_after data insertions occur_**. 

The reason for this is that the intake API code executes and runs validations on the list of Bundles then builds list of dicts for data inserts so that if some of the payload raises a `Badrequest` then so we don't have discrepancies in the data. We are already having issues with the test payload we have been receiving so this is why I refactored this to work with validation first, then data insertions. 

Since you have this cloud task here there could be inconsistent data with what is getting stored in the participant summary and what we have in the intake events.

dtharpeuno (changes requested):

kenny-skaggs (approved):

rdr_service/api/nph_intake_api.py
```diff
@@ -267,4 +277,11 @@ def post(self):
             if dao_event_objs:
                 dao.insert_bulk(dao_event_objs)
 
+        if GAE_PROJECT == 'localhost':
+            pass
+        else:
+            _task = GCPCloudTask()
+            for summary_update in summary_updates:
+                _task.execute(endpoint='update_participant_summary_for_nph_task', payload=summary_update, queue='nph')
+
```
dtharpeuno, line 286: :nit instead of having the full conditional we can just do:
```
if GAE_PROJECT != 'localhost':
... proceeding iteration
```

dtharpeuno (approved):

PR #3390, opened 2023-03-08
[PDR-1625] Add support for EHR SUBMITTED_NOT_VALIDATED/INVALID statuses

pegbertsch:
RDR will now roll up EHR status details into `participant summary.consent_for_electronic_health_records` which can include new statuses (SUBMITTED_NOT_VALIDATED, SUBMITTED_INVALID) to indicate the EHR PDF has not yet been validated, or failed validation and we don't have a good PDF copy of the consent on file yet.

PDR generators must determine the EHR consent status (SUBMITTED, SUBMITTED_NO_CONSENT, ...) for each `questionnaire_response` in the participant's history.  This used to be based solely on the consent question answer found for the response (e.g., `ConsentPermission_Yes`).  Now it will have to include checking the PDF validation status, using data from the `consent_file` and `consent_response` tables.

Older consent payloads will essentially skip the new logic, since in RDR it only applies to new incoming consents received after the RDR release supporting the new SUBMITTED_NOT_VALIDATED and SUBMITTED invalid statuses for `participant_summary.consent_for_electronic_health_records`.  

## Tests
- Did dry run PDR participant data builds (nothing saved to the DB) for a number of participants who had EHR validation results in RDR prod and verified the generated data.

[PDR-1625]: https://precisionmedicineinitiative.atlassian.net/browse/PDR-1625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -64,6 +64,9 @@ class ModuleLookupEnum(enum.Enum):
     LAST = 2
 
 
+# TODO in new RDR-PDR pipeine:  See if we need to create separate participant_module records for CABOR consent, if
+# the extra CABOR question code/signature answer code is part of a ConsentPII payload.
+
 _consent_module_question_map = {
     # { module: question code string }
     'ConsentPII': None,
@@ -813,7 +816,7 @@ class ParticipantSummaryGenerator(generators.BaseGenerator):
                         if (module_name == 'EHRConsentPII' and module_status == BQModuleStatusEnum.SUBMITTED
                                 and not prior_ehr_submitted_status):
                             module_status = self.get_pdf_validation_status(row.questionnaireResponseId,
-                                                                           row.authored, ro_session)
+                                                                           row.authored, module_name, ro_session)
                             prior_ehr_submitted_status = module_status == BQModuleStatusEnum.SUBMITTED
 
                         module_data['status'] = module_status.name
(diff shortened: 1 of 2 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -1829,6 +1839,47 @@ def _calculate_ubr(self, p_id, summary, ro_session):
...
+
+        # Find consent validation result associated with this questionnaire response.  Note, many older consent
+        # validation results do not have a consent_response record because that relation/table was added later
+        sql = """
+             select cf.created, cf.sync_status
+             from consent_response cr
+             join consent_file cf on cr.id = cf.consent_response_id
+             where cr.questionnaire_response_id = :qr_id
+
+        """
```
robabram: Peggy, can there be more than one 'consent_response' record per questionnaire_response_id?  If so, should there be an 'order by xxxx desc' clause in the sql to get the most recent record?

pegbertsch: I guess this is possible in normal circumstances for other consents that are not yet part of this PDF validation  logic (EHR specific for now) -- e.g., a Primary consent and a CABOR will have separate `consent_file` validation records / `consent_response` records but have the same `questionnaire_response_id`

There is also what looks like one case of unintended duplication in the `consent_file` table for one EHR consent, where four validation records were created for the same consent (all details in the `consent_file` record match), and four `consent_response` records for that `questionnaire_response_id` were created as well.  

@robabram addressed in latest commit

kenny-skaggs (approved):

robabram (approved):

PR #3391, opened 2023-03-09
[PDR-1585] Add id verification data to PDR

pegbertsch:
Adds interim / old RDR-PDR pipeline support for OnSite and Remote ID verification data.  Because these come in via different sources to RDR (`POST /OnSite/Id/Verification` endpoint and `POST /QuestionnaireResponse` for the `remote_id_verification` module, they are handled by different paths in the existing RDR-PDR pipeline:

- OnsiteIdRemoteVerification PDR data flows through RDR /resource API (NiFi GET requests), and the PDR generator will upsert records into the `resource_data` table in RDR.   Cloud task endpoints are added to handle the record rebuilds.  The RDR `default` service `post` handler for the OnsiteIdVerification API will queue the cloud task PDR build for the new RDR record before exiting.

- QuestionnaireResponse data for `remote_id_verification` module will flow through the existing module data build/ BigQuery sync process  (records built for `bigquery_sync` RDR table).    The new BigQuery `v_pdr_mod_remote_id_verification` view will use custom SQL to convert the `value_string` strings from the `remote_id_verification` questionnaire_response payloads for `remote_identity_verified` and `remote_identity_verified_on` into a smallint (0/1 instead of "true" or "false") and a datetime (from the seconds since epoch string value), respectively.

The `resource` tool which rebuilds PDR records on demand was refactored to consolidate builds for RDR `hpo`, `organization`, `site` and `code` table records into a generic `MiscResource` tool class. This class will also handle the new `onsite_id_verification` record rebuilds.  Batch rebuild requests via cloud tasks were enabled for the `hpo`, `site` and `organization` tables, when rebuilding all ids or large id sets (previously could only execute those table record rebuilds on the local server, which was very slow when large sets of records were being rebuilt).

## Tests
- [X] unit tests
Extended API unit tests for testing POST /OnSite/Id/Verification to also verify the PDR generator activity
Added `task_api` unit tests to verify the new task endpoints.

[PDR-1585]: https://precisionmedicineinitiative.atlassian.net/browse/PDR-1585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/api/cloud_tasks_api.py
+++ b/rdr_service/api/cloud_tasks_api.py
@@ -147,7 +147,7 @@ class OnSiteIdVerificationBatchRebuildTaskApi(Resource):
         if not ov_ids:
             raise NotFound('OnSiteIdVerification record id list invalid')
 
-        logging.info(f'Rebuilding onsite_id_verification records')
+        logging.info('Rebuilding onsite_id_verification records')
         logging.debug(f'Rebuild id list: {ov_ids}')
         onsite_id_verification_batch_rebuild_task(ov_ids)
         logging.info('Complete.')
--- a/rdr_service/api/cloud_tasks_api.py
+++ b/rdr_service/api/cloud_tasks_api.py
@@ -8,6 +8,9 @@ from rdr_service.api.data_gen_api import generate_samples_task
 from rdr_service.api_util import parse_date, returns_success
 from rdr_service.app_util import task_auth_required
 from rdr_service.dao.bq_code_dao import rebuild_bq_codebook_task
+from rdr_service.dao.bq_hpo_dao import bq_hpo_update_all
+from rdr_service.dao.bq_organization_dao import bq_organization_update_all
+from rdr_service.dao.bq_site_dao import bq_site_update_all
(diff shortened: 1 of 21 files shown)


rdr_service/api/cloud_tasks_api.py
```diff
@@ -67,6 +72,86 @@ def post(self):
...
+    """
+    @task_auth_required
+    def post(self):
+        log_task_headers()
+        data = request.get_json(force=True)
+        ov_ids = data.get('onsite_verification_id_list', [])
+        if not ov_ids:
+            raise NotFound('OnSiteIdVerification record id list invalid')
+
+        logging.info(f'Rebuilding onsite_id_verification records')
```
ayazamlani:
```suggestion
        logging.info('Rebuilding onsite_id_verification records')
```

ayazamlani (approved):

rdr_service/api/cloud_tasks_api.py
pegbertsch: Done

robabram (approved):

PR #3394, opened 2023-03-13
Improving memory efficiency for monthly consent sync

kenny-skaggs:
## Partially Resolves *[DA-3325](https://precisionmedicineinitiative.atlassian.net/browse/DA-3325)*
This refactors the consent sync code a bit to help with readability, and also adjusts how memory is used in a way that may reduce the memory footprint and keep GCP from killing the server.

When processing consent files to zip. They were being "downloaded" to a temp folder. In GAE, anything stored to "disk" this way was actually being maintained in memory. This PR updates the code by removing that middle step of storing the files locally, and instead using the remote cloud path as a source for creating the zip file.

Note: the ZipFile library appears to be storing the file locally anyway (when building the PDF), but this change atleast prevents us from doubling the memory that we'd be using by storing them again.

## Tests
- [x] unit tests

[DA-3325]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/offline/sync_consent_files.py
+++ b/rdr_service/offline/sync_consent_files.py
@@ -3,6 +3,7 @@ Sync Consent Files
 
 Organize all consent files from PTSC source bucket into proper awardee buckets.
 """
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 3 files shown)


ayazamlani (approved):

dtharpeuno (approved):
Nice

PR #3396, opened 2023-03-15
[DA-3432] Upgrade offline to F4_1G

j-kanuch:
Upgrading offline app instances to F4_1G. This instance class supports automatic scaling.

## Tests
- [x] unit tests

[DA-3432]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/offline.yaml
+++ b/rdr_service/offline.yaml
@@ -4,12 +4,9 @@ runtime: python37
 service: offline
 entrypoint: gunicorn -c rdr_service/services/gunicorn_config.py --timeout 18000 --max-requests 10 --max-requests-jitter 3 rdr_service.offline.main:app
 
-# Default AppEngine configuration (B1) has only 128 MB of RAM and 600 MhZ CPU, which makes
-# the metrics pipeline run very slowly. Bumping up to 1 GB of RAM and 2.4 GHz to speed things up.
-# https://cloud.google.com/appengine/docs/standard/
+# Because we are changing to an auto-scaled class, need to specify automatic_scaling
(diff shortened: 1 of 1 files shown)


rdr_service/offline.yaml
```diff
@@ -4,12 +4,10 @@ runtime: python37
...
-instance_class: B4
-# We need to specify basic scaling in order to use a backend instance class.
-basic_scaling:
-  max_instances: 10
-  idle_timeout: 1m
+# Because we are changing to an auto-scaled class, need to specify automatic_scaling
+# https://cloud.google.com/appengine/docs/standard/reference/app-yaml?tab=python#scaling_elements
+instance_class: F4_1G
+automatic_scaling:
+  min_idle_instances: automatic
```
pegbertsch, line 11: We're covered on the "Warning" in the documentation related to `min_idle_instances` because we have `inbound_services` warmup configured for all of our services (part of the base config), is that correct?

I could not find where it talked about what `automatic` meant with respect to `min_idle_instances`, although it did say it was optional to specify a value at all?

pegbertsch (approved):
Approved assuming I interpreted the documentation you linked correctly.

rdr_service/offline.yaml
j-kanuch: Yes, the default is `automatic`. I specified just so that it is apparent what we are doing. Same with latency. We're going to monitor it, it may require some "tuning".

kenny-skaggs (approved):

PR #3416, opened 2023-03-28
Updates to finding validated EHR for enrollment status calculation

kenny-skaggs:
## Resolves *no ticket*
The enrollment status calculation now depends on finding a validated EHR consent file, but there are some cases where that doesn't work well.
- Occasionally the questionnaire response linked to the consent file that passes validation is marked as a duplicate, and is then ignored when looking for EHR consent responses
- For any responses received before the ConsentResponse functionality was added, the link between the EHR QuestionnaireResponse and a validated consent PDF wouldn't be found.

This PR updates the check for validated files: also checking for duplicated responses and not checking for validation on older files.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/repository/questionnaire_response_repository.py
+++ b/rdr_service/repository/questionnaire_response_repository.py
@@ -167,7 +167,12 @@ class QuestionnaireResponseRepository:
                 code_constants.CONSENT_FOR_DVEHR_MODULE,
                 code_constants.CONSENT_FOR_ELECTRONIC_HEALTH_RECORDS_MODULE
             ],
-            participant_ids=[participant_id]
+            participant_ids=[participant_id],
+            classification_types=[
+                # The EHR response linked to a validated EHR file might be marked as duplicate of another response
+                enums.QuestionnaireResponseClassificationType.COMPLETE,
+                enums.QuestionnaireResponseClassificationType.DUPLICATE
+            ]
         ).get(participant_id)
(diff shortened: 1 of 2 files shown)


rdr_service/repository/questionnaire_response_repository.py
```diff
@@ -198,7 +203,12 @@ def get_interest_in_sharing_ehr_ranges(cls, participant_id, session: Session):
                 consent_answer = response.get_single_answer_for(code_constants.EHR_CONSENT_QUESTION_CODE)
                 if consent_answer:
                     # ignore any EHR responses that are not validated
-                    if response.id not in validated_ehr_id_list and not skip_validation_check:
+                    # Note: only check for validated EHR if the consent was authored after the response->consent data
```
pegbertsch, line 206: @kenny-skaggs FYI, and fixable with a `consent_response` backfill? I found 491 EHRConsentPII responses authored since this date that do not have a corresponding `consent_response` table record.  (Was not filtering on test pids or duplicate `questionnaire_response_id` classifications).

pegbertsch (approved):

rdr_service/repository/questionnaire_response_repository.py
kenny-skaggs: @pegbertsch, do you have the query handy for that? (just to give me a jump-start for looking into it)

PR #3441, opened 2023-04-18
Fixes removing and creating constraints when auto-generating nph schemas

j-kanuch:
Fixes bug when auto-generating NPH schema changes. Additionally, changed `BiobankFileExport` to align with current DB on Stable.

## Tests
- Current Tests should pass

### DIFF
--- a/rdr_service/model/study_nph.py
+++ b/rdr_service/model/study_nph.py
@@ -310,7 +310,7 @@ class BiobankFileExport(NphBase):
     id = Column("id", BigInteger, autoincrement=True, primary_key=True)
     created = Column(UTCDateTime)
     file_name = Column(String(256))
-    crc32c_checksum = Column(String(64), nullable=False)
+    crc32c_checksum = Column(String(64))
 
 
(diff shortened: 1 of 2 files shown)


rdr_service/model/study_nph.py
```diff
@@ -310,7 +310,7 @@ class BiobankFileExport(NphBase):
     id = Column("id", BigInteger, autoincrement=True, primary_key=True)
     created = Column(UTCDateTime)
     file_name = Column(String(256))
-    crc32c_checksum = Column(String(64), nullable=False)
+    crc32c_checksum = Column(String(64))
```
ayazamlani, line 313: @j-kanuch When changing this field in the model, does another alembic file need to be generated, or the old alembic file edited to match this change?

dtharpeuno (approved):

ayazamlani (approved):

rdr_service/model/study_nph.py
j-kanuch: Good question. The DB is out of sync with the model. It seems an alembic file was edited without editing the model. This change fixes that.

PR #3469, opened 2023-05-08
Ensuring a participant summary always has an email or login phone number

kenny-skaggs:
## Resolves *no ticket*
Participants are now able to clear their email from the participant summary through the ProfileUpdate API. But unless they already have a login phone number set up, this will cause later issues with the API. Logic is built in to the QuestionnaireResponse API to prevent a participant from having an empty login phone number and empty email address (https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/dao/questionnaire_response_dao.py#L1082).

This worked when QuestionnaireResponse was the only way to set these fields, but this check is missed when updating through the ProfileUpdate code.

This PR adds model level validation to be sure each ParticipantSummary has an email or login phone number.

## Tests
- [x] unit tests


ayazamlani (approved):

rdr_service/model/participant_summary.py
```diff
@@ -1648,6 +1648,15 @@ def enrollmentSiteId(cls):
 Index("participant_summary_login_phone_number", ParticipantSummary.loginPhoneNumber)
 
 
+def validate_participant_summary(_, __, summary: ParticipantSummary):
+    if not summary.email and not summary.loginPhoneNumber:
```
dtharpeuno, line 1652: Does this need to be an `or`?

dtharpeuno (approved):

rdr_service/model/participant_summary.py
kenny-skaggs: No. One of them can be empty, just not both.

PR #3473, opened 2023-05-10
Nph Intake DOB updates and NPH participant filtering with prefixes

dtharpeuno:
Need to update the NPH intake process for saving DOB when multiple are present, added prefix filtering for Biobank ID in NPH participant API

## Tests
- [x] unit tests

### DIFF
--- a/tests/api_tests/test_nph_participant_api.py
+++ b/tests/api_tests/test_nph_participant_api.py
@@ -380,12 +380,11 @@ class NphParticipantAPITest(BaseTestCase):
 
         result_participant_list = result.get('participant').get('edges')
         self.assertEqual(1, len(result_participant_list))
-        self.assertTrue(
-            result_participant_list[0].get('node').get('biobankId') == f'T{current_nph_participant.biobank_id}'
+        self.assertEqual(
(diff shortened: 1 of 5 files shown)


j-kanuch (approved):

tests/api_tests/test_nph_participant_api.py
```diff
@@ -366,6 +366,54 @@ def test_nph_dob_filter_parameter(self):
...
+
+        result_participant_list = result.get('participant').get('edges')
+        self.assertEqual(1, len(result_participant_list))
+        self.assertTrue(
+            result_participant_list[0].get('node').get('biobankId') == f'T{current_nph_participant.biobank_id}'
+        )
+        self.assertTrue(
+            result_participant_list[0].get('node').get('participantNphId') ==
+            str(current_nph_participant.id)
+        )
```
kenny-skaggs, line 388: What do you think of using `assertEqual` in cases like this instead? That way the values get printed out in a easy to debug way if the assertion fails.

dtharpeuno: Makes sense.
fc8d4f032bd320e1a6bcd8134e765c2d79a02380

kenny-skaggs (approved):

PR #3479, opened 2023-05-11
[DA-3543] Cron job to check for missing/unacknowledged DNA samples

pegbertsch:
Adds a cron job (production only) to alert if ordered DNA samples have not been acknowledged / had a `biobank_stored_sample` record created within a week since the order was finalized.   Will only look at orders finalized within the last month; does not continue to generate weekly alerts for older orders/missing samples.

Final determination of what internal/RDR-only slack channel to use for the slack alerts is still TBD, so the final step will be to make any slack changes and update the production configuration to add a `rdr_biobank_missing_samples_webhook` key/value to enable slack notifications.

## Tests
- [x] unit tests

[DA-3543]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/offline/biobank_samples_pipeline.py
+++ b/rdr_service/offline/biobank_samples_pipeline.py
@@ -17,9 +17,10 @@ from rdr_service import config
 from rdr_service.api_util import list_blobs
 from rdr_service.code_constants import PPI_SYSTEM, RACE_AIAN_CODE, RACE_QUESTION_CODE, WITHDRAWAL_CEREMONY_YES,\
     WITHDRAWAL_CEREMONY_NO, WITHDRAWAL_CEREMONY_QUESTION_CODE
-from rdr_service.config import BIOBANK_SAMPLES_DAILY_INVENTORY_FILE_PATTERN,\
-    BIOBANK_SAMPLES_MONTHLY_INVENTORY_FILE_PATTERN
+from rdr_service.config import BIOBANK_SAMPLES_DAILY_INVENTORY_FILE_PATTERN, \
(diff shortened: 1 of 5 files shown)


kenny-skaggs (approved):

rdr_service/offline/biobank_samples_pipeline.py
```diff
@@ -796,3 +798,52 @@ def in_past_n_days(result, now, n_days, ordered_before=None):
     if max_time:
         return (now - max_time).days <= n_days
     return False
+
+
+# DA-3543: Ordered DNA samples where order was finalized at least a week ago, but haven't been acknowledged by biobank
+# Intended for weekly check/alerts. Orders from more than 30 days ago will not be included in the alerts
+MISSING_STORED_SAMPLES_SQL = f"""
```
dtharpeuno: We try not to write much raw SQL can you convert this a SQLAlchemy method and add it to the dao itself?

dtharpeuno (changes requested):

rdr_service/offline/biobank_samples_pipeline.py
pegbertsch: I am following the raw SQL usage pattern in this `biobank_samples_pipeline` file.  Is that okay?

dtharpeuno: Yes I see it now. I understand your approach, since this pattern already exists here.

dtharpeuno (approved):

PR #3483, opened 2023-05-17
[PDR-1795] Fixes to legacy PDR enrollment status calculator

pegbertsch:
During a PDR investigation involving differences between the legacy enrollment status PDR and RDR calculations, several issues with the legacy PDR calculations were discovered:

1.  Some recently added EHR validation checks were returning SUBMITTED_NOT_VALIDATED incorrectly.  PDR filters out DUPLICATE questionnaire_responses,  but a DUPLICATE `questionnaire_response_id` can still be referenced in the `consent_file` / `consent_response` validation results.  Updating the generator to account for including DUPLICATE `questionnaire_response_id` values when querying for validation results.
2. PDR was using `confirmed` timestamps for biobank samples that were not DNA samples (but were part of the same order).  Revised logic to include check of test type.  This affects/resolves many mismatches between PDR calculator `enrl_core_participant_time` and RDR `enrollment_status_core_stored` timestamps.
3.  RDR has stopped downgrading from FULLY_CONSENTED (MEMBER) if an EHR consent is revoked or expired, but PDR would still prevent elevation to FULLY_CONSENTED (or above).  Updated PDR calculation to match RDR.  

## Tests
Generated a list of mismatched enrollment statuses from PDR with this query and ran hundreds of examples through the RDR `resource participant` tool with `--qc` option.  That acts as a dry run to be able to compare the PDR vs. RDR calculated values.
```
select participant_id from drc.pdr.mv_participant_all
where (enrollment_status_legacy_v2 != enrollment_status)
;
```

[PDR-1795]: https://precisionmedicineinitiative.atlassian.net/browse/PDR-1795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/resource/calculators/participant_enrollment_status.py
+++ b/rdr_service/resource/calculators/participant_enrollment_status.py
@@ -3,7 +3,7 @@
 # file 'LICENSE', which is part of this source code package.
 #
 import datetime
-
+#pylint: disable=unused-import
 from rdr_service.code_constants import (
     CONSENT_PERMISSION_YES_CODE,
     CONSENT_PERMISSION_NO_CODE,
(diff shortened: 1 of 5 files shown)


pegbertsch (commented):
I've left some old PDR calculator logic visible / commented out.  This will enable an easy way to recalculate the participant details using the old logic locally if necessary, for any data investigations that may involve comparing old PDR calculator vs. new logic.

robabram (approved):

kenny-skaggs (approved):

PR #3489, opened 2023-05-23
[PDR-1604] (Partial) Generate PDR pipeline pubsub delete events for d…

pegbertsch:
…uplicate responses

Updates the `ResponseDuplicationDetector.clean_pdr_data()` logic to now generate pipeline pubsub `delete` events for the records being marked as duplicate responses.

Also, during other pipeline development for [PDR-1803](https://precisionmedicineinitiative.atlassian.net/browse/PDR-1803), had observed some issues where consents that do have an explicit consent question should not default to SUBMITTED status if their consent answer value is missing from the response payload.  This default should only apply to consent modules that do not have an explicit consent question.

Updated a reference to SUBMITTED_INVALID that should have been changed to SUBMITTED_UNSET after the meaning of SUBMITTED_INVALID was redefined to refer to consent pdf validation status.

## Tests
- [x] unit tests 

[PDR-1604]: https://precisionmedicineinitiative.atlassian.net/browse/PDR-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PDR-1803]: https://precisionmedicineinitiative.atlassian.net/browse/PDR-1803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/resource/generators/participant.py
+++ b/rdr_service/resource/generators/participant.py
@@ -736,12 +736,13 @@ class ParticipantSummaryGenerator(generators.BaseGenerator):
                 if (module_name == 'EHRConsentPII'
                         and row.classificationType == QuestionnaireResponseClassificationType.PARTIAL):
                     continue
-                # Non-consent modules or modules with no explicit consent question configured default to SUBMITTED
-                if module_name not in _consent_module_question_map or not _consent_module_question_map[module_name]:
-                    module_status = BQModuleStatusEnum.SUBMITTED
-                else:
-                    # Consent modules with an explicit consent question start as UNSET
+
+                # Consent modules with a configured consent question start in UNSET status pending answer evaluation
+                if module_name in _consent_module_question_map and _consent_module_question_map[module_name]:
                     module_status = BQModuleStatusEnum.UNSET
+                else:
+                    # Other modules (PPI, consents like ConsentPII w/o explicit consent questions) default to SUBMITTED
+                    module_status = BQModuleStatusEnum.SUBMITTED
 
                 module_data = {
                     'module': module_name,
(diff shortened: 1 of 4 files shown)


rdr_service/resource/generators/participant.py
```diff
@@ -1662,7 +1666,7 @@ def _find_consent_response(self, response_rec: BQRecord, module: str):
                   consent_status:  BQModuleStatusEnum
         """
         # If a module doesn't have defined consent questions that determine status, it defaults to SUBMITTED status
-        if _consent_module_question_map[module] is None:
+        if module in _consent_module_question_map and _consent_module_question_map[module] is None:
```
kenny-skaggs, line 1659: Should this be `module not in _consent_module_question_map` (similar to above on line 740)?

pegbertsch: @kenny-skaggs This line was intended to be specific to modules that must be a key in the map (must be a recognized consent module), but have a None value because there is no explicit consent question (e.g., ConsentPII, ProgramUpdate).   On line 740, the initialization was intending to default any non-consent modules (e.g., PPI modules), or consent modules without a question code, to SUBMITTED.

But I'll revisit this to see if I can make this less confusing.   This`find_consent_response()`  function  did not exist originally, but was added in a prior refactor and appears to have overlapped with the logic on line 740.

kenny-skaggs: My mistake!

kenny-skaggs (approved):

rdr_service/resource/generators/participant.py
pegbertsch: @kenny-skaggs You pointed out some redundant logic that should have been  handled better in previous refactorings.  We're avoiding any major updates to this code and deferring to the new pipeline development, but I've made some minor changes that I hope will make the logic a little more clear.  Thanks!

robabram (approved):

PR #3498, opened 2023-06-06
[DA-3546] Adding core data flag to physical measurements

kenny-skaggs:
## Partially Resolves *[DA-3546](https://precisionmedicineinitiative.atlassian.net/browse/DA-3546)*
As part of the enrollment status 3.2 calculation, we need to determine if we have height and weight for a participant (or at least a modifier for height and weight). And we need to make that easily accessible to PDR.

This adds a new column to the physical measurements table that indicates whether we have a participant's height and weight (or modifications detailing the exception).

## Tests
- [x] unit tests


rdr_service/dao/physical_measurements_dao.py
```diff
@@ -974,3 +980,20 @@ def _restore_record(self, record, resource):
...
+    def update_measurement_core_data_flag(cls, measurement_collection: PhysicalMeasurements):
+        """
+        Analyzes the physical measurement data to determine if the Core Data flag should be set to True
+        """
+        has_height = False
+        has_weight = False
+        for measurement in measurement_collection.measurements:
+            # Measurement should count if there is a value recorded, or if there are modifications/qualifiers
+            is_valid_value = measurement.valueDecimal is not None or measurement.qualifiers
+            if measurement.codeValue == 'height':
```
pegbertsch: Believe we'll need to account for the `pre-pregnancy-weight` code here as well (counts as a valid weight?).  For older PM records (e.g., when backfilling the new column) , the` loinc` system codes `8302-2` (height) and `29463-7` (weight) may be the only ones in the `measurement` data instead.

kenny-skaggs: Thanks. I wasn't aware of those codes. I'll get it updated.

dtharpeuno (approved):

ayazamlani (approved):

PR #3500, opened 2023-06-07
Fix for finding consent data for retention calculation

kenny-skaggs:
## Resolves *no ticket*
The retention calculation import code is currently crashing when trying to get consent information for participants. This fixes bugs in the process so it can continue to work again.

## Tests
- [ ] unit tests

### DIFF
--- a/rdr_service/offline/retention_eligible_import.py
+++ b/rdr_service/offline/retention_eligible_import.py
@@ -314,7 +314,10 @@ def _get_earliest_intent_for_ehr_datetime(session, participant_id) -> datetime:
     if not date_range_list:
         return None
 
-    return min(date_range.start for date_range in date_range_list)
+    return Consent(
+        is_consent_provided=True,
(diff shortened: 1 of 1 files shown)


dtharpeuno (approved):

rdr_service/offline/retention_eligible_import.py
```diff
@@ -314,7 +314,10 @@ def _get_earliest_intent_for_ehr_datetime(session, participant_id) -> datetime:
     if not date_range_list:
         return None
 
-    return min(date_range.start for date_range in date_range_list)
+    return Consent(
```
brianmkimmel, line 317: The return type hint for this method should match the type being returned.

brianmkimmel (approved):

rdr_service/offline/retention_eligible_import.py
kenny-skaggs: Good catch, thanks!

PR #3501, opened 2023-06-07
DA-3608 NPH API latest DOB

dtharpeuno:
## Resolves *[ticket DA-3608]*
https://precisionmedicineinitiative.atlassian.net/browse/DA-3608

Need to add filtering on ops data elements since we are having to contend with multiple data records

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/api/nph_participant_api_schemas/schema.py
+++ b/rdr_service/api/nph_participant_api_schemas/schema.py
@@ -383,7 +383,7 @@ class ParticipantQuery(ObjectType):
         **filter_kwargs
     ):
         pm2 = aliased(PairingEvent)
-        participant_date_of_birth = aliased(ParticipantOpsDataElement)
+        participant_dob = aliased(ParticipantOpsDataElement)
 
         nph_participant_dao = NphParticipantDao()
         consent_subquery = nph_participant_dao.get_consents_subquery()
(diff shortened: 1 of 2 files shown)


rdr_service/api/nph_participant_api_schemas/schema.py
```diff
@@ -433,6 +435,15 @@ def resolve_participant(
...
                 ParticipantMapping.ancillary_participant_id == ParticipantOpsDataElement.participant_id
+            ).outerjoin(
+                participant_date_of_birth,
+                and_(
+                    ParticipantOpsDataElement.participant_id == participant_date_of_birth.participant_id,
+                    ParticipantOpsDataElement.source_data_element == ParticipantOpsElementTypes.BIRTHDATE,
+                    ParticipantOpsDataElement.source_value.isnot(None),
+                    ParticipantOpsDataElement.ignore_flag != 1,
+                    ParticipantOpsDataElement.id < participant_date_of_birth.id
+                )
```
kenny-skaggs, line 450: Is the ParticipantOpsElementTypes table going to store other elements for a participant, ones that wouldn't overwrite a DOB? If I'm reading this correctly, the DOB record would left join to something else that has a higher id and then would have the whole participant result filtered out.

dtharpeuno: It doesn't now but It theoretically can in the future. What are you suggesting? Move the constraints to the initial join?

kenny-skaggs: Yeah, maybe as a subquery or something like that (select the latest DOB record and then join to that).

kenny-skaggs (approved):

j-kanuch (approved):

PR #3505, opened 2023-06-08
[DA-3546] Updating enrollment status calculation to 3.2

kenny-skaggs:
## Partially Resolves *[DA-3546](https://precisionmedicineinitiative.atlassian.net/browse/DA-3546)*
This updates the code to calculate enrollment status 3.2 (replacing 3.1). That involves:
- Replacing all definitions/references of 3.1 with 3.2
- Updating the calculation with changes for 3.2 (renaming a status, removing BASELINE)
- Adding a separate Core Data flag (that uses data from the WGS AW4 and the new physical measurements field indicating that we have height and weight)

## Tests
- [x] unit tests


rdr_service/dao/participant_summary_dao.py
```diff
@@ -725,6 +735,20 @@ def update_enrollment_status(self, summary: ParticipantSummary, session, allow_d
...
+            genome_type=config.GENOME_TYPE_WGS
+        )
+        if genomic_set_member and genomic_set_member.qcStatus == GenomicQcStatus.PASS:
+            # Determine when the job ran to get the rough date of the pass
+            aw4_job_run: GenomicJobRun = GenomicJobRunDao.get_with_session(
+                session=session,
+                obj_id=genomic_set_member.aw4ManifestJobRunID
+            )
+            wgs_sequencing_time = aw4_job_run.startTime
+
```
dtharpeuno: With this I think it might easier to build a util method in the genomic dao with the joins already instead of having to do multiple db calls here.

kenny-skaggs: Sounds good to me. Which dao should I use: GenomicSetMemberDao, GenomicJobRunDao, or another one?

dtharpeuno: Since it deals with a member obj, I would use `GenomicSetMemberDao `.

kenny-skaggs: Sounds good. Updated in latest commit. Let me know what you think.

ayazamlani (approved):

dtharpeuno (approved):

PR #3506, opened 2023-06-08
[DA-3546] Updating enrollment status based on aw4

kenny-skaggs:
## Resolves *[DA-3546](https://precisionmedicineinitiative.atlassian.net/browse/DA-3546)*
Changes in another PR are introducing a need to update the enrollment status (specifically the Core Data flag) when we receive an AW4 WGS for participants. This adds a dispatch to have the enrollment status updated when we receive an AW4.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -31,7 +31,6 @@ from rdr_service.api_util import (
     open_cloud_file,
     copy_cloud_file,
     delete_cloud_file,
-    dispatch_task,
     list_blobs,
     get_blob
 )
@@ -550,7 +549,7 @@ class GenomicFileIngester:
 
                 self.member_dao.update(member)
 
-                dispatch_task(
+                self.controller.execute_cloud_task(
                     endpoint='update_enrollment_status',
                     payload={
                         'participant_id': member.participantId
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -553,7 +553,8 @@ class GenomicFileIngester:
                     endpoint='update_enrollment_status',
                     payload={
                         'participant_id': member.participantId
-                    }
+                    },
+                    task_queue='resource-tasks'
                 )
 
             return GenomicSubProcessResult.SUCCESS
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -31,8 +31,10 @@ from rdr_service.api_util import (
     open_cloud_file,
     copy_cloud_file,
     delete_cloud_file,
+    dispatch_task,
     list_blobs,
-    get_blob)
+    get_blob
+)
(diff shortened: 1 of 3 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -548,6 +550,13 @@ def _ingest_aw4_manifest(self, rows):
 
                 self.member_dao.update(member)
 
+                dispatch_task(
```
j-kanuch: The AW4 job runs when either an Array or a WGS AW4 comes through. The AW4 itself will be for one of these two genome types. Do you want to update the enrollment status on either the Array or WGS, or do you only want to update it if the WGS sample passes QA. If the latter is the case, you'll want to make sure the `self.job_id == GenomicJob.AW4_WGS_WORKFLOW`

rdr_service/genomic/genomic_job_components.py
```diff
@@ -548,6 +550,13 @@ def _ingest_aw4_manifest(self, rows):
...
+                    endpoint='update_enrollment_status',
+                    payload={
+                        'participant_id': member.participantId
+                    }
+                )
```
dtharpeuno: Sorry, this works but would you mind changing this to the one that is attached to the controller for consistency in the genomics system. You will just need to pass the queue that this is defaulting to. 

`self.controller.execute_cloud_task`

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: My thinking was that I didn't want to put duplicate business logic here for when the task should be fired, but instead have the event triggered and the calculation code determine what it reads for the calculation. That way, if something changes later, it's simple enough to remember that an AW4 triggers the calculation and the calculation is the only thing that needs to change. Rather than needing to remember that the AW4 only triggers the calculation if it's a WGS, and then both the calculation code and the triggering code would need to be updated.

But I know that's debatable, so if you want it changed then I can go ahead and make the update.

j-kanuch: Sounds good, I didn't know the calculation handled that.

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: No worries, thanks for letting me know about the method. Updated!

dtharpeuno: Hey last thing, this method defaults to the `genomics` queue. You will need to update this to be in the `resource-tasks` queue if you want it there, but I dont think the queue really matter for which endpoints can be accessed from it.

dtharpeuno (approved):

rdr_service/genomic/genomic_job_components.py
kenny-skaggs: Definitely meant to have that go to another queue. It seems cleaner to leave the genomics queue as genomics tasks. Thanks for catching that.

j-kanuch (approved):

PR #3509, opened 2023-06-08
[DA-3546] Storing a snapshot of data used for finding enrollment statuses

kenny-skaggs:
## Resolves *[DA-3546](https://precisionmedicineinitiative.atlassian.net/browse/DA-3546)*
This stores a JSON snapshot in the enrollment status history table of all the data used to upgrade a participant's status. This should help when investigating issues with enrollment status.

## Tests
- [ ] unit tests


rdr_service/dao/participant_summary_dao.py
```diff
@@ -739,24 +739,23 @@ def update_enrollment_status(self, summary: ParticipantSummary, session, allow_d
...
+            basics_authored_time=summary.questionnaireOnTheBasicsAuthored,
+            overall_health_authored_time=summary.questionnaireOnOverallHealthAuthored,
+            lifestyle_authored_time=summary.questionnaireOnLifestyleAuthored,
+            earliest_ehr_file_received_time=summary.ehrReceiptTime,
+            earliest_mediated_ehr_receipt_time=summary.firstParticipantMediatedEhrReceiptTime,
+            earliest_physical_measurements_time=earliest_physical_measurements_time,
+            earliest_biobank_received_dna_time=earliest_biobank_received_dna_time,
+            ehr_consent_date_range_list=ehr_consent_ranges,
+            dna_update_time=revised_consent_time,
+            earliest_core_physical_measurement_time=min_or_none(meas.finalized for meas in core_measurements),
```
pegbertsch, line 755: I think this can potentially be addressed in a separate PR if it's an actual issue.  I assume core_measurements is the required height/weight values, but I'm wondering how this works if there are cases where the height and weight values were submitted in different `physical_measurements` payloads with different finalized time (e.g., a remote PM rec with just a height followed later by an in-clinic PM rec with a weight as well).  Does it need to be the max of those two finalized dates?

pegbertsch (approved):
Approved but please note question on `earliest_core_physial_measurement_time` in case it needs to be revisited later

rdr_service/dao/participant_summary_dao.py
kenny-skaggs: I guess I was imagining that the height and weight would need to be measured at the same time (in the same measurement), but that's a good point. I'll ping Mara in the ticket.

dtharpeuno (approved):

kenny-skaggs:
> Approved but please note question on `earliest_core_physial_measurement_time` in case it needs to be revisited later

Brought this up in the ticket. Will merge this for now and follow up with Mara.

PR #3511, opened 2023-06-14
Fixing bug when determining core data validity of physical measurements

kenny-skaggs:
## Resolves *no ticket*
When a physical measurement has height and/or weight qualifiers, the resulting value could unintentionally be a list rather than a boolean value.

This adds an explicit cast to give a boolean based on the list rather than the list itself.

## Tests
- [x] unit tests


pegbertsch (approved):

dtharpeuno (approved):
Thanks for fixing this

dtharpeuno (approved):

PR #3512, opened 2023-06-15
Refactor/optimization of retention import and calculation

kenny-skaggs:
## Resolves
This refactors and optimizes the retention data process. The code looks through the records of the retention data being imported, and updates the retention metrics objects that need changes. When adding code for calculating the retention data internally, I had set up the code to calculate the values for every participant in the file. The files appear to list all participants, so this process will need to be updated to only recalculate as needed.

This refactors the code that checks if the retention data matches what we have in the database. The upsert method will now act on data (inserting or updating it), the check for whether the upsert should happen was moved outside the upsert method and can be used now to determine if the internal calculations should be run.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/retention_eligible_metrics_dao.py
+++ b/rdr_service/dao/retention_eligible_metrics_dao.py
@@ -1,11 +1,9 @@
 
 from sqlalchemy.orm import Session
 
-from rdr_service.config import GAE_PROJECT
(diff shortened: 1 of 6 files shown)


ayazamlani (approved):
this is cool. I like how you are using the csv_reader to build out the files. These comments are really helpful. Thanks!

brianmkimmel (approved):

PR #3531, opened 2023-07-06
[DA-3670] Move consent error reporting to its own cron job

pegbertsch:
Now that the consent validation cron job is running hourly, the consent error reporting tasks will be split out into their own cron job that only runs once a day.  It is only intended to run in production.

## Tests
- [x] unit tests

[DA-3670]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/cron_job_tests/test_offline_app_setup.py
+++ b/tests/cron_job_tests/test_offline_app_setup.py
@@ -1,6 +1,6 @@
 import http.client
 import mock
-
+from unittest.mock import call
 from rdr_service.services.gcp_config import RdrEnvironment
 from tests.helpers.unittest_base import BaseTestCase
 
@@ -83,7 +83,6 @@ class OfflineAppTest(BaseTestCase):
 
     @mock.patch('rdr_service.offline.main.dispatch_check_consent_errors_task')
     def test_consent_error_report_route(self, mock_checker):
-        from unittest.mock import call
         # Do not expect the reporting tasks to be dispatched except when on PROD
         self.send_cron_request('ConsentErrorReport')
         mock_checker.assert_not_called()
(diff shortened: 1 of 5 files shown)


ayazamlani (approved):

tests/cron_job_tests/test_offline_app_setup.py
```diff
@@ -79,3 +80,18 @@ def test_genomics_config_disables_jobs(self, pipeline_mock):
     def test_biobank_missing_samples_check_route(self, mock_checker):
         self.send_cron_request('BiobankMissingSamplesCheck')
         mock_checker.assert_called()
+
+    @mock.patch('rdr_service.offline.main.dispatch_check_consent_errors_task')
+    def test_consent_error_report_route(self, mock_checker):
+        from unittest.mock import call
```
kenny-skaggs: Does this work if the import is at the top of the file? (mostly just curious why the import is in the method)

pegbertsch: I'd seen other instances where narrowly used imports were local to the specific place where they were needed.  I can move it if you want.

@kenny-skaggs Pushed the change

kenny-skaggs: I can't speak for all the code, but I think when I've done an import in the method it was only when needed (to avoid a circular dependency when trying to load files). Otherwise it seems to be a best practice to have the imports at the top (at the very least for having a standard place for knowing what dependencies a file has).

Alternatively, you could use `mock.call` to take advantage of something already present.

kenny-skaggs (approved):

tests/cron_job_tests/test_offline_app_setup.py
pegbertsch: Resolved in latest commit

PR #3542, opened 2023-07-13
[DA-3678] Handle SendGrid errors for GCR Outreach notification emails

brianmkimmel:
Adds a try/except block around the SendGrid api call so succesfully sent emails can be saved to the database.

Adds a slack notification if any attempts to send an email raise an error.

## Tests
- [x] unit tests

[DA-3678]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/genomic/genomic_job_controller.py
```diff
@@ -1968,22 +1968,35 @@ def check_gcr_14day_escalation(self):
...
 
             notified_dao.insert_bulk(notified_participants)
-            self.job_result = GenomicSubProcessResult.SUCCESS
+            if error_pids:
+                self.genomic_alert_slack.send_message_to_webhook(
+                    message_data={'text': f"GCR Outreach Escalation email had errors for PIDs: {error_pids}"}
+                )
+                self.job_result = GenomicSubProcessResult.ERROR
+            else:
+                self.job_result = GenomicSubProcessResult.SUCCESS
```
dtharpeuno: I think a different approach for this would be better. Since you are already writing the records and pids for this to a table maybe we can add a message_sent bool to the table to denote that the API request for Sendgrid has been successful and if it wasn't tag it with False, so that way we have a way to keep track of the pids that were not sent to Color on the escalation emails instead of keeping them in slack. 

Also I would edit the slack message to say that the Scheduled escalation email was unsuccessful with sending the payload of the pids since you already have them in db now.

rdr_service/genomic/genomic_job_controller.py
```diff
@@ -1968,22 +1968,42 @@ def check_gcr_14day_escalation(self):
...
```
dtharpeuno: :nit I think we should not send the pids list, since this can be a indeterminate amount and could be longer than the Slack API lets us send. Maybe a more general message would work here just to denote the failure of the sending.

rdr_service/alembic/versions/e31a8f514602_add_message_sent_column_to_genomic_gcr_.py
```diff
@@ -0,0 +1,51 @@
...
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+def upgrade_rdr():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('genomic_gcr_outreach_escalation_notified', sa.Column('message_sent', sa.Boolean(), nullable=True))
+    op.execute("UPDATE genomic_gcr_outreach_escalation_notified SET message_sent=TRUE")
```
dtharpeuno: Can we make this a smallInt similar to how we handle the ignore flags?

ayazamlani (approved):

j-kanuch (approved):

PR #3555, opened 2023-07-26
Adding participant id column to profile update archive

kenny-skaggs:
## Resolves *no ticket*
More frequently we're needing to look through profile updates for a particular participant. To make that easier, this adds a participant id column to the table that archives the profile update json that we receive.

## Tests
- [x] unit tests


robabram:
Hi Kenny, I noticed that the JSON payload column in this table is named 'json' instead of 'resource', is there any interest in renaming the column to be consistent with the other DB tables and rename the column to 'resource'?  I find it helpful knowing the JSON payload is named 'resource' on tables.

kenny-skaggs:
> Hi Kenny, I noticed that the JSON payload column in this table is named 'json' instead of 'resource', is there any interest in renaming the column to be consistent with the other DB tables and rename the column to 'resource'? I find it helpful knowing the JSON payload is named 'resource' on tables.

I think that could be good as a best-practice moving forward. But I don't think we have plans in place to update it here (or in the few other tables that it's set as JSON). But if you want to suggest it to the RDR product team they could add it to the backlog to be prioritized in.

ayazamlani (approved):

j-kanuch (approved):

PR #3607, opened 2023-09-13
Reverting Remote ID Verification fields back to a Boolean from Strings

ayazamlani:
Reverting the following fields back to a boolean from a string for Remote ID Verification

    - 'remoteIdVerificationStatus'
    - 'everIdVerified'

## Tests
- [] unit tests

[DA-3808]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/dao/participant_summary_dao.py
+++ b/rdr_service/dao/participant_summary_dao.py
@@ -1271,9 +1271,9 @@ class ParticipantSummaryDao(UpdatableDao):
             if result.get(field_name) is None:
                 result[field_name] = UNSET
             elif result[field_name]:
-                result[field_name] = 'True'
+                result[field_name] = True
             else:
-                result[field_name] = 'False'
+                result[field_name] = False
 
         # Map demographic Enums if TheBasics was submitted and Skip wasn't in use
         if is_the_basics_complete and not should_clear_fields_for_withdrawal:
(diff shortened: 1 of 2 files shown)


rdr_service/dao/participant_summary_dao.py
```diff
@@ -1271,9 +1271,9 @@ def build_default_obj_response(self, obj: ParticipantSummary):
             if result.get(field_name) is None:
                 result[field_name] = UNSET
             elif result[field_name]:
-                result[field_name] = 'True'
+                result[field_name] = True
             else:
-                result[field_name] = 'False'
+                result[field_name] = False
```
kenny-skaggs: These are coming from the database as booleans, right? It seems like it'd be good to keep it simple by removing the elif and else, just changing the value if it's None. What do you think?

ayazamlani: Yes, @kenny-skaggs  you are right. That makes complete sense. I am removing this redundant code. Thanks 🍻

kenny-skaggs (approved):

pegbertsch (approved):

PR #3630, opened 2023-09-28
[DA-3782] API result sorting for Environmental Health survey

kenny-skaggs:
## Resolves *[DA-3782](https://precisionmedicineinitiative.atlassian.net/browse/DA-3782)*
To allow Hpro to sort participant summaries by the environmental health fields, we'd need to be able to modify the query to sort by the related entries in the pediatric data. This adds functionality to detect when sorting by the Environmental health fields and add the order_by information to the query.

## Tests
- [x] unit tests

[DA-3782]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/dao/participant_summary_dao.py
```diff
@@ -542,14 +548,27 @@ def _add_order_by_state(order_by, query):
...
+
     def _make_query(self, session, query_definition):
         query, order_by_field_names = super(ParticipantSummaryDao, self)._make_query(session, query_definition)
         # Note: leaving for future use if we go back to using a relationship to PatientStatus table.
         # query.options(selectinload(ParticipantSummary.patientStatus))
         # sql = self.query_to_text(query)
 
         if not query_definition.attributes:  # temporarily skip any joinloads if using result field filters
-            query.options(self._default_api_query_options())
+            query = query.options(*self._default_api_query_options())
```
dtharpeuno, line 571: Was this not working before since the added assignment variable? Also can you explain the usage of the asterisk in with the parameter?

kenny-skaggs: Yeah, it wasn't working without the assignment. I missed testing the list endpoint functionality for this until looking for the env health fields.

`_default_api_query_options` returns a list, and the asterisk expands the list so that each element is passed in as a separate parameter. It looks like the `options` method can take a list and do the same anyway. So the change was mostly left over from me playing around with it and seeing whether there was a difference.

dtharpeuno: Nice thanks for explaining.

dtharpeuno (approved):

ayazamlani (approved):

PR #3631, opened 2023-09-28
[DA-3777] Add fields to participant_summary for enrollment status and retention eligibility details

pegbertsch:
Fields added to participant_summary to show activity relevant to enrollment status and retention:
- WEAR consent activity (active retention qualifying activity)
- Availability of valid height and weight measurements (related to Goal 1 `has_core_data` status calculation)
- Latest EtM task activity (active retention qualifying activity)

The auto-generated alembic migration (individual `add_column` ops) was replaced with a raw SQL ALTER TABLE since it may help minimize migration time.  This may have had some side effects like whether the timestamps are considered "naive" in Python code when used in comparison operations.

Needed to reorder some logic in handling of `physical_measurements` data ingestion including when data was committed, so that the queries executed during enrollment status calculation  (during the POST request path) would see the new PM data.

Fixed up the EtM task payload test data to simplify expected authored time comparisons, since the test data `authored` value originally had milliseconds (>500) that impacted how it was rounded up/stored in the DATETIME column.  Confirmed that no existing production EtM authored timestamps contain milliseconds.

## Tests
- [x] unit tests

[DA-3777]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -922,6 +923,20 @@ def _update_participant_summary(
...
+                            participant_summary.consentForWearStudyAuthored = authored
+                            if answer_value.lower() == WEAR_YES_ANSWER_CODE:
+                                participant_summary.consentForWearStudy = QuestionnaireStatus.SUBMITTED
+                            elif answer_value.lower() == WEAR_NO_ANSWER_CODE:
+                                participant_summary.consentForWearStudy = QuestionnaireStatus.SUBMITTED_NO_CONSENT
+                            else:
+                                logging.warning(
+                                    f'Unrecognized/no answer value for WEAR consent question: {answer_value} ',
+                                )
+                                participant_summary.consentForWearStudy = QuestionnaireStatus.UNSET
```
kenny-skaggs, line 939: In this case, should we set the consent status to SUBMITTED_INVALID instead?

pegbertsch: @kenny-skaggs SUBMITTED_INVALID can now imply a problem with consent PDFs (even if it's only used for EHR right now).  When that was use case was introduced, the PDR generators started using UNSET if we couldn't figure out how a participant responded to a consent question.  Do you think it's preferable to leave it null?

kenny-skaggs: I still think an INVALID would be good in this case. But this works too.

kenny-skaggs (approved):

dtharpeuno (approved):
Godspeed!

j-kanuch (approved):

rdr_service/dao/questionnaire_response_dao.py
pegbertsch: I thought that our data dictionary/reference documentation now defined SUBMITTED_INVALID as "did not pass PDF validation", but maybe it can be a broader error indication.

PR #3642, opened 2023-10-04
Updating last modified for remote idv

ayazamlani:
updating the last modified when remote IDV updates are made so PTSC can see the changes on their end.

## Tests
- [] unit tests

### DIFF
--- a/rdr_service/dao/questionnaire_response_dao.py
+++ b/rdr_service/dao/questionnaire_response_dao.py
@@ -1018,6 +1018,7 @@ class QuestionnaireResponseDao(BaseDao):
                 participant_summary.remoteIdVerificationOrigin = participant_summary.participantOrigin
                 participant_summary.remoteIdVerificationStatus = False
                 participant_summary.remoteIdVerifiedOn = None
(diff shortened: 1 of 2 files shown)


rdr_service/dao/questionnaire_response_dao.py
```diff
@@ -1018,6 +1018,7 @@ def _update_participant_summary(
                 participant_summary.remoteIdVerificationOrigin = participant_summary.participantOrigin
                 participant_summary.remoteIdVerificationStatus = False
                 participant_summary.remoteIdVerifiedOn = None
+            participant_summary.lastModified = datetime.now()
```
shahwaiz14: I think lastModified should be in UTC, so maybe we can use clock.CLOCK.now()...that way we are also consistent with how lastModified is changed throughout this file 😃

ayazamlani: thanks!

kenny-skaggs (approved):

shahwaiz14 (approved):

PR #3649, opened 2023-10-09
[DA-3892][DA-3891] Fixes for issues discovered after deploy of RDR 1.157.1 release to prod

pegbertsch:
Remediations for issues that were observed after the RDR 1.157.1 release was deployed to prod:

- Comparisons between the existing `consentForWearStudyAuthored` value in `participant_summary` and the authored timestamp from a subsequent WEAR consent payload were raising an exception because of mishandling the timezone aware vs. naive timestamp discrepancy.  Fixed the code to use the proper `authored` value from the payload that had already had the tzinfo removed up on lines 744-748 of `_update_participant_summary()` 
-  Added a new unit test case to reproduce the exception before fix was applied, then confirm fix resolved the bug
- Updated the timestamp comparison handling in EtM response POST handler to use the same logic, for consistency
-  Adjusted the query in PDR generator that tried to filter on `biobank_order.ignore_flag`, in case any null values exist
-  Added workaround to the PDR generator when it tries to figure out the EHR PDF consent validation status for a given  EHR consent response, to account for potential manual overrides of the `consentForElectronicHealthRecords` status in RDR `participant_summary` (to resolve discrepancies in PDR vs. pre-Goal 1 RDR enrollment status values).  
- Adjusted PDR generator and unit tests to use the ENROLLMENT_STATUS_SKIP_VALIDATION config setting as appropriate.

## Tests
- [x] unit tests

[DA-3892]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/resource/generators/participant.py
```diff
@@ -1892,52 +1893,68 @@ def get_consent_pdf_validation_status(p_id, consent_response_rec,
...
-                select questionnaire_response_id from questionnaire_response
+        if module_name == 'EHRConsentPII':
+            if not ro_session:
+                dao = ResourceDataDao(backup=True)
+                ro_session = dao.session()
+
+            # Check for RDR overrides to SUBMITTED_NOT_VALIDATED/SUBMITTED_INVALID, which may have been manually
+            # applied to participant_summary to allow participants to continue to PM&B.   If the consent_response
+            # being checked matches what's in participant_summary on authored date, use participant_summary status
+            ps_sql = f"""
```
pegbertsch, line 1907: FYI:  The code has been reorganized so the diff appears more extensive than it is.  The main change is inserting this new `ps_sql` query and potential early return if this check determines we should use the `consentForElectronicHealthRecords` status value from the `participant_summary` record, instead of forcing a match on a  `consent_file` validation record that indicates successful validation.

robabram (approved):

kenny-skaggs (approved):

PR #3650, opened 2023-10-10
[DA-3840] Surface new NPH consent info in NPH participant API

shahwaiz14:

Update consents subquery to fetch only the most recent event types for each participant.

(note: mysql5.7 does not support window functions, so I had to write subqueries to achieve the same result)

## Tests
- ☑️ unit tests


j-kanuch (approved):

rdr_service/dao/study_nph_dao.py
```diff
@@ -73,26 +73,46 @@ def from_client_json(self):
         pass
 
     def get_consents_subquery(self):
+        """
+        Returns a subquery that gets the latest consents for each participant grouped by event_type_id.
+        """
+        ce2 = aliased(ConsentEvent)
```
dtharpeuno: :nit Can we rename this variable to something like `consent_event_alias`?

rdr_service/dao/study_nph_dao.py
```diff
@@ -73,26 +73,46 @@ def from_client_json(self):
...
+                    ConsentEventType.id == ConsentEvent.event_type_id,
+                )
+                .outerjoin(
+                    ce2,
+                    and_(
+                        ce2.participant_id == ConsentEvent.participant_id,
+                        ce2.event_type_id == ConsentEvent.event_type_id,
+                        ce2.id > ConsentEvent.id,
+                    ),
+                )
```
dtharpeuno: :nit For better readability that will mimic the rest of these types of outerjoins in this dao, can we move the alias to the left side of the assignment?
ex: `ConsentEvent.participant_id == ce2.participant_id`

tests/api_tests/test_nph_participant_api.py
```diff
@@ -951,6 +990,7 @@ def tearDown(self):
         self.clear_table_after_test("nph.enrollment_event_type")
         self.clear_table_after_test("nph.participant_ops_data_element")
         self.clear_table_after_test("nph.consent_event")
+        self.clear_table_after_test("nph.consent_event_type")
```
dtharpeuno, line 993: Does this need to be here? I dont see this data changing from test to test?

shahwaiz14: the data does not change, but the setUp methods keep adding the same data to the consent_event_type table after every test function. So to avoid having duplicate event types in that table, we added it here.

dtharpeuno: Sounds good

dtharpeuno (approved):

PR #3667, opened 2023-10-18
[DA 3796] Implement Site Hierarchy API

shahwaiz14:
- Implemented API for site hierarchy according to this [tech design document](https://docs.google.com/document/d/1nhTgYkehrjp7uzsMEm7mj2Ccln9dNWHDJ1wzOztiyPE/edit)

## Tests
- ✅  unit tests

### DIFF
--- a/rdr_service/api/site_hierarchy_api.py
+++ b/rdr_service/api/site_hierarchy_api.py
@@ -3,19 +3,19 @@ from flask import request
 from rdr_service import app_util
 from rdr_service.api.base_api import BaseApi
 from rdr_service.dao.site_hierarchy_dao import SiteHierarchyDao
-from rdr_service.api_util import CURATION
+from rdr_service.api_util import CURATION, RDR
 
 
 class SiteHierarchyApi(BaseApi):
     def __init__(self):
         super().__init__(SiteHierarchyDao())
 
-    @app_util.auth_required(CURATION)
-    def get(self, id_=None, p_id=None):
+    @app_util.auth_required([CURATION, RDR])
+    def get(self, id_=None, participant_id=None):
         """This will call the list method below"""
-        return super().get(id_=id_, participant_id=p_id)
+        return super().get(id_=id_, participant_id=participant_id)
 
-    @app_util.auth_required(CURATION)
+    @app_util.auth_required([CURATION, RDR])
     def list(self, participant_id=None):
         """Handle a GET request."""
         kwargs = request.args
--- a/rdr_service/api/site_hierarchy_api.py
+++ b/rdr_service/api/site_hierarchy_api.py
@@ -1,0 +1,14 @@
+from flask import request
+
+from rdr_service.api.base_api import BaseApi
+from rdr_service.dao.site_hierarchy_dao import SiteHierarchyDao
+
+
+class SiteHierarchyApi(BaseApi):
+    def __init__(self):
+        super().__init__(SiteHierarchyDao())
+
+    def list(self, participant_id=None):
+        """Handle a GET request."""
+        kwargs = request.args
+        return SiteHierarchyDao().handle_list_queries(**kwargs)
--- a/rdr_service/api/site_hierarchy_api.py
+++ b/rdr_service/api/site_hierarchy_api.py
@@ -1,13 +1,21 @@
 from flask import request
 
+from rdr_service import app_util
 from rdr_service.api.base_api import BaseApi
 from rdr_service.dao.site_hierarchy_dao import SiteHierarchyDao
+from rdr_service.api_util import CURATION
 
 
 class SiteHierarchyApi(BaseApi):
     def __init__(self):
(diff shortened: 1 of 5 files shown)


rdr_service/api/site_hierarchy_api.py
```diff
@@ -0,0 +1,14 @@
...
+
+from rdr_service.api.base_api import BaseApi
+from rdr_service.dao.site_hierarchy_dao import SiteHierarchyDao
+
+
+class SiteHierarchyApi(BaseApi):
+    def __init__(self):
+        super().__init__(SiteHierarchyDao())
+
+    def list(self, participant_id=None):
```
kenny-skaggs, line 14: Do we know who access for this should be restricted to? We'd want to use a decorator here to make sure any requests are authenticated. And maybe the role should be curation, but I'm not sure if the EHR ops team is using the same service account.

`BaseApi`'s `get` method may also need to be overridden just to add in a auth check there too.

shahwaiz14: Makes sense. Yes, I think this will mainly be called by curation.

shahwaiz14: @kenny-skaggs I added the `get` method in there,  but I think it's redundant since it calls the `list` method, and the list method is protected with `@app_util.auth_required([...])`.

kenny-skaggs: Yeah, looking at it closer you're right. I wasn't quite sure on how that fit together.

kenny-skaggs (approved):

rdr_service/dao/site_hierarchy_dao.py
```diff
@@ -0,0 +1,83 @@
...
+        response = {"data": [model._asdict() for model in result]}
+        return jsonify(response)
+
+    def _validate_query_params(self, **kwargs: Dict[str, str]):
+        for key, _ in kwargs.items():
+            if key not in self._QUERY_PARAM_MAPPING:
+                raise BadRequest(
+                    f"Invalid query parameter(s). "
+                    f"Supported parameters are: {', '.join(self._QUERY_PARAM_MAPPING.keys())}"
+                )
```
dtharpeuno: Since we are not utilizing the secondary arg in items can we just change this to keys?

tests/api_tests/test_site_hierarchy_api.py
```diff
@@ -0,0 +1,178 @@
...
+        self.site_dao.insert(
+            Site(
+                siteId=OBSOLETE_ID,
+                siteName="Illinois Site 2",
+                googleGroup="illinois-site-2",
+                organizationId=ILLINOIS_ORG_ID,
+                hpoId=ILLINOIS_HPO_ID,
+                isObsolete=ObsoleteStatus.OBSOLETE,
+            )
+        )
```
dtharpeuno: So in the `BaseTestCase` we have a instantiation to the `DataGenerator` class which we use for data generation mainly in the unit tests. There is a method for site that you can utilize in that class that will help clean some of these unneeded instance variables.

shahwaiz14: Done!

tests/api_tests/test_site_hierarchy_api.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def test_empty_response_no_records(self):
+        awardee_id = "no_awardee"
+        expected_response = _make_expected_response(awardee_id=awardee_id)
+        response = self.send_get(f"SiteHierarchy?awardee_id={awardee_id}")
+        self.assertEqual(
+            expected_response,
+            response,
+            msg="Response structure does not match expected structure. Response should be empty.",
+        )
```
dtharpeuno, line 163: Can we also add a test for the validation that was included in the API, for testing against a bad key being sent in the params?

rdr_service/dao/site_hierarchy_dao.py
```diff
@@ -0,0 +1,84 @@
...
+from rdr_service.model.hpo import HPO
+from rdr_service.model.site_enums import ObsoleteStatus
+
+
+class SiteHierarchyDao(BaseDao):
+    _QUERY_PARAM_MAPPING = {
+        "awardee_id": {"col": HPO.name},
+        "organization_id": {"col": Organization.externalId},
+        "google_group": {"col": Site.googleGroup},
+    }
```
dtharpeuno: :nit Can we also flatten the values in this map? If you had more than one key in the dictionaries I would understand but I think it would be better to default it to the model attribute instead to make the dict accessing cleaner.
`_QUERY_PARAM_MAPPING.get(key)`

shahwaiz14: sure. I renamed the variable to `_QUERY_PARAM_TO_COLUMN_MAPPING`, so its clear the values in the dict are column names in model classes.

dtharpeuno (approved):

PR #3675, opened 2023-10-25
Pediatric updates for last modified time

kenny-skaggs:
## Resolves *no ticket*
Some pediatric data is stored and loaded into a participant summary response in a slightly different way, so the `lastModified` would not get updated in these cases. But the summary would be providing new information, so the timestamp of the new information should be set in the `lastModified` field.

This updates the code to set the `lastModified` timestamp when something would change in the participant summary because of pediatric data.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/pediatric_data_log_dao.py
+++ b/rdr_service/dao/pediatric_data_log_dao.py
@@ -3,7 +3,9 @@ from typing import Optional
 
 from sqlalchemy.orm import Session
 
+from rdr_service.clock import CLOCK
 from rdr_service.dao.base_dao import with_session
+from rdr_service.model.participant_summary import ParticipantSummary
 from rdr_service.model.pediatric_data_log import PediatricDataLog, PediatricDataType
 from rdr_service.participant_enums import PediatricAgeRange
 
@@ -26,7 +28,8 @@ class PediatricDataLogDao:
                 data_type=PediatricDataType.AGE_RANGE,
                 value=age_range_str
             ),
-            session=session
+            session=session,
+            callback=cls._check_new_pediatric_participant
         )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


ayazamlani (approved):
👍

rdr_service/dao/pediatric_data_log_dao.py
```diff
@@ -26,7 +28,8 @@ def record_age_range(cls, participant_id: int, age_range_str: str, session: Sess
                 data_type=PediatricDataType.AGE_RANGE,
                 value=age_range_str
             ),
-            session=session
+            session=session,
+            callback=cls._check_new_pediatric_participant
```
dtharpeuno, line 32: Haha haven't seen that use of that noun outside of JS before, works for this though.

rdr_service/dao/pediatric_data_log_dao.py
```diff
@@ -61,3 +64,17 @@ def insert(cls, data: PediatricDataLog, session: Optional[Session] = None):
...
+    def _check_new_pediatric_participant(cls, session: Session, existing_data: PediatricDataLog, participant_id: int):
+        # If we're receiving our first age_range value for the participant, then they would switch to is_pediatric.
+        # So then the last_modified date should be updated.
+        if not existing_data:
+            summary: ParticipantSummary = session.query(ParticipantSummary).filter(
+                ParticipantSummary.participantId == participant_id
+            ).one_or_none()
+
+            if summary:
+                summary.lastModified = CLOCK.now()
```
dtharpeuno, line 80: When does this get commited?

dtharpeuno (approved):

rdr_service/dao/pediatric_data_log_dao.py
kenny-skaggs: That can vary. `record_age_range` is called by the API code without passing in a session. So the decorator sets one up and then it would commit when that context manager ends.

PR #3702, opened 2023-11-10
Fixing Validation CRON Job 'ValueError'

vmoniqb:
The offline/ResponseValidation CRON job failed recently with the following error:
`ValueError('not enough values to unpack (expected 6, got 2)')` because keys & values were not being unpacked properly when iterating through the error list. 

This PR resolves the error and updates the code to only update the RDR table when the job is run offline. This allows developers to run the tool locally without affecting data in the RDR.

## Tests
- [X] unit tests
- Successfully ran response validation tool locally

### DIFF
--- a/rdr_service/offline/response_validation.py
+++ b/rdr_service/offline/response_validation.py
@@ -113,7 +113,7 @@ class ResponseValidationController:
         self._output_result(result_text)
 
     def _insert_data(self):
-        for key in self._error_list.items():
+        for key, error_list in self._error_list.items():  # pylint: disable=unused-variable
             survey_code, question_code, error_str, error_type, pid, answer_id = key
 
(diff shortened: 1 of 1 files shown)


rdr_service/offline/response_validation.py
```diff
@@ -113,7 +113,7 @@ def _send_error_message(self):
         self._output_result(result_text)
 
     def _insert_data(self):
-        for key in self._error_list.items():
+        for key, error_list in self._error_list.items():  # pylint: disable=unused-variable
```
shahwaiz14, line 116: Do you think `for key in self._error_list.keys():` will be better here, so you don't have an unused variable?

shahwaiz14 (approved):

PR #3723, opened 2023-12-04
Automating Query to Load Tactis Data to BigQuery

vmoniqb:
In order to make data available to Tactis in BigQuery, we have created a table in the RDR called `participant_data_to_tactis` that will track all participants that have changes in Tactis relevant fields. The `participant_data_to_tactis` table is queried by the CRON job and all participants with changes in the last day are synced to Big Query.

This PR includes the following changes to set up the automatic BigQuery sync for Tactis:
* Creating a CRON job to run on a daily cadence and sync the 7 requested fields of participant data to BigQuery
  * The job will be disabled initially using the decorator `@check_genomic_cron_job()` until Tactis is ready on their end
  * The `participant_data_to_tactis` table will store all changes over time, while the BigQuery table will only provide Tactis with updated fields from the past 24 hours
* Creating a BigQuery Tactis tool to be able to manually sync Tactis data in the case of backfills or requests for specific data from Tactis

## Tests
- Tested in Sandbox & Stable

[DA-3803]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/offline/main.py
+++ b/rdr_service/offline/main.py
@@ -44,6 +44,7 @@ from rdr_service.offline.response_validation import ResponseValidationController
 from rdr_service.offline.service_accounts import ServiceAccountKeyManager
 from rdr_service.offline.sync_consent_files import ConsentSyncController
 from rdr_service.offline.table_exporter import TableExporter
+from rdr_service.offline.tactis_bq_sync import TactisBQDataSync
 from rdr_service.repository.obfuscation_repository import ObfuscationRepository
 from rdr_service.resource.tasks import dispatch_check_consent_errors_task
 from rdr_service.services.consent.validation import ConsentValidationController, ReplacementStoringStrategy,\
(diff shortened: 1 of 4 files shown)


rdr_service/offline/tactis_bq_sync.py
```diff
@@ -0,0 +1,65 @@
...
+        "modified_participants": """
+                        SELECT * FROM
+                          EXTERNAL_QUERY("{external_connection}",
+                            "SELECT t.participant_id, ps.first_name, ps.middle_name, ps.last_name, ps.email, """
+                                 """ps.phone_number, ps.participant_origin FROM participant_data_to_tactis t """
+                                 """LEFT JOIN participant_summary ps ON ps.participant_id = t.participant_id """
+                                 """LEFT JOIN participant p ON p.participant_id = ps.participant_id """
+                                 """WHERE p.is_test_participant = 0 AND p.is_ghost_id is null """
+                                 """AND p.hpo_id not in (21) AND t.created >= '{since_date}'");"""
+    }
```
dtharpeuno, line 33: We try to mainly lean into the ORM for SQL queries,  so I think it would be best to adhere to that and create a DAO class for this and have this query in that. So that way you can initialize the dao and pull the designated query from that.

rdr_service/offline/main.py
```diff
@@ -888,6 +889,22 @@ def nph_sms_n1_generation():
     return '{"success": "true"}'
 
 
+@app_util.auth_required_cron
+@check_genomic_cron_job('tactis_participants_to_bq')
```
dtharpeuno: While I like that you are using this decorator we wrote, I can see the use-case now for extending it outside the realm of genomics, but I think it might be better to change the schedule in cron_default.yaml to `schedule: 1 of jan 12:00` and remove the use of this decorator since its specific to Genomics workflows, that way it nevers get executed and we have scripts for updating the cron scheduler when tactis is ready, and you can add a subsequent PR to the cron schedule to make it permanent.

vmoniqb: Updated!

dtharpeuno (approved):

j-kanuch (approved):

PR #3724, opened 2023-12-05
[DA-4017] Fix Slack alert when withdrawn PIDs received in NPH Intake API

shahwaiz14:
- There was a bug where we didn't convert NPH participant ids to AOU participant ids when searching for withdrawn ids used to send slack alert. This PR fixes that.
- Moved the logic for converting pids to cloud tasks instead of `handle_data_from_payload` in the NPH Intake API.

## Tests
✅  unit tests

### DIFF
--- a/rdr_service/api/ancillary_study_cloud_tasks_api.py
+++ b/rdr_service/api/ancillary_study_cloud_tasks_api.py
@@ -211,6 +211,7 @@ class WithdrawnParticipantNotifierTaskApi(BaseAncillaryTaskApi):
 
     def post(self):
         super().post()
+        bundle_id = self.data.get("bundle_id")
         nph_pids = self.data.get("nph_pids")
         aou_pids = self._convert_nph_pids_to_aou_pids(nph_pids=nph_pids)
         withdrawn_pids: list[str] = self.participant_dao.get_withdrawn_participant_ids(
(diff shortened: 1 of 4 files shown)


rdr_service/api/ancillary_study_cloud_tasks_api.py
```diff
@@ -202,14 +203,36 @@ class WithdrawnParticipantNotifierTaskApi(BaseAncillaryTaskApi):
...
-        msg = (f"NPH Intake API Payload received on {ts} contains {len(withdrawn_pids)} withdrawn participant IDs:"
-               f" {', '.join(withdrawn_pids)}")
-        create_nph_incident(slack=True, message=msg)
+        if withdrawn_pids:
+            ts = CLOCK.now().strftime('%Y-%m-%d %H:%M:%S')
+            msg = (
+                f"NPH Intake API Payload received on {ts} contains {len(withdrawn_pids)} withdrawn participant IDs:"
+                f" {', '.join(withdrawn_pids)}"
+            )
+            create_nph_incident(slack=True, message=msg)
```
dtharpeuno: I like the idea of making the alert more granular, but utilizing timestamps for searching will be arduous to track down if an investigation needs to be done. We are sent and store a uuid  that is generated from our partners that would be better here. Should be an instance variable called `bundle_identifier`.

j-kanuch (approved):

dtharpeuno (approved):

PR #3738, opened 2023-12-13
Updating AW Manifest Naming Conventions to Accept "_v2"

vmoniqb:
* Updated the AW1F and AW2 manifests to accept "_v2". All other AW manifests that are validated can accept "_v2" in the naming conventions with no changes.
* Added unit tests to verify that manifests with and without "_v2" will be accepted

## Tests
- [X] unit tests

[DA-3996]: https://precisionmedicineinitiative.atlassian.net/browse/DA-3996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -1952,7 +1952,7 @@ class GenomicFileValidator:
         def gem_a2_manifest_name_rule():
             """GEM A2 manifest name rule: i.e. AoU_GEM_A2_manifest_2020-07-11-00-00-00.csv"""
             return (
-                len(filename_components) <= 6 and
+                len(filename_components) == 5 and
                 filename_components[0] == 'aou' and
                 filename_components[1] == 'gem' and
                 filename_components[2] == 'a2' and
--- a/rdr_service/genomic/genomic_job_components.py
+++ b/rdr_service/genomic/genomic_job_components.py
@@ -1939,20 +1939,20 @@ class GenomicFileValidator:
         def aw1f_manifest_name_rule():
             """Biobank to GCs Failure (AW1F) manifest name rule"""
             return (
-                len(filename_components) == 5 and
+                len(filename_components) <= 6 and
                 filename_components[0] in self.VALID_GENOME_CENTERS and
                 filename_components[1] == 'aou' and
                 filename_components[2] in ('seq', 'gen') and
                 re.search(r"pkg-[0-9]{4}-[0-9]{5,}$",
                           filename_components[3]) is not None and
-                filename_components[4] == 'failure.csv' and
+                filename_components[4] in ('failure', 'failure.csv') and
                 filename.lower().endswith('csv')
             )
 
         def gem_a2_manifest_name_rule():
             """GEM A2 manifest name rule: i.e. AoU_GEM_A2_manifest_2020-07-11-00-00-00.csv"""
             return (
-                len(filename_components) == 5 and
+                len(filename_components) <= 6 and
                 filename_components[0] == 'aou' and
                 filename_components[1] == 'gem' and
                 filename_components[2] == 'a2' and
--- a/tests/genomics_tests/test_genomic_file_ingester.py
+++ b/tests/genomics_tests/test_genomic_file_ingester.py
@@ -117,36 +117,48 @@ class GenomicFileIngesterTest(BaseTestCase):
         self.assertEqual(copy_member.blockResearch, 1)
 
     def test_validate_filenames(self):
-        job_controller = GenomicJobController(job_id=GenomicJob.AW1F_MANIFEST)
+        job_controller = GenomicJobController(job_id=0)
 
         file_validator = GenomicFileValidator(
             job_id=job_controller.job_id,
             controller=job_controller
         )
 
-        short_read_job_ids = [GenomicJob.AW1F_MANIFEST,
-                              GenomicJob.AW1_MANIFEST,
-                              GenomicJob.METRICS_INGESTION,
-                              GenomicJob.AW4_ARRAY_WORKFLOW,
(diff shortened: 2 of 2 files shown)


rdr_service/genomic/genomic_job_components.py
```diff
@@ -1939,20 +1939,20 @@ def bb_to_gc_manifest_name_rule():
...
             )
 
         def gem_a2_manifest_name_rule():
             """GEM A2 manifest name rule: i.e. AoU_GEM_A2_manifest_2020-07-11-00-00-00.csv"""
             return (
-                len(filename_components) == 5 and
+                len(filename_components) <= 6 and
                 filename_components[0] == 'aou' and
                 filename_components[1] == 'gem' and
                 filename_components[2] == 'a2' and
```
dtharpeuno: Since the GEM A2 is not part of the short-read workflows and is not a contract we have with that partner, think we should revert this change.

tests/genomics_tests/test_genomic_file_ingester.py
```diff
@@ -116,6 +116,44 @@ def test_replating_copy(self):
...
+                    True, True,
+                    True, True]
+        actual = []
+
+        for f in aw1f_files:
+            actual.append(file_validator.validate_filename(f))
+
+        file_validator.job_id = GenomicJob.GEM_A2_MANIFEST
+        for f in aw2_files:
+            actual.append(file_validator.validate_filename(f))
```
dtharpeuno: Sorry this job is actually the GEM A2, which we don't need a v2 of anyway, need to focus on the `AW2` job which is  `GenomicJob.METRICS_INGESTION`.

tests/genomics_tests/test_genomic_file_ingester.py
```diff
@@ -116,6 +116,44 @@ def test_replating_copy(self):
...
+                      'UW_AoU_GEN_PKG-1234-567890_FAILURE_v2.csv',
+                      'UW_AoU_GEN_PKG-1234-567890_FAILURE-v2.csv',
+                      'UW_AoU_GEN_PKG-1234-567890_FAILURE.v2.csv']
+        aw2_files = ['AoU_GEM_A2_manifest_2020-07-11.csv',
+                     'AoU_GEM_A2_manifest_2020-07-11_v2.csv']
+        aw4_arr_files = ['AoU_DRCB_GEN_2020-07-11-00-00-00.csv',
+                         'AoU_DRCB_GEN_2020-07-11-00-00-00_v2.csv']
+        aw4_wgs_files = ['AoU_DRCB_SEQ_2020-07-11-00-00-00.csv',
+                         'AoU_DRCB_SEQ_2020-07-11-00-00-00_v2.csv']
+
```
dtharpeuno: While this works with the `assertEqual` method, I think it might be better to structure the tests a little differently. 
Maybe we could do something like looping thru the designated short-read job and lean into the targeted testing boolean methods like `assertTrue`/`assertFalse` and pass in the filename str as an argument.
Ex: 
```
for job_id in short_read_job_ids:
        file_validator.job_id = job_id
        self.assertTrue(file_validator.validate_filename(<filename_string>) // if true
        self.assertTrue(file_validator.validate_filename(<filename_string>) // if false
```

dtharpeuno (changes requested):

tests/genomics_tests/test_genomic_file_ingester.py
vmoniqb: Yeah sorry I keep mixing up A2 & AW2. Will fix this!

tests/genomics_tests/test_genomic_file_ingester.py
vmoniqb: Updated to use targeted testing boolean methods! Let me know if I missed anything

tests/genomics_tests/test_genomic_file_ingester.py
```diff
@@ -116,6 +116,37 @@ def test_replating_copy(self):
...
+                                       'UW_AoU_SEQ_PKG-1234-567890_FAILURE_v2.csv'],
+            GenomicJob.AW1_MANIFEST: ['UW_AoU_SEQ_PKG-1234-567890.csv',
+                                      'UW_AoU_GEN_PKG-1234-567890_v2.csv'],
+            GenomicJob.METRICS_INGESTION: ['UW_AoU_GEN_DataManifest_01234567_890.csv',
+                                           'UW_AoU_SEQ_DataManifest_01234567_890_v2.csv'],
+            GenomicJob.AW4_ARRAY_WORKFLOW: ['AoU_DRCB_GEN_2020-07-11-00-00-00.csv',
+                                            'AoU_DRCB_GEN_2020-07-11-00-00-00_v2.csv'],
+            GenomicJob.AW4_WGS_WORKFLOW: ['AoU_DRCB_SEQ_2020-07-11-00-00-00.csv',
+                                          'AoU_DRCB_SEQ_2020-07-11-00-00-00_v2.csv']
+        }
```
dtharpeuno: So you already have your jobs id in the keys for this dictionary, no need to have them in the above list as well for the  main iteration in the test. 

Also I think it would be best to add valid and non-valid examples to the dictionary for iteration instead of only testing the correct filenames on most and incorrect ones on a select few. 

Maybe we can add those examples to the aforementioned dictionary and use the nested keys to determine if we should be asserting True of False.
Ex:
```
short_read_validation_map = {
   <job_id>: {
         'valid': [<valid examples>],
         'invalid': [<invalid examples>]
    }
}
```
This would encompass all your checks in a single iteration.

dtharpeuno (changes requested):

dtharpeuno (approved):

brianmkimmel (approved):

PR #3746, opened 2023-12-18
[DA 4040] Add validation for Pull List and N1

shahwaiz14:
- Contains migration file to change bmi string length from 4 to 5 chars.
- Added validation checks for ingesting pull list and generating N1.      
      - Ingesting Pull List Validation rule: Avoid writing to the table if bmi value length is greater than 5 characters. 
      - Generating N1 Validation rule: Avoid generating N1 if biobank_id exists, but other values, like race, sample_id, etc., is None.
- Added slack alert if validation fails.

## Tests
- ✅  unit tests


rdr_service/model/study_nph_sms.py
```diff
@@ -64,7 +64,7 @@ class SmsSample(NphBase):
     sample_identifier = Column(String(32))
     diet = Column(String(32))
     sex_at_birth = Column(String(32))
-    bmi = Column(String(4))
+    bmi = Column(String(5))
```
dtharpeuno: TBH knowing the BB, I would definitely extend this to at least `32`, just to be on the safe side, if they decide to start sending floats for this. I see you have validation for that but sometimes its easier to defensively program against what the partners are going to send.

rdr_service/workflow_management/nph/sms_validation.py
```diff
@@ -0,0 +1,68 @@
...
+        if bmi_str_len > max_length_allowed:
+            msg = (
+                f"Pull list Validation Error: BMI value exceeds the max length allowed of {max_length_allowed} "
+                f"and contains {bmi_str_len} characters for sample id {row.get('sample_id')}."
+                f"Check the data at: {self.file_path}."
+            )
+            create_nph_incident(slack=True, message=msg)
+            raise ValueError(
+                f"BMI str length of {bmi_str_len} exceeds the limit of {max_length_allowed} characters."
+            )
```
dtharpeuno: Made a comment concerning the expected length, but with this you can get a manifest that can have 1000 rows and this will alert up to the `max num of times` if the bmi length is too long. I would say its better to aggregate the alert in this case but might be better to extend the length of the value expected.

rdr_service/workflow_management/nph/sms_validation.py
```diff
@@ -0,0 +1,68 @@
...
+        if row.biobank_id:
+            if not all([col for col in required_columns]):
+                msg = (
+                    f"N1 manifest for package id {row.package_id} contains biobank id "
+                    f"but is missing other value(s) for recipient {self.recipient}."
+                )
+                create_nph_incident(slack=True, message=msg)
+                raise ValueError(
+                    "Biobank ID exists in N1, but other required values are missing."
+                )
```
dtharpeuno: What happens if you dont have a `biobank_id`? Maybe include that in the req cols and validate against all of them?

shahwaiz14: Yep, I think including `biobank_id` in the list will have the same result. 

I think I need to add a comment in the code describing the reasoning why other values should exist if there is a biobank_id. The way I understand it is that part of the N1 manifest contains information related to each individual's sample within each compartment (well) of a plate. So if a `biobank_id` is present in a row in N1 manifest, the corresponding well box position must also contain a `sample_id` and other details of that individual. (We noticed that in some cases we had a biobank_id but not other values.) Do you think keeping it the way it is now and adding comment will make more sense here?

rdr_service/workflow_management/nph/sms_validation.py
shahwaiz14: This will only send 1 slack alert for each pull list it tries to ingest - as soon as it finds the first occurrence of bmi string length greater than 5, it will terminate, won't write to the `sms_sample` table, and send us an alert, so we can look at the file, and report back to the client.

Josh found out about this last week from the client that they would like BMI values to be a maximum of 5 characters.

rdr_service/workflow_management/nph/sms_validation.py
dtharpeuno: Sounds good, sorry thought this was an ingestion.

dtharpeuno (approved):

j-kanuch (approved):

PR #3771, opened 2024-01-19
[DA-3887] Added email cron job for participant cleanup notification.

DaveK-14:
Adds a cron job to automate an email notifying PTSC that they should begin cleanup of test participants.


rdr_service/config.py
```diff
@@ -291,6 +291,8 @@
 PM_HEIGHT_CODES = 'pm_height_codes'
 PM_WEIGHT_CODES = 'pm_weight_codes'
 
+PARTICIPANT_CLEANUP_CC_LIST = 'participant_cleanup_cc_list'
```
ayazamlani: is this code constant needed? I am not sure if it is being pulled on L925 of rdr_service/offline/main.py

ayazamlani (commented):
Do we need to write any tests for this new cron job?

tests/cron_job_tests/test_offline_app_setup.py
```diff
@@ -94,3 +94,9 @@ def test_consent_error_report_route(self, mock_checker):
         mock_checker.assert_has_calls([call(origin='vibrent', in_seconds=30),
                                        call(origin='careevolution', in_seconds=180)])
         config.GAE_PROJECT = saved_config_project
+
+    @mock.patch('rdr_service.offline.main.ptsc_test_participant_cleanup_request')
+    def test_ptsc_test_participant_cleanup_request(self, mock_checker):
+        response = self.send_cron_request('PTSCTestParticipantCleanupRequest')
+        self.assertEqual(response, {"success": "true"})
+        mock_checker.assert_not_called()
```
ayazamlani: shouldn't the mock_checker be getting called at least once since self.send_cron_request('PTSCTestParticipantCleanupRequest') is working and the response is coming back as "success": "true" ? I think it should be getting called but I could be mistaken here

DaveK-14: I thought that as well, but could not get it to agree. It is getting the correct output from the program, which is all I really wanted. the assert on mock_checker is to prevent circleci from reporting an unused variable since mock_checker is passed in when the test is called.

kenny-skaggs: I think the timing of the mock is causing this. The `ptsc_test_participant_cleanup_request` method is getting replaced when the test starts, but by the time the mock is applied we've already set up the offline flask app in the setUp method (in the `from rdr_service.offline.main import app` line). So `setUp` processes the offline main.py file, and while doing that still has access to the real definition of the `ptsc_test_participant_cleanup_request` method. Only after that is the mock applied (replacing the method in the offline's main module, but not replacing the handle that the API has for the real definition of the method).

Maybe a target for mocking is to mock the `EmailService.send_email` method and check that it's called (and, even better, something about what was sent matches up with the wording of the reminder email).

DaveK-14: I have the EmailService.send_email in a block that only runs on prod to avoid sending emails during testing, so it wouldn't be called. I can add the message to the return and assert on that as well, but the closest thing I can think of to check with the mocker in that case would be maybe the Email constructor?

kenny-skaggs: Yeah, that sounds like it could work.

rdr_service/offline/main.py
```diff
@@ -906,6 +907,37 @@ def nph_biobank_inventory_file_import():
     return '{"success": "true"}'
 
 
+@app_util.auth_required_cron
+def ptsc_test_participant_cleanup_request():
+    date = datetime.today()
+    if date.month == 12:
+        month = 1
+    else:
+        month = date.month + 1
```
kenny-skaggs: Importing and using `relativedelta` from the `dateutil.relativedelta` module could be helpful here. `date = datetime.today() + relativedelta(months=1)` would be a little more concise, and would increment the year when needed too.

kenny-skaggs (changes requested):
Please merge the recent changes from devel into this branch (to get the Safety check to pass).

kenny-skaggs:
> Please merge the recent changes from devel into this branch (to get the Safety check to pass).

Nevermind, you're already on it :D

kenny-skaggs (approved):

ayazamlani:
please get 2 approvals before merging

PR #3772, opened 2024-01-19
added slack alert

DaveK-14:
Added code to create a slack alert for NPH when files are submitted in non-csv format.


rdr_service/workflow_management/nph/sms_workflows.py
```diff
@@ -146,6 +159,15 @@ def job_ingestion(self):
         Main method for ingestion jobs.
         """
 
+        # Ensure file is CSV
+        if not str(self.file_path).endswith(".csv"):
+            nph_slack = get_slack_message_handler()
+            nph_slack.send_message_to_webhook(
+                message_data="Error ingesting nph file of type {0}. "
+                             "File does not conform to csv format.".format(self.file_type)
```
ayazamlani, line 167: I think the self.file_type is using the SmsFileTypes (A1, B2, C3), not the actual ending of the file (.csv or .pdf). you may want to grab that from the file_path.

ayazamlani: f-strings pls

ayazamlani (changes requested):

PR #3778, opened 2024-01-23
[DA-3959] NPH Sample Management System (SMS) Slack Alert

DaveK-14:
Added nph slack alert for mismatched filetypes.

## Tests
[] unit tests


rdr_service/workflow_management/nph/sms_workflows.py
```diff
@@ -3,14 +3,27 @@
...
+    slack_config = config.getSettingJson(config.NPH_SLACK_WEBHOOKS, {})
+    if slack_config is None:
+        logging.warning("'slack_config' for 'NPH_SLACK_WEBHOOKS' is empty")
+
+    webhook_url = slack_config.get('rdr_nph_alerts', None)
+    if webhook_url is None:
+        logging.warning("'rdr_nph_alerts' is not available in slack config. 'webhook_url' is None")
+    return SlackMessageHandler(webhook_url=webhook_url)
+
+
```
dtharpeuno: This needs to be moved to a cloud task call that we use for all other NPH incidents.  `NphIncidentTaskApi`

rdr_service/workflow_management/nph/sms_workflows.py
```diff
@@ -146,6 +159,16 @@ def job_ingestion(self):
         Main method for ingestion jobs.
         """
 
+        # Ensure file is CSV
+        if not str(self.file_path).endswith(".csv"):
+            file_end = self.file_path.split(".")[-1]
+            nph_slack = get_nph_slack_message_handler()
+            nph_slack.send_message_to_webhook(
+                message_data=f"Error ingesting nph file of type {file_end}. "
+                             "File does not conform to csv format."
```
j-kanuch: Can you add the `file_path` to the message as well? Something like "File {file_path} does not conform to csv format."

dtharpeuno (changes requested):
Not seeing any tests? Please add them.

rdr_service/workflow_management/nph/sms_workflows.py
```diff
@@ -146,6 +147,23 @@ def job_ingestion(self):
         Main method for ingestion jobs.
         """
 
+        # Ensure file is CSV
+        if not str(self.file_path).endswith(".csv"):
+            file_end = self.file_path.split(".")[-1]
```
j-kanuch: nit: you could probably move `file_end` up a couple lines and then test `if file_end != 'csv'`

rdr_service/rdr_client/pairing_assigner.py
```diff
@@ -105,6 +105,7 @@ def main(client):
             if client.args.dry_run:
                 logging.info("Dry run, would update participant[%r] to %r.", pairing_key, new_pairing)
             else:
+                print(participant)
```
j-kanuch: I'm not sure why the pairing_assigner.py is in this PR, but can you remove this print statement?

rdr_service/api/ancillary_study_cloud_tasks_api.py
```diff
@@ -197,6 +199,15 @@ def post(self):
...
 
+    def get_nph_slack_message_handler(self) -> SlackMessageHandler:
+        slack_config = config.getSettingJson(config.NPH_SLACK_WEBHOOKS, {})
+        if slack_config is None:
+            logging.warning("'slack_config' for 'NPH_SLACK_WEBHOOKS' is empty")
+
+        webhook_url = slack_config.get('rdr_nph_alerts', None)
+        if webhook_url is None:
+            logging.warning("'rdr_nph_alerts' is not available in slack config. 'webhook_url' is None")
+        return SlackMessageHandler(webhook_url=webhook_url)
```
dtharpeuno: Can you remove this in favor of the comment for using the cloud task?

rdr_service/workflow_management/nph/sms_workflows.py
```diff
@@ -146,6 +147,23 @@ def job_ingestion(self):
...
+                    "message": f"Error ingesting nph file of type {file_end}. "
+                               f"File '{self.file_path}' does not conform to csv format."
+                }
+                create_nph_incident(**kwargs)
+            else:
+                logging.warning(
+                    msg=f"Error ingesting nph file of type {file_end}. "
+                        f"File '{self.file_path}' does not conform to csv format."
+                )
+            return
```
dtharpeuno: Can this move to `NphIncidentTaskApi`, this is not the cloud task call.

dtharpeuno (changes requested):

rdr_service/rdr_client/pairing_assigner.py
DaveK-14: oh i might have added that while dealing with the nph repair i did yesterday, will get that out of there.

tests/workflow_tests/test_nph_sms_workflows.py
```diff
@@ -457,3 +457,25 @@ def test_sms_pipeline_n1_function(self, task_mock):
                                      payload=data,
                                      queue='nph')
 
+
+    def test_sms_ingestion_filetype_failure(self):
```
ayazamlani: im not sure how this is testing a filetype failure. If this is testing that an error is occuring pls add that to the name like in L386. If you can also add comments. I am not sure what you are trying to test for here.

tests/workflow_tests/test_nph_sms_workflows.py
```diff
@@ -457,3 +457,25 @@ def test_sms_pipeline_n1_function(self, task_mock):
...
+        ingestion_data = {
+            "job": "FILE_INGESTION",
+            "file_type": "SAMPLE_LIST",
+            "file_path": f"{self.test_bucket}/test_sample_list_fail.txt"
+        }
+        workflow = SmsWorkflow(ingestion_data)
+        workflow.execute_workflow()
+
+        from rdr_service.resource import main as resource_main
+        with self.assertLogs(level="WARNING") as cm:
```
ayazamlani: sry, what is cm? I also am seeing it in some other functions but I am not sure what it is.

ayazamlani (changes requested):

tests/workflow_tests/test_nph_sms_workflows.py
DaveK-14: My understanding of the with statement there is to store the logging output so it can be asserted against for level and content. cm is the variable used elsewhere, but I can update that to something more suitable to this. I am using it to check that I got a warning(not error) log and that it contains the proper warning that shows it reached the point in the code where a slack message would be sent and the ingestion process would stop.

tests/workflow_tests/test_nph_sms_workflows.py
```diff
@@ -457,3 +457,25 @@ def test_sms_pipeline_n1_function(self, task_mock):
...
+        with self.assertLogs(level="WARNING") as slack_warning:
+            self.send_post(
+                local_path='NphSmsIngestionTaskApi',
+                request_data=ingestion_data,
+                prefix="/resource/task/",
+                test_client=resource_main.app.test_client(),
+            )
+            # Check that the logs contain a noticeable part of the warning indicating a slack message would be sent.
+            self.assertIn('does not conform', slack_warning.output[0])
+
```
dtharpeuno: Can you add a test here for when the config item is set, that a mock is called for the cloud task endpoint.

rdr_service/workflow_management/nph/sms_workflows.py
```diff
@@ -146,6 +147,23 @@ def job_ingestion(self):
         Main method for ingestion jobs.
         """
 
+        file_end = self.file_path.split(".")[-1]
+        # Ensure file is CSV
+        if not file_end == "csv":
```
dtharpeuno: Maybe a good idea to cast this a lower() and do the equivalence check then.
Also a  better variable could be helpful here like `file_extension`.

dtharpeuno (approved):

ayazamlani (approved):

PR #3787, opened 2024-02-05
[DA-4065] NPH visitPeriod backfill script

kenny-skaggs:
## Resolves *[DA-4065](https://precisionmedicineinitiative.atlassian.net/browse/DA-4085)*
Now that we're switching from visitType to visitPeriod for NPH orders, we'll need to fill in visitPeriods for the existing orders. This script reads data from a file provided by RTI that details what the visitPeriod name should be for each existing order.

The script will load all StudyCategories from the database and keep track of existing and newly created visitPeriod categories. When we encounter an order to backfill, we check to see if we have already created the visitPeriod and timepoint needed. If not, we create new ones.

## Tests
- [x] unit tests

[DA-4065]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/tools/tool_libs/backfill_nph_study_category.py
+++ b/rdr_service/tools/tool_libs/backfill_nph_study_category.py
@@ -30,7 +30,7 @@ class TimepointManager:
         }
 
         self.visit_name_map = {}
-        for timepoint in self.tp_visit_name_map.values:
+        for timepoint in self.tp_visit_name_map.values():
             visit = timepoint.parent
             self.visit_name_map[self.visit_name_key(visit.parent.id, visit.name)] = visit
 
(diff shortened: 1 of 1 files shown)


ayazamlani (approved):
can we please leave a comment with an example of the file .csv format for if ppl need to run the tool and create a .csv file. I think this one is taking orderID, studyID, and visitID

dtharpeuno (approved):

kenny-skaggs:
> can we please leave a comment with an example of the file .csv format for if ppl need to run the tool and create a .csv file. I think this one is taking orderID, studyID, and visitID

I think that's a good idea for scripts we plan to run multiple times. But this backfill will only happen once in this way. If something comes up in the future like this, then it's best to create a new script or re-work this one rather than rerun exactly what this is doing.

PR #3798, opened 2024-02-08
MIGRATION - Add diet_name_str diet name string to the `nph.diet_event` table

j-kanuch:
This PR adds the string value for the `DietType` enum to the NPH `diet_event` table. Intake API was updated to write this field. An additional back-fill ticket will need to be completed once this is on Stable and Production. This will aid in testing, investigations, and analysis.

## Tests
- [x] unit tests


rdr_service/alembic/nph/versions/04817233e930_add_diet_string_to_diet_event.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+def upgrade_nph():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('diet_event', sa.Column('diet_name_str', sa.String(length=128), nullable=True))
+    # ### end Alembic commands ###
```
dtharpeuno, line 29: Should we go ahead and add a SQL backfill on this migration to take care of the subsequent backfill you mentioned?

j-kanuch: That's a smart idea, and I usually like do the smart ideas. But for this one, since it's not super high priority, I wanted to give a newer dev the backfill query as a separate ticket.

dtharpeuno (approved):

vmoniqb (approved):

PR #3815, opened 2024-02-29
Update date on safety check

dtharpeuno:

Moving safety check date -> 4mos

## Tests
- [] unit tests
* all tests should pass

### DIFF
--- a/rdr_service/genomic/validation.py
+++ b/rdr_service/genomic/validation.py
@@ -2,6 +2,7 @@ import collections
 import datetime
 import functools
 import operator
+from dateutil.relativedelta import relativedelta
 
 from rdr_service import clock
 from rdr_service.dao.genomics_dao import GenomicSetDao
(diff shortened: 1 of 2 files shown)


kenny-skaggs (approved):

j-kanuch (approved):

j-kanuch (approved):

rdr_service/genomic/validation.py
```diff
@@ -24,7 +24,7 @@ def validate_and_update_genomic_set_by_id(genomic_set_id, dao=None):
   :type dao: GenomicSetDao or None
   """
     now = clock.CLOCK.now()
-    date_of_birth_cutoff = datetime.date(year=now.year - GENOMIC_VALID_AGE, month=now.month, day=now.day)
+    date_of_birth_cutoff = datetime.date(year=now.year - GENOMIC_VALID_AGE, month=now.month, day=28)
```
kenny-skaggs: :nit Just in case the day might end up mattering for a given test (current or future), what about using a leap year aware method? I think this would give the correct date (after adding `from dateutil.relativedelta import relativedelta` to the imports):
```suggestion
    date_of_birth_cutoff = now - reltativedelta(years=GENOMIC_VALID_AGE)
```

dtharpeuno: That works thanks! Had to convert it back to a date object, since relativedelta  was returning as a datetime obj, for a later comparison though.

kenny-skaggs (approved):

PR #3820, opened 2024-03-04
[DA-4141] Automatically detecting duplicate accounts

kenny-skaggs:
## Resolves *[DA-4141](https://precisionmedicineinitiative.atlassian.net/browse/DA-4141)*
RDR will need to implement a process for automatically detecting duplicate accounts.

This implements a cron job that will load any participants that have been modified in the last day (with a few extra hours to make sure there's overlap rather than a gap). If these participants have been modified, they may have had their name or birthday updated to show that they match another participant.

With the recently modified participants loaded, we then compare their data to all other participants in the system. That could be costly in terms of time, so we gather all participant data into a set of dictionaries to make the comparison run quicker.

Then we save any duplicates that we detected (ignoring them if they're already recorded).

## Tests
- [x] unit tests

[DA-4141]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/services/duplicate_detection.py
+++ b/rdr_service/services/duplicate_detection.py
@@ -1,0 +1,80 @@
+from collections import defaultdict
+from datetime import datetime
+from typing import List
+
+from sqlalchemy.orm import Session
+
+from rdr_service.dao.base_dao import with_session
+from rdr_service.dao.duplicate_account_dao import DuplicateAccountDao, DuplicationSource, DuplicateExistsException
+from rdr_service.model.participant_summary import ParticipantSummary
+
+
+class _SummaryCache:
+    def __init__(self, session: Session):
+        self._session = session
+        self._name_dob_cache = defaultdict(         # top level dict for keying by first name
+            lambda: defaultdict(                    # next level dict for keying by last name
+                lambda: defaultdict(                # next level dict for keying by dob
+                    list                            # list to hold any participant ids that have the above data
+                )
+            )
+        )
+        self._email_cache = defaultdict(list)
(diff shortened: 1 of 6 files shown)


rdr_service/services/duplicate_detection.py
```diff
@@ -0,0 +1,80 @@
...
+                    list                            # list to hold any participant ids that have the above data
+                )
+            )
+        )
+        self._email_cache = defaultdict(list)
+        self._phone_cache = defaultdict(list)
+        self._load_data()
+
+    def _load_data(self):
+        for summary in DuplicateAccountDao.query_participant_duplication_data(self._session):
```
ayazamlani, line 27: should we throw in a try, except here incase any exceptions occur when trying to load the data we can log, and handle accordingly?

ayazamlani (approved):

ayazamlani (commented):
awesome tests!

brianmkimmel (approved):

rdr_service/services/duplicate_detection.py
kenny-skaggs: If any exceptions happen when loading this data we'd need to kill the whole run anyway. Since we wouldn't need to do any other cleanup, and since we'd re-raise the exception we catch here (so it can stop the rest of the process and so it can get logged by the error handler) it seems like it makes more sense to not catch it here.

PR #3823, opened 2024-03-06
[DA-4159] Database table for enrollment status data

kenny-skaggs:
## Partially Resolves *[DA-4159](https://precisionmedicineinitiative.atlassian.net/browse/DA-4159)*
This PR creates a database table that will store copies of all the enrollment status data points.

I also added a copy of the ParticipantCohort enum to make it easier to work with database enums

## Tests
- [ ] unit tests

[DA-4159]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/model/database.py
```diff
@@ -29,6 +29,7 @@
 from rdr_service.model.deceased_report import DeceasedReport
 from rdr_service.model.deceased_report_import_record import DeceasedReportImportRecord
 from rdr_service.model.ehr import EhrReceipt, ParticipantEhrReceipt
+from rdr_service.model.enrollment_dependencies import EnrollmentDependencies
```
shahwaiz14, line 32: What does importing models in database.py do?

shahwaiz14 (approved):

rdr_service/model/database.py
kenny-skaggs: We subclass the sqlalchemy declarative_base class to signal that a python class should be reflected in the database. We add it here so Alembic will see that there's a new class that needs to be reflected in the migration files.

When we run any of the alembic scripts something will need to load the file that defines our model class because if the file never gets parsed, then from Alembic's perspective there wasn't anything new that subclassed declarative base.

`rdr_service/alembic/env.py` is loaded any time we run alembic. And that file in turn loads `rdr_service/model/database.py`. So anything that the database.py file loads would also get loaded. So we add them here to get something to load the files when we run alembic.

shahwaiz14: ah ok, makes sense. Thanks!

brianmkimmel (approved):

PR #3830, opened 2024-03-13
[DA-2560] Questionnaire_Response Duplication Detection Tool

DaveK-14:
Added a Tool which can run existing questionnaire_response duplication detection over a long period of time.
Per Peggy this was done via a tool using existing logic rather than a backfill query.

[DA-2560]: https://precisionmedicineinitiative.atlassian.net/browse/DA-2560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/tools/tool_libs/questionnaire_response_duplicate_cleanup.py
```diff
@@ -0,0 +1,31 @@
...
+tool_desc = "Tool to run the questionnaire response duplicate detection logic over a long period of time."
+
+
+def run():
+
+    # Starting from year of oldest observed duplicates
+    start = datetime(2017, 1, 31, 23, 59, 59, 999999, pytz.UTC)
+    rdd = ResponseDuplicationDetector()
+
+    while start < datetime.now(tz=pytz.UTC):
```
dtharpeuno: `start` will always be less than `now` so not sure if we need this conditional.

rdr_service/tools/tool_libs/questionnaire_response_duplicate_cleanup.py
```diff
@@ -0,0 +1,31 @@
...
+# Remember to add/update bash completion in 'tool_lib/tools.bash'
+tool_cmd = "qr_duplicate_fix"
+tool_desc = "Tool to run the questionnaire response duplicate detection logic over a long period of time."
+
+
+def run():
+
+    # Starting from year of oldest observed duplicates
+    start = datetime(2017, 1, 31, 23, 59, 59, 999999, pytz.UTC)
+    rdd = ResponseDuplicationDetector()
```
j-kanuch: I'm not familiar with `ResponseDuplicationDetector()`. How is the class aware of which GCP project to flag the records in?
Using the other tools in tool_libs as a template, you can see how arguments are passed to the tool.

dtharpeuno: Please always name your variable adequately for what they describe, maybe `response_detector` would be better?

dtharpeuno (commented):
Judging by the ticket I dont think this encompasses what is intended might be better to check with product or maybe do this thru SQL instead of a tool?

rdr_service/tools/tool_libs/questionnaire_response_duplicate_cleanup.py
DaveK-14: Start is the initial date, but it is being added to 1 month at a time until it catches up to current. the conditional just stops it from running indefinitely.

rdr_service/tools/tool_libs/questionnaire_response_duplicate_cleanup.py
DaveK-14: The class does have a default, which I realize looking close will use the run location so localhost in my case. I will add the logic to take a project param and use that, thanks for catching.

dtharpeuno: Also can you make sure this inherits from the ToolBase class to fit the same pattern as the rest of the tools?

ayazamlani (commented):
do we need to count the number of duplicates flagged and reveal a total flagged at the end? And/Or perhaps print out when something is flagged? so the dev running the tool knows it is working

DaveK-14:
> do we need to count the number of duplicates flagged and reveal a total flagged at the end? And/Or perhaps print out when something is flagged? so the dev running the tool knows it is working

I do not believe this tool will have access to that information, it is running existing logic which is normally called by a cloud task that expects no output. This is also expected to be run exactly once, as the normal task should take care of all future occurrences.

ayazamlani:
I am thinking we could create the logic in the flag_duplicate_response function
perhaps adding a param
```
 return_questionnaire_response_ids = False
```
and then at the end of the function
```
if return_questionnaire_response_ids and questionnaire_ids_to_mark_as_duplicates: return questionnaire_ids_to_mark_as_duplicates
```

ayazamlani:
> Judging by the ticket I dont think this encompasses what is intended might be better to check with product or maybe do this thru SQL instead of a tool?

I am thinking [this function](https://github.com/all-of-us/raw-data-repository/blob/d232d2e3fd4617e22aa9ec9424cbc14d9eb5d44a/rdr_service/services/response_duplication_detector.py#L31) is getting called in the tool to help handle what is required in the ticket. 

It looks like the [default duplication threshold](https://github.com/all-of-us/raw-data-repository/blob/d232d2e3fd4617e22aa9ec9424cbc14d9eb5d44a/rdr_service/services/response_duplication_detector.py#L20) is 2.

but my question is the first part says complete in the ticket. If the default is 2, then the cron job could still be 10. are we updating that somewhere?  perhaps it was updated in the config or something?

DaveK-14:
The function being called is the bottom function, flag_duplicate_responses [this function](https://github.com/all-of-us/raw-data-repository/blob/d232d2e3fd4617e22aa9ec9424cbc14d9eb5d44a/rdr_service/services/response_duplication_detector.py#L128) which covers the entire process of finding and marking QR records in a given timeframe(found by the days_before, defaulted to 2 days before the cloud task calls it). With this tool I am passing it a larger range of days and going back a long time on repeated runs to cover all of the duplicates found. 

As for the cloud task, that section of the ticket was already complete, and the only thing needed now is to mark the duplicates that evaded detection in the past.

ayazamlani (changes requested):
> > do we need to count the number of duplicates flagged and reveal a total flagged at the end? And/Or perhaps print out when something is flagged? so the dev running the tool knows it is working
> 
> I do not believe this tool will have access to that information, it is running existing logic which is normally called by a cloud task that expects no output. This is also expected to be run exactly once, as the normal task should take care of all future occurrences.

You should be able to check [the logs](https://github.com/all-of-us/raw-data-repository/blob/d232d2e3fd4617e22aa9ec9424cbc14d9eb5d44a/rdr_service/services/response_duplication_detector.py#L146) and ensure that this ran properly after you run it, but if u want to confirm in the DB after, you may want to get those questionnaire_response_ids back when you run this tool along w/ a total count so you can easily confirm in the db after you run the tool that all of the appropriate questionnaire responses have been marked as duplicate

ayazamlani (approved):
no test? 

the tool looks good to me. Approved, if a test is not required here by a senior dev.

dtharpeuno (approved):

PR #3832, opened 2024-03-14
Fix for CEBYOK - PUT - 400 BAD REQUEST 

ayazamlani:
CEBYOK is running in to a similar issue with PUT requests that we just resolved with GET requests. Applying the same fix to the PUT requests to look for the bypass origin check flag.

## Tests
- [x] unit tests

### DIFF
--- a/rdr_service/dao/participant_dao.py
+++ b/rdr_service/dao/participant_dao.py
@@ -186,11 +186,12 @@ class ParticipantDao(UpdatableDao):
                     base_name = "example"  # TODO: This is a hack because something sets up configs different
                     # when running all tests and it doesnt have the clientId key.
             base_name = base_name.lower()
-            if base_name in ORIGINATING_SOURCES and base_name != existing_obj.participantOrigin:
-                logging.warning(f"{base_name} tried to modify participant from \
-                        {existing_obj.participantOrigin}")
-                raise BadRequest(f"{base_name} not able to update participant from \
-                        {existing_obj.participantOrigin}")
+            if not user_info.get("bypassOriginCheck"):
+                if base_name in ORIGINATING_SOURCES and base_name != existing_obj.participantOrigin:
+                    logging.warning(f"{base_name} tried to modify participant from \
+                            {existing_obj.participantOrigin}")
+                    raise BadRequest(f"{base_name} not able to update participant from \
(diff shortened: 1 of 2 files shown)


tests/api_tests/test_participant_api.py
```diff
@@ -848,17 +848,47 @@ def test_participant_post_skips_summary_update(self, summary_update_mock):
...
-        participant_id = response["participantId"]
-        # Expecting a Failed Response without the bypass origin check  flag
-        BaseTestCase.switch_auth_user("example@cebyok.com", "example")
-        bad_get_response = self.send_get("Participant/%s" % participant_id, expected_status=http.client.BAD_REQUEST)
-        self.assertEqual(bad_get_response.json['message'], 'Can not retrieve participant from a different origin')
-        # Expecting a passing response with the bypass origin check flag
-        BaseTestCase.switch_auth_user("example@cebyok.com", "example", bypass_origin_check=True)
-        get_response = self.send_get("Participant/%s" % participant_id)
-        self.assertEqual(response, get_response)
+        with FakeClock(TIME_1):
```
dtharpeuno: Don't think we need this.

dtharpeuno (approved):

tests/api_tests/test_participant_api.py
ayazamlani: You are correct, it is not absolutely necessary, It's making the response last modified time the same for the assertEqual.

 I will just del that field when checking the response.

rdr_service/dao/participant_dao.py
```diff
@@ -186,11 +186,12 @@ def _validate_update(self, session, obj, existing_obj):
                     base_name = "example"  # TODO: This is a hack because something sets up configs different
                     # when running all tests and it doesnt have the clientId key.
             base_name = base_name.lower()
```
j-kanuch, line 188: If not `base_name`, and `email` doesn't equal `DEV_MAIL`, this line is going to throw an exception.

ayazamlani: @j-kanuch cebyok does have a clientId set in the config, so there should be a base_name. I am not sure why this is going to throw an exception.

j-kanuch (approved):

PR #3837, opened 2024-03-19
[DA-4160] Adding backfill script for enrollment dependencies

kenny-skaggs:
## Resolves *[DA-4160](https://precisionmedicineinitiative.atlassian.net/browse/DA-4160)*
The enrollment status calculations will be using a new table as a source for the data. The new table will start receiving new data as it comes in to the API, but we'll need this script to copy the fields we already have in the database over to this new table.

The intention of this script is to fill in as a raw sql backfill. Writing the raw sql for this could work, but we wouldn't be able to reuse alot of the code we have in place (most everything here is copied from how ParticipantSummaryDao loads the data when getting ready to calculate). Testing will involve monitoring and validating the data a it's going into the table, and then once it's used the script will be obsolete. Since it doesn't need to be maintained moving forward, unit tests haven't been added.

## Tests
- [ ] unit tests

[DA-4160]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/tools/tool_libs/backfill_enrollment_dependencies.py
+++ b/rdr_service/tools/tool_libs/backfill_enrollment_dependencies.py
@@ -1,0 +1,127 @@
+from typing import List
+
+from dateutil.parser import parse
+from sqlalchemy.orm import joinedload
+
+from rdr_service import code_constants
+from rdr_service.dao.biobank_stored_sample_dao import BiobankStoredSampleDao
+from rdr_service.dao.genomics_dao import GenomicSetMemberDao
+from rdr_service.dao.physical_measurements_dao import PhysicalMeasurementsDao
+from rdr_service.model.enrollment_dependencies import EnrollmentDependencies
+from rdr_service.model.participant_summary import ParticipantSummary
+from rdr_service.model.pediatric_data_log import PediatricDataLog, PediatricDataType
+from rdr_service.participant_enums import ParticipantCohortEnum, QuestionnaireStatus, SampleStatus
+from rdr_service.repository.questionnaire_response_repository import QuestionnaireResponseRepository
+from rdr_service.services.system_utils import min_or_none
+from rdr_service.tools.tool_libs.tool_base import cli_run, ToolBase
+
+tool_cmd = 'enrollment-dependencies-backfill'
+tool_desc = 'Backfill enrollment dependencies'
+
+
+class EnrollmentDependenciesBackfill(ToolBase):
+    def run(self):
+        super().run()
+
+        with self.get_session() as session:
+            summary_list = session.query(ParticipantSummary).options(
+                joinedload(ParticipantSummary.guardianParticipants).load_only(),
+                joinedload(ParticipantSummary.pediatricData)
+            ).yield_per(500)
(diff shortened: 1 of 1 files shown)


rdr_service/tools/tool_libs/backfill_enrollment_dependencies.py
```diff
@@ -0,0 +1,127 @@
...
+            overall_health_survey_authored_time=summary.questionnaireOnOverallHealthAuthored,
+            lifestyle_survey_authored_time=summary.questionnaireOnLifestyleAuthored,
+            exposures_survey_authored_time=cls.get_exposures_time(summary.pediatricData),
+            biobank_received_dna_time=cls.get_biobank_dna_time(summary=summary, session=session),
+            wgs_sequencing_time=GenomicSetMemberDao.get_wgs_pass_date(session=session, biobank_id=summary.biobankId),
+            first_ehr_file_received_time=min_or_none([
+                summary.ehrReceiptTime, summary.firstParticipantMediatedEhrReceiptTime
+            ]),
+            first_mediated_ehr_received_time=summary.firstParticipantMediatedEhrReceiptTime,
+            physical_measurements_time=min_or_none([
```
dtharpeuno, line 61: TIL

dtharpeuno (approved):

rdr_service/tools/tool_libs/backfill_enrollment_dependencies.py
```diff
@@ -0,0 +1,127 @@
...
+            summary_list = session.query(ParticipantSummary).options(
+                joinedload(ParticipantSummary.guardianParticipants).load_only(),
+                joinedload(ParticipantSummary.pediatricData)
+            ).yield_per(500)
+            for summary in summary_list:
+                self.create_dependency_data(summary, session)
+
+    @classmethod
+    def create_dependency_data(cls, summary: ParticipantSummary, session):
+        core_measurements = PhysicalMeasurementsDao.get_core_measurements_for_participant(
```
ayazamlani, line 36: since testing manually, should we throw in a log each time dependency data for a participant is created, and log if anything were to go wrong when creating the dependency data?

ayazamlani (approved):

rdr_service/tools/tool_libs/backfill_enrollment_dependencies.py
kenny-skaggs: My plan was to go over the data created checking general integrity (things like: every participant summary should have a dependency object, and there should only be one for each participant). And then go over each field: comparing directly against things like participant summary when the field has essentially been copied, and taking a handful of participants and verifying the data when the field is derived from something (like environment exposures).

Logging when one is created could be helpful if the script was going to be used more than once, but for this once I think that could be alot to look through and redundant with manual checks. Logging if something goes wrong could be good, but it'll likely be easier to spot something that went wrong when looking through the data.

PR #3838, opened 2024-03-19
Create Slack Alert for PTSC Retention File Failures

vmoniqb:
* Updated the retention file import code to catch and track exceptions during import. An alert will be sent to Slack with the file path and the number of rows that failed to import.
* A new Slack channel `rdr-retention-import-alerts` has been created for the PTSC retention file alerts
* Created a new `rdr_retention_status_webhook` to send alerts to Slack and added it to the config file

## Tests
- [X] unit tests

[DA-4169]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/offline/retention_eligible_import.py
```diff
@@ -43,36 +46,62 @@ def import_retention_eligible_metrics_file(task_data):
...
+                continue
 
         if records:
-            upsert_count += dao.upsert_all_with_session(session, records)
+            try:
+                upsert_count += dao.upsert_all_with_session(session, records)
+            except (IntegrityError, InvalidRequestError):
+                failed_records_count += batch_count
+
+        _send_slack_alert(f'gs://{csv_file_cloud_path}', failed_records_count)
```
dtharpeuno: Do we only want this to send it there are alerts? If so not seeing the conditional.

dtharpeuno (commented):
So might be a good idea since this is an issue to store the file name and the counts in a table for historical reference instead of just relying on the slack alert. 
Probably a good idea to ask if this is something we need to accomplish with this feature.

rdr_service/offline/retention_eligible_import.py
```diff
@@ -43,36 +46,62 @@ def import_retention_eligible_metrics_file(task_data):
...
 
     logging.info(f"Updating participant summary retention eligible flags for {upsert_count} participants...")
     ParticipantSummaryDao().bulk_update_retention_eligible_flags(upload_date)
     logging.info(f"Import and update completed for gs://{csv_file_cloud_path}")
 
 
+def _send_slack_alert(file_name, failed_records):
```
dtharpeuno: Since this isnt part of a class dont think this needs to be denoted as a protected method.

rdr_service/offline/retention_eligible_import.py
vmoniqb: I'm going to add in a table to store the errors and added the conditional I forgot. Thanks for that call out!

rdr_service/offline/retention_eligible_import.py
```diff
@@ -43,36 +46,63 @@ def import_retention_eligible_metrics_file(task_data):
...
+                continue
 
         if records:
-            upsert_count += dao.upsert_all_with_session(session, records)
+            try:
+                upsert_count += dao.upsert_all_with_session(session, records)
+            except (IntegrityError, InvalidRequestError):
+                failed_records_count += batch_count
+
+        if failed_records_count > 0:
```
dtharpeuno: :nit `failed_records_count` is initialized as `0` which is a "Falsey" value. So  `if failed_records_count` works here.

rdr_service/dao/retention_status_import_failures_dao.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class RetentionStatusImportFailuresDao(BaseDao):
+
+    def __init__(self):
+        super(RetentionStatusImportFailuresDao, self).__init__(RetentionStatusImportFailures)
+
+    def insert(self, obj: RetentionStatusImportFailures):
+        with self.session() as session:
+            self.insert_with_session(session, obj)
```
dtharpeuno: `insert` is part of the BaseDao so no need to have this here.

rdr_service/dao/retention_status_import_failures_dao.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def get(self, failure_id):
```
dtharpeuno: Can we rename this to signify what the method is doing... ex `get_obj_from_failure_id` as to not override the standard `get` from the base class.

dtharpeuno (approved):

pegbertsch (approved):

PR #3853, opened 2024-04-05
[DA-4204] Adding Pagination to workbench/audit/workspace/snapshots API

vmoniqb:
RH requested the ability to call the `workbench/audit/workspace/snapshots` API and return all workspaces. Previously, these calls have been failing due to the number of records in the response. 

This PR adds pagination to the API through both `workbench/audit/workspace/snapshots` and  `workbench/audit/workspace/snapshots?last_snapshot_id=<ID>` so that both can be used to fetch all workspaces.

## Tests
- [X] unit tests

[DA-4204]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/api/base_api.py
+++ b/rdr_service/api/base_api.py
@@ -22,7 +22,6 @@ from rdr_service.cloud_utils.gcp_cloud_tasks import GCPCloudTask
 
 
 DEFAULT_MAX_RESULTS = 100
-MAX_MAX_RESULTS = 6000
 
 
 def log_api_request(log: RequestsLog = None, model_obj=None):
@@ -130,6 +129,7 @@ class BaseApi(Resource, ApiUtilMixin):
     def __init__(self, dao, get_returns_children=False):
         self.dao = dao
         self._get_returns_children = get_returns_children
+        self.max_max_results = 10000
 
     def _get_request_arg_bool(self, key, default=False):
         """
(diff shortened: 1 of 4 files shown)


kenny-skaggs (approved):

dtharpeuno (approved):

rdr_service/api/base_api.py
```diff
@@ -22,7 +22,7 @@
 
 
 DEFAULT_MAX_RESULTS = 100
-MAX_MAX_RESULTS = 10000
+MAX_MAX_RESULTS = 6000
```
kenny-skaggs: This update affects the Paricipant Summary API too. Is that intended, or should the code be updated to only set this as the max for the research workbench API's?

kenny-skaggs (approved):

rdr_service/api/base_api.py
vmoniqb: This was intended, but after discussion in Slack, I will override the value instead to avoid any errors for awardees currently calling the PaarticipantSummary API with `_count > 6000`

PR #3858, opened 2024-04-12
[ROC-1878] Fixing cache bug in enrollment dependencies

kenny-skaggs:
## Resolves *[ROC-1878](https://precisionmedicineinitiative.atlassian.net/browse/ROC-1878)*
This addresses a bug in the code for storing enrollment dependency data. Some participants have authored dates in their participant summary for surveys and consents, but their enrollment dependencies record has nulls for those values instead. It appears that the caching mechanism is to blame in this. A record can be in the cache, but the session associated with the record may have already been committed and closed. So if this record is retrieved from the cache and used again, any further modifications wouldn't persist (since the session wouldn't be told to commit again).

This updates the code to reinitialize any cached objects on the new session. This way any further modifications to the record's data will be persisted when the new session commits.

## Tests
- [x] unit tests

[ROC-1878]: https://precisionmedicineinitiative.atlassian.net/browse/ROC-1878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/dao/enrollment_dependencies_dao.py
+++ b/rdr_service/dao/enrollment_dependencies_dao.py
@@ -15,8 +15,12 @@ class EnrollmentDependenciesDao:
     def get_enrollment_dependencies(cls, participant_id: int, session: Session) -> Optional[EnrollmentDependencies]:
         if participant_id in cache:
             cached_record = cache[participant_id]
-            merge_result = session.merge(cached_record)
-            return merge_result
+            if Session.object_session(cached_record) == session:
+                return cached_record
+            else:
+                merge_result = session.merge(cached_record)
+                cache[participant_id] = merge_result
+                return merge_result
 
         result = session.query(EnrollmentDependencies).filter(
             EnrollmentDependencies.participant_id == participant_id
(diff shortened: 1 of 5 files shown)


rdr_service/dao/enrollment_dependencies_dao.py
```diff
@@ -14,7 +14,13 @@ class EnrollmentDependenciesDao:
     @classmethod
     def get_enrollment_dependencies(cls, participant_id: int, session: Session) -> Optional[EnrollmentDependencies]:
         if participant_id in cache:
-            return cache[participant_id]
+            cached_record = cache[participant_id]
+            if Session.object_session(cached_record) == session:
```
dtharpeuno, line 18: Nice, whats data type does this use to evaluate for equivalency?

dtharpeuno (approved):

brianmkimmel (approved):

rdr_service/dao/enrollment_dependencies_dao.py
kenny-skaggs: I'm not sure. I'm guessing it's checking if they're both references to the same Session object, but I could be wrong on that.

PR #3884, opened 2024-05-14
[DA-4295] Adding asynchronicity to BigQueryJob

kenny-skaggs:
## Resolves *[DA-4295](https://precisionmedicineinitiative.atlassian.net/browse/DA-4295)*
The UpdateEhrStatusOrganization cron job is failing because of a timeout. It is likely from the BigQuery job taking just a bit long enough compared to when it was working. When submitting the BigQuery request, we have a timeout default of 10 minutes, meaning the request wouldn't return until the 10 minutes are up or it's complete. I believe the timeout we're seeing is from a default in the library being used in the Google SDK to make the HTTP request.

Asynchronous requests are possible with the BigQuery API. Setting a short timeout (20 seconds for the UpdateEhrStatusOrganization) would give us a response that the query isn't yet complete, but would give us the job id so we can make another request to check on it again. Checking for results uses the same API endpoint we're using for getting the next page of results.

## Tests
- [ ] unit tests

[DA-4295]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/cloud_utils/bigquery.py
+++ b/rdr_service/cloud_utils/bigquery.py
@@ -70,18 +70,33 @@ class BigQueryJob(object):
             self._service.jobs().query(projectId=self.project_id, body=job_body).execute(num_retries=self.retry_count)
         )
         self._job_ref = self.JobReference(**result["jobReference"])
+
+        if not result['jobComplete'] and 'errors' not in result:
+            result = self._retrieve_query_results()
+
         page_token = result.get("pageToken")
         if page_token:
             self._page_token = page_token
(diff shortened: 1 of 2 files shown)


ayazamlani (approved):
👍

dtharpeuno (approved):

PR #3893, opened 2024-05-29
Fix for the ParticipantCountsOverTime Job

vmoniqb:
* The PublicMetrics API is returning incorrect metrics due to the  ParticipantCountsOverTime Job excluding participants with `is_ghost_id = null`. This PR updates the `participant_sql` query that populates the `temp_tables` to include all participants with `is_ghost_id = null`. This should correct the metrics cache tables and the metric counts for the PublicMetrics API.
* Previously, the job would DROP all temp_tables at the beginning and at the end of each job run. To allow for easier troubleshooting in the future, we want to only delete the temp_tables at the beginning of the job. This PR removes the dropping of tables at the end of the job.

## Tests
- [X] unit tests
- Ran the updated `participant_sql` query manually on BigQuery in AoU Warehouse PreProd. It now includes participants with `is_ghost_id = null` as expected. 1,171,640 records were returned when not filtering for a specific HPO id.
- Manually ran the `init_tmp_tables` portion of the job with AoU Warehouse PreProd to ensure that the temp tables are populated as expected. In total, 1,171,640 rows were populated across all temp tables which matches the number of rows when the `participant_sql` query was run in BigQuery

### DIFF
--- a/rdr_service/dao/participant_counts_over_time_service.py
+++ b/rdr_service/dao/participant_counts_over_time_service.py
@@ -43,7 +43,7 @@ class ParticipantCountsOverTimeService(BaseDao):
             LEFT JOIN `{participant_summary_bq_table}` AS ps
             ON p.participant_id = ps.participant_id
             WHERE p.hpo_id != @test_hpo_id
-            AND p.is_ghost_id != 1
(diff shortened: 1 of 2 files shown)


j-kanuch (approved):

shahwaiz14 (approved):
Thanks for fixing this :-)

PR #3903, opened 2024-06-11
[DA-4229] Writing sample status data to new tables

kenny-skaggs:
## Partially Resolves *[DA-4229](https://precisionmedicineinitiative.atlassian.net/browse/DA-4229)*
To reduce the number of columns in the participant summary table, we're moving the sample status values out into their own tables. 

This creates two new tables to hold the status values for sample order and receipt. The API code is updated to also write incoming data to these tables while leaving the current process of writing data to the participant summary in place. The next step will be to backfill the new tables with existing data and compare after some time to be sure the new system of writing the statuses matches the participant summary workflow.

## Tests
- [x] unit tests

[DA-4229]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/dao/sample_summary_dao.py
```diff
@@ -0,0 +1,94 @@
...
+        if receipt_data:
+            sample_list = session.query(BiobankStoredSample).filter(
+                BiobankStoredSample.biobankId == new_sample.biobankId,
+                BiobankStoredSample.test == new_sample.test
+            ).all()
+            sample_list = [*sample_list, new_sample]
+
+            status_sample = None
+            non_disposed_samples = [
+                sample for sample in sample_list if sample.status < SampleStatus.SAMPLE_NOT_RECEIVED
```
ayazamlani, line 49: Neat way to filter for the good and bad samples, but if more good sampleStatus Enums are added this may break later. It may be better to put the good and bad sample status enums in separate lists?

ayazamlani (approved):
Looks great

vmoniqb (approved):

rdr_service/dao/sample_summary_dao.py
kenny-skaggs: Yeah, it'd be good to deal with this kind of thing before it becomes a problem. But for this work I'm just looking to replicate the existing logic in the SQL that processes the statuses (just to reduce the chances of unintended changes/bugs 😅).

PR #3904, opened 2024-06-12
[DA-4293] Create ETM Response Duplicate Detector

brianmkimmel:
Creates a cron job to identify and set the classification type of EtM responses which duplicate an existing response as DUPLICATE.

## Tests
- [x] unit tests

[DA-4293]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kenny-skaggs (approved):

rdr_service/offline/etm_duplicate_detector.py
```diff
@@ -0,0 +1,146 @@
...
+                pk_columns=["etm_questionnaire_response_id"],
+                pk_values=duplicate_responses,
+                project=project,
+            )
+
+            logging.info(
+                f"Sent PubSub notifications to mark {len(duplicate_responses)} records for deletion."
+            )
+
+        # TODO: Once the new pipeline is fully operational, the following code can be deleted:
```
pegbertsch: @brianmkimmel  Because the EtM data was never transferred through our old RDR-PDR pipeline and only implemented in our new pipeline, you can skip everything beyond the TODO.   That activity is still needed for regular `questionnaire_response` records that get marked as duplicate since those were part of the old pipeline, which is why it exists in the other response duplication detector code.

pegbertsch (changes requested):
Should remove some PDR pipeline related code that does not apply to `etm_questionnaire_response` records (only needed for updates to `questionnaire_response` records)

rdr_service/offline/etm_duplicate_detector.py
brianmkimmel: @pegbertsch Thanks for explaining that. I've updated to remove that part of the code.

pegbertsch (approved):

PR #3919, opened 2024-06-27
PPSC Intake API initial implementation

j-kanuch:
This PR implements the first iteration of the PPSC Intake API. 
[Tech Design](https://docs.google.com/document/d/1W-Qxc1gOk40HYIzDwDWmwU1dwvOe0RJ1BMxlSJPd47c/edit?usp=sharing)
Includes the Consent, Survey Completion, and Profile Updates participant activities, per PPSC's request. 
Specific data elements for these activities are still being negotiated.
The specific values for Activities and their related EventTypes are defined in the base_config. The API validates against these values.
Future iterations will include:
- More activity types
- Cleaner validation using a library or tool such as ValidX or Cerberus once the data element array for each activity is better defined.
- The tool currently accepts all values for data_elements. Specific event types should have specific payload schemas which will be enforced, once these are defined/negotiated.

## Tests
- [x] unit tests

[DA-4353]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/dao/ppsc_dao.py
```diff
@@ -22,5 +24,16 @@ def __init__(self, model_type):
     def from_client_json(self):
         pass
 
+    def to_client_json(self, payload):
+        # pylint: disable=unused-argument
+        return "Event Record Created"
```
dtharpeuno: Should we return the participant Id with this?
"Event record created for: P<pid>"

rdr_service/api/ppsc_intake_api.py
```diff
@@ -0,0 +1,123 @@
...
+            'resource': req_data
+        }
+
+        # Validate participant ID and insert
+        try:
+            participant_event_activity = self.participant_event_activity_dao.insert(
+                self.participant_event_activity_dao.model_type(**participant_event_activity_dict)
+            )
+        except IntegrityError:
+            raise NotFound(f"Participant with ID {req_data['participantId']} not found")
```
dtharpeuno, line 97: Will this be thrown on the insert if the FK does not exist? Should we add this check to the validate_payload method so nothing gets thru to this point?

tests/api_tests/test_ppsc_intake_api.py
```diff
@@ -0,0 +1,460 @@
...
+        self.assertEqual('The Basics', survey_events[1].event_type_name)
+        self.assertEqual('activity_date_time', survey_events[1].data_element_name)
+        self.assertEqual("2024-05-20T14:30:00Z", survey_events[1].data_element_value)
+
+    def test_intake_profile_updates_event_type_validation(self):
+        participant = self.ppsc_data_gen.create_database_participant()
+
+        payload = {
+            "activity": "Profile Updates",
+            "eventType": "Pepperoni",
```
dtharpeuno, line 368: 🍕

dtharpeuno (approved):
Nice job, left a few comments.

rdr_service/dao/ppsc_dao.py
j-kanuch: Good suggestion. I'll add that.

rdr_service/api/ppsc_intake_api.py
j-kanuch: For performance reasons, I was trying to avoid a second call to the database to lookup the PID. It may not be that big of a performance hit. This method here does feel a little messy. This is on the insert to `participant_event_activity`, where the FK does exist, so for now it should work. You suggestion is something I  could try on a future iteration when I revisit the validation.

shahwaiz14 (approved):

PR #3931, opened 2024-07-19
Creating Questionnaire Answers Backfill Tool for DA-4318

vmoniqb:
Creating a backfill tool to add `questionnaire_response_answer` records that are missing for the ips_6 concept code. The tool takes in `questionnaire_ids` for the questionnaires that have the missing concept code and inserts the corresponding `questionnaire_response_answers` that are included in  each participant's questionnaire response.

## Tests
- [] unit tests
- Tested in Sandbox
    - Ran queries to correct the `questionnaire`, `questionnaire_history`, and `questionnaire_question` tables
    - Ran the tool to insert missing records into the `questionnaire_response_answer` table
    - Sent a new POST QuestionnaireResponse request to the corrected questionnaire & the `questionnaire_response_answer` table was updated correctly with the response to the ips_6 question

[DA-4318]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kenny-skaggs (approved):

rdr_service/tools/tool_libs/backfill_questionnaire_answers.py
```diff
@@ -0,0 +1,169 @@
...
+
+                        # Now add the child answers, using the IDs in code_id_map
+                        answers_to_insert = self._add_answers(code_id_map, answers,
+                                                              response_data.questionnaireResponseId)
+                        session.add_all(answers_to_insert)
+
+                    latest_id = response_data.questionnaireResponseId
+
+                if found_responses:
+                    print(f'got to {latest_id}')
```
pegbertsch: May want to move these print statements down after the `commit`, to confirm it was successful and nothing got rolled back?

If the tool script fails at any point (can happen with long-running scripts due to failed network connections, etc.), I assume before any subsequent run, the starting value of `latest_id` will be changed in the code and instead of -10, will be set to the last value printed out on the previous run.   Need to make sure the same `questionnaire_response_id` can't be processed by this script twice, or we may end up with duplicate/extra `questionnaire_response_answer` records.

pegbertsch (commented):
Will likely take a while to get through the `found_responses` while loop since we expect ~25K records to need processing.   Sometimes long-running tool scripts can terminate unexpectedly due to closed/lost connections, etc.

I assume if the tool has to be restarted, the starting value of the `latest_id` variable will be updated first to reflect that last value printed out on the previous run

rdr_service/tools/tool_libs/backfill_questionnaire_answers.py
vmoniqb: Good call out! I will update to make sure we don't reprocess any `questionnaire_response_ids`

pegbertsch (approved):

PR #3934, opened 2024-07-23
Added new PPSC intake models, tests.

j-kanuch:
This PR adds the following Activities in the PPSC Intake API:
- Withdrawal
- Deactivation
- Participant Status
- Site Attribution

Models and tests were updated to include the above Activities. 

## Tests
- [x] unit tests


dtharpeuno (approved):
Nice job. 
I am working on the design for the site api, so maybe after that is implemented we can add an extra validation step to verify the site exists if they send us a Site Attribution payload.

vmoniqb (approved):

PR #3981, opened 2024-09-23
[DA-4497] Add PUT to dlwdosage API

shahwaiz14:
- Added PUT request in dlwdosage api
- Updated POST to send the unique id in the response json when it adds a new row to the table

## Tests
✅  unit tests

[DA-4497]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/dao/base_dao.py
+++ b/rdr_service/dao/base_dao.py
@@ -774,7 +774,7 @@ class UpdatableDao(BaseDao):
     that it matches.
     """
         if not existing_obj:
-            raise NotFound(f"{self.model_type.__name__} with id {id} does not exist")
+            raise NotFound(f"{self.model_type.__name__} with id {obj.id} does not exist")
         if self.validate_version_match and existing_obj.version != obj.version:
             raise PreconditionFailed(
                 f"Expected version was {obj.version}; stored version was {existing_obj.version}"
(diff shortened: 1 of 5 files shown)


rdr_service/dao/base_dao.py
```diff
@@ -774,7 +774,7 @@ def _validate_update(self, session, obj, existing_obj):
     that it matches.
     """
         if not existing_obj:
-            raise NotFound(f"{self.model_type.__name__} with id {id} does not exist")
+            raise NotFound(f"{self.model_type.__name__} with id {getattr(obj, 'id', id)} does not exist")
```
shahwaiz14, line 777: I initially updated this line to use `obj.id`. But Participant DAO gave an error because the participant table does not have a column named `id`. I’ve updated the code to check for the presence of an id in the model. If it exists, the code will use that; if not, it will continue with the previous behavior.

j-kanuch (approved):

kenny-skaggs (approved):

PR #4032, opened 2024-12-31
[DA-4616] AwardeeInSite Backend Job & API

shahwaiz14:
This PR includes:
1. **Backend Job** that reads from the necessary tables in BQ, stores that data to an awardee_insite staging table in BQ, and then streams new and updated data to MySQL table.
2. **API**: Contains support for pagination, and allows retrieval of participant since a specified date. It also enforces awardee level participant return.
3. Publishes Awardee InSite Fields under Read the Docs

## Tests
- ✅  unit tests

- Also tested in Staging

[DA-4616]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/dao/awardee_insite_dao.py
```diff
@@ -0,0 +1,165 @@
...
+    def __init__(self):
+        super().__init__(AwardeeInSite)
+        self.total_result = None
+
+    @staticmethod
+    def snake_to_camel(string_value: str) -> str:
+        mod_string = string_value.split('_')
+        string = mod_string[0] + ''.join(x.title() for x in mod_string[1:])
+        # Replace sal with SAL
+        return re.sub(r'(\d)sal(\d)', r'\1SAL\2', string, flags=re.IGNORECASE)
```
dtharpeuno: This already exists in the BaseDao, maybe call super for the initial value and add later processing logic for SAL?

rdr_service/dao/awardee_insite_dao.py
```diff
@@ -0,0 +1,165 @@
...
+                    self._make_pagination_token(
+                        items[-1][0].asdict(), order_by_field_names
+                    )
+                )
+                if query_definition.always_return_token
+                else None
+            )
+            return Results(
+                items, pagination_token, more_available=False, total=self.total_result
+            )
```
dtharpeuno: How is this different from the BaseDao’s method? Can you not leverage that instead of repeating this?

dtharpeuno (approved):

rdr_service/dao/awardee_insite_dao.py
shahwaiz14: Good point, there are some differences between the two but the main difference is that base dao’s method calculates the total count in this method while mine doesn’t calculate the total count. The total count is calculated else where after all the filters are applied to the query. Also the structure of the `items` variable is a little different. Since I am adding hpo.name to the query, the `items` variable contains [awardee_model, HPO.name]. That’s why I would like to keep my method here.

rdr_service/dao/awardee_insite_dao.py
shahwaiz14: thanks. Done!

rdr_service/dao/awardee_insite_dao.py
```diff
@@ -0,0 +1,164 @@
...
+class AwardeeInSiteDao(UpsertableDao):
+    def __init__(self):
+        super().__init__(AwardeeInSite)
+        self.total_result = None
+
+    @staticmethod
+    def snake_to_camel(string_value: str) -> str:
+        string = UpsertableDao.snake_to_camel(string_value)
+        # Replace sal with SAL
+        return re.sub(r'(\d)sal(\d)', r'\1SAL\2', string, flags=re.IGNORECASE)
```
dtharpeuno: If you lean into the instance you can just use `self.snake_to_camel`

shahwaiz14: Done!

dtharpeuno (approved):

shahwaiz14:
I have added code to create awardee insite doc at https://all-of-us-raw-data-repository.readthedocs.io/en/latest/api_workflows/api_resource_ref.html

Tested it by running `python -m sphinx -T -E -b html -d _build/doctrees -D language=en . ../_readthedocs/html`

j-kanuch (approved):

PR #4035, opened 2025-01-06
Modify gunicorn entrypoint for all RDR APIs to add keep-alive

j-kanuch:
This PR updates ~app_base.yaml~ gunicorn_config.py with a `keep-alive` time of 700s. 
This is an attempt to mitigate the intermittent 502 bad gateway errors. The source behind this modification comes from Google's documentation on the subject of [App Engine 502 Errors](https://cloud.google.com/appengine/docs/flexible/troubleshooter/serving#app-container-unreachable).
The keep-alive time is set higher than the Nginx time of 600s. The number of workers is set to 8 per Google's instance class recommendations.

## Tests
- [x] unit tests should pass
- Tested locally and on Sandbox.

### DIFF
--- a/rdr_service/app_base.yaml
+++ b/rdr_service/app_base.yaml
@@ -8,7 +8,7 @@ env_variables:
 
 runtime: python311
 
-entrypoint: python3 main.py --service
+entrypoint: gunicorn -w 5 -b :$PORT --keep-alive 700 main:app
 
 # Send warmup requests to instances on startup, which will initialize the app and load configs.
 inbound_services:
--- a/rdr_service/app_base.yaml
+++ b/rdr_service/app_base.yaml
@@ -8,7 +8,7 @@ env_variables:
 
 runtime: python311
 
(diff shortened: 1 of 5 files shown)


kenny-skaggs:
It could be good to update the gunicorn config and reference it with a -c flag (like supervisor's doing at https://github.com/all-of-us/raw-data-repository/blob/devel/rdr_service/services/supervisor.conf#L42). Maybe the bind parameter in there would need to be changed (since we wouldn't be putting gunicorn behind supervisor anymore)? But there are some other settings in there we may want to continue using (like the max_requests, threads, and memory monitoring function).

j-kanuch:
Thanks @kenny-skaggs , I'll take a look at the supervisor conf. I'll probably revert the entrypoint in the yaml and then just append the keep-alive to the conf's command, keeping the rest of the parameters as is.

dtharpeuno (approved):

kenny-skaggs (approved):

PR #4038, opened 2025-01-08
[DA-4588] Migrating rdr schema fields ppsc schema

kenny-skaggs:
## Resolves *[DA-####](https://precisionmedicineinitiative.atlassian.net/browse/DA-4588)*
Adding script for migrating data from tables in the rdr schema to event tables in the ppsc schema. Also reads an exported ubr CSV (not included) for creating ubr data in the ppsc schema.

## Tests
- [ ] unit tests


dtharpeuno (commented):
Sorry if I am not seeing this here but are we moving the participants records over as well to the `ppsc.participant` table? If we do I think we should use something to tag them as legacy also in that table.

@j-kanuch I think we should do this, if it already hasn't been done.

kenny-skaggs:
> Sorry if I am not seeing this here but are we moving the participants records over as well to the `ppsc.participant` table? If we do I think we should use something to tag them as legacy also in that table.
> 
> @j-kanuch I think we should do this, if it already hasn't been done.

I had originally had code to migrate them over, but I think Josh already loaded them into the PPSC schema when we started getting payloads for existing participants. I don't know if there's something in place currently to mark them as legacy/migrated.

dtharpeuno:
> > Sorry if I am not seeing this here but are we moving the participants records over as well to the `ppsc.participant` table? If we do I think we should use something to tag them as legacy also in that table.
> > @j-kanuch I think we should do this, if it already hasn't been done.
> 
> I had originally had code to migrate them over, but I think Josh already loaded them into the PPSC schema when we started getting payloads for existing participants. I don't know if there's something in place currently to mark them as legacy/migrated.

Understood. thanks!

dtharpeuno (approved):

j-kanuch (approved):

PR #4049, opened 2025-01-17
Resolve intermittent RebuildParticipantTaskApi failures

pegbertsch:
We're no longer guaranteed (post-PPSC launch) every `participant_summary` record has default content in the enrollment status fields.  Null values lead to the old PDR generator code throwing a TypeError trying to cast expected Enum class fields, and the record won't be built/saved

We've mostly transitioned to getting these particular values from our new pipeline generators, so we don't care if the fields end up null.   This is just to allow the old generator to succeed so other elements/nested records still get built and synced to BigQuery.

## Tests
- [x] unit tests


robabram (approved):

pegbertsch:
Closing, should be superceded by #4072

PR #4062, opened 2025-02-24
[DA-4657] Updating datafeed query to use physical measurement table

kenny-skaggs:
## Resolves *[DA-4657](https://precisionmedicineinitiative.atlassian.net/browse/DA-4657)*
Updates the AwardeeInSite datafeed query to reference the physical measurement table rather than the participant summary. Clinic fields are set based on the most recently received measurement that is not cancelled. If all measurements have been cancelled then we set the status as "cancelled". Self-reported measurements use the data from the latest self-reported measurement.

## Tests
- [ ] unit tests

[DA-4657]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rdr_service/workflow_management/ppsc/data_feed_queries.py
```diff
@@ -689,18 +754,16 @@ def insert_awardee_insite_data(
...
-                  WHEN 2 THEN 'cancelled'
+              , CASE
+                  WHEN clinic_physical_measurements_id IS NOT NULL THEN 'completed'
+                  WHEN cancelled_measurement_id IS NOT NULL THEN 'cancelled'
                   ELSE 'unset'
                 END AS clinic_physical_measurements_status
               , clinic_physical_measurements_finalized_time
               , clinic_physical_measurements_finalized_site
-              , CASE self_reported_physical_measurements_status
+              , CASE self_reported_physical_measurements_id IS NOT NULL
```
shahwaiz14, line 764: I think we need to take out `IS NOT NULL` since it gives an error in BQ and add back ELSE 'unset' in the CASE statement :)

kenny-skaggs: Ah, yeah I just tried in the BQ web page for running queries and get an error saying I can't compare the boolean to an int like that. Would something like this work? (it seems like this works in the webpage)
```
              , CASE self_reported_physical_measurements_id IS NOT NULL
                WHEN False then 'unset'
                WHEN True then 'completed'
```

Or is there something about how this would would be run that has more constraints?

If we just took out the `IS NOT NULL` part I don't think it would function correctly. My goal with this is to check if there is a self-reported measurement, if there is then the id wouldn't necessarily be 1 but if there isn't then it's definitely NULL.

shahwaiz14: yep that should work!!

kenny-skaggs: Updated!

dtharpeuno (approved):

shahwaiz14 (approved):

PR #4069, opened 2025-03-11
[DA-4729] Update Participant Generator to set `participant_origin` to PPSC

vmoniqb:
The participant generator is being used to create Participant records in the RDR when payloads are received from PPSC. Initially, the `participant_origin` value was not updated to be `ppsc` for these new records because the change created countless unit tests to break. However, the Biobank daily received report is now being created with `participant_origin = 'example'` for PPSC participants which needs to be corrected.

These changes update the participant generator to set `participant_origin = 'ppsc'` for newly created participants except for when the generator is run for unit tests.

## Tests
- [X] unit tests

[DA-4729]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/data_gen/generators/data_generator.py
+++ b/rdr_service/data_gen/generators/data_generator.py
@@ -45,7 +45,8 @@ from rdr_service.participant_enums import PatientStatusFlag, QuestionnaireRespon
 
 
 class DataGenerator:
-    def __init__(self, session, faker):
+    def __init__(self, session, faker, project):
+        self.project = project
         self.session = session
         self.faker = faker
         self._next_unique_participant_id = 900000000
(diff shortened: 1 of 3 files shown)


j-kanuch (approved):
Thanks for fixing this!

rdr_service/data_gen/generators/data_generator.py
```diff
@@ -327,12 +328,16 @@ def _participant_with_defaults(self, **kwargs):
...
+            'participantOrigin': 'ppsc',
             'version': 1,
             'lastModified': datetime.now(),
             'signUpTime': datetime.now(),
             'isTestParticipant': False
         }
+
+        if self.project == 'test':
+            defaults.update(participantOrigin='example')
+
```
dtharpeuno: There is no need here to add another variable to the constructor, we have env variables for unit test flags. 

```
if os.environ.get('UNITTEST_FLAG') == “1":
             defaults.update(participantOrigin='example
```

dtharpeuno (approved):

PR #4077, opened 2025-04-04
[DA-4785]  Ensure Test PIDs Won't Go Out on AW0s Moving Forward

lancealopez:
Resolves [DA-4785](https://precisionmedicineinitiative.atlassian.net/browse/DA-4785)

Changed criteria when selecting new biobank participants to ensure that we do not send test participants on the AW0

## Tests
Modified unit tests to create a test participant

[DA-4785]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/genomic/genomic_data.py
+++ b/rdr_service/genomic/genomic_data.py
@@ -416,6 +416,7 @@ class GenomicQueryClass:
         WHERE TRUE
             AND ss.test in ('1ED04', '1ED10', '1SAL2')
             AND m.id IS NULL
+            AND p.is_test_participant != 1
         """
 
     # BEGIN Data Quality Pipeline Report Queries
--- a/rdr_service/genomic/genomic_data.py
+++ b/rdr_service/genomic/genomic_data.py
@@ -413,10 +413,14 @@ class GenomicQueryClass:
             LEFT JOIN genomic_set_member m ON m.participant_id = ps.participant_id
                     AND m.genomic_workflow_state <> :ignore_param
             LEFT JOIN biobank_mail_kit_order mk ON mk.participant_id = p.participant_id
+            LEFT JOIN ppsc.participant_status_event pse on pse.participant_id = p.participant_id
+                    AND pse.event_type_name = "Test Account"
+                    AND pse.data_element_name = "activity_status"
+                    AND pse.data_element_value = "test"
         WHERE TRUE
             AND ss.test in ('1ED04', '1ED10', '1SAL2')
             AND m.id IS NULL
-            AND p.is_test_participant != 1
+            AND pse.id IS NULL
         """
 
     # BEGIN Data Quality Pipeline Report Queries
--- a/tests/genomics_tests/test_genomic_pipeline.py
+++ b/tests/genomics_tests/test_genomic_pipeline.py
@@ -6,6 +6,7 @@ import operator
 import pytz
 import random
 import time
+import http.client
 
 from copy import deepcopy
 from dateutil.parser import parse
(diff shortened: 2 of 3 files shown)


rdr_service/genomic/genomic_data.py
```diff
@@ -416,6 +416,7 @@ def new_biobank_samples():
         WHERE TRUE
             AND ss.test in ('1ED04', '1ED10', '1SAL2')
             AND m.id IS NULL
+            AND p.is_test_participant != 1
```
dtharpeuno: @lancealopez I think this works for now, but in the PPSC schema we denote the participant as being test in a different way. 

I would verify first with @j-kanuch about our plan to turn off the backend sync job to the participant summary tables.

Also another pitfall is things keep changing from PPSC on how they are sending us test participants so not sure of the backend sync job os denoting that correctly to the legacy data.

j-kanuch: Right now, we're marking `rdr.participant.is_test_participant` as `true` if `ppsc.participant_status_event` records match the following criteria:  
- `event_type_name` = "Test Account"  
- `data_element_name` = "activity_status"
- `data_element_value` = "test"

The long-term goal is to deactivate this feedback to the legacy system. If we could future proof this section of the code, that would be preferable. We'd want to add an outer left join to the `ppsc.particpant_status_event` table with the above filters. This is similar to what we David just implemented in the NPH Optin Sync (see https://github.com/all-of-us/raw-data-repository/pull/4076)

lancealopez: @j-kanuch @dtharpeuno I have changed the filter criteria and test case to use the participant status event table instead of test flag.

tests/genomics_tests/test_genomic_pipeline.py
```diff
@@ -1401,6 +1407,60 @@ def test_new_participant_workflow(self):
...
+                    "Consent",
+                    "Survey Completion",
+                    "Profile Updates",
+                    "Withdrawal",
+                    "Deactivation",
+                    "Participant Status",
+                    "Attribution",
+                    "NPH Opt In",
+                    "Account Linkage"
+                ]
```
dtharpeuno: Can you only include the activity you need?

tests/genomics_tests/test_genomic_pipeline.py
```diff
@@ -1401,6 +1407,60 @@ def test_new_participant_workflow(self):
...
+                        'id': 10,
+                        'biobank_id': 100010,
+                    }
+                )
+                payload = {
+                    'participantId': 'P10',
+                    'biobankId': 'T100010',
+                    'registeredDate': '2024-03-26T13:24:03.935Z'
+                }
+                self.send_post('createParticipant', request_data=payload,expected_status=http.client.FORBIDDEN)
```
dtharpeuno: Why is the expected status FORBIDDEN here, doesn't that mean the participant record isn't getting created from the endpoint? I think the other method, before this call, is creating the record.

tests/genomics_tests/test_genomic_pipeline.py
lancealopez: @dtharpeuno I have removed the extra activities.

tests/genomics_tests/test_genomic_pipeline.py
lancealopez: @dtharpeuno I have removed the unneeded post call.

dtharpeuno (approved):

PR #4080, opened 2025-04-15
[DA-4653] Primary Consent Intake API Validation

DaveK-14:
Added validation for Primary Consent, checking whether the current event is a primary consent and if not, checking whether one exists as lightweight as possible.

## Tests
- [tests/api_tests/test_ppsc_intake_api.py] unit tests

[DA-4653]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/api/ppsc_intake_api.py
+++ b/rdr_service/api/ppsc_intake_api.py
@@ -88,7 +88,9 @@ class PPSCIntakeAPI(BaseApi):
 
         # Check for Primary Consent
         if req_data['eventType'] != "Primary Consent":
-            if not self.check_primary_consent(req_data['participantId'].split('P')[1]):
+            if not self.check_primary_consent(req_data['participantId'].split('P')[1],
+                                              'Primary Consent',
+                                              'activity_status', 'Yes'):
                 raise BadRequest("No Primary Consent record found.")
 
     def handle_event_insert(self, *, req_data: dict) -> dict:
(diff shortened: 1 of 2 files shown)


rdr_service/api/ppsc_intake_api.py
```diff
@@ -86,6 +86,17 @@ def validate_payload(self, *, req_data: dict):
         else:
             raise BadRequest("No activity_date_time_value provided.")
 
+        # Check for Primary Consent
+        if req_data['eventType'] != "Primary Consent":
+            with self.dao.session() as session:
+                is_primary = session.query(ConsentEvent).filter(
+                    ConsentEvent.participant_id == req_data['participantId'].split('P')[1],
+                    ConsentEvent.event_type_name == 'Primary Consent'
```
j-kanuch: Can you add two more conditions?
`ConsentEvent.data_element_name == 'activity_status'`
and 
`ConsentEvent.data_element_value == 'yes'`

Also, can you put this logic into its own method?

j-kanuch: Also, if this is in its own method, we might want to parameterize the `event_type_name`, `data_element_name` and `data_element_value` so we can check other consents if the need arised, i.e. EHR, DICOM, etc.

rdr_service/api/ppsc_intake_api.py
```diff
@@ -90,7 +90,8 @@ def validate_payload(self, *, req_data: dict):
         if req_data['eventType'] != "Primary Consent":
             if not self.check_primary_consent(req_data['participantId'].split('P')[1],
```
j-kanuch: Thanks, this looks good. I just thought of a couple minor tweaks. Can you rename the method more generically? Something like "check_consent" since we can check more than just Primary.
Additionally, the DE value is actually "yes" (all lower case). MySQL values aren't case sensitive by default, but Python values are. You can get around this by using `ConsentEvent.data_element_value == ilike(data_element_value)`

DaveK-14: Thanks, that's done. I checked the sql tables to be sure first and was seeing "Yes" instead of "yes" but maybe there was some issue in pycharm. regardless it should be good with the ilike.

j-kanuch (approved):
Thanks for making those changes.

kenny-skaggs (approved):

PR #4096, opened 2025-05-23
Update Intake to PS Datafeed for AIAN

j-kanuch:
This PR modifies the survey completion query. Previously, if a participant had multiple race answers, only the latest was taken into account when calculating AIAN status. This update marks the participant as AIAN if any of their race answers were `WhatRaceEthnicity_AIAN`.
Also cleaned up a variable that wasn't needed for the "sent table" name.

## Tests
- [x] unit tests; tested in BigQuery console as well

[DA-4866]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/workflow_management/ppsc/ppsc_intake_to_ps_queries.py
+++ b/rdr_service/workflow_management/ppsc/ppsc_intake_to_ps_queries.py
@@ -331,8 +331,7 @@ GROUP BY participant_id, event_id, event_type_name
 
 def get_survey_completion_activity_to_stream(project: str,
                                               source_dataset: str,
-                                              temp_table_name: str,
-                                              sent_table_name: str) -> str:
+                                              temp_table_name: str) -> str:
     return f"""
 CREATE OR REPLACE TABLE `{project}.{source_dataset}.{temp_table_name}` AS
 WITH ranked_events AS (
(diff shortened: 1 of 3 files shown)


rdr_service/workflow_management/ppsc/ppsc_intake_to_ps_queries.py
```diff
@@ -368,21 +367,29 @@ def get_survey_completion_activity_to_stream(project: str,
...
+  SELECT
+    participant_id,
+    MAX(CASE
+      WHEN data_element_name = 'race_whatraceethnicity'
+           AND data_element_value = 'WhatRaceEthnicity_AIAN'
+      THEN 1 ELSE 0 END) AS is_aian
+  FROM `{project}.{source_dataset}.ppsc_survey_completion_event`
+  WHERE ignore_flag = 0
+  GROUP BY participant_id
+)
```
kenny-skaggs, line 380: @j-kanuch, maybe I'm reading this wrong. But would this flag a participant as AIAN if they've _ever_ provided that as an answer? If so, should their AIAN status be unset if they ever answer that they're not?

j-kanuch: @kenny-skaggs Yes, you are reading this correctly. I'll need to follow up on what the requirement is, whether we should be updating this if they remove the AIAN answer from their list of race/ethnicity answers.

j-kanuch: @kenny-skaggs According to Mara, we should use the same logic here as what was used with PTSC, the logic being that once set, the `aian` flag is never unset.

kenny-skaggs (approved):

DaveK-14 (approved):

PR #4118, opened 2025-08-15
[DA-4926] Add method to default Duke's null sample_ids to parent

DaveK-14:
Added method to stored_sample intake process that gives child samples their parent's sample_id if not provided.

## Tests
- [x] unit tests

[DA-4926]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/offline/study_nph_biobank_import_inventory_file.py
+++ b/rdr_service/offline/study_nph_biobank_import_inventory_file.py
@@ -93,10 +93,26 @@ def _convert_csv_row_to_stored_sample_object(csv_obj: Dict[str, Union[str, int]]
 
 def import_biobank_inventory_into_stored_samples(csv_filepath: str):
     nph_stored_sample_dao = NphStoredSampleDao()
-    for row in read_nph_biobank_inventory_file(csv_filepath):
+    rows = read_nph_biobank_inventory_file(csv_filepath)
+    parent_fields = {}
+
+    for row in rows:
+        parent_id = row.get('LIMS_PARENT_SAMPLE_ID')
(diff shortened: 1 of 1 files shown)


rdr_service/offline/study_nph_biobank_import_inventory_file.py
```diff
@@ -93,11 +93,27 @@ def _convert_csv_row_to_stored_sample_object(csv_obj: Dict[str, Union[str, int]]
 
 def import_biobank_inventory_into_stored_samples(csv_filepath: str):
     nph_stored_sample_dao = NphStoredSampleDao()
-    for row in read_nph_biobank_inventory_file(csv_filepath):
+    rows = read_nph_biobank_inventory_file(csv_filepath)
+    parent_fields = {}
+
+    for row in rows:
+        parent_id = row.get('LIMS_PARENT_SAMPLE_ID')
```
kenny-skaggs, line 101: When reading cells from a CSV that are empty, I thought we usually get an empty string but it looks like the rest of the code is expecting `None`. Are we definitely going to get `None` here when parsing the file?

DaveK-14: That is a good point, I will cover both cases to be sure

kenny-skaggs (approved):

j-kanuch (approved):

PR #4127, opened 2025-09-16
[DA-4970] Morehouse org updates: Intake and Site Attribution

shahwaiz14:
Morehouse org was historically paired to Awardee SEEC, however this is no longer going to be the case and Morehouse is now going to be under awardee DREF. However due to limitations in PPSC systems, they cannot make a change to send us the right naming conventions. 

This PR updates the the Intake API. When PPSC sends us participant attribution events with `SEEC_MOREHOUSE` as the org pairing, we will transform this so we instead insert `DREF_MOREHOUSE` for the Org pairing.

Also when PPSC sends us Site hierarchy payloads in site API, `SEEC_MOREHOUSE` will be transformed to `DREF_MOREHOUSE`

## Tests
✅  unit tests

[DA-4970]: https://precisionmedicineinitiative.atlassian.net/browse/DA-4970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rdr_service/api/ppsc_site_api.py
+++ b/rdr_service/api/ppsc_site_api.py
@@ -75,7 +75,9 @@ class PPSCSiteAPI(BaseApi):
             raise BadRequest(f'Payload for Site is invalid: Required keys - {response_string}')
 
     def handle_site_updates(self, *, site_data: dict) -> Site:
-        # site data
+        if site_data.get("org_id").upper() == "SEEC_MOREHOUSE":
+            site_data["org_id"] = "DREF_MOREHOUSE"
+
         site_record = self.dao.upsert(self.dao.model_type(**site_data))
 
         # event tracking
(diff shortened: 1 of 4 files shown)


j-kanuch (approved):

rdr_service/api/ppsc_site_api.py
```diff
@@ -75,7 +75,9 @@ def validate_request(cls, req_data):
             raise BadRequest(f'Payload for Site is invalid: Required keys - {response_string}')
 
     def handle_site_updates(self, *, site_data: dict) -> Site:
-        # site data
+        if site_data.get("org_id").upper() == "SEEC_MOREHOUSE":
+            site_data["org_id"] = "DREF_MOREHOUSE"
+
```
dtharpeuno: This is not ideal that we have to do this and this might not be the last time, so would you add a dictionary that we can add to in the future to handle this.

dtharpeuno (approved):

PR #4138, opened 2025-10-30
[DA-5000] Add "pediatric" field on AW0

lancealopez:
Resolves *[DA-####](https://precisionmedicineinitiative.atlassian.net/browse/DA-5000)*

These changes are for the AW0 pipeline to change to use PPSC tables instead of legacy participant summary values. The other changes are to allow Pediatric test codes, and add the pediatric flag when generating an AW0. Old validation code was also removed. 
## Tests
- [] unit tests


rdr_service/genomic/genomic_data.py
```diff
@@ -148,7 +144,7 @@ def remaining_c2_participants():
                       SELECT ra.participant_id
                       FROM participant_race_answers ra
                           JOIN code cr ON cr.code_id = ra.code_id
-                              AND SUBSTRING_INDEX(cr.value, "_", -1) = "AIAN"
+                              AND SUBSTRING_INDEX(cr.value, '_', -1) = 'AIAN'
```
j-kanuch, line 147: Is participant_race_answers still getting updated?

lancealopez: No, I do not believe it is. I just did not change the cohort 1 and cohort 2 workflows as we will only be processing cohort 3 participants now. Think I should still change this?

j-kanuch: I think this might actually be a formatting change. I don't think `remaining_c2_participants` is executed at all anymore.

lancealopez:
Moved to draft as we will change the source to bigquery.

PR #4209, opened 2026-07-14
Schedule awardee insite data file drop to sites

shahwaiz14:
- Setting up an airflow dag to drop awardee insite data to some sites in their GCS buckets.

## Tests
- [] unit tests (Tested by running the dag)


rdr_service/tools/tool_libs/export_awardee_insite_data.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        rows = [dict(row.items()) for row in self.client.query(query)]
+        return rows
+
+    def get_data(self, awardee: str) -> list[dict]:
+        job_config = bigquery.QueryJobConfig(
+            query_parameters=[bigquery.ScalarQueryParameter("awardee", "STRING", awardee)]
+        )
+        query_job = self.client.query(awardee_insite_query, job_config=job_config)
+        rows = [dict(row.items()) for row in query_job.result()]
```
j-kanuch: How much data is expected to be included in these scheduled files? It looks like all records for each awardee.
Whatever is returned in the query will need to fit into memory here. It might not be an issue, depending on the volume of data. 
If memory is a concern, it might be better to move this process to Airflow and use the `BigQueryToGCSOperator` to directly export the files from BQ to GCS.
I'm not sure if the timing on this update allows for a refactor in Airflow, but in the future we'll probably want to move tasks like these to Airflow. Let me know if you are interested in doing that here; I have a working process for genomics that does a very similar thing and can be adapted.

shahwaiz14: oh sure. Let me do this in Airflow.

shahwaiz14: I’ve converted this to an Airflow DAG and moved it to `rdr_service/operations_pipeline_airflow/components/dags`. The DAG ran successfully without any issues in Prod.

kenny-skaggs (approved):

j-kanuch (approved):

